PR #1215, opened 2024-05-21
Release tag workflow fixes

kian99:
## Description

When pushing the latest tag v3.1.0 tag for JIMM there were several issues evident.
- We are still running charm workflows. The k8s charm has moved to a separate repo and the machine charm requires updates before we should carry on releasing it.
- The git version string used in the snap names was incorrect, the issue is the same one that has cropped up previously where the tag used was 1 older than the most recent, [see here](https://github.com/canonical/jimm/actions/runs/9171726083/job/25216681038#step:4:1347).

This PR:
- Removes the charm workflows. The actual charm code will be removed separately.
- Finally identifies the issue that was causing the tag issue (see https://github.com/actions/checkout/issues/882 and https://github.com/actions/checkout/issues/290)
- Fixes some naming in the workflows.

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [ ] Covered by unit tests

## Test instructions
I've tested that the tag fix works by running the workflows in my fork.

## Notes for code reviewers
NB! Before merging, there is one more issue where the publishing of the snaps [is failing](https://github.com/canonical/jimm/actions/runs/9171726083/job/25217141999#step:3:6). It looks like a secret is missing (which can be fixed outside of this PR) but if the issue is more than that I'd like to fix it in this PR. We just need to confirm whether the secret is missing or not.


alesstimec (approved):

ale8k (approved):
Ty for removing old crap too

babakks (approved):

PR #1251, opened 2024-07-01
CSS-9328 stream connector

kian99:
## Description
**NB**: This PR now incorporates the changes from #1269, this is necessary to ensure tests pass. But those changes will cause the Juju CLI to stop working. So this PR should only land after Juju CLI 3.5.4 is released.

This PR addresses an issue where users cannot use commands like `juju debug-log` with JAAS. I have explained the issue in more detail in https://github.com/juju/juju/pull/17606, copy-pasting from there.

>Running `juju debug-log` does the following:
>1. Connects to the Juju controller at `/model/<uuid>/api` creating an API client in the normal fashion, establishing a websocket connection and performing a login request.
>2. It then goes on to establish a new separate connection to the `/model/<uuid>/log` endpoint, eventually establishing a websocket but instead of sending an RPC login request, it uses the apiClient's `ConnectStream` method where an HTTP basic auth header is sent that includes the user's password and includes the macaroons as cookies.
>3. Proceeds to read from the websocket to stream logs.
>
>With JAAS' introduction of login providers and particularly the session login provider, it was realised that there is no mechanism in place to send the session token used with OIDC login in the basic auth header used in the `ConnectStream` method. Because JAAS acts as a proxy layer for all controller and model level calls, JAAS has no way of authenticating requests to `/model/<uuid>/log`.

This PR introduces a new type of proxy called the `streamProxy` that allows for any Juju streaming endpoints to be proxied. Currently only the `/log` endpoint is allowed to be accessed provided that the user has model `read` permission.

This PR is currently a draft until the following Juju changes have landed, at which point we can update JIMM to use the latest Juju version. https://github.com/juju/juju/pull/17606 and https://github.com/juju/juju/pull/17615

Fixes [CSS-9328](https://warthogs.atlassian.net/browse/CSS-9328)

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [x] Covered by unit tests

## Test instructions
To test the changes,
- Checkout Juju 3.5 and merge the two branches mentioned above.
- Run `go mod edit -replace github.com/juju/juju=<path-to-juju-repo>`
- `docker compose --profile dev pull` (making sure to grab the latest container images for tools like Air which is required for the bump in Go.mod version.
- Add a controller and model to JIMM and try `juju debug-log`, ensuring you build the Juju CLI from source to incorporate the fixes.

## Notes for code reviewers
I recommend reviewing the commits in order. The second has mechanical changes where I've moved the API proxy into a separate file.

[CSS-9328]: https://warthogs.atlassian.net/browse/CSS-9328?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


alesstimec (changes requested):

ale8k (commented):
Half a review so far, only a little left and will review separately

alesstimec (dismissed):
LGTM once ale8k's comments are addressed

PR #1336, opened 2024-09-02
Improve errors and edge case handling in model proxy

kian99:
## Description

This PR modifies the model proxy to improve the error logging and simplify logic in some edge cases.

The reason for this change was that when running JIMM in a test setup there were constant debug logs about "error reading from client" or "error reading from controller" which were completely expected, every time the client closes the websocket our routines waiting to read from the websocket stop and return. In these cases, there were debug and error level logs which were not necessary because this was the normal mode of operation. 

The first commit in this PR simply makes the client/controller proxy return a nil error when they fail to read from their respective websocket connections.

The second commit in this PR was done after taking a look at whether we could simplify the complicated logic when handling `ctx.Done()` involving a mutex. It was thankfully easy to clean this up because we already handle things gracefully when the client connection is closed so we can simply close the client connection to clean things up.

Finally, I noticed an edge case that wasn't handled and added a test for it. If the connection to the controller was closed unexpectedly, the model proxy would not handle this gracefully. Now if the controller proxy routine ends early, we handle this by stopping the model proxy.

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [x] Covered by unit tests

## Test instructions
Ran the relevant tests with `-race` and/or `-count=1000`. E.g.
```
go test -race -timeout 30s github.com/canonical/jimm/v3/internal/rpc -count=10
go test -timeout 30s -run ^TestProxySockets$ github.com/canonical/jimm/v3/internal/rpc -count=1000
```
## Notes for code reviewers
I suggest reviewing this PR by each commit individually to see the changes more clearly.

### DIFF
--- a/internal/rpc/proxy.go
+++ b/internal/rpc/proxy.go
@@ -121,8 +121,6 @@ func ProxySockets(ctx context.Context, helpers ProxyHelpers) error {
 	}()
 	var err error
 	select {
-	// No cleanup is needed on error, when the client closes the connection
-	// all go routines will proceed to error and exit.
 	case err = <-errChan:
 		if err != nil {
 			zapctx.Debug(ctx, "Proxy error", zap.Error(err))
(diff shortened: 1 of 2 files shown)


internal/rpc/proxy.go
```diff
@@ -339,8 +332,8 @@ func (p *clientProxy) start(ctx context.Context) error {
 		zapctx.Debug(ctx, "Reading on client connection")
 		msg := new(message)
 		if err := p.src.readJson(&msg); err != nil {
```
SimoneDutto, line 334: reading at gorilla/websocket/json.go can return an error for two reasons:
- error on marshaling the json
- error on closing the connection.

I think your PR wants to hide error on closing connection, and that is fine from what I understand, but we are also masking errors on json marshaling, which could potentially lead to situation where the clients sends us non-json messages and we are not seeing the error.

Wouldn't be better to check here for marshaling error? And if not ignoring?

kian99: Yes good point

kian99: Fixed in #1345, I tried updating this PR but Github just hangs, I believe because this PR is still from my old private fork of JIMM.

PR #1343, opened 2024-09-03
fixes for composite action

kian99:
## Description

This PR makes some tweaks to JIMM composite action.
- Use the `working_dir` keyword when referencing scripts in the repo alongside the `${{ github.action_path }}` context variable.
- Ensure service account in the integration test is a JIMM admin.

I've tested these changes [here](https://github.com/kian99/terraform-provider-juju/actions/runs/10672951389/job/29581320745).

### DIFF
--- a/.github/actions/test-server/action.yaml
+++ b/.github/actions/test-server/action.yaml
@@ -41,15 +41,21 @@ runs:
         username: ${{ github.actor }}
         password: ${{ inputs.ghcr-pat }}
 
+    # We can't use a make target here because a composite action
+    # doesn't have a .git folder when checked out.
     - name: Start server based on released version
       if: ${{ inputs.jimm-version != 'dev' }}
-      run: make integration-test-env
+      run: |
+        cd local/traefik/certs; ./certs.sh; cd - && \
+        docker compose --profile test up -d --wait
       shell: bash
+      working-directory: ${{ github.action_path }}/../../..
       env:
         JIMM_VERSION: ${{ inputs.jimm-version }}
 
     - name: Start server based on development version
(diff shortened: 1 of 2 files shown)


.github/actions/test-server/action.yaml
```diff
@@ -41,15 +41,21 @@ runs:
...
+    # We can't use a make target here because a composite action
+    # doesn't have a .git folder when checked out.
     - name: Start server based on released version
       if: ${{ inputs.jimm-version != 'dev' }}
-      run: make integration-test-env
+      run: |
+        cd local/traefik/certs; ./certs.sh; cd - && \
+        docker compose --profile test up -d --wait
       shell: bash
+      working-directory: ${{ github.action_path }}/../../..
```
ale8k, line 52: Interesting...

ale8k (approved):

.github/actions/test-server/action.yaml
SimoneDutto, line 52: Second answer mentions that it should be possible to set a default work dir for each step in a action:
https://stackoverflow.com/questions/58139175/running-actions-in-another-directory

kian99: Unfortunately it's not supported in composite actions, see Github's [design doc here](https://github.com/actions/runner/blob/main/docs/adrs/0549-composite-run-steps.md#defaults).

SimoneDutto: in a follow up PR we will try the default on top of the job list, like so:
```
name: CI

on:
 ...

defaults:
 run:
  working-directory: ./app
```

SimoneDutto (approved):

PR #1344, opened 2024-09-03
CSS-9328 stream connector

kian99:
## Description

Recreates #1251

**NB**: This PR now incorporates the changes from #1269, this is necessary to ensure tests pass. But those changes will cause the Juju CLI to stop working. So this PR should only land after Juju CLI 3.5.4 is released.

This PR addresses an issue where users cannot use commands like `juju debug-log` with JAAS. I have explained the issue in more detail in https://github.com/juju/juju/pull/17606, copy-pasting from there.

>Running `juju debug-log` does the following:
>1. Connects to the Juju controller at `/model/<uuid>/api` creating an API client in the normal fashion, establishing a websocket connection and performing a login request.
>2. It then goes on to establish a new separate connection to the `/model/<uuid>/log` endpoint, eventually establishing a websocket but instead of sending an RPC login request, it uses the apiClient's `ConnectStream` method where an HTTP basic auth header is sent that includes the user's password and includes the macaroons as cookies.
>3. Proceeds to read from the websocket to stream logs.
>
>With JAAS' introduction of login providers and particularly the session login provider, it was realised that there is no mechanism in place to send the session token used with OIDC login in the basic auth header used in the `ConnectStream` method. Because JAAS acts as a proxy layer for all controller and model level calls, JAAS has no way of authenticating requests to `/model/<uuid>/log`.

This PR introduces a new type of proxy called the `streamProxy` that allows for any Juju streaming endpoints to be proxied. Currently only the `/log` endpoint is allowed to be accessed provided that the user has model `read` permission.

Fixes [CSS-9328](https://warthogs.atlassian.net/browse/CSS-9328)

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [x] Covered by unit tests

## Test instructions
To test the changes,
- Install the juju client snap from 3.5/edge and bootstrap a controller.
- Add the controller to JIMM and add a model to JIMM and try `juju debug-log`.

## Notes for code reviewers
I recommend reviewing the commits in order.

The CLI test changes were necessitated by [this PR](https://github.com/juju/juju/pull/17667/files#diff-6a09735ead260521301d78800279f9c2c610289ba71c3b362ae95e0dd85a50f4R122-R124) in Juju which enforced more strictness on how the Juju CLI would handle login.

[CSS-9328]: https://warthogs.atlassian.net/browse/CSS-9328?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/jujuapi/api.go
```diff
@@ -36,9 +36,17 @@ func APIHandler(ctx context.Context, jimm *jimm.JIMM, p Params) http.Handler {
...
 	mux := http.NewServeMux()
-	mux.Handle("/", &jimmhttp.WSHandler{
+	mux.Handle("/{uuid}/api", &jimmhttp.WSHandler{
 		Upgrader: websocketUpgrader,
-		Server:   modelProxyServer{jimm: jimm},
+		Server: &apiProxier{apiServer: apiServer{
+			jimm: jimm,
+		}},
+	})
+	mux.Handle("/{uuid}/", &jimmhttp.WSHandler{
```
SimoneDutto: is this endpoint the correct one for the stream? if not I would suggest to be more specific with the path (e.g. `/{uuid}/stream_path`).
This is to avoid confusion when juju try to use some missing endpoint via jimm and it fallbacks to this

internal/jujuapi/streamproxy.go
```diff
@@ -0,0 +1,150 @@
...
+		writeError(err.Error(), errors.CodeModelNotFound)
+		return
+	}
+	if ok, err := checkPermission(ctx, finalPath, user, model.ResourceTag()); err != nil {
+		writeError(err.Error(), errors.CodeUnauthorized)
+		return
+	} else if !ok {
+		writeError(fmt.Sprintf("unauthorized access to endpoint: %s", finalPath), errors.CodeUnauthorized)
+		return
+	}
```
SimoneDutto: why didn't you put this in the `Authenticate`?

internal/jujuapi/streamproxy.go
```diff
@@ -0,0 +1,150 @@
...
+			String: modelUUID,
+			Valid:  modelUUID != "",
+		},
+	}
+	if err := s.jimm.Database.GetModel(context.Background(), &model); err != nil {
+		zapctx.Error(ctx, "failed to find model", zap.String("uuid", modelUUID), zap.Error(err))
+		return dbmodel.Model{}, fmt.Errorf("failed to find model: %s", err.Error())
+	}
+	return model, nil
+}
```
SimoneDutto: i feel like this method belongs to JIMM and not in the API layer.
I have created it in JIMM in my branch, if you want it i can refactor yours as well when i create my pr

internal/jujuapi/streamproxy.go
```diff
@@ -0,0 +1,150 @@
...
+
+// proxyStreams starts a simple proxy for 2 websockets.
+// After starting the proxy we listen for the first error
+// returned and then close both connections before waiting
+// for an error from the second connection.
+func proxyStreams(src, dst base.Stream) {
+	errChan := make(chan error, 2)
+	go func() { errChan <- proxy(src, dst) }()
+	go func() { errChan <- proxy(dst, src) }()
+	<-errChan
```
SimoneDutto: shouldn't we catch and log these errors?

internal/jujuapi/streamproxy.go
```diff
@@ -0,0 +1,150 @@
...
+
+func proxy(src base.Stream, dst base.Stream) error {
+	for {
+		var data map[string]any
+		err := src.ReadJSON(&data)
+		if err != nil {
+			return err
+		}
+		err = dst.WriteJSON(data)
+		if err != nil {
```
SimoneDutto: i think you should use the same approach you used in a previous pr, where you don't return errors for closed connections, but you do for json errors.

internal/jujuapi/streamproxy_test.go
```diff
@@ -0,0 +1,57 @@
...
+	ofganames "github.com/canonical/jimm/v3/internal/openfga/names"
+)
+
+type streamProxySuite struct {
+	websocketSuite
+}
+
+var _ = gc.Suite(&streamProxySuite{})
+
+func (s *streamProxySuite) TestDebugLogs(c *gc.C) {
```
SimoneDutto, line 23: i suggest to expand the debug log good test with some actual debug logs, if that's possible

internal/jujuapi/api.go
kian99: It's the fallback for any model endpoint which we haven't defined, which is valid behaviour for now. If the handler receives an endpoint it doesn't recognise it will return an appropriate error.

internal/jujuapi/streamproxy.go
kian99: Because this is authorization so it deserves to go somewhere else.

internal/jujuapi/streamproxy.go
kian99: I agree, but the current apiServer accepts a concrete JIMM struct rather than a JIMM interface. We need to do a larger refactor to pass in the interface then we can refactor this. So I think it's best left for a follow-up.

internal/jujuapi/streamproxy.go
kian99: I'm inclined to say no because we can't do anything with these errors besides possibly logging them.

internal/jujuapi/streamproxy.go
kian99: I would agree, but I don't think it's really possible for the marshal/unmarshal parts to trigger an error in this scenario. But I'll take another look at this.

internal/jujuapi/streamproxy_test.go
kian99: For some reason the Juju controller deployed as part of the conn suite doesn't return debug logs. Maybe because there is nothing in the model.

internal/jujuapi/api.go
SimoneDutto: i would rather prefer the mux to give a 404 then the wshandler to give back some "cannot upgrade" error.
Don't you think it's best to have the full path here also for clarity and when in the future we need to add new endpoints under this model handler?

internal/jujuapi/streamproxy.go
SimoneDutto: i do think logging the error in this case is valid.
Just imagine the stream proxy not working without any error logs.

internal/jujuapi/streamproxy_test.go
SimoneDutto: i think that if you issue some command like "deploy application" it should log something.

internal/jujuapi/api.go
kian99: Yeah that's fair, I'll switch to being explicit in the handler

internal/jujuclient/dial.go
```diff
@@ -331,7 +340,22 @@ func (c *Connection) Context() context.Context {
...
+	if err != nil {
+		return nil, errors.E(op, err)
+	}
+	requestHeader := jujuhttp.BasicAuthHeader(names.NewUserTag(user).String(), pass)
+
+	modelTag, ok := c.ModelTag()
+	if !ok {
+		return nil, errors.E(op, "no model found")
+	}
+	conn, err := rpc.Dial(c.ctx, c.ctl, modelTag, path, requestHeader)
```
ale8k, line 355: I think a rpc.DialWithBasicAuth would be a nice wrapper instead, and another one called something like rpc.DialNoAuth (or better naming plz aha), then you can put request header into it.

Also just move line 348 above this if you don't choose to do this change , because we compute the requestheader then get the model tag, but we waste the request header.

internal/jujuapi/streamproxy.go
```diff
@@ -0,0 +1,150 @@
...
+	controllerStream, err := api.ConnectStream(finalPath, nil)
+	if err != nil {
+		zapctx.Error(ctx, "failed to connect stream", zap.Error(err))
+		writeError(fmt.Sprintf("failed to connect stream: %s", err.Error()), errors.CodeConnectionFailed)
+		return
+	}
+	proxyStreams(clientConn, controllerStream)
+}
+
+func (s streamProxier) getModel(ctx context.Context, modelUUID string) (dbmodel.Model, error) {
```
ale8k: This doesn't feel right to be done here, we're going to db layer in jujuapi?

internal/jujuapi/api_test.go
```diff
@@ -32,7 +32,7 @@ func (s *apiSuite) SetUpTest(c *gc.C) {
 
 	mux := http.NewServeMux()
 	mux.Handle("/api", jujuapi.APIHandler(ctx, s.JIMM, s.Params))
-	mux.Handle("/model/", jujuapi.ModelHandler(ctx, s.JIMM, s.Params))
+	mux.Handle("/model/", http.StripPrefix("/model", jujuapi.ModelHandler(ctx, s.JIMM, s.Params)))
```
ale8k, line 35: Why strip prefix?

internal/jujuapi/streamproxy.go
```diff
@@ -0,0 +1,150 @@
...
+			Message: msg,
+			Code:    string(code),
+		}
+		err := clientConn.WriteJSON(errResult)
+		if err != nil {
+			zapctx.Error(ctx, "failed to write error message to client", zap.Error(err), zap.Any("client message", errResult))
+		}
+	}
+	user, err := s.jimm.UserLogin(ctx, auth.SessionIdentityFromContext(ctx))
+	if err != nil {
```
ale8k, line 62: Nit: you don't tend to leave any space between lines, i.e.,:
```go
f, err := something()
if err != nil {
}
l, err := something2()
if err != nil {
}
g, err := something3()
if err != nil {
}
```
and it can be kinda difficult to read when blocked together like this.

Where as:
```go
f, err := something()
if err != nil {
}

l, err := something2()
if err != nil {
}

g, err := something3()
if err != nil {
}
```

Reads a bit easier.

ale8k (dismissed):
lgtm, some minor comments

internal/jujuapi/streamproxy_test.go
```diff
@@ -0,0 +1,57 @@
...
+	conn := s.open(c, &api.Info{ModelTag: s.Model.ResourceTag()}, "bob")
+	defer conn.Close()
+	_, err := common.StreamDebugLog(context.TODO(), conn, common.DebugLogParams{})
+	c.Assert(err, gc.IsNil)
```
ale8k, line 29: I don't think I fully understand this test, does no error really mean its working?

internal/jujuapi/api_test.go
kian99: It's because of the change in [jujuapi/api.go](https://github.com/canonical/jimm/pull/1344/files#diff-fe36fdbf3a8c97b34cc281bca11eb3bba8ff8fe8bfc80b84e06abf48c5870391R39), before we just matched `/` onto the api proxy. But now we have some different endpoints, so we just define the relative paths in there and the at the top level here (and in service.go where we also do this), we strip the prefix before passing on the rest of the path. This [SO post](https://stackoverflow.com/questions/27945310/why-do-i-need-to-use-http-stripprefix-to-access-my-static-files) also probably explains it better than I did.

internal/jujuapi/streamproxy.go
kian99: Will change.

internal/jujuapi/streamproxy.go
kian99: I agree, Simone mentioned the same thing, pasting my comment here
>I agree, but the current apiServer accepts a concrete JIMM struct rather than a JIMM interface. We need to do a larger refactor to pass in the interface then we can refactor this. So I think it's best left for a follow-up.

internal/jujuapi/streamproxy.go
kian99: Adding logging for them.

internal/jujuapi/streamproxy_test.go
kian99: It actually does funnily enough, because it tests that we can go all the way from the client through to the Juju controller without any auth errors, etc. It would be better to also test there are some debug logs, but I was always getting an empty list of debug logs (probably because there is no application deployed but I'll take another look at that).

internal/jujuclient/dial.go
kian99: I like the idea for `rpc.DialWithBasicAuth` but I think what we could do is refactor the parameter list into an object, then you can manipulate that object before passing it to dial potentially. 

I'd like to leave this for a follow-up, I'll implement the second suggestion.

ale8k: Not a fan of a params object in this scenario, we have two different behaviour dials. Lets just have two dial? I would think wrapping a struct up in params is really more for large amounts of params.

internal/jujuapi/streamproxy.go
ale8k: You can create the method on jimm and call it on concrete for now. Beats doing this here honestly.

kian99: Will do

internal/jujuapi/api_test.go
ale8k: Doesn't stripprefix just remove r.path.URL prefixes? Your handler would work all the same? Are we using r.path.URL in the handler anywhere?

internal/jujuapi/streamproxy.go
ale8k: Thanks, I don't mean to be picky it just makes things a bit difficult if we don't have a bit of cushion between things aha.

internal/jujuapi/streamproxy_test.go
ale8k: Yeah let's add an app and try get some logs somehow... If it's super difficult, leave a comment explaining why we aren't testing any logs coming through please.

internal/jujuapi/api_test.go
kian99: I think if we don't use stripprefix then we need to put /model/ in the handler, this just strips the parts that are common.

internal/jujuapi/streamproxy.go
kian99: Moved the stream proxy to the rpc package where it feels more appropriate and now I can reuse that work.

internal/jujuapi/api_test.go
ale8k: Can you try without please?

internal/jujuapi/api.go
alesstimec: there are other endpoints served by the juju contoller at /{uuid}/...

internal/jujuclient/dial.go
kian99: Will leave this up to a follow-up PR. But moved the code around so that the request header is not created unnecessarily.

internal/jujuapi/api_test.go
kian99: It would work but it would mean adding /model/ to the routes in `jujuapi.ModelHandler`, I did it like this intentionally.

internal/jujuapi/streamproxy_test.go
kian99: Deploying an app seems tricky, I took a brief look at the [Deploy Client](https://github.com/juju/juju/blob/3.5/api/client/application/client.go#L126) and I don't want to deploy an actual charm from charmhub in these tests. So I've opted to leave a comment.

internal/jujuapi/api.go
kian99: True, I've changed this to be explicit about the endpoint.

internal/jimm/model.go
```diff
@@ -686,6 +687,20 @@ func (j *JIMM) AddModel(ctx context.Context, user *openfga.User, args *ModelCrea
 	return mi, nil
 }
 
+func (j *JIMM) GetModel(ctx context.Context, uuid string) (dbmodel.Model, error) {
```
alesstimec, line 690: godoc, please

internal/jujuapi/streamproxy_test.go
kian99: Will leave this for the follow up as discussed.

internal/auth/oauth2.go
```diff
@@ -334,7 +334,7 @@ func (as *AuthenticationService) MintSessionToken(email string) (string, error)
 func (as *AuthenticationService) VerifySessionToken(token string) (_ jwt.Token, err error) {
 	const op = errors.Op("auth.AuthenticationService.VerifySessionToken")
 	errorFn := func(message string) error {
-		return errors.E(op, message, errors.CodeUnauthorized)
+		return errors.E(op, message, errors.CodeSessionTokenInvalid)
```
alesstimec, line 340: afaik this is also an important change, right? because it will tell juju cli that the token is invalid and differentiate it from the "regular" unauthorized error coming from the juju controller.. perhaps it deserves a godoc, so that somebody doesn't just change it in the future..

internal/cmdtest/jimmsuite.go
```diff
@@ -150,6 +154,17 @@ func (s *JimmCmdSuite) AddAdminUser(c *gc.C, email string) {
 	c.Assert(err, gc.IsNil)
 }
 
+// SetupCLIAccess should be run at the start of all CLI tests that want to communicate with JIMM.
+// It will ensure a user is created in JIMM's in-memory accounts store that matches the username passed in.
+// This is necessary as Juju's CLI library validates the logged-in user matches the expected user from the accounts store.
+// This function will return a login provider that can be used to handle authentication.
+func (s *JimmCmdSuite) SetupCLIAccess(c *gc.C, username string) api.LoginProvider {
```
alesstimec, line 161: this test change could possibly be done in a separate PR, because it just makes the current PR harder to read and does not seem really related to the main topic of the PR.

internal/jimm/model.go
alesstimec: i suppose authorization check for this call is made at the jujuapi level?

internal/jimmhttp/websocket.go
```diff
@@ -65,74 +68,16 @@ func (h *WSHandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
-	data := websocket.FormatCloseMessage(websocket.CloseNormalClosure, "")
-	if err := conn.WriteControl(websocket.CloseMessage, data, time.Time{}); err != nil {
-		zapctx.Error(ctx, "cannot write close message", zap.Error(err))
-	}
-}
-
 func writeInternalServerErrorClosure(ctx context.Context, conn *websocket.Conn, err any) {
 	data := websocket.FormatCloseMessage(websocket.CloseInternalServerErr, fmt.Sprintf("%v", err))
 	if err := conn.WriteControl(websocket.CloseMessage, data, time.Time{}); err != nil {
 		zapctx.Error(ctx, "cannot write close message", zap.Error(err))
```
alesstimec, line 77: hmm.. error or a warning, i don't know..

internal/jujuapi/streamproxy.go
```diff
@@ -0,0 +1,113 @@
+// Copyright 2024 Canonical.
+package jujuapi
```
alesstimec, line 3: newline between copyright and package, otherwise we'll get copyright as package godoc..

internal/jujuapi/streamproxy.go
```diff
@@ -0,0 +1,113 @@
...
+	"go.uber.org/zap"
+
+	"github.com/canonical/jimm/v3/internal/auth"
+	"github.com/canonical/jimm/v3/internal/errors"
+	"github.com/canonical/jimm/v3/internal/jimmhttp"
+	"github.com/canonical/jimm/v3/internal/openfga"
+	jimmRPC "github.com/canonical/jimm/v3/internal/rpc"
+)
+
+// A streamProxier serves all HTTP endpoints by proxying
```
alesstimec: due to implementation of the checkPermission, this only serves the /log endpoint atm

internal/jujuapi/streamproxy.go
```diff
@@ -0,0 +1,113 @@
...
+		return
+	}
+
+	model, err := s.jimm.GetModel(ctx, uuid)
+	if err != nil {
+		writeError(err.Error(), errors.CodeModelNotFound)
+		return
+	}
+
+	if ok, err := checkPermission(ctx, finalPath, user, model.ResourceTag()); err != nil {
```
alesstimec: we could do this permission check before fetching the controller on L74 since we have the model UUID

internal/jujuclient/dial.go
```diff
@@ -331,7 +340,23 @@ func (c *Connection) Context() context.Context {
...
+	modelTag, ok := c.ModelTag()
+	if !ok {
+		return nil, errors.E(op, "no model found")
+	}
+
+	user, pass, err := c.dialer.ControllerCredentialsStore.GetControllerCredentials(c.ctx, c.ctl.Name)
+	if err != nil {
+		return nil, errors.E(op, err)
+	}
+	requestHeader := jujuhttp.BasicAuthHeader(names.NewUserTag(user).String(), pass)
```
alesstimec, line 353: please investigate if we can use the JWT to authenticate with the juju controller instead of basic auth..

internal/rpc/streamproxy.go
```diff
@@ -0,0 +1,47 @@
+// Copyright 2024 Canonical.
+package rpc
```
alesstimec, line 3: same here, we need a new line between copyright and package

internal/jujuapi/streamproxy_test.go
ale8k: Sure sure

internal/jujuapi/api_test.go
ale8k: Are you sure?

ale8k (dismissed):

internal/jujuapi/api_test.go
kian99: That's what I recall happening, but this was like 2 months ago. I'll try it again now and let you know.

internal/auth/oauth2.go
kian99: Yes I'll add a godoc

internal/cmdtest/jimmsuite.go
kian99: I agree, and that's my bad. I think I'd initially kept these changes separated into a specific commit but there are now a lot of commits.
I made the change because the CLI tests were broken when I upgraded the Juju dependency.

internal/jimm/model.go
kian99: Yes

internal/jujuclient/dial.go
kian99: Looking at this with Simone, it's not implemented on Juju's end. So we'll need to keep this as is for now until we have Juju controller's that support JWT for HTTP auth.

SimoneDutto (approved):

alesstimec (approved):

PR #1345, opened 2024-09-04
Cleanup model proxy

kian99:
## Description

This PR recreates #1336 (that PR was coming from my old fork)

This PR modifies the model proxy to improve the error logging and simplify logic in some edge cases.

The reason for this change was that when running JIMM in a test setup there were constant debug logs about "error reading from client" or "error reading from controller" which were completely expected, every time the client closes the websocket our routines waiting to read from the websocket stop and return. In these cases, there were debug and error level logs which were not necessary because this was the normal mode of operation. 

The first commit in this PR simply makes the client/controller proxy return a nil error when they fail to read from their respective websocket connections.

The second commit in this PR was done after taking a look at whether we could simplify the complicated logic when handling `ctx.Done()` involving a mutex. It was thankfully easy to clean this up because we already handle things gracefully when the client connection is closed so we can simply close the client connection to clean things up.

Finally, I noticed an edge case that wasn't handled and added a test for it. If the connection to the controller was closed unexpectedly, the model proxy would not handle this gracefully. Now if the controller proxy routine ends early, we handle this by stopping the model proxy.

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [x] Covered by unit tests

## Test instructions
Ran the relevant tests with `-race` and/or `-count=1000`. E.g.
```
go test -race -timeout 30s github.com/canonical/jimm/v3/internal/rpc -count=10
go test -timeout 30s -run ^TestProxySockets$ github.com/canonical/jimm/v3/internal/rpc -count=1000
```
## Notes for code reviewers
I suggest reviewing this PR by each commit individually to see the changes more clearly.


internal/rpc/proxy.go
```diff
@@ -316,16 +313,22 @@ func (p *modelProxy) auditLogMessage(msg *message, isResponse bool) error {
 	return nil
 }
 
+func unexpectedReadError(err error) bool {
+	closeError := websocket.IsUnexpectedCloseError(err,
+		websocket.CloseNormalClosure,
+		websocket.CloseNoStatusReceived,
+		websocket.CloseAbnormalClosure)
+	_, unmarshalError := err.(*json.InvalidUnmarshalError)
+	return closeError || unmarshalError
```
ale8k, line 322: I don't like this logic, don't do this. Just check if we have a close error, if we do return else check a marshal error. The name is also confusing, as this is an unexpected closure versus unmarshalling.

internal/rpc/proxy.go
```diff
@@ -316,16 +313,22 @@ func (p *modelProxy) auditLogMessage(msg *message, isResponse bool) error {
...
 type clientProxy struct {
 	modelProxy
 	wg                   sync.WaitGroup
 	errChan              chan error
 	createControllerConn func(context.Context) (WebsocketConnectionWithMetadata, error)
-	// mu synchronises changes to closed and modelproxy.dst, dst is is only created
-	// at some unspecified point in the future after a client request.
-	mu     sync.Mutex
-	closed bool
+	connectController    sync.Once
```
ale8k, line 331: I like this, much clearer.

ale8k (approved):
Looks good, addition of once is nice.

SimoneDutto (approved):
LGTM, i think i added comments in the other pr and it seems you have already addressed them in this new one

internal/rpc/proxy.go
kian99: Will fix this in a follow-up.

PR #1347, opened 2024-09-04
Css 9402/resources

SimoneDutto:
## Description

This pr adds the resources endpoint.
The expected response is a paginated list of Resources (clouds, controllers, models, service_accounts, application offers).

We implemented this with a raw query with multiple union for the different tables.
The reason behind a raw query instead of a view is that view doesn't provide any cons in our situation.

> Fly by update of the db client to enable prepared statements, it should improve the performance caching request and it was a simple flag in the gorm client [Doc](https://gorm.io/docs/performance.html#Caches-Prepared-Statement)

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [ ] Covered by unit tests


internal/common/pagination/pagination.go
```diff
@@ -62,6 +62,22 @@ func CreatePagination(sizeP, pageP *int, total int) (int, *int, LimitOffsetPagin
 	return page, nextPage, NewOffsetFilter(pageSize, offset)
 }
 
+// CreatePagination returns the current page, the page size, and the pagination.LimitOffsetPagination.
+// This method is different approach to the method `CreatePagination` when we don't have the total number of records.
+func CreatePaginationWithoutTotal(sizeP, pageP *int) (int, int, LimitOffsetPagination) {
```
kian99: We should probably return a struct here and in `CreatePagination` because I'm very confused looking at this function and even the test, what `int, int` represent.

internal/db/resources.go
```diff
@@ -0,0 +1,108 @@
...
+import (
+	"context"
+	"database/sql"
+
+	"github.com/canonical/jimm/v3/internal/errors"
+	"github.com/canonical/jimm/v3/internal/servermon"
+)
+
+// MULTI_TABLES_RAW_SQL contains the raw query fetching entities from multiple tables, with their respective entity parents.
+const MULTI_TABLES_RAW_SQL = `
```
ale8k: This name isn't ok lol

internal/common/pagination/pagination_test.go
```diff
@@ -130,6 +130,18 @@ func TestCreatePagination(t *testing.T) {
 	}
 }
 
+// test the requested size is 1 more than then page size.
+func TestCreatePaginationWithoutTotal(t *testing.T) {
+	c := qt.New(t)
+	pPage := utils.IntToPointer(0)
+	pSize := utils.IntToPointer(10)
+	page, size, pag := pagination.CreatePaginationWithoutTotal(pSize, pPage)
```
kian99, line 138: `page` is the offset? 
`size` is the number of returned elements? Why is it the page size + 1? 
And `pag` if the filter, maybe the name could be changed.

internal/db/resources.go
```diff
@@ -0,0 +1,108 @@
...
+	Name       string
+	ParentId   sql.NullString
+	ParentName string
+	ParentType string
+}
+
+// ListResources returns a list of models, clouds, controllers, service accounts, and application offers, with its respective parents.
+// It has been implemented with a raw query because this is a specific implementation for the ReBAC Admin UI.
+func (d *Database) ListResources(ctx context.Context, limit, offset int) (_ []Resource, err error) {
+	const op = errors.Op("db.GetMultipleModels")
```
ale8k: Op is wrong

internal/db/resources_test.go
```diff
@@ -0,0 +1,95 @@
+// Copyright 2024 Canonical.
```
ale8k: We call the db files resources, but the jimm pkg resource, can we make it consistent?

internal/db/resources.go
```diff
@@ -0,0 +1,108 @@
...
+const MULTI_TABLES_RAW_SQL = `
+(
+	SELECT 'application_offer' AS type, 
+			application_offers.uuid AS id, 
+			application_offers.name AS name, 
+			models.uuid AS parent_id,
+			models.name AS parent_name,
+			'model' AS parent_type
+	FROM application_offers
+	JOIN models ON application_offers.model_id = models.id
```
kian99: Why do we need the parent objects? Does the rebac-admin spec require that?

internal/db/resources_test.go
```diff
@@ -0,0 +1,95 @@
...
+	"database/sql"
+
+	qt "github.com/frankban/quicktest"
+	"github.com/juju/juju/state"
+
+	"github.com/canonical/jimm/v3/internal/db"
+	"github.com/canonical/jimm/v3/internal/dbmodel"
+)
+
+func (s *dbSuite) SetupDB(c *qt.C) (dbmodel.Model, dbmodel.Controller, dbmodel.Cloud) {
```
ale8k: Do we need this on the suite? Also bit surprised it didn't exist already

internal/jimm/resource.go
```diff
@@ -0,0 +1,22 @@
...
+
+	"github.com/canonical/jimm/v3/internal/common/pagination"
+	"github.com/canonical/jimm/v3/internal/db"
+	"github.com/canonical/jimm/v3/internal/errors"
+	"github.com/canonical/jimm/v3/internal/openfga"
+)
+
+// ListResources returns a list of resources known to JIMM with a pagination filter.
+func (j *JIMM) ListResources(ctx context.Context, user *openfga.User, filter pagination.LimitOffsetPagination) ([]db.Resource, error) {
+	const op = errors.Op("jimm.GetResources")
```
ale8k: Op is wrong

internal/rebac_admin/resources_integration_test.go
```diff
@@ -0,0 +1,158 @@
...
+			size:         utils.IntToPointer(2),
+			page:         utils.IntToPointer(2),
+			wantPage:     2,
+			wantSize:     1,
+			wantNextpage: nil,
+			ids:          []testEntity{ids[4]},
+		},
+	}
+	for _, t := range testCases {
+
```
ale8k: Random new line?

internal/rebac_admin/resources_integration_test.go
```diff
@@ -0,0 +1,158 @@
...
+  region: test-cloud-region
+  cloud-credential: cred-1
+  owner: alice@canonical.com
+`
+
+// TestPatchIdentityEntitlements tests the patching of entitlements for a specific identityId,
+// adding and removing relations after the setup.
+// Setup: add user to a group, and add models to the user.
+func (s *resourcesSuite) TestPatchIdentityEntitlements(c *gc.C) {
+	// initialization
```
ale8k: No need for this comment

internal/rebac_admin/resources_integration_test.go
```diff
@@ -0,0 +1,158 @@
...
+  controller: controller-1
+  cloud: test-cloud
+  region: test-cloud-region
+  cloud-credential: cred-1
+  owner: alice@canonical.com
+`
+
+// TestPatchIdentityEntitlements tests the patching of entitlements for a specific identityId,
+// adding and removing relations after the setup.
+// Setup: add user to a group, and add models to the user.
```
ale8k: ??

ale8k (commented):
Initial run over, will go over again, looking more closely at query and pagination. Not keen on having test with integration in the name...  You can definitely unit test just the handler with mocks to test the handlers logic. This probably something to discuss at standup.

internal/common/pagination/pagination.go
kian99: Or I like the idea of named returns here.

internal/rebac_admin/resources.go
```diff
@@ -0,0 +1,60 @@
...
+	// Otherwise next page is empty.
+	var nextPage *int
+	if len(res) == pageSize {
+		nPage := page + 1
+		nextPage = &nPage
+		res = res[:len(res)-1]
+	}
+	rRes := make([]resources.Resource, len(res))
+	for i, u := range res {
+		rRes[i] = utils.FromDbResourcesToResources(u)
```
kian99: nit: what about `ToRebacResources`.
Falls in line with the other conversion methods we have.

internal/rebac_admin/resources.go
```diff
@@ -0,0 +1,60 @@
...
+		return nil, err
+	}
+	// We fetch one record more than the page size. Then, we set the next page if we have this many records.
+	// Otherwise next page is empty.
+	var nextPage *int
+	if len(res) == pageSize {
+		nPage := page + 1
+		nextPage = &nPage
+		res = res[:len(res)-1]
+	}
```
kian99: I get it, but it's confusing to read this.

internal/rebac_admin/resources.go
```diff
@@ -0,0 +1,60 @@
...
+	return &resources.PaginatedResponse[resources.Resource]{
+		Data: rRes,
+		Meta: resources.ResponseMeta{
+			Page:  &page,
+			Size:  len(rRes),
+			Total: nil,
+		},
+		Next: resources.Next{
+			Page: nextPage,
+		},
```
kian99, line 52: Shouldn't this whole object be nil if `nextPage` is nil?

internal/rebac_admin/resources_integration_test.go
```diff
@@ -0,0 +1,152 @@
...
+type resourcesSuite struct {
+	jimmtest.JIMMSuite
+}
+
+var _ = gc.Suite(&resourcesSuite{})
+
+// patchIdentitiesEntitlementTestEnv is used to create entries in JIMM's database.
+// The rebacAdminSuite does not spin up a Juju controller so we cannot use
+// regular JIMM methods to create resources. It is also necessary to have resources
+// present in the database in order for ListRelationshipTuples to work correctly.
```
kian99: docstring is c/p, missing changes.

SimoneDutto:
> Initial run over, will go over again, looking more closely at query and pagination. Not keen on having test with integration in the name... You can definitely unit test just the handler with mocks to test the handlers logic. This probably something to discuss at standup.

In the rebac admin folder we use this approach of having `<name>_test.go` and `<name>_integration_test.go`.
Yeah, you can totally mock out Jimm methods and we did for some endpoints and these tests are usually in the `<name>_test.go`, but for this one I think it's best to have the integration tests directly

kian99 (dismissed):
Very nice, I think this is pretty much good to go, just left some comments where things were a little unclear.

internal/db/resources_test.go
SimoneDutto: there is but in other pkgs, I don't think we necessarly need this in the suite because it's useful for this specific use case.

internal/rebac_admin/resources.go
ale8k: Just make a func maybe call geTnextPageAndResponse or something

internal/db/resources.go
SimoneDutto: yes, you can find in the issue on Jira the entities who need the parent as well https://warthogs.atlassian.net/browse/CSS-9402

internal/common/pagination/pagination.go
SimoneDutto: yeah, you are right on this. It's pretty confusing without names

internal/rebac_admin/resources.go
SimoneDutto: it could, but i think it complicates the code to add an `if nextPage == nil` to exclude the `Next` argument.
What do you think?

SimoneDutto: Also, we used the same approach for all the other endpoints

ale8k (dismissed):
lgtm

internal/db/resources_test.go
ale8k: Then off suite it goes aha

SimoneDutto: fair enough, lol

internal/common/pagination/pagination.go
SimoneDutto: changed to named returns and improved godoc

internal/rebac_admin/resources.go
SimoneDutto: did it, thank you for the suggestion

internal/rebac_admin/resources.go
```diff
@@ -58,3 +52,17 @@ func (s *resourcesService) ListResources(ctx context.Context, params *resources.
 		},
 	}, nil
 }
+
+// We fetch one record more than the page size.
+// Then, we set the next page if we have this many records.
+// If it does we return the records minus 1 and advide the consumer there is another page.
```
kian99:
```suggestion
// If it does we return the records minus 1 and advise the consumer there is another page.
```

internal/rebac_admin/resources_integration_test.go
```diff
@@ -19,7 +19,7 @@ type resourcesSuite struct {
 
 var _ = gc.Suite(&resourcesSuite{})
 
-// patchIdentitiesEntitlementTestEnv is used to create entries in JIMM's database.
+// resourcesTestEnv is used to create entries in JIMM's database.
 // The rebacAdminSuite does not spin up a Juju controller so we cannot use
 // regular JIMM methods to create resources. It is also necessary to have resources
 // present in the database in order for ListRelationshipTuples to work correctly.
```
kian99:
```suggestion
// present in the database in order for ListResources to work correctly.
```

internal/rebac_admin/resources.go
kian99: The wording is confusing here,
Reading this I see
- If we have as many records as requested, we set the next page.
- If we don't have as many records as request, we return records minus 1 and tell client there is another page.
- Else we return the records and set the next page as empty.


Aren't points 1 and 3 saying the same thing?

SimoneDutto: you are right, i made a mess with c/p. Now it's updated.

kian99 (approved):

ale8k (approved):

PR #1349, opened 2024-09-04
Add CORS middleware

kian99:
## Description

This PR does a few things around CORS to fix issues raised by the Juju dashboard.
1. Added CORS middleware to return valid CORS headers - I've opted to use the https://github.com/rs/cors module which was already a dependency in our go.sum because it is pulled in by a Juju dependency.
2. Fixed bug with cookies differing when calling login and whoami endpoints. The options set on a cookie when calling /login were wiped when calling /whoami because the options are not saved to the database. So we set them before we return a cookie to the user.
3. Changed the cookie's "sameSite" policy from **lax** to **none** to allow the Juju dashboard to send the cookie in cross-origin requests.

There are some things worth mentioning here:
1. Setting the cors allowed origin option to the wildcard "*" to accept all origins is not valid when also setting the `Access-Control-Allow-Credentials` header to true. See [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#examples_of_access_control_scenarios).
>If a request includes a credential (most commonly a Cookie header) and the response includes an Access-Control-Allow-Origin: * header (that is, with the wildcard), the browser will block access to the response, and report a CORS error in the devtools console.

2. A "sameSite" policy of **none** is necessary because **strict** and **lax** do not permit sending a cookies cross-origin via AJAX. See [here](https://stackoverflow.com/questions/59990864/what-is-the-difference-between-samesite-lax-and-samesite-strict) for more info.

Fixes [CSS-10514](https://warthogs.atlassian.net/browse/CSS-10514)

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [ ] Covered by unit tests

## Test instructions
I tested the changes with a react app and made a request from the app, running at `localhost:3006` to JIMM's /auth/whoami endpoint. After configuring CORS there was a 403 error because the browser did not send JIMM's session cookie in the request until the cookie's sameSite policy was changed to **none**.

[CSS-10514]: https://warthogs.atlassian.net/browse/CSS-10514?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cmd/jimmsrv/service/service.go
+++ b/cmd/jimmsrv/service/service.go
@@ -20,6 +20,7 @@ import (
 	"github.com/juju/names/v5"
 	"github.com/juju/zaputil/zapctx"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"github.com/rs/cors"
 	"go.uber.org/zap"
 	"gorm.io/driver/postgres"
 	"gorm.io/gorm"
@@ -81,6 +82,10 @@ type OAuthAuthenticatorParams struct {
 
 	// SessionCookieMaxAge holds the max age for session cookies in seconds.
 	SessionCookieMaxAge int
+
+	// SecureSessionCookies determines if HTTPS must be enabled in order for JIMM
+	// to set cookies when creating browser based sessions.
+	SecureSessionCookies bool
 
 	// JWTSessionKey holds the secret key used for signing/verifying JWT tokens.
 	// See internal/auth/oauth2.go AuthenticationService.SessionSecretkey for more details.
@@ -175,14 +180,14 @@ type Params struct {
 	// the /callback in an authorisation code OAuth2.0 flow to finish the flow.
 	DashboardFinalRedirectURL string
 
-	// SecureSessionCookies determines if HTTPS must be enabled in order for JIMM
-	// to set cookies when creating browser based sessions.
-	SecureSessionCookies bool
-
 	// CookieSessionKey is a randomly generated secret passed via config used for signing
 	// cookie data. The recommended length is 32/64 characters from the Gorilla securecookie lib.
 	// https://github.com/gorilla/securecookie/blob/main/securecookie.go#L124
 	CookieSessionKey []byte
+
+	// CorsAllowedOrigins represents all addresses that are valid for cross-origin
+	// requests. A wildcard '*' is accepted to allow all cross-origin requests.
+	CorsAllowedOrigins []string
 }
 
 // A Service is the implementation of a JIMM server.
@@ -347,6 +352,7 @@ func NewService(ctx context.Context, p Params) (*Service, error) {
 			SessionTokenExpiry:  p.OAuthAuthenticatorParams.SessionTokenExpiry,
 			SessionCookieMaxAge: p.OAuthAuthenticatorParams.SessionCookieMaxAge,
 			JWTSessionKey:       p.OAuthAuthenticatorParams.JWTSessionKey,
+			SecureCookies:       p.OAuthAuthenticatorParams.SecureSessionCookies,
 			Store:               &s.jimm.Database,
 			SessionStore:        sessionStore,
 			RedirectURL:         redirectUrl,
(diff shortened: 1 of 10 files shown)


internal/auth/oauth2.go
```diff
@@ -420,13 +426,23 @@ func (as *AuthenticationService) VerifyClientCredentials(ctx context.Context, cl
 	return nil
 }
 
+// sessionCrossOriginSafe sets parameters on the session that allow its use in cross-origin requests.
+// Options are not saved to the database so this must be called whenever a session cookie will be returned to a client.
+//
+// Note browsers require cookies with the same-site policy as 'none' to additionally have the secure flag set.
+func sessionCrossOriginSafe(session *sessions.Session, secure bool) *sessions.Session {
+	session.Options.Secure = secure                  // Ensures only sent with HTTPS
+	session.Options.HttpOnly = false                 // Allow Javascript to read it
```
SimoneDutto: why are we allowing javascript to read it? Do they manipulate the cookies in some way?

cmd/jimmsrv/service/service.go
SimoneDutto, line 1: i think you can add some requests forged to trigger cors policies in service_test.go

internal/auth/oauth2.go
ale8k: No they call whoami to get details

ale8k (commented):
Just so I'm understanding correctly,

now we same site none of on our session cookies allowing them to go anywhere? And the server has cors enabled with GET only?

internal/auth/oauth2.go
kian99: Good point @SimoneDutto I went and tried my test again, setting HttpOnly = true and the browser still sends the cookie. Had a google and I see from [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies)
>A cookie with the HttpOnly attribute can't be modified by JavaScript, for example using [Document.cookie](https://developer.mozilla.org/en-US/docs/Web/API/Document/cookie); it can only be modified when it reaches the server. Cookies that persist user sessions for example should have the HttpOnly attribute set — it would be really insecure to make them available to JavaScript. This precaution helps mitigate cross-site scripting ([XSS](https://developer.mozilla.org/en-US/docs/Web/Security/Types_of_attacks#cross-site_scripting_xss)) attacks.

So I'll change this to true.

SimoneDutto: yes, this is more secure! thank you for this

kian99:
@ale8k 
>now we same site none of on our session cookies allowing them to go anywhere? And the server has cors enabled with GET only?

I'm not sure I understand the first part of your message. But with "samesite" set to **none**, the browser will include the cookie the cookie in all requests to the domain that set the cookie, including JS requests and via top-level navigation (i.e. a user clicking a link). And the server now has CORS only for GET requests, correct.

internal/auth/oauth2.go
```diff
@@ -420,13 +426,23 @@ func (as *AuthenticationService) VerifyClientCredentials(ctx context.Context, cl
...
 }
 
+// sessionCrossOriginSafe sets parameters on the session that allow its use in cross-origin requests.
+// Options are not saved to the database so this must be called whenever a session cookie will be returned to a client.
+//
+// Note browsers require cookies with the same-site policy as 'none' to additionally have the secure flag set.
+func sessionCrossOriginSafe(session *sessions.Session, secure bool) *sessions.Session {
+	session.Options.Secure = secure                  // Ensures only sent with HTTPS
+	session.Options.HttpOnly = true                  // Don't allow Javascript to modify cookie
+	session.Options.SameSite = http.SameSiteNoneMode // Allow cross-origin requests via Javascript
```
SimoneDutto, line 436: i think it's possible to set a domain for cookies, like `.canonical`. I can imagine in a production environment having the dashboard and jimm under the same subdomain.
In this way we are more secure.

SimoneDutto:
> @ale8k
> 
> > now we same site none of on our session cookies allowing them to go anywhere? And the server has cors enabled with GET only?
> 
> I'm not sure I understand the first part of your message. But with "samesite" set to **none**, the browser will include the cookie the cookie in all requests to the domain that set the cookie, including JS requests and via top-level navigation (i.e. a user clicking a link). And the server now has CORS only for GET requests, correct.

I think the attack vector @ale8k is worried about is having a malicious js snippet in the victim browser using the dashboard able to send out (to `malicious.server.com` for example) our cookies

internal/auth/oauth2.go
kian99: The domain automatically gets set to the domain that set the cookie. [Doc source](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#domaindomain-value)
image

SimoneDutto (approved):

ale8k (approved):

PR #1353, opened 2024-09-05
Css 10530/filterning resources

SimoneDutto:
## Description

Follow-up to [1347](https://github.com/canonical/jimm/pull/1347) to enable filtering based of resource name, resource type.
- refactor the big raw query into multiple gorm query to be able to select them based of filtering mechanism.
- Add unit test file, thanks @ale8k because I caught a nasty bug thanks to them.

> fly by entitlements parametrization because it was useful to validate entityType filter.

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [ ] Covered by unit tests

### DIFF
--- a/internal/db/resource.go
+++ b/internal/db/resource.go
@@ -116,31 +116,27 @@ func (d *Database) ListResources(ctx context.Context, limit, offset int, namePre
 // namePrefixFilter: used to match resources name prefix.
 // typeFilter: used to match resources type. If this is not empty the resources are fetched from a single table.
 func buildQuery(db *gorm.DB, offset, limit int, namePrefixFilter, typeFilter string) (*gorm.DB, error) {
-	// if namePrefixEmpty set to true we have a 'TRUE IS TRUE' in the SQL WHERE statement, which disable the filtering.
-	namePrefixEmpty := namePrefixFilter == ""
-	namePrefixFilter += "%"
-
 	applicationOffersQuery := db.Select(selectApplicationOffers).
 		Model(&dbmodel.ApplicationOffer{}).
-		Where("? IS TRUE OR application_offers.name LIKE ?", namePrefixEmpty, namePrefixFilter).
+		Where("(CASE WHEN ? = '' THEN TRUE ELSE application_offers.name LIKE ? END)", namePrefixFilter, namePrefixFilter+"%").
 		Joins("JOIN models ON application_offers.model_id = models.id")
 
 	cloudsQuery := db.Select(selectClouds).
 		Model(&dbmodel.Cloud{}).
-		Where("? IS TRUE OR clouds.name LIKE ?", namePrefixEmpty, namePrefixFilter)
+		Where("(CASE WHEN ? = '' THEN TRUE ELSE clouds.name LIKE ? END)", namePrefixFilter, namePrefixFilter+"%")
 
 	controllersQuery := db.Select(selectControllers).
 		Model(&dbmodel.Controller{}).
-		Where("? IS TRUE OR controllers.name LIKE ?", namePrefixEmpty, namePrefixFilter)
+		Where("(CASE WHEN ? = '' THEN TRUE ELSE controllers.name LIKE ? END)", namePrefixFilter, namePrefixFilter+"%")
 
 	modelsQuery := db.Select(selectModels).
 		Model(&dbmodel.Model{}).
-		Where("? IS TRUE OR models.name LIKE ?", namePrefixEmpty, namePrefixFilter).
+		Where("(CASE WHEN ? = '' THEN TRUE ELSE models.name LIKE ? END)", namePrefixFilter, namePrefixFilter+"%").
 		Joins("JOIN controllers ON models.controller_id = controllers.id")
 
 	serviceAccountsQuery := db.Select(selectIdentities).
 		Model(&dbmodel.Identity{}).
-		Where("name LIKE '%@serviceaccount' AND (? IS TRUE OR identities.name LIKE ?)", namePrefixEmpty, namePrefixFilter)
+		Where("name LIKE '%@serviceaccount' AND (CASE WHEN ? = '' THEN TRUE ELSE identities.name LIKE ? END)", namePrefixFilter, namePrefixFilter+"%")
 
 	// if the typeFilter is set we only return the query for that specif entityType, otherwise the union.
 	if typeFilter == "" {
--- a/internal/db/resource.go
+++ b/internal/db/resource.go
@@ -12,6 +12,7 @@ import (
 	"github.com/canonical/jimm/v3/internal/servermon"
 )
 
+const ApplicationOffersQueryKey = "application_offers"
 const SELECT_APPLICATION_OFFERS = `
 'application_offer' AS type, 
 application_offers.uuid AS id, 
@@ -21,6 +22,7 @@ models.name AS parent_name,
 'model' AS parent_type
 `
 
+const CloudsQueryKey = "clouds"
 const SELECT_CLOUDS = `
 'cloud' AS type, 
 clouds.name AS id, 
@@ -30,6 +32,7 @@ clouds.name AS name,
 '' AS parent_type
 `
 
+const ControllersQueryKey = "controllers"
 const SELECT_CONTROLLERS = `
 'controller' AS type, 
 controllers.uuid AS id, 
@@ -39,6 +42,7 @@ controllers.name AS name,
 '' AS parent_type
 `
 
+const ModelsQueryKey = "models"
 const SELECT_MODELS = `
 'model' AS type, 
 models.uuid AS id, 
@@ -48,6 +52,7 @@ controllers.name AS parent_name,
 'controller' AS parent_type
 `
 
+const ServiceAccountQueryKey = "identities"
 const SELECT_IDENTITIES = `
 'service_account' AS type, 
 identities.name AS id, 
@@ -100,6 +105,9 @@ func (d *Database) ListResources(ctx context.Context, limit, offset int, nameFil
 	return resources, nil
 }
 
+// buildQuery is an utility function to build the database query according to two optional parameters.
+// nameFilter: used to match resources name.
+// typeFilter used to match resources type. If this is not empty the resources are fetched from a single table.
 func buildQuery(db *gorm.DB, offset, limit int, nameFilter, typeFilter string) (*gorm.DB, error) {
 	query := `
 	? UNION ? UNION ? UNION ? UNION ?
@@ -126,11 +134,11 @@ func buildQuery(db *gorm.DB, offset, limit int, nameFilter, typeFilter string) (
 		Where("name LIKE '%@serviceaccount'")
 
 	queries := map[string]*gorm.DB{
-		"application_offers": applicationOffersQuery,
-		"clouds":             cloudsQuery,
-		"controllers":        controllersQuery,
-		"models":             modelsQuery,
-		"identities":         serviceAccountsQuery,
+		ApplicationOffersQueryKey: applicationOffersQuery,
+		CloudsQueryKey:            cloudsQuery,
+		ControllersQueryKey:       controllersQuery,
+		ModelsQueryKey:            modelsQuery,
+		ServiceAccountQueryKey:    serviceAccountsQuery,
 	}
 	// we add the where clause only if the nameFilter is filled
 	if nameFilter != "" {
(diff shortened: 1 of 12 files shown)


internal/db/resource.go
```diff
@@ -5,67 +5,61 @@ import (
...
-			'' AS parent_name,
-			'' AS parent_type
-	FROM identities
-	WHERE name LIKE '%@serviceaccount'
-)
-ORDER BY type, id
-OFFSET ?
-LIMIT  ?;
+const ApplicationOffersQueryKey = "application_offers"
+const SELECT_APPLICATION_OFFERS = `
```
babakks: I'd rather name them like this:

```
const ApplicationOffersQueryKey
const ApplicationOffersQuery
```

What do you think?

internal/db/resource.go
```diff
@@ -106,3 +104,68 @@ func (d *Database) ListResources(ctx context.Context, limit, offset int) (_ []Re
 	}
 	return resources, nil
 }
+
+// buildQuery is an utility function to build the database query according to two optional parameters.
```
babakks: s/an/a

internal/db/resource.go
```diff
@@ -106,3 +104,68 @@ func (d *Database) ListResources(ctx context.Context, limit, offset int) (_ []Re
...
+// nameFilter: used to match resources name.
```
babakks: Can you please rename this to `namePrefixFilter` to make it clear this is not an exact match? Also, please include prefix search in the godoc.

internal/db/resource.go
```diff
@@ -106,3 +104,68 @@ func (d *Database) ListResources(ctx context.Context, limit, offset int) (_ []Re
...
+	queries := map[string]*gorm.DB{
+		ApplicationOffersQueryKey: applicationOffersQuery,
+		CloudsQueryKey:            cloudsQuery,
+		ControllersQueryKey:       controllersQuery,
+		ModelsQueryKey:            modelsQuery,
+		ServiceAccountQueryKey:    serviceAccountsQuery,
+	}
+	// we add the where clause only if the nameFilter is filled
+	if nameFilter != "" {
+		nameFilter = "%" + nameFilter + "%"
```
babakks: It's a prefix search, so this should be:

```
nameFilter = nameFilter + "%"
```

internal/db/resource.go
```diff
@@ -106,3 +104,68 @@ func (d *Database) ListResources(ctx context.Context, limit, offset int) (_ []Re
...
+			Raw(query,
+				applicationOffersQuery,
+				cloudsQuery,
+				controllersQuery,
+				modelsQuery,
+				serviceAccountsQuery,
+				offset,
+				limit,
+			), nil
+	} else {
```
babakks: You can get rid of `else` here because you have an early return up in L152.

internal/db/resource.go
```diff
@@ -106,3 +104,68 @@ func (d *Database) ListResources(ctx context.Context, limit, offset int) (_ []Re
...
+		CloudsQueryKey:            cloudsQuery,
+		ControllersQueryKey:       controllersQuery,
+		ModelsQueryKey:            modelsQuery,
+		ServiceAccountQueryKey:    serviceAccountsQuery,
+	}
+	// we add the where clause only if the nameFilter is filled
+	if nameFilter != "" {
+		nameFilter = "%" + nameFilter + "%"
+		for k := range queries {
+			queries[k] = queries[k].Where(k+".name LIKE ?", nameFilter)
```
babakks: To be honest, I don't like playing with queries like this. I'd rather have the const queries richer with parameters and here just decide on the arguments to pass to them. For example, we can define the const queries like this:

```
SELECT * ...
WHERE ? IS TRUE OR name LIKE ?
```

Where first parameter is a boolean that we set to `false` for prefix matching to take place.

internal/db/resource.go
```diff
@@ -106,3 +104,68 @@ func (d *Database) ListResources(ctx context.Context, limit, offset int) (_ []Re
...
+func buildQuery(db *gorm.DB, offset, limit int, nameFilter, typeFilter string) (*gorm.DB, error) {
+	query := `
+	? UNION ? UNION ? UNION ? UNION ?
+	ORDER BY type, id
+	OFFSET ?
+	LIMIT  ?;
+	`
+	applicationOffersQuery := db.Select(SELECT_APPLICATION_OFFERS).
+		Model(&dbmodel.ApplicationOffer{}).
+		Joins("JOIN models ON application_offers.model_id = models.id")
```
babakks, line 122: I'd have the `JOIN`s already set in the const queries. Basically we don't get type-safe benefits from using GORM here (except for type mapping of the result).

internal/db/resource.go
SimoneDutto: it's a key indeed, not a query. So i don't know about this change

internal/db/resource.go
SimoneDutto: yes, thank you

internal/db/resource.go
SimoneDutto: i was debating with myself this too, maybe you are right!

internal/db/resource.go
SimoneDutto: so you are suggesting to get rid of gorm all together and using raw queries?
i think using gorm + the where clause as you suggested you suggested it's clearer

SimoneDutto: but i do agree that gorm here doesn't provide type safety but just a tidier organization than manipulating SQL strings

internal/db/resource.go
kian99: +1 Babak's suggestion, I don't think the all caps naming for constants is very helpful, I'd reserve it for env vars.

internal/db/resource.go
```diff
@@ -106,3 +104,68 @@ func (d *Database) ListResources(ctx context.Context, limit, offset int) (_ []Re
 	}
 	return resources, nil
 }
+
+// buildQuery is an utility function to build the database query according to two optional parameters.
+// nameFilter: used to match resources name.
+// typeFilter used to match resources type. If this is not empty the resources are fetched from a single table.
```
kian99: For consistency with the previous line.
```suggestion
// typeFilter: used to match resources type. If this is not empty the resources are fetched from a single table.
```

internal/jujuapi/controllerroot.go
```diff
@@ -66,7 +66,7 @@ type JIMM interface {
 	InitiateMigration(ctx context.Context, user *openfga.User, spec jujuparams.MigrationSpec) (jujuparams.InitiateMigrationResult, error)
 	ListApplicationOffers(ctx context.Context, user *openfga.User, filters ...jujuparams.OfferFilter) ([]jujuparams.ApplicationOfferAdminDetailsV5, error)
 	ListIdentities(ctx context.Context, user *openfga.User, filter pagination.LimitOffsetPagination) ([]openfga.User, error)
-	ListResources(ctx context.Context, user *openfga.User, filter pagination.LimitOffsetPagination) ([]db.Resource, error)
+	ListResources(ctx context.Context, user *openfga.User, filter pagination.LimitOffsetPagination, nameFilter, typeFilter string) ([]db.Resource, error)
```
kian99: What do you think about using stronger types for nameFilter and typeFilter that provide validation?
Like
```go
type NameFilter struct{
   name string
}
```
Then we can have
```go
func (n NameFilter) Validate() bool{}
```

kian99 (commented):
looks good with some minor comments

internal/jujuapi/controllerroot.go
SimoneDutto: the thing is that validation is only for entity and it should be done at API layer before entering Jimm world.
I'll be constructing a strong just for one of the two, so i don't see many benefits. 
But I am open to be convinced

internal/db/resource.go
SimoneDutto: i deleted my comment because i though it was referring to the key and not the query itself

SimoneDutto: i deleted my comment, because i though babak was talking about changing the key name

internal/db/resource.go
SimoneDutto: @babakks i tried using the RAW query for each table but the code was more confusing.
I think this iteration provides clarity and ease of manipulation.

internal/db/resource.go
SimoneDutto: using the where in this way simplified the code, thanks for the suggestion

internal/db/resource.go
```diff
@@ -106,3 +111,65 @@ func (d *Database) ListResources(ctx context.Context, limit, offset int) (_ []Re
...
+// namePrefixFilter: used to match resources name prefix.
+// typeFilter: used to match resources type. If this is not empty the resources are fetched from a single table.
+func buildQuery(db *gorm.DB, offset, limit int, namePrefixFilter, typeFilter string) (*gorm.DB, error) {
+	// if namePrefixEmpty set to true we have a 'TRUE IS TRUE' in the SQL WHERE statement, which disable the filtering.
+	namePrefixEmpty := namePrefixFilter == ""
+	namePrefixFilter += "%"
+
+	applicationOffersQuery := db.Select(selectApplicationOffers).
+		Model(&dbmodel.ApplicationOffer{}).
+		Where("? IS TRUE OR application_offers.name LIKE ?", namePrefixEmpty, namePrefixFilter).
```
babakks: Thanks for applying this suggestion, but I felt a bit suspicious about PostgreSQL expression evaluation (and short-circuiting of condition), and found out there is no such guaranty in PostgreSQL (See [here](https://www.postgresql.org/docs/14/sql-expressions.html#SYNTAX-EXPRESS-EVAL)). The gist is we have to use the `CASE` statement. With that we don't need the extra boolean argument:

```sql
WHERE (CASE WHEN ? = '' THEN TRUE ELSE application_offers.name LIKE ? END)
```

I tried this with PostgreSQL and it seems working.

babakks (changes requested):

internal/db/resource.go
SimoneDutto: i was not aware of that.
Updated

babakks (approved):

internal/db/resource.go
```diff
@@ -106,3 +111,61 @@ func (d *Database) ListResources(ctx context.Context, limit, offset int) (_ []Re
...
 	return resources, nil
 }
+
+// buildQuery is a utility function to build the database query according to two optional parameters.
+// namePrefixFilter: used to match resources name prefix.
+// typeFilter: used to match resources type. If this is not empty the resources are fetched from a single table.
+func buildQuery(db *gorm.DB, offset, limit int, namePrefixFilter, typeFilter string) (*gorm.DB, error) {
+	applicationOffersQuery := db.Select(selectApplicationOffers).
+		Model(&dbmodel.ApplicationOffer{}).
+		Where("(CASE WHEN ? = '' THEN TRUE ELSE application_offers.name LIKE ? END)", namePrefixFilter, namePrefixFilter+"%").
```
kian99, line 121:
```suggestion
		Where("(CASE WHEN ? = '' THEN TRUE ELSE application_offers.name LIKE ?% END)", namePrefixFilter, namePrefixFilter).
```
? and elsewhere

internal/rebac_admin/resources.go
```diff
@@ -30,7 +32,15 @@ func (s *resourcesService) ListResources(ctx context.Context, params *resources.
...
 	}
 	currentPage, expectedPageSize, pagination := pagination.CreatePaginationWithoutTotal(params.Size, params.Page)
-	res, err := s.jimm.ListResources(ctx, user, pagination)
+	namePrefixFilter, typeFilter := utils.GetNameAndTypeResourceFilter(params.EntityName, params.EntityType)
+	if typeFilter != "" {
+		typeFilter, err = validateAndConvertResourceFilter(typeFilter)
+		if err != nil {
+			return nil, v1.NewInvalidRequestError(err.Error())
+		}
+	}
```
kian99, line 41: For a follow-up, but I think this should be calling an application layer function to create custom types. Like the idea I was suggesting earlier, the API layer can pass the strings it has received from the client into application layer validators that return app layer types that can be used in ListResources().

kian99: Because a string doesn't tell me much, but a `type resource struct{ resource string }` tells me a lot more.

kian99 (approved):

internal/db/resource.go
SimoneDutto: it doesn't work with the SQL replacements

PR #1356, opened 2024-09-09
Add version method to JIMM facade

kian99:
## Description

This PR adds a `version` method to the `JIMM` facade. This was added because client's like the Terraform Provider require a way to differentiate between a Juju controller and JIMM controller.

When an API connection to a controller is made, the controller returns a list of facades that the controller supports. The Juju client can return the best version that the controller supports and which the client supports. In the case of JIMM, the Juju client is not aware of the JIMM facade and doesn't expose direct access to the facade methods. With this new method, a client can call the `Version` method which should never return an error and can indicate that the controller is a JIMM controller.

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [x] Covered by unit tests


SimoneDutto (approved):
lgtm

ale8k (approved):

PR #1357, opened 2024-09-09
add http proxy for charms and resources

SimoneDutto:
## Description

This pr addresses the fixes needed to enable the deployment of locally built charms.
For this task, JIMM needs to proxy to the controllers two new HTTP endpoints:
- /model/uuid/charms
- /model/uuid/resources (only if we need external resources to build the charm)

The implementation is along the lines of the debug-logs streaming, I defined two new endpoint handlers and those requests are proxied to their relative controllers.

> I moved the http.go to utils.go, and http.go is used to implement the actual HTTP serving. So if you see some code deleted from the utils. go is not deleted but just moved into utils.go

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [ ] Covered by unit tests


internal/jimmhttp/http.go
```diff
@@ -1,56 +1,37 @@
...
-// StripPathElement returns a handler that serves HTTP requests by removing
-// the first element from the request path and invoking the handler h.
-//
-// If a key is specified the removed element will be stored in the context
-// attached to the request such that it can be retrieved using
-// PathElementFromContext.
-//
-// If the request URL contains a RawPath field then it will be cleared
-// before calling h.
-func StripPathElement(key string, h http.Handler) http.Handler {
```
kian99, line 31: I can see that these have just been moved to utils.go but I'd suggest doing mechanical/moving changes in a specific commit or separate PR to make it easier to see the changes.

internal/jujuapi/api.go
```diff
@@ -36,17 +36,24 @@ func APIHandler(ctx context.Context, jimm *jimm.JIMM, p Params) http.Handler {
 // ModelHandler creates an http.Handler for "/model" endpoints.
 func ModelHandler(ctx context.Context, jimm *jimm.JIMM, p Params) http.Handler {
 	mux := http.NewServeMux()
-	mux.Handle("/{uuid}/api", &jimmhttp.WSHandler{
+	mux.Handle("/model/{uuid}/api", http.StripPrefix("/model", &jimmhttp.WSHandler{
```
kian99, line 39: Why are you including the `/model/` here? Previously I removed it from here and just did stripPrefix outside of this function.

internal/jujuapi/http_proxy.go
```diff
@@ -0,0 +1,73 @@
...
+	// extract auth token
+	_, password, ok := req.BasicAuth()
+	if !ok {
+		return errors.E(errors.CodeUnauthorized, "authentication missing")
+	}
+	user, err := s.jimm.LoginWithSessionToken(ctx, password)
+	if err != nil {
+		return errors.E(errors.CodeUnauthorized, err)
+	}
+	// extract model uuid anche check permission
```
kian99:
```suggestion
	// extract model uuid and check permission
```

internal/jujuapi/http_proxy.go
```diff
@@ -0,0 +1,73 @@
...
+	sPath, _ := strings.CutPrefix(req.URL.EscapedPath(), "/model")
```
kian99: We wouldn't need to do this if we strip the /model before we reach the `httpProxier`.

internal/jujuapi/http_proxy.go
```diff
@@ -0,0 +1,73 @@
...
+	uuid, _, err := modelInfoFromPath(sPath)
+	if err != nil {
+		return errors.E(errors.CodeUnauthorized, "cannot parse path")
+	}
+	access, err := s.jimm.GetUserModelAccess(ctx, user, names.NewModelTag(uuid))
+	if err != nil || (access != "admin" && access != "writer") {
+		return errors.E(errors.CodeUnauthorized, "unauthorized")
+	}
```
kian99: I would split these lines up for clarity.

internal/jujuapi/http_proxy.go
```diff
@@ -0,0 +1,73 @@
...
+func (s *httpProxier) ServeHTTP(ctx context.Context, w http.ResponseWriter, req *http.Request) {
+	writeError := func(msg string, code int) {
+		w.WriteHeader(code)
+		_, err := w.Write([]byte(msg))
+		if err != nil {
+			zapctx.Error(ctx, "cannot write to connection", zap.Error(err))
+		}
+	}
+	sPath, _ := strings.CutPrefix(req.URL.EscapedPath(), "/model")
+	uuid, _, err := modelInfoFromPath(sPath)
```
kian99: We don't need to extract the path segment because the full path is in the request?

internal/jujuapi/http_proxy_integration_test.go
```diff
@@ -0,0 +1,228 @@
...
+	c.Assert(err, gc.ErrorMatches, "unauthorized")
+}
+
+func (s *httpProxySuite) TestCharmHTTPServe(c *gc.C) {
+	ctx := context.Background()
+	httpProxier := jujuapi.HTTPProxier(s.JIMM)
+	expectU, expectP, err := s.JIMM.GetCredentialStore().GetControllerCredentials(ctx, s.Model.Controller.Name)
+	c.Assert(err, gc.IsNil)
+	// we expect the controller to respond with TLS
+	fakeController := httptest.NewTLSServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
kian99: If we're using a fake controller then we shouldn't be using the websocketSuite (that spins up a real Juju controller), we can use one of the simpler suites that just spins up JIMM.

internal/jujuapi/http_proxy_integration_test.go
```diff
@@ -0,0 +1,228 @@
...
+
+	tests := []struct {
+		description    string
+		setup          func()
+		url            string
+		statusExpected int
+	}{
+		{
+			description: "good",
+			setup: func() {
```
kian99: Some comments on these tests,
- We're misusing the websocket suite by modifying the Juju controller that is created and replacing its details with the fake controller. This is unclear and we should avoid it.
- We're not really testing the /model/UUID/charms endpoint here so much as we are testing connections to a Juju controller. E.g. Why would we test here what happens when trying to connect to a controller if the controller is unreachable. I can see the value is testing that the http proxy connects to the Juju controller but I'm not sure about these tests. I'm inclined to think you could delete them.

internal/jujuapi/utils.go
```diff
@@ -0,0 +1,26 @@
...
+	"context"
+	"strings"
+
+	"github.com/juju/names/v5"
+
+	"github.com/canonical/jimm/v3/internal/errors"
+	"github.com/canonical/jimm/v3/internal/openfga"
+)
+
+func checkPermission(ctx context.Context, path string, u *openfga.User, mt names.ModelTag) (bool, error) {
```
kian99: I would also rename this function since it's in the utils.go file.

internal/jujuapi/websocket_test.go
```diff
@@ -52,7 +52,7 @@ func (s *websocketSuite) SetUpTest(c *gc.C) {
 
 	mux := http.NewServeMux()
 	mux.Handle("/api", jujuapi.APIHandler(ctx, s.JIMM, s.Params))
-	mux.Handle("/model/", http.StripPrefix("/model", jujuapi.ModelHandler(ctx, s.JIMM, s.Params)))
+	mux.Handle("/model/", jujuapi.ModelHandler(ctx, s.JIMM, s.Params))
```
kian99, line 55: I think it would be better to revert this

internal/rpc/proxy_http.go
```diff
@@ -0,0 +1,100 @@
...
+		err := proxyHTTP(ctx, w, req, HTTPClientOptions{
+			TLSConfig: tlsConfig,
+			URL:       httpUrl(*req.URL, ctl.PublicAddress),
+		})
+		if err == nil {
+			return
+		}
+	}
+	for _, hps := range ctl.Addresses {
+		for _, hp := range hps {
```
kian99: I'm wondering if we shouldn't leverage some of the logic that we already have in the dialer for this.

internal/jujuapi/http_proxy.go
kian99, line 1: I think the file name should be just `httpproxy.go` and the same for others. Though we're not super consistent in this area.

internal/rpc/proxy_http.go
kian99, line 1: filename as proxyhttp.go ?

internal/rpc/proxy_http.go
```diff
@@ -0,0 +1,100 @@
...
+	w.WriteHeader(resp.StatusCode)
+	// copy body
+	_, err = io.Copy(w, resp.Body)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func httpUrl(reqUrl url.URL, host string) url.URL {
```
kian99: Consider a different name, maybe `ensureHttpsScheme()` or `ensureHttpsURL()`?

internal/rpc/proxy_http.go
```diff
@@ -0,0 +1,100 @@
...
+	reqUrl.Scheme = "https"
+	reqUrl.Host = host
+	return reqUrl
```
kian99: Also if we're replacing the hot and scheme, what part of `reqUrl` are we keeping? The path? Maybe the function could make that clearer too.

internal/rpc/proxy_http.go
kian99: Also this file isn't tested, though I understand it's tested as part of the e2e tests.

internal/jujuapi/http_proxy_integration_test.go
kian99: These tests feel like they're better suited for proxy_http.go

internal/jujuapi/http_proxy_test.go
kian99, line 1: The authentication on the `jujuapi` could be tweaked, let me know what you think about the below.

What about using a middleware for the http proxy endpoints that handles auth, we can use the one we created for the rebac-admin-ui. Then the httpproxy Authenticate just checks that a user is present in the context. Then our HTTP handler are starting to look the same across the rebac-admin and http proxy.

internal/jimmhttp/http.go
SimoneDutto: for some reason I though the file rename would have been handled by git.
Guess I broke the algorithm by renaming another file with the previous name

internal/jujuapi/api.go
SimoneDutto: the reason is that we want to proxy the req with the model in its path, and it doesn't make sense to strip it outside and then add it inside the proxy HTTP function.

internal/jujuapi/http_proxy.go
SimoneDutto: See https://github.com/canonical/jimm/pull/1357#discussion_r1751582926

internal/jujuapi/http_proxy.go
SimoneDutto: see https://github.com/canonical/jimm/pull/1357#discussion_r1751582926

internal/jujuapi/websocket_test.go
SimoneDutto: see https://github.com/canonical/jimm/pull/1357#discussion_r1751582926

internal/rpc/proxy_http.go
SimoneDutto: I thought about that, but to me it seems like they behave differently enough that try to use the same function make it more confusing

internal/rpc/proxy_http.go
SimoneDutto: yeah, you're right. A better name should do the trick

internal/jujuapi/http_proxy_test.go
SimoneDutto: two things:
- to use the middleware I think it's best to extract them from modelHandler, loosing a little bit of coupling with other model methods (though is not a big deal)
- this middleware is model specific I feel like, because we need to extract the user and then check if it has permission for models he is trying to access.

I don't really have an answer, just putting my thoughts out here

internal/jujuapi/api.go
```diff
@@ -36,17 +36,24 @@ func APIHandler(ctx context.Context, jimm *jimm.JIMM, p Params) http.Handler {
...
 		Upgrader: websocketUpgrader,
 		Server: &apiProxier{apiServer: apiServer{
 			jimm: jimm,
 		}},
+	}))
+	mux.Handle("/model/{uuid}/charms", &jimmhttp.HTTPHandler{
```
kian99, line 46: Add a comment explaining why we strip on some endpoints and not others.

internal/jujuapi/http_proxy_test.go
SimoneDutto: as discussed, to avoid expanding the scope of this pr I created a [card](https://warthogs.atlassian.net/browse/CSS-10627) to rework the authorization and authentication part for HTTP endpoints. 

Both these two and debug-logs

internal/jimmhttp/http.go
```diff
@@ -1,56 +1,37 @@
...
+		w.WriteHeader(http.StatusUnauthorized)
+		_, err := w.Write([]byte(authErr.Error()))
+		zapctx.Error(ctx, "failed to write authentication error", zap.Error(err))
+		return
 	}
-	return s, ""
+	h.HTTPProxier.ServeHTTP(ctx, w, req)
+}
+
+type HTTPProxier interface {
```
alesstimec, line 39: godoc, please

internal/jujuapi/http_proxy_integration_test.go
SimoneDutto: i split the test between the networking ones (put into httpproxt.go) and logic one (left here).
thanks for the suggestion!

internal/rpc/proxy_http.go
SimoneDutto: now it's tested

internal/jujuapi/httpproxy.go
```diff
@@ -28,14 +28,17 @@ func (s *httpProxier) Authenticate(ctx context.Context, w http.ResponseWriter, r
...
 		return errors.E(errors.CodeUnauthorized, err)
 	}
-	// extract model uuid anche check permission
+	// extract model uuid and check permission
 	sPath, _ := strings.CutPrefix(req.URL.EscapedPath(), "/model")
 	uuid, _, err := modelInfoFromPath(sPath)
 	if err != nil {
 		return errors.E(errors.CodeUnauthorized, "cannot parse path")
 	}
 	access, err := s.jimm.GetUserModelAccess(ctx, user, names.NewModelTag(uuid))
```
kian99: Actually we can simplify this further, just check that the user has write permission on the model.

internal/jujuapi/httpproxy_integration_test.go
```diff
@@ -16,45 +17,93 @@ import (
...
+  cloud: test-cloud
+  region: test-cloud-region
+  cloud-credential: cred-1
+  owner: alice@canonical.com
+users:
+- username: alice@canonical.com
+  access: admin
+`
+
 var _ = gc.Suite(&httpProxySuite{})
```
kian99: nit: Move this above `testEnv`.

internal/jujuapi/utils.go
kian99: Thoughts on this one @SimoneDutto ?

internal/rpc/httpproxy.go
```diff
@@ -0,0 +1,101 @@
...
+
+	"github.com/canonical/jimm/v3/internal/dbmodel"
+)
+
+type HTTPClientOptions struct {
+	TLSConfig *tls.Config
+	URL       url.URL
+}
+
+func ProxyHTTP(ctx context.Context, ctl *dbmodel.Controller, w http.ResponseWriter, req *http.Request) {
```
kian99, line 26: Missing docstring

internal/rpc/httpproxy.go
```diff
@@ -0,0 +1,101 @@
...
+				return
+			}
+		}
+	}
+
+	zapctx.Error(ctx, "couldn't find a valid address for controller")
+	http.Error(w, "Gateway timeout", http.StatusGatewayTimeout)
+}
+
+func proxyHTTP(ctx context.Context, w http.ResponseWriter, req *http.Request, opt HTTPClientOptions) error {
```
kian99: What about naming this just `doRequest` or something?

kian99 (commented):
Looking good

internal/jujuapi/httpproxy.go
SimoneDutto: i've used the utility function directly

internal/jujuapi/httpproxy.go
```diff
@@ -0,0 +1,74 @@
+// Copyright 2024 Canonical.
+package jujuapi
```
alesstimec, line 2: new line, please

internal/jujuapi/httpproxy.go
```diff
@@ -0,0 +1,74 @@
...
+
+	"github.com/canonical/jimm/v3/internal/errors"
+	"github.com/canonical/jimm/v3/internal/rpc"
+)
+
+type httpProxier struct {
+	jimm JIMM // interface
+}
+
+func (s *httpProxier) Authenticate(ctx context.Context, w http.ResponseWriter, req *http.Request) error {
```
alesstimec: godoc, please

internal/jujuapi/httpproxy.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+	if ok, err := checkModelAccessForUser(ctx, finalPath, user, names.NewModelTag(uuid)); err != nil {
+		return errors.E(errors.CodeUnauthorized, "unauthorized")
+	} else if !ok {
+		return errors.E(errors.CodeUnauthorized, "unauthorized")
+	}
+	return nil
+}
+
+func (s *httpProxier) ServeHTTP(ctx context.Context, w http.ResponseWriter, req *http.Request) {
```
alesstimec, line 48: godoc, please

internal/jujuapi/httpproxy.go
```diff
@@ -0,0 +1,74 @@
...
+		writeError("cannot parse path", http.StatusUnprocessableEntity)
+		return
+	}
+	// retrieving credentials from controller
+	model, err := s.jimm.GetModel(ctx, uuid)
+	if err != nil {
+		writeError("cannot get model", http.StatusNotFound)
+		return
+	}
+	u, p, err := s.jimm.GetCredentialStore().GetControllerCredentials(ctx, model.Controller.Name)
```
alesstimec, line 68: we never check if the user is authorized to access the model?

internal/jujuapi/httpproxy.go
```diff
@@ -0,0 +1,74 @@
...
+	sPath, _ := strings.CutPrefix(req.URL.EscapedPath(), "/model")
+	uuid, finalPath, err := modelInfoFromPath(sPath)
+	if err != nil {
+		return errors.E(errors.CodeUnauthorized, "cannot parse path")
+	}
+	if ok, err := checkModelAccessForUser(ctx, finalPath, user, names.NewModelTag(uuid)); err != nil {
+		return errors.E(errors.CodeUnauthorized, "unauthorized")
+	} else if !ok {
+		return errors.E(errors.CodeUnauthorized, "unauthorized")
+	}
```
kian99, line 43: I would do this in `ServeHTTP` in a function possibly called `AuthorizeUser`. Since it authorisation and not authentication.

kian99 (dismissed):
lgtm with 1 change

internal/jimmhttp/http_test.go
```diff
@@ -12,58 +15,49 @@ import (
...
-			resp := rr.Result()
-			defer resp.Body.Close()
-			c.Check(resp.StatusCode, qt.Equals, http.StatusOK)
-		})
+type testHTTPServer struct {
+	t testing.TB
+}
+
+func (s testHTTPServer) Authenticate(ctx context.Context, w http.ResponseWriter, req *http.Request) error {
+	if auth := req.Header.Get("Authorization"); auth == "wrong" {
```
ale8k, line 53: Why no quicktest test?

internal/jimmhttp/http.go
```diff
@@ -1,56 +1,37 @@
...
 	}
-	if n := strings.IndexByte(s, '/'); n >= 0 {
-		return s[:n], s[n:]
+
+	authErr := h.HTTPProxier.Authenticate(ctx, w, req)
+	if authErr != nil {
+		zapctx.Error(ctx, "authentication error", zap.Error(authErr))
+		w.WriteHeader(http.StatusUnauthorized)
+		_, err := w.Write([]byte(authErr.Error()))
+		zapctx.Error(ctx, "failed to write authentication error", zap.Error(err))
```
ale8k, line 29: This will always print? We don't check the error?

internal/jujuapi/api.go
ale8k: These should probably be their own handlers, what if I make a HTTPHandler in the future that doesn't need the proxier? Please look at ./internal/debugapi/api.go for the pattern I was aiming for for us and refactor into this, perhaps sub-packages within jujuapi? And a nice package doc explaining it's a HTTP API.

internal/jimmhttp/http.go
```diff
@@ -1,56 +1,37 @@
...
-		req2.URL.Path = p
-		// clear the RawPath if it was previously set.
-		req2.URL.RawPath = ""
-		if key != "" {
-			req2 = req2.WithContext(context.WithValue(req2.Context(), contextPathKey(key), v))
-		}
-		h.ServeHTTP(w, req2)
-	})
+type HTTPHandler struct {
+	HTTPProxier HTTPProxier
```
ale8k, line 14: What if we don't use a proxier?... Then it's on this struct and it's nil, dangerous code.

internal/jimmhttp/http.go
```diff
@@ -1,56 +1,37 @@
...
+func (h *HTTPHandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
+	ctx := req.Context()
+	if h.HTTPProxier == nil {
+		w.WriteHeader(http.StatusInternalServerError)
+		return
 	}
-	if n := strings.IndexByte(s, '/'); n >= 0 {
-		return s[:n], s[n:]
+
+	authErr := h.HTTPProxier.Authenticate(ctx, w, req)
```
ale8k: Why aren't we doing authenticate as a middleware? We're using chi?

internal/jimmhttp/http.go
```diff
@@ -1,56 +1,37 @@
...
+		_, err := w.Write([]byte(authErr.Error()))
+		zapctx.Error(ctx, "failed to write authentication error", zap.Error(err))
+		return
 	}
-	return s, ""
+	h.HTTPProxier.ServeHTTP(ctx, w, req)
+}
+
+type HTTPProxier interface {
+	Authenticate(ctx context.Context, w http.ResponseWriter, req *http.Request) error
```
ale8k: We don't need an interface if it's just a middleware, you can test the middleware independently and register it on endpoints that need it. This also makes the code clearer when grouping routes together as to how they're authenticated / authorised.

internal/jujuapi/httpproxy.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/canonical/jimm/v3/internal/errors"
+	"github.com/canonical/jimm/v3/internal/rpc"
+)
+
+type httpProxier struct {
+	jimm JIMM // interface
+}
+
+func (s *httpProxier) Authenticate(ctx context.Context, w http.ResponseWriter, req *http.Request) error {
+	// extract auth token
```
ale8k, line 24: This could be 2 middlewares, one is authenticate, one is authorise. This does too much and is kinda misleading, you're checking permissions and specifically model permissions...

internal/jujuapi/utils.go
```diff
@@ -0,0 +1,26 @@
...
+	"context"
+	"strings"
+
+	"github.com/juju/names/v5"
+
+	"github.com/canonical/jimm/v3/internal/errors"
+	"github.com/canonical/jimm/v3/internal/openfga"
+)
+
+func checkModelAccessForUser(ctx context.Context, path string, u *openfga.User, mt names.ModelTag) (bool, error) {
```
ale8k, line 14: The fact this takes a path is confusing and not documented, we're splitting on "/" and taking just [0], it doesn't really explain what you're doing. I also think because it's looking for log/charms/applications it's a mysterious method, log/charms/applications could be anything, it just so happens they're path segments for authorisation. We need to do this better, I'd recommend the middleware approach and register it per each endpoint for the kind of access needed. For example, mux.Use(ModelReaderAuthoriser()) on a handler just for the log endpoint.

internal/jujuapi/httpproxy.go
SimoneDutto: we do in the AuthenticateAndAuthorize method

internal/rpc/httpproxy.go
```diff
@@ -0,0 +1,103 @@
+// Copyright 2024 Canonical.
+
```
ale8k, line 2: Why is this in rpc? We aren't using anything RPC related no?

ale8k (changes requested):

ale8k (changes requested):
Let's try follow the handler conventions I set in place a while back and use middleware instead for handling authentication / authorisation.

SimoneDutto:
Closing this in favor of splitting it into multiple PRs.

ale8k:
@CodiumAI-Agent /review

ale8k:
@CodiumAI-Agent /review --pr_reviewer.require_score_review=true --pr_reviewer.require_can_be_split_review=true

PR #1360, opened 2024-09-10
Update copyright header in linter with year regexp

pkulik0:
## Description

This PR updates the header template in golangci.yaml to allow any year in file headers not only the current one which was hard-coded. It is not legally required to update the year and only causes unnecessary changes to the files.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -69,8 +69,11 @@ linters-settings:
   gocognit:
     min-complexity: 30
   goheader:
+    values:
+      regexp:
+        year: (\d{4})
     template: |-
-      Copyright 2024 Canonical.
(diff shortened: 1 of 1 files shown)


kian99 (approved):
Very nice!

.golangci.yaml
```diff
@@ -69,8 +69,11 @@ linters-settings:
   gocognit:
     min-complexity: 30
   goheader:
+    values:
+      regexp:
+        year: (\d{4})
```
kian99, line 74: Will break in the year 10000

alesstimec: somebody else's problem :)

alesstimec (approved):

PR #1361, opened 2024-09-11
add basic auth middleware

SimoneDutto:
## Description

Add auth middleware to authenticate user via basic auth header

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [ ] Covered by unit tests


internal/middleware/auth.go
```diff
@@ -13,6 +14,9 @@ import (
 	"github.com/canonical/jimm/v3/internal/jujuapi"
 )
 
+// BasicAuthUser is the unique key to extract user from context for basic-auth authentication
+type BasicAuthUser struct{}
```
kian99: I think it should be the same key for any middleware that extracts/adds a user from/to the context.

kian99 (dismissed):

internal/middleware/auth.go
```diff
@@ -13,6 +14,9 @@ import (
 	"github.com/canonical/jimm/v3/internal/jujuapi"
 )
 
+// UserContext is the unique key to extract user from context for basic-auth authentication
+type UserContext struct{}
```
kian99: Sorry what I should have said was, isn't there an existing key used for putting the user into the context? Used by the other middlewares.

SimoneDutto: there is in oauth2.go, but it implies it is a string identity in the context and not a openfga.User

Other than that we are not really using oauth identities here, another key should be fine?

pkulik0: You can take a look at `export_test.go` files in other packages which we use to keep things unexported but still use them in _test equivalents of the package.

SimoneDutto: even better, i set some utilities function to get and set user in the context

kian99: We should make that consistent sooner rather than later. I think an openfga.User should be in the context rather than a string.

SimoneDutto: i think is better to listen to @ale8k before moving forward

kian99 (dismissed):

internal/middleware/auth.go
```diff
@@ -3,16 +3,22 @@
...
 	"go.uber.org/zap"
 
 	"github.com/canonical/jimm/v3/internal/auth"
+	"github.com/canonical/jimm/v3/internal/errors"
 	"github.com/canonical/jimm/v3/internal/jujuapi"
+	"github.com/canonical/jimm/v3/internal/openfga"
 )
 
+// UserContext is the unique key to extract user from context for basic-auth authentication
+type userContext struct{}
```
alesstimec: suggestion: identityContextKey

internal/middleware/auth.go
SimoneDutto: the rebac admin middleware puts in a openfga user for example using an utility function

internal/middleware/auth.go
```diff
@@ -57,3 +63,39 @@ func AuthenticateRebac(next http.Handler, jimm jujuapi.JIMM) http.Handler {
...
+	identity := ctx.Value(userContext{})
+	user, ok := identity.(*openfga.User)
+	if !ok {
+		return nil, errors.E("cannot extract user from context")
+	}
+	return user, nil
+}
+
+// WithUserContext sets the user into the context and return the context
+func WithUserContext(ctx context.Context, user *openfga.User) context.Context {
```
alesstimec: suggestion: withIdentity

- user might not be a user, but also a service account
- we can omit the "context" bit in the name as we know the return type
- does not need to be exported as this is the only package that will use the unexported identityContextKey to put the identity in the context

internal/middleware/auth.go
```diff
@@ -57,3 +63,39 @@ func AuthenticateRebac(next http.Handler, jimm jujuapi.JIMM) http.Handler {
...
+			w.WriteHeader(http.StatusUnauthorized)
+			_, _ = w.Write([]byte("error authenticating the user"))
+			return
+		}
+		next.ServeHTTP(w, r.WithContext(WithUserContext(ctx, user)))
+	})
+}
+
+// GetUserFromContext extracts the user from the context.
+func GetUserFromContext(ctx context.Context) (*openfga.User, error) {
```
alesstimec: suggestion: IdentityFromContext

ale8k:
@CodiumAI-Agent /review

ale8k:
@CodiumAI-Agent /review --pr_reviewer.require_score_review=true --pr_reviewer.require_can_be_split_review=true

internal/middleware/auth.go
```diff
@@ -57,3 +63,39 @@ func AuthenticateRebac(next http.Handler, jimm jujuapi.JIMM) http.Handler {
 		next.ServeHTTP(w, r.WithContext(ctx))
 	}), jimm)
 }
+
+// AuthenticateViaBasicAuth performs basic auth authentication and puts an identity in the request's context
+func AuthenticateViaBasicAuth(next http.Handler, jimm jujuapi.JIMM) http.Handler {
```
alesstimec: suggestion: AuthenticateWithSessionTokenViaBasicAuth

naming is hard, but i'd like to convey the message that we're using the session token instead of username/password..

alesstimec (commented):
A few comments. Looks like an XS to me.

SimoneDutto:
> A few comments. Looks like an XS to me.

@alesstimec names refactored

alesstimec (dismissed):

internal/middleware/auth.go
```diff
@@ -57,3 +63,39 @@ func AuthenticateRebac(next http.Handler, jimm jujuapi.JIMM) http.Handler {
 		next.ServeHTTP(w, r.WithContext(ctx))
 	}), jimm)
 }
+
+// AuthenticateWithSessionTokenViaBasicAuth performs basic auth authentication and puts an identity in the request's context
```
kian99: With the change in the name it made me realise, the doc string should also reflect that we are (mis)using basic auth slightly and the session token is in the password with an empty user.

kian99 (dismissed):

alesstimec (approved):

kian99 (approved):

PR #1364, opened 2024-09-13
Css 10675/authz middleware

SimoneDutto:
## Description

This pr implements a authorization middleware for user/model permissions.

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [ ] Covered by unit tests


internal/middleware/authz.go
```diff
@@ -0,0 +1,55 @@
...
+		}
+		modelTag := names.NewModelTag(modelUUID)
+		switch accessNeeded {
+		case ofganames.ReaderRelation:
+			ok, err := user.IsModelReader(ctx, modelTag)
+			if !ok || err != nil {
+				w.WriteHeader(http.StatusForbidden)
+				_, _ = w.Write([]byte("no access to the resource"))
+				return
+			}
```
kian99, line 40: You could wrap this up in a closure to reduce the duplication but like this good too.

internal/middleware/authz_test.go
```diff
@@ -0,0 +1,116 @@
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			c := qt.New(t)
+			req := httptest.NewRequest(http.MethodGet, "/", nil)
+			w := httptest.NewRecorder()
+			handler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				w.WriteHeader(http.StatusOK)
+			})
+			var h http.Handler
+			if tt.permissionRequired == "reader" {
```
kian99: Why not just make `tt.permissionRequired` of type ofganames.Relation?

kian99 (dismissed):

internal/middleware/authz_test.go
SimoneDutto: lol, you are right

kian99 (dismissed):
Size: S

internal/middleware/authz.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/canonical/jimm/v3/internal/jujuapi"
+	ofganames "github.com/canonical/jimm/v3/internal/openfga/names"
+)
+
+// AuthorizeUserForModelAccess extract the user from the context, and checks for permission on the model uuid extracted from the path.
+func AuthorizeUserForModelAccess(next http.Handler, jimm jujuapi.JIMM, accessNeeded cofga.Relation) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		ctx := r.Context()
+		user, err := IdentityFromContext(ctx)
+		if err != nil {
```
ale8k, line 21: Some space between blocks would be nice, it's frustrating to read when all bulked together... i.e.,:

```go
		modelUUID := chi.URLParam(r, "uuid")
		if modelUUID == "" {
			w.WriteHeader(http.StatusUnauthorized)
			_, _ = w.Write([]byte("cannot find uuid in URL path"))
			return
		}

		modelTag := names.NewModelTag(modelUUID)

		switch accessNeeded {
		case ofganames.ReaderRelation:
```

ale8k (approved):
lgtm

ale8k:
size s

internal/middleware/authz.go
```diff
@@ -0,0 +1,55 @@
...
+			if !ok || err != nil {
+				w.WriteHeader(http.StatusForbidden)
+				_, _ = w.Write([]byte("no access to the resource"))
+				return
+			}
+		case ofganames.WriterRelation:
+			ok, err := user.IsModelWriter(ctx, modelTag)
+			if !ok || err != nil {
+				w.WriteHeader(http.StatusForbidden)
+				_, _ = w.Write([]byte("no access to the resource"))
```
kian99, line 45: nit: technically they might have access to the resource but without the right level.
Maybe "insufficient access to the resource".
And if you really want to take it far, if they have no access at all, for more security we should return a 404 and if they don't have enough access we should return a 403. But I'm happy with the way it is currently.

kian99 (approved):

PR #1366, opened 2024-09-16
use consistent permissions when adding/import models

kian99:
## Description

This PR fixes an inconsistency when creating vs importing models.

When a model is created we grant the user access to the model and set the backing controller as an admin of the model.
The latter step was missing when importing the model. 

This PR encapsulates the logic of adding permissions for a new model into a function that can be called from both places.

Partially fixes [CSS-10592](https://warthogs.atlassian.net/browse/CSS-10592)

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [x] Covered by unit tests

[CSS-10592]: https://warthogs.atlassian.net/browse/CSS-10592?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/jimm/model.go
+++ b/internal/jimm/model.go
@@ -661,29 +661,35 @@ func (j *JIMM) AddModel(ctx context.Context, user *openfga.User, args *ModelCrea
 
 	mi := builder.JujuModelInfo()
 
-	if err := j.OpenFGAClient.AddControllerModel(
-		ctx,
-		builder.controller.ResourceTag(),
-		builder.model.ResourceTag(),
-	); err != nil {
+	ownerUser := openfga.NewUser(owner, j.OpenFGAClient)
+	modelTag := names.NewModelTag(mi.UUID)
+	controllerTag := builder.controller.ResourceTag()
+
+	return mi, j.addModelPermissions(ctx, ownerUser, modelTag, controllerTag)
+}
+
+// addModelPermissions grants a user access to a model and sets the relation between the controller and model.
+// Call this when adding/importing a model to set the necessary permissions.
+func (j *JIMM) addModelPermissions(ctx context.Context, owner *openfga.User, mt names.ModelTag, ct names.ControllerTag) error {
+	if err := j.OpenFGAClient.AddControllerModel(ctx, ct, mt); err != nil {
 		zapctx.Error(
 			ctx,
 			"failed to add controller-model relation",
-			zap.String("controller", builder.controller.UUID),
-			zap.String("model", builder.model.UUID.String),
+			zap.String("controller", ct.Id()),
+			zap.String("model", mt.Id()),
 		)
+		return err
 	}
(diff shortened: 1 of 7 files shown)


SimoneDutto (dismissed):

SimoneDutto:
size: S

internal/jimm/model.go
```diff
@@ -661,29 +661,35 @@ func (j *JIMM) AddModel(ctx context.Context, user *openfga.User, args *ModelCrea
...
-	if err := j.OpenFGAClient.AddControllerModel(
-		ctx,
-		builder.controller.ResourceTag(),
-		builder.model.ResourceTag(),
-	); err != nil {
+	ownerUser := openfga.NewUser(owner, j.OpenFGAClient)
+	modelTag := names.NewModelTag(mi.UUID)
+	controllerTag := builder.controller.ResourceTag()
+
+	return mi, j.addModelPermissions(ctx, ownerUser, modelTag, controllerTag)
```
babakks: I think the returned error should be wrapped with the `op` const, like the other returns in this method:

```go
if err := j.addModelPermissions(ctx, ownerUser, modelTag, controllerTag); err != nil {
	return nil, errors.E(op, err)
}
```

internal/jimm/model.go
```diff
@@ -661,29 +661,35 @@ func (j *JIMM) AddModel(ctx context.Context, user *openfga.User, args *ModelCrea
...
+			zap.String("model", mt.Id()),
 		)
+		return err
 	}
-	err = openfga.NewUser(owner, j.OpenFGAClient).SetModelAccess(ctx, names.NewModelTag(mi.UUID), ofganames.AdministratorRelation)
-	if err != nil {
+	if err := owner.SetModelAccess(ctx, mt, ofganames.AdministratorRelation); err != nil {
 		zapctx.Error(
 			ctx,
 			"failed to add administrator relation",
```
babakks: Now that we're here, maybe we change this log message into something a bit clearer, like `"failed to add user-model relation (admin)"`?

internal/openfga/openfga.go
```diff
@@ -238,6 +238,21 @@ func (o *OFGAClient) AddControllerModel(ctx context.Context, controller names.Co
 	return nil
 }
 
+// AddControllerModel removes a relation between a controller and a model.
+func (o *OFGAClient) RemoveControllerModel(ctx context.Context, controller names.ControllerTag, model names.ModelTag) error {
```
babakks: Please fix the method name in godoc.

babakks (dismissed):
LGTM (S)

babakks (approved):

ale8k (approved):

ale8k:
size s

PR #1367, opened 2024-09-16
return model summaries with new model owner

kian99:
## Description

This PR intends to fix an inconsistency in JIMM between the `ModelInfo` and `ModelSummaries` methods. The latter takes each JIMM dbmodel.Model object that a user has access to and converts into into a model summary.
The former makes an API call to the Juju controller and returns the modelInfo object to the user after replacing some JIMM specific fields like `userAccess`.

One issue is that the modelOwner field was not being replaced in the `modelInfo` call. But this value is unique when using JIMM. In a similar vein, the modelOwner may be different in JIMM versus the Juju controller, particularly when importing a model.

Full list of changes:
- refactor ModelInfo slightly to use a separate function where we merge the juju model info with JIMM specific data
- refactor the test for ModelInfo to reduce duplication
- tweak the test dialer to have less surprising behavior

Partially fixes [CSS-10592](https://warthogs.atlassian.net/browse/CSS-10592)

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [x] Covered by unit tests

[CSS-10592]: https://warthogs.atlassian.net/browse/CSS-10592?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/jimm/model.go
+++ b/internal/jimm/model.go
@@ -701,13 +701,7 @@ func (j *JIMM) ModelInfo(ctx context.Context, user *openfga.User, mt names.Model
 		return nil, errors.E(op, err)
 	}
 
-	modelAccess, err := j.GetUserModelAccess(ctx, user, mt)
-	if err != nil {
-		return nil, errors.E(op, err)
-	}
-	if modelAccess == "" {
-		// If the user doesn't have any access on the model return an
-		// unauthorized error
+	if ok, err := user.IsModelReader(ctx, mt); !ok || err != nil {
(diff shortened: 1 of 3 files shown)


babakks (approved):
LGTM. It's an S.

pkulik0 (approved):
lgtm, size s

PR #1370, opened 2024-09-17
Css 10676/add http proxy

SimoneDutto:
## Description

This pr implements the http proxy methods.

> fly-by moved utils functions to a more appropiate file

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [ ] Covered by unit tests


internal/rpc/httpproxy_test.go
```diff
@@ -0,0 +1,97 @@
...
+	"github.com/canonical/jimm/v3/internal/rpc"
+)
+
+func TestProxyHTTP(t *testing.T) {
+	c := qt.New(t)
+	ctx := context.Background()
+	// we expect the controller to respond with TLS
+	fakeController := httptest.NewTLSServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		_, err := w.Write([]byte("OK"))
+		c.Assert(err, qt.IsNil)
```
kian99: I'd like to see one test where the controller doesn't return an HTTP 200 and the client gets that back.

kian99 (dismissed):
LGTM with a request for extra test

internal/rpc/httpproxy.go
```diff
@@ -0,0 +1,103 @@
...
+	"net/http"
+	"net/url"
+
+	"github.com/juju/zaputil"
+	"github.com/juju/zaputil/zapctx"
+
+	"github.com/canonical/jimm/v3/internal/dbmodel"
+)
+
+type HTTPClientOptions struct {
```
alesstimec: godoc, please

internal/rpc/httpproxy_test.go
SimoneDutto: added

kian99 (dismissed):

internal/rpc/httpproxy.go
```diff
@@ -0,0 +1,103 @@
...
+	TLSConfig *tls.Config
```
alesstimec, line 21: we could pass in an HTTP client instead?

internal/rpc/httpproxy.go
alesstimec: also these seem more like `HTTPOptions` not just `client` options: e.g. the URL is a property of the request. and i don't think this type even needs to be exported

internal/rpc/httpproxy.go
```diff
@@ -0,0 +1,103 @@
...
+	http.Error(w, "Gateway timeout", http.StatusGatewayTimeout)
+}
+
+func doRequest(ctx context.Context, w http.ResponseWriter, req *http.Request, opt HTTPClientOptions) error {
+	client := &http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: opt.TLSConfig,
+		},
+	}
+	req.RequestURI = ""
```
alesstimec, line 75: suggestion: also do `req = req.Clone()`

internal/rpc/httpproxy_test.go
```diff
@@ -0,0 +1,112 @@
+// Copyright 2024 Canonical.
+package rpc_test
```
alesstimec, line 3: new line between copyright and package, please

alesstimec (dismissed):
LGTM modulo a few comments

internal/rpc/httpproxy.go
SimoneDutto: tls config must be used when creatinig the struct. I don't like the idea of creating the httpclient outside of the doRequest function, however i don't really have a strong opinion on that.
Why do you think it's better?

kian99 (approved):

ale8k (approved):

internal/rpc/httpproxy.go
ale8k: Generally speaking you'd have clients specific for different purposes with different roundtrippers

PR #1374, opened 2024-09-20
remove duplicated appoffer tag

kian99:
## Description

This PR removes the applicationOffer tag from Jimm. The same tag with the same parsing logic already existed in Juju so we can use that instead.


alesstimec (approved):
LGTM (XXS)

SimoneDutto (approved):

PR #1377, opened 2024-09-24
Ignore auth for `/swagger.json` endpoint

babakks:
## Description

This PR disables authentication on `/swagger.json` in the ReBAC Admin API.

Fixes CSS-10918

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [ ] Covered by unit tests

### DIFF
--- a/cmd/jimmsrv/service/service.go
+++ b/cmd/jimmsrv/service/service.go
@@ -408,7 +408,7 @@ func NewService(ctx context.Context, p Params) (*Service, error) {
 
 	s.mux.Mount("/metrics", promhttp.Handler())
 
-	s.mux.Mount("/rebac", middleware.AuthenticateRebac(rebacBackend.Handler(""), &s.jimm))
+	s.mux.Mount("/rebac", middleware.AuthenticateRebac("/rebac", rebacBackend.Handler(""), &s.jimm))
 
 	mountHandler(
 		"/debug",
--- a/internal/middleware/auth.go
+++ b/internal/middleware/auth.go
@@ -4,6 +4,7 @@ package middleware
 
 import (
 	"net/http"
+	"strings"
 
 	rebac_handlers "github.com/canonical/rebac-admin-ui-handlers/v1"
 	"github.com/juju/zaputil/zapctx"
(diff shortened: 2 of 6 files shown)


internal/middleware/auth.go
```diff
@@ -56,4 +59,13 @@ func AuthenticateRebac(next http.Handler, jimm jujuapi.JIMM) http.Handler {
 		ctx = rebac_handlers.ContextWithIdentity(ctx, user)
 		next.ServeHTTP(w, r.WithContext(ctx))
 	}), jimm)
+
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		relativePath, _ := strings.CutPrefix(r.URL.Path, baseURL)
+		if relativePath == "/v1/swagger.json" {
```
alesstimec: perhaps define that one endpoint as a constant.. or better maybe
```
var unauthenticatedEndpoints []string = []string{
   "/v1/swagger.json"
}

for _, path := range unauthenticatedEndpoints {
  if relativePath == path {
    next.ServeHTTP(w, r)
  }
}
```

alesstimec (dismissed):

internal/middleware/auth.go
babakks: How about a map?

pkulik0 (dismissed):
lgtm with Ales's comment, gives us more flexibility
size S

cmd/jimmsrv/service/service.go
```diff
@@ -408,7 +408,7 @@ func NewService(ctx context.Context, p Params) (*Service, error) {
 
 	s.mux.Mount("/metrics", promhttp.Handler())
 
-	s.mux.Mount("/rebac", middleware.AuthenticateRebac(rebacBackend.Handler(""), &s.jimm))
+	s.mux.Mount("/rebac", middleware.AuthenticateRebac("/rebac", rebacBackend.Handler(""), &s.jimm))
```
ale8k, line 411: Can we make this follow the pattern of the other httphandlers in jimmhttp? I asked @SimoneDutto to do the same on his PR/ Maybe we can't with this because of the lib but would be nice to try

ale8k (dismissed):
if can't do my comment, lgtm

cmd/jimmsrv/service/service.go
babakks: What do you mean? To make it independent of the base URL?

internal/middleware/auth.go
```diff
@@ -27,11 +28,18 @@ func AuthenticateViaCookie(next http.Handler, jimm jujuapi.JIMM) http.Handler {
 	})
 }
 
-// AuthenticateRebac is a layer on top of AuthenticateViaCookie
-// It places the OpenFGA user for the session identity inside the request's context
-// and verifies that the user is a JIMM admin.
-func AuthenticateRebac(next http.Handler, jimm jujuapi.JIMM) http.Handler {
-	return AuthenticateViaCookie(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+// Set of ReBAC Admin API endpoints that do not require authentication.
+var unauthenticatedEndpoints = map[string]any{
```
pkulik0: the nittiest of nit-picks: I would prefer struct{} as the value type, because any requires some memory for the type information. It's not like we have a hot path here, so feel free to keep it this way

pkulik0 (dismissed):

internal/middleware/auth.go
babakks: Thanks. That's actually a nice idea. Will do.

alesstimec (approved):

internal/middleware/auth.go
alesstimec: wouldn't map[string]bool be the most "efficient"?

babakks: I don't think so. Since the `struct{}` requires no memory space, that would be the most efficient (in terms of size, of course). I couldn't find an official guideline/post on this, but this subreddit is interesting:
https://www.reddit.com/r/golang/comments/qzha6o/for_sets_maptstruct_vs_maptbool/

kian99 (approved):
Nice ty

PR #1378, opened 2024-09-26
Get group by name

pkulik0:
## Description

This PR makes it possible to fetch groups by name using the GetGroup endpoint. This change is required for the jaas group data source of the juju terraform provider. 

[CSS-10995](https://warthogs.atlassian.net/browse/CSS-10995)

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [x] Covered by unit tests

[CSS-10995]: https://warthogs.atlassian.net/browse/CSS-10995?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/jimm/access.go
```diff
@@ -717,14 +717,14 @@ func (j *JIMM) CountGroups(ctx context.Context, user *openfga.User) (int, error)
 	return count, nil
 }
 
-// GetGroup returns a group based on the provided UUID.
-func (j *JIMM) GetGroupByID(ctx context.Context, user *openfga.User, uuid string) (*dbmodel.GroupEntry, error) {
+// GetGroup returns a group based on the provided UUID or name.
+func (j *JIMM) GetGroup(ctx context.Context, user *openfga.User, uuid, name string) (*dbmodel.GroupEntry, error) {
```
kian99: I suggest a separate method, so we have `GetGroupByID` and `GetGroupByName`

pkulik0: Needed to sleep on this but at the end I still wasn't sure about this, but both ways are fine. 
Just noticed the op for the function is set to AddGroup too so one more change.

alesstimec: also nitpick: `GetGroupByUUID` and `GetGroupByName`

internal/jimm/access.go
```diff
@@ -717,18 +717,27 @@ func (j *JIMM) CountGroups(ctx context.Context, user *openfga.User) (int, error)
...
-	if err := j.Database.GetGroup(ctx, &group); err != nil {
+	if err := j.Database.GetGroup(ctx, group); err != nil {
 		return nil, errors.E(op, err)
 	}
-	return &group, nil
+	return group, nil
+}
+
+// GetGroupByUUID returns a group based on the provided UUID.
+func (j *JIMM) GetGroupByUUID(ctx context.Context, user *openfga.User, uuid string) (*dbmodel.GroupEntry, error) {
```
kian99, line 734: Thanks!

internal/jujuapi/jimm.go
```diff
@@ -41,7 +41,8 @@ func init() {
 		addCloudToControllerMethod := rpc.Method(r.AddCloudToController)
 		removeCloudFromControllerMethod := rpc.Method(r.RemoveCloudFromController)
 		addGroupMethod := rpc.Method(r.AddGroup)
-		getGroupMethod := rpc.Method(r.GetGroup)
```
kian99: This is not backwards compatible and would break existing Terraform provider code that has landed. We could do work to go and remove it from their codebase before is shipped, but that is not how we should build.

I think in the case of the public API, we can keep it as just `GetGroup` and we can add a field to `GetGroupRequest` so it has a UUID and a name, then in the jujuapi handler, if both are set, return an error.

OR we can keep `GetGroup` which queries by UUID and add `GetGroupByName` which queries by name.

internal/jimm/access.go
```diff
@@ -717,18 +717,27 @@ func (j *JIMM) CountGroups(ctx context.Context, user *openfga.User) (int, error)
 	return count, nil
 }
 
-// GetGroup returns a group based on the provided UUID.
-func (j *JIMM) GetGroupByID(ctx context.Context, user *openfga.User, uuid string) (*dbmodel.GroupEntry, error) {
-	const op = errors.Op("jimm.AddGroup")
+// getGroup returns a group based on the provided UUID or name.
+func (j *JIMM) getGroup(ctx context.Context, user *openfga.User, group *dbmodel.GroupEntry) (*dbmodel.GroupEntry, error) {
+	const op = errors.Op("jimm.getGroup")
```
SimoneDutto, line 722: `
```suggestion
        if group != nil && group.name != "" {
	   const op = errors.Op("jimm.GetGroupByName")
	} else {
	...
	}
```
otherwise it will be very hard to tell from the logs which API is failing.

internal/jimm/access_test.go
```diff
@@ -834,9 +834,13 @@ func TestGetGroupByID(t *testing.T) {
 	c.Assert(err, qt.IsNil)
 	c.Assert(groupEntry.UUID, qt.Not(qt.Equals), "")
 
-	gotGroup, err := j.GetGroupByID(ctx, u, groupEntry.UUID)
+	gotGroupUuid, err := j.GetGroupByUUID(ctx, u, groupEntry.UUID)
 	c.Assert(err, qt.IsNil)
-	c.Assert(gotGroup, qt.DeepEquals, groupEntry)
+	c.Assert(gotGroupUuid, qt.DeepEquals, groupEntry)
+
+	gotGroupName, err := j.GetGroupByName(ctx, u, groupEntry.Name)
```
SimoneDutto, line 841: add also some tests when the group is not found

internal/jimm/access.go
pkulik0: In this case I'd rather duplicate the code and keep them separate. But since ByName and ByUUID are just syntactic sugar on top of the actual method that accepts a GroupEntry struct I wouldn't change it.

internal/jimm/access_test.go
pkulik0: Already covered by tests for the API client, but just to be sure I'll add this here too.

internal/jujuapi/jimm.go
pkulik0: Ahh true, then I'll just revert one commit, because at some point I wanted to have only one endpoint. Thanks for pointing this out

pkg/api/client.go
```diff
@@ -123,17 +123,10 @@ func (c *Client) AddGroup(req *params.AddGroupRequest) (params.AddGroupResponse,
 	return resp, err
 }
 
-// GetGroupByUUID returns the group with the given UUID.
-func (c *Client) GetGroupByUUID(req *params.GetGroupByUUIDRequest) (params.GetGroupResponse, error) {
+// GetGroup returns the group with the given UUID.
```
kian99: Update the godoc, mention that only the UUID or the name should be provided

internal/jujuapi/access_control_test.go
```diff
@@ -70,18 +70,15 @@ func (s *accessControlSuite) TestGetGroup(c *gc.C) {
...
-	retrievedName, err := client.GetGroupByName(&apiparams.GetGroupByNameRequest{Name: created.Name})
+	retrievedName, err := client.GetGroup(&apiparams.GetGroupRequest{Name: created.Name})
 	c.Assert(err, jc.ErrorIsNil)
 	c.Assert(retrievedName.Group, gc.DeepEquals, created.Group)
 
-	_, err = client.GetGroupByUUID(&apiparams.GetGroupByUUIDRequest{UUID: "non-existent"})
-	c.Assert(err, gc.ErrorMatches, ".*not found.*")
-
-	_, err = client.GetGroupByName(&apiparams.GetGroupByNameRequest{Name: "non-existent"})
+	_, err = client.GetGroup(&apiparams.GetGroupRequest{UUID: "non-existent"})
```
kian99, line 81: Add a test where you include both a UUID and name

kian99 (dismissed):
lgtm with some minor comments

SimoneDutto (approved):

kian99 (approved):

PR #1379, opened 2024-10-01
Adds security.md.

alesstimec:
## Description

Adds security.md as part of the vulnerability and response.

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [ ] Covered by unit tests

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,18 @@
+## Reporting a vulnerability
+
+The easiest way to report a security issue is through
+[GitHub](https://github.com/canonical/jimm/security/advisories/new). See
+[Privately reporting a security
+vulnerability](https://docs.github.com/en/code-security/security-advisories/guidance-on-reporting-and-writing/privately-reporting-a-security-vulnerability)
+for instructions.
+
+The LXD GitHub admins will be notified of the issue and will work with you
+to determine whether the issue qualifies as a security issue and, if so, in
+which component. We will then handle figuring out a fix, getting a CVE
+assigned and coordinating the release of the fix to the various Linux
+distributions.
+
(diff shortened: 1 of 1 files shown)


kian99 (commented):
Going to https://github.com/Canonical/jimm/security/advisories/new returns a 404

kian99 (dismissed):

ale8k:
I think this is missing detail, this is really just the temlate they given us but lxd is more in depth.

https://github.com/canonical/lxd/blob/main/SECURITY.md

This is much nicer and also the file name should be capitlised I think...

We should say what versions we're willing to fix. I.e., juju last major latest minor, current, and maybe next major?...

ale8k (changes requested):
requesting changes on file name and some details on what we're willing to fix (i.e., versioning)

SECURITY.md
```diff
@@ -0,0 +1,18 @@
...
+[GitHub](https://github.com/canonical/jimm/security/advisories/new). See
+[Privately reporting a security
+vulnerability](https://docs.github.com/en/code-security/security-advisories/guidance-on-reporting-and-writing/privately-reporting-a-security-vulnerability)
+for instructions.
+
+The LXD GitHub admins will be notified of the issue and will work with you
+to determine whether the issue qualifies as a security issue and, if so, in
+which component. We will then handle figuring out a fix, getting a CVE
+assigned and coordinating the release of the fix to the various Linux
+distributions.
```
kian99, line 13: This needs fixing to be about JAAS not LXD and the bit at the end about "various Linux distributions" doesn't apply.

ale8k (approved):
bar kians comment

PR #1380, opened 2024-10-02
remove bakery v2 dependency

kian99:
## Description

When investigating the use of Macaroons in JIMM I found a few things that could be removed.
- The bakery v2 dependency is no longer needed.
- The BakeryLogger is no longer needed.


alesstimec (approved):
RIP macaroons

SimoneDutto (approved):

PR #1381, opened 2024-10-04
add cors protection for websocket endpoints

kian99:
## Description

This PR introduces protection against CSRF attacks on our websocket endpoints.
The following links share details on how this vulnerability can be exploited:
- https://web.archive.org/web/20240715011125/https://www.christian-schneider.net/blog/cross-site-websocket-hijacking/
- https://portswigger.net/web-security/websockets/cross-site-websocket-hijacking

In summary, browsers do not enforce the Same-Origin Policy (SOP) policy when it comes to websocket requests and if a websocket handler uses cookies or HTTP-auth for authentication during the websocket upgrade, then this can be exploited. The browser will send cookies (provided the **SameSite** policy is None) and a malicious page can exploit this to hijack a user's cookies on a websocket handler.

The fix implemented is to introduce a websocketCors middleware that will verify the `Origin` header. If the incoming origin matches the allowed origins then the websocket handshake is allowed to continue otherwise we return a unauthorised error. Only browsers are expected to send the origin header so the Juju CLI and other clients will not be affected.

Fixes [CSS-10930](https://warthogs.atlassian.net/browse/CSS-10930)

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [x] Covered by unit tests

[CSS-10930]: https://warthogs.atlassian.net/browse/CSS-10930?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/middleware/websocketcors.go
```diff
@@ -0,0 +1,47 @@
...
+	corsOpts := cors.New(cors.Options{
+		AllowedOrigins: allowedOrigins,
+	})
+	return &WebsocketCors{cors: corsOpts}
+}
+
+// Handler implements CORS validation for websocket handlers.
+// Any methods beside GET are rejected as a bad request.
+// If the origin is not in the allow list, a status forbidden is returned.
+func (c *WebsocketCors) Handler(h http.Handler) http.Handler {
```
SimoneDutto, line 27: nit pick:
h -> next

internal/middleware/websocketcors.go
```diff
@@ -0,0 +1,47 @@
...
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.Method != http.MethodGet {
+			w.WriteHeader(http.StatusBadRequest)
+			_, _ = w.Write([]byte("unexpected method"))
+			return
+		}
+
+		// We can allow cross-origin requests for those requests that don't
```
SimoneDutto, line 35: requests coming from juju clients what origin do they have? From your code I imagine nothing.

SimoneDutto (approved):
LGTM, as soon as we merge the http-proxy branch we need to add the same checks for the HTTP methods

internal/middleware/websocketcors.go
kian99: Nothing, our tests prove as much, otherwise CLI requests would start failing. And as far as I understand the `origin` header is added by browsers.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Origin
https://developer.mozilla.org/en-US/docs/Glossary/Forbidden_header_name

SimoneDutto: it also makes sense, otherwise it would be "localhost" or something like that. 
I just want to make sure

ale8k (approved):
lgtm, clear

PR #1382, opened 2024-10-04
Security scan weekly

ale8k:
## Description

Security Scan Weekly

Uploads scans weekly as artifacts and github security. Currently only runs against v3, but to be updated later with more branches we maintain.

This enables us to track vulnerabilites every monday in our security tab. Doesn't include any notifications but gonna look into it.

It also uses a new composite workflow in comsys-build-tools dedicated to this.


kian99 (approved):

kian99:
I suggest cleaning up the commits fwiw

alesstimec (approved):

.github/workflows/security-scan-weekly.yaml
```diff
@@ -0,0 +1,25 @@
...
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ matrix.branch }}
+
+      - name: Security scan
+        uses: canonical/comsys-build-tools/.github/actions/security-scan-upload@main
+        with: 
+          upload-github-security: true
+          upload-sarif-artifact: true
```
babakks, line 25: These are defined as string in the action, so they should be quoted here.

babakks (changes requested):

.github/workflows/security-scan-weekly.yaml
ale8k: I ran a scan and it worked fine, must be parsing them to strings?

PR #1387, opened 2024-10-07
Feature http proxy

kian99:
## Description

Merges `feature-http-proxy` into `v3`.

This bumps the Juju dependency and introduces a break change with the Juju CLI requiring Juju v3.5.4+.

This PR lands the changes for the following features:
- Support `juju debug-log` thanks to the additions to the Juju CLI to send the JWT session token to JIMM.
- Support uploading local charm by adding support in JIMM for these endpoints (proxied to the underlying Juju controller).


SimoneDutto (approved):

alesstimec (approved):
LGTM with a note: we are proxying HTTP endpoints using admin basic auth for controllers. Are we ok with that?

kian99:
> LGTM with a note: we are proxying HTTP endpoints using admin basic auth for controllers. Are we ok with that?

I think we have to be okay with it for now, the debug log endpoints is authenticated via a header value (with username/password) and Simone updated Juju to support auth with a JWT token but that is only in Juju 3.x (I think 3.4.5) so if we switched to a JWT token auth here then we wouldn't support support older controllers.

PR #1388, opened 2024-10-09
add qa script and readme

SimoneDutto:
## Description

While qa locally ReBAC Admin I made some change to the original rebac-admin-ui-handlers script to adapt it to jimm's deployment.
After discussing with @babakks we decided to add it to Jimm's repo for reproducibility.

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [ ] Covered by unit tests

### DIFF
--- a/local/rebac-admin/README.md
+++ b/local/rebac-admin/README.md
@@ -1,0 +1,52 @@
+# QA ReBAC Admin Backend
+
+To QA locally the ReBAC Admin Backend you need to:
+
+## Change the rebac authentication middleware at `internal/middleware/authn.go` to:
+```go
+func AuthenticateRebac(baseURL string, next http.Handler, jimm JIMMAuthner) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		ctx := r.Context()
+		relativePath, _ := strings.CutPrefix(r.URL.Path, baseURL)
+		if _, found := unauthenticatedEndpoints[relativePath]; found {
+			next.ServeHTTP(w, r)
+			return
+		}
+		identity := "admin@canonical.com"
+		user, err := jimm.UserLogin(ctx, identity)
+		if err != nil {
+			zapctx.Error(ctx, "failed to get openfga user", zap.Error(err))
+			http.Error(w, "internal authentication error", http.StatusInternalServerError)
+			return
+		}
+		user.JimmAdmin = true
+		if !user.JimmAdmin {
+			w.WriteHeader(http.StatusUnauthorized)
+			_, _ = w.Write([]byte("user is not an admin"))
+			return
+		}
+
+		ctx = rebac_handlers.ContextWithIdentity(ctx, user)
+		next.ServeHTTP(w, r.WithContext(ctx))
+	})
+}
+```
+This change will create and authenticate an admin user used by the QA script.
+
+## Start JIMM
+
+Follow [local/README.md](../README.md)
+
+## Launch the script
+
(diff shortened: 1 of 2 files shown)


babakks (dismissed):

local/rebac-admin/qa.sh
```diff
@@ -0,0 +1,296 @@
...
+    function onexit {
+        curl "$_host/shutdown"
+        wait $_PID1
+        echo "Server shut down"
+    }
+    trap onexit EXIT
+
+    ## Run server in background
+    go build -o server ./cmd
+    bash -c 'sleep 1 && ./server' &
```
kian99: Is this used? I don't see how it would work without setting a bunch of env vars for Postgres, Vault, etc.

local/rebac-admin/README.md
```diff
@@ -0,0 +1,52 @@
+# QA ReBAC Admin Backend
+
+To QA locally the ReBAC Admin Backend you need to:
+
+## Change the rebac authentication middleware at `internal/middleware/authn.go` to:
```
kian99, line 5: This feels a bit flimsy because the code below will not stay up to date as the actual code changes.
Also, we don't mention what the change is doing?
Could we not run the tests with the default admin user in the docker compose?

kian99 (dismissed):
To understand better, does testing like this offer us anything beyond what our integration tests do?

local/rebac-admin/qa.sh
SimoneDutto: this is to be removed

local/rebac-admin/README.md
SimoneDutto: this change is the easiest way to remove the authentication but keep the user in the context for the rebac-admin stuff to work.
Otherwise we need to create a cookie and use it in the bash script, or something like taht

SimoneDutto:
@kian99 to answer you: it's not adding much to the whole integration tests. In fact it is not run in a action, it is just because @babakks mentioned that this script could be useful so he suggested to add to Jimm's repo

kian99:
>@kian99 to answer you: it's not adding much to the whole integration tests. In fact it is not run in a action, it is just because @babakks mentioned that this script could be useful so he suggested to add to Jimm's repo

In that case I'd question whether we should put it in the repo with the additional instructions on having to modify the source code to get it to run. It feels like "dead code" something that won't get used and potentially only cause confusion to someone that looks at it in a few years time. Maybe instead we could include in the readme a link to the test script in the other repo, etc.

SimoneDutto:
> > @kian99 to answer you: it's not adding much to the whole integration tests. In fact it is not run in a action, it is just because @babakks mentioned that this script could be useful so he suggested to add to Jimm's repo
> 
> In that case I'd question whether we should put it in the repo with the additional instructions on having to modify the source code to get it to run. It feels like "dead code" something that won't get used and potentially only cause confusion to someone that looks at it in a few years time. Maybe instead we could include in the readme a link to the test script in the other repo, etc.

Since it is temporary until we have a proper QA from the Web Team we can even leave this PR open for reference and close the PR as soon as the QA have been done.

@babakks what do you think?

local/rebac-admin/README.md
ale8k: I think the best way to QA this is honestly a set of Cryprus tests running against a local UI. Not like this... ANd the tests are the responsibility of the webteam.

ale8k (changes requested):
Let's just QA this via cyprus from the web team. Happy to write CI that runs that in JIMM or calls their workflow. This adds extra load on us for the QAing when really it needs to be e2e from the UI..

PR #1389, opened 2024-10-10
Cleanup go mod

ale8k:
Our go mod had a couple require blocks for some reason and indirects in the wrong place, this tidies it up. I also noticed we were using v4 names in a single places still, removed this.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,69 +3,62 @@ module github.com/canonical/jimm/v3
 go 1.22.5
 
 require (
-	github.com/canonical/go-dqlite v1.21.0 // indirect
+	github.com/antonlindstrom/pgstore v0.0.0-20220421113606-e3a6e3fed12a
 	github.com/canonical/go-service v1.0.0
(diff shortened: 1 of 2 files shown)


kian99 (commented):
What triggered these changes? Just a `go mod tidy`?

SimoneDutto (commented):
Are you sure we were not using functionality from v4, i remember @kian99 mentioned something once about using v4 to validate something about service account

ale8k:
> What triggered these changes? Just a `go mod tidy`?

The separate replace blocks? No idea... I moved it back manually and moved the indirects out manually and then ran a go mod tidy. It stayed in order now, not sure how it ended up like this.

> Are you sure we were not using functionality from v4, i remember @kian99 mentioned something once about using v4 to validate something about service account

Uhm no no, its just creating a user and controller tag via v4 in an admin test of jujuapi. Only place, everything else is v5.

SimoneDutto (approved):

kian99 (approved):

PR #1395, opened 2024-10-17
remove soft-deletes

kian99:
## Description

This PR removes the gorm.Model type from ~~most of our db model structs~~ groups to avoid the issue of soft-deletes. Removing it from all models break several tests (it seems because we ignore gorm.Model in these tests).

Fixes [CSS-11335](https://warthogs.atlassian.net/browse/CSS-11335)

[CSS-11335]: https://warthogs.atlassian.net/browse/CSS-11335?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


SimoneDutto (commented):
i am curious, how is this not requiring a db migration (gorm magic)? Are we fine with removing the soft delete from controllers? Killing a controller on juju can fail and it is async, right?

kian99:
> i am curious, how is this not requiring a db migration (gorm magic)? Are we fine with removing the soft delete from controllers? Killing a controller on juju can fail and it is async, right?

Ah good point, we do need a migration to remove the column. Though it should work fine without the migration because Gorm will just ignore the field.
As for removing soft delete from controllers, that was already done previously, this PR just removes the soft-delete from `CloudRegionControllerPriority` and `ControllerConfig` controller related tables (among others). When you remove a Juju controller from JIMM you basically just detach it.

Edit: But I've edited this now to only remove soft delete for groups. See the description.

internal/dbmodel/group.go
```diff
@@ -6,15 +6,17 @@ import (
...
 
 	apiparams "github.com/canonical/jimm/v3/pkg/api/params"
 	jimmnames "github.com/canonical/jimm/v3/pkg/names"
 )
 
 // A GroupEntry holds information about a user group.
 type GroupEntry struct {
-	gorm.Model
+	// Note this doesn't use the standard gorm.Model to avoid soft-deletes.
+	ID        uint `gorm:"primarykey"`
```
SimoneDutto, line 17: i could be wrong, but when creating `GroupEntry` this ID is always set to 0. Right?
Or the form struct tag does something to it?

SimoneDutto (approved):

SimoneDutto (commented):
wrong approve

internal/dbmodel/group.go
kian99: The default value is 0 but gorm will assign the database ID to it. (I hope that answers your question?)

SimoneDutto: yes, i was worried we were using that ID for something in the tests. But I've checked and it is alright

ale8k (approved):
lgtm

PR #1397, opened 2024-10-17
make logging SQL queries optional at debug log level

kian99:
## Description

Currently test logs are very verbose and in production, when the log level is set to "debug" we print all SQL queries. These queries may contains sensitive information like OAuth token secrets. This PR makes the most logging of SQL queries optional and gates it behind an environment variable.

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [ ] Covered by unit tests


SimoneDutto (dismissed):
can you add this env var to the docker compose as well?

SimoneDutto (approved):
ty

ale8k (approved):
I did this a while ago but never proposed aha, wanted to fix both this and jujuconnsuite but could never do connsuite... Ty!

babakks (approved):
LGTM. But as a sidenote, this still can be seen as a security vulnerability. If someone gets access to the running JIMM instance (the Pebble binary, I mean) then they can enable SQL logging.

If it's just for testing or local debugging, I think we can turn the flag into a compile time constant and enable it in our docker compose or tests. Have a look at [this](https://stackoverflow.com/a/49530118) Q/A as an example.

kian99:
>LGTM. But as a sidenote, this still can be seen as a security vulnerability. If someone gets access to the running JIMM instance (the Pebble binary, I mean) then they can enable SQL logging.

But surely if they can access the running JIMM instance they can already do more harm?

SimoneDutto:
> LGTM. But as a sidenote, this still can be seen as a security vulnerability. If someone gets access to the running JIMM instance (the Pebble binary, I mean) then they can enable SQL logging.
> 
> If it's just for testing or local debugging, I think we can turn the flag into a compile time constant and enable it in our docker compose or tests. Have a look at [this](https://stackoverflow.com/a/49530118) Q/A as an example.

This is very interesting, we can use this approach for the disable signature verification in the livepatch client

PR #1400, opened 2024-10-21
WD-16026 - fix: improve local JIMM setup

huwshimi:
## Description

- Improve sys-deps instructions.
- Fix jimmctl in add-controller.sh script.

Fixes https://warthogs.atlassian.net/browse/WD-16026

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [ ] Covered by unit tests


local/jimm/add-controller.sh
```diff
@@ -26,6 +26,7 @@ if [ $jimmctlAvailable -ne 0 ]; then
     echo "Building jimmctl..."
     # Build jimmctl so we may add a controller.
     go build ./cmd/jimmctl
+    sudo ln -s $PWD/jimmctl /usr/local/bin/jimmctl
```
huwshimi: Not sure if this is the right way to make the binary available, but with out this the following calls to `jimmctl` (and the `which jimmctl` above) don't use the local `./jimmctl`

ale8k: Ah is this in the case of the snap being available? If so, coudl you update this PR to do "./" infront of all jimmctl calls?

ale8k (changes requested):
Requesting changes if I'm understanding the jimmctl issue right

Makefile
```diff
@@ -138,9 +138,9 @@ ifeq ($(APT_BASED),0)
 # yq is necessary for some scripts that process controller-info yaml files.
 	@$(call check_dep,yq,Missing yq - install with 'sudo snap install yq')
 # Microk8s is required if you want to start a Juju controller on Microk8s.
-	@$(call check_dep,microk8s,Missing microk8s - install with 'sudo snap install microk8s')
+	@$(call check_dep,microk8s,Missing microk8s - install with 'sudo snap install microk8s --classic')
```
babakks: Since JIMM v3 works with with Juju 3.X, it should be an strict snap installation. For example:

```sh
sudo snap install microk8s --channel=1.30-strict/stable
```

Makefile
```diff
@@ -138,9 +138,9 @@ ifeq ($(APT_BASED),0)
...
 # Docker is required to start the test dependencies in containers.
-	@$(call check_dep,docker,Missing Docker - install from https://docs.docker.com/engine/install/)
+	@$(call check_dep,docker,Missing Docker - install from https://docs.docker.com/engine/install/  or 'sudo snap install docker')
```
babakks: Please remove extra whitespace before "or".

babakks (changes requested):

local/jimm/add-controller.sh
huwshimi: This is actually the inverse, without this fix it will build jimmctl but then it'll display an error about jimmctl not existing as it's trying to use the global command.

However, if you prefer we could symlink the snap jimmctl into the project directory if it exists and then use `./jimmctl` which wouldn't pollute the global bin.

huwshimi: Actually, symlinking the snap won't work as then the `which jimmctl` lines won't work.

ale8k: I think just doing ./ before each jimmctl is fine and clearer

huwshimi: > I think just doing ./ before each jimmctl is fine and clearer

This would mean it would always use the locally built jimmctl and not the jimmctl from the snap (if available) so if we switch to `./jimmctl` I suppose we should also remove the other parts of the script that handle the snap jimmctl.

huwshimi: I've removed the snap handling from the script.

kian99: I'm ooo, but saw this come through in an email. Don't remove the snap handling because the Github Action in .github/actions/test-server spins up a JIMM test env and doesn't require Go to be installed, rather opting to use the jimmctl snap and uses this add-controller script.

kian99: I suggest either using the original symlink idea or making a bash variable that becomes either "./" or empty string and placing it in front of calls to jimmctl.

babakks (dismissed):

local/jimm/add-controller.sh
huwshimi: Thanks @kian99 I've undone my changes to remove the snap handling and instead this now checks for the existence of `./jimmctl` instead of just relying on `which jimmctl` and swaps out the jimmctl command (stored in a variable).

kian99 (approved):

ale8k (approved):
lgtm!

PR #1401, opened 2024-10-21
CSS-11554: Improve local docker startup time

SimoneDutto:
## Description

This pr tries to cut the local startup time of JIMM.

_Tests have been done on my machine, the times referees to the second run after the volumes have been populated_

- Start: 85 seconds
- + volume for go modules: 50 seconds
- + dockerfile to install delve to cache between multiple run: 45 seconds
- + volume for go build: 25 seconds

Final time: 25 seconds (-70% startup time)

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [ ] Covered by unit tests


Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM cosmtrek/air:latest
+
+RUN go install github.com/go-delve/delve/cmd/dlv@latest
+
+ENTRYPOINT [ "air" ]
```
ale8k: New line ;)

docker-compose.yaml
```diff
@@ -133,3 +131,7 @@ services:
       interval: 5s
       timeout: 10s
       retries: 30
+
+volumes:
+  go-modules:
+  go-cache:
```
ale8k: New line ;)

docker-compose.yaml
```diff
@@ -42,23 +42,21 @@ services:
...
-      - -c
-      - >-
-        go install github.com/go-delve/delve/cmd/dlv@latest
-        && air
     ports:
       - 17070:80
       - 2345:2345
     volumes:
       - ./:/jimm/
+      - go-modules:/go/pkg/mod
```
ale8k: Leave a comment here explaining we just wanna shove mods and build caches to speed up restarts please.

ale8k (dismissed):
Seems good to me, not sure it was necessary as when we do QA JIMM, it's a long process no matter what aha :P

P.S., add the newlines ;)

Dockerfile
babakks: Normally the `Dockerfile` file at the root has something to do with the underlying codebase. Since, this is just the `air` image (with `delve` installation), I'd rather put it in the `local/jimm` directory.

Also, a comment might be helpful on why we use this image. Maybe just comment before line 3, saying *"Installing delve in the image to avoid running `go install` in running containers"*.

babakks (dismissed):

ale8k (approved):

babakks (approved):

PR #1411, opened 2024-10-22
CSS-11598: Improve logging

SimoneDutto:
## Description

This pr includes:
- request logger middleware
- local logger with colours
- local logger configured for suites
- production logger default config 

Some examples:

image

image

Updated examples:
image

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [ ] Covered by unit tests


cmd/jimmsrv/service/service.go
```diff
@@ -273,6 +278,12 @@ func NewService(ctx context.Context, p Params) (*Service, error) {
 	s := new(Service)
 	s.mux = chi.NewRouter()
 
+	// chi middlewares
+	if p.LogLevel == "debug" {
+		s.mux.Use(chimiddleware.RequestLogger(&logger.HTTPLogFormatter{}))
+	}
+	s.mux.Use(chimiddleware.Recoverer)
```
babakks: Why do we need the `Recoverer` middleware? Is that for recovering from panics? If so, it'll mask panics and prevents us/users from catching them.

internal/logger/default_logger.go
```diff
@@ -0,0 +1,38 @@
...
+	"github.com/juju/zaputil/zapctx"
+	"github.com/mattn/go-colorable"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+)
+
+// SetupLogger sets up the default logger.
+// The local logger is a colorized plain text logger.
+// The production logger is a JSON structured logger.
+func SetupLogger(ctx context.Context, logLevel string, logLocal bool) {
```
ale8k: Let's not call it log local, perhaps development mode?

babakks: I'd rather name `logLocal` to somthing like `development`.

internal/logger/http_logger.go
```diff
@@ -0,0 +1,47 @@
...
+	"net/http"
+	"time"
+
+	"github.com/go-chi/chi/v5/middleware"
+	"github.com/juju/zaputil/zapctx"
+	"go.uber.org/zap"
+)
+
+// HTTPLogFormatter is an implementation of chimiddleware.LogFormatter. It formats logs for http requests.
+type HTTPLogFormatter struct{}
```
ale8k, line 15: Can we have a constructor for this? We've been trying to keep to constructors recently

cmd/jimmsrv/main.go
```diff
@@ -38,15 +39,13 @@ func main() {
 //
 //nolint:gocognit // Start function to be ignored.
 func start(ctx context.Context, s *service.Service) error {
+	logLevel := os.Getenv("JIMM_LOG_LEVEL")
+	logLocal, _ := strconv.ParseBool(os.Getenv("JIMM_LOG_LOCAL"))
```
babakks: And the same change for the env var name, because it's a bit unclear what does it mean. Maybe something like `JIMM_LOG_DEVELOPMENT`?

docker-compose.common.yaml
```diff
@@ -3,6 +3,7 @@
 services:
   jimm-base:
     environment:
+      JIMM_LOG_LOCAL: true
```
babakks: Please assign it with a string.

internal/logger/default_logger.go
```diff
@@ -0,0 +1,38 @@
...
+	pLogLevel, err := zapcore.ParseLevel(logLevel)
+	if err != nil {
+		fmt.Printf("ERROR: log level '%s' cannot be parsed, defaulting to info\n", logLevel)
```
babakks: **Hint:** You can use `%q` to automatically quote the value:

```go
fmt.Printf("... %q ...", logLevel)
```

internal/logger/default_logger.go
```diff
@@ -0,0 +1,38 @@
...
+		pLogLevel = zap.InfoLevel
+	}
+	if logLocal {
+		devConfig := zap.NewDevelopmentEncoderConfig()
```
ale8k, line 26: Can we remove the tabs before hand? You can ask AI for build a pretty logging format for you also...

in internal/testutils/jimmtest/logging.go

we have a test logger, 

```go
func (s *LoggingSuite) setUp(c *gc.C) {
	output := gocheckZapWriter{c}
	logger := zap.New(zapcore.NewCore(
		zapcore.NewConsoleEncoder(zapcore.EncoderConfig{
			LevelKey:    "level",
			MessageKey:  "msg",
			EncodeLevel: zapcore.CapitalLevelEncoder,
			EncodeTime:  zapcore.ISO8601TimeEncoder,
		}),
		output,
		zap.DebugLevel,
	))

	zapctx.Default = logger

```

Ideally, they'd share the same development logger so it's consistent. Let's revisit the test part in this PR, I see in jimmsuite you're doing it on our suites setuptest, but then we're just setting two loggers up for no reason. So move the development config (with the amendments) to a centralised location that can be shared between the two:
1. The running server in dev mode
2. Our LoggingSuite updated with the same dev log config

internal/logger/default_logger.go
```diff
@@ -0,0 +1,38 @@
...
+		developLogger := zap.New(zapcore.NewCore(
+			zapcore.NewConsoleEncoder(devConfig),
+			zapcore.AddSync(colorable.NewColorableStdout()),
+			pLogLevel,
+		))
+		zapctx.Default = developLogger
+	} else {
+		prodConfig := zap.NewProductionConfig()
+		prodConfig.Level = zap.NewAtomicLevelAt(pLogLevel)
+		proLogger := zap.Must(prodConfig.Build()) // this panic is an error is encountered during Build
```
babakks: s/this panic is/this panics if

ale8k (changes requested):
See my comment on default_logger as to why I'm requesting changes. Good work so far!

internal/logger/default_logger_test.go
```diff
@@ -0,0 +1,57 @@
...
+			logLocal: true,
+		},
+		{
+			logLevel: "not_exisiting_level",
+			logLocal: true,
+		},
+	}
+
+	for _, t := range tests {
+		logger.SetupLogger(ctx, t.logLevel, t.logLocal)
```
babakks: Can you please run the test cases in subtests (i.e., `t.Run(...)`)?

babakks (dismissed):

cmd/jimmsrv/service/service.go
SimoneDutto: it catches panic and pretty print the stack trace

docker-compose.common.yaml
ale8k: Could also just use same env var for level, but if "development" is sent, it'll use the dev config? So no need for another var?

SimoneDutto: mmm, it could be a shortcut to have debug + pretty + requests

SimoneDutto: i think it's nice, but maybe too opinionated

internal/logger/default_logger.go
SimoneDutto: i'll miss your hints babak <3

internal/logger/http_logger.go
SimoneDutto: we don't gain anything from a constructor with empty arguments

cmd/jimmsrv/service/service.go
babakks: That's my point. The godoc says:

> Recoverer is a middleware that recovers from panics, logs the panic (and a backtrace), and returns a HTTP 500 (Internal Server Error) status if possible. Recoverer prints a request ID if one is provided.

This means if there is a bug in JIMM's code (e.g., a nil reference) then we'll never see it panics, and the issue might be left unnoticed or mistaken.

babakks (dismissed):

cmd/jimmsrv/service/service.go
SimoneDutto: well, it is logged and we should have rules allerting us on ERROR.
I dont' feel like crashing the whole server on a route panicking is the right thing to do.

SimoneDutto:
@ale8k @babakks the production config with zap is to log into stderr. Is it a problem for charms/rock?

cmd/jimmsrv/service/service.go
babakks: Yeah, I agree that an API call should never cause the server to terminate, but in some ways recovering from panic has quite the same effect as how we ignored the error in the Livepatch server and we never knew we had it. Since JAAS is still under debugging and test deployment, maybe it's better to let these issues surface. But it's my personal opinion and I don't want to push it. Feel free to resolve this comment if you deem fit.

internal/logger/http_logger.go
alesstimec: agreed.. no need for a constructor.. it is an exported type. If we needed certain parameters, then, yes, constructor would make sense.

cmd/jimmsrv/service/service.go
alesstimec: also, let's please not do this in this PR.. this change has nothing to do with logging and might feel like you're trying to sneak in a change ;)

internal/logger/default_logger_test.go
```diff
@@ -0,0 +1,68 @@
+// Copyright 2024 Canonical.
+package logger_test
```
alesstimec, line 2: missing new line

internal/logger/http_logger.go
```diff
@@ -0,0 +1,47 @@
+// Copyright 2024 Canonical.
+package logger
```
alesstimec, line 3: new line

cmd/jimmsrv/service/service.go
SimoneDutto: fair enough

SimoneDutto: i've removed the middleware in favor of discussing that in an another pr

internal/testutils/loggingsuite/loggingsuite.go
```diff
@@ -34,27 +34,14 @@ func (s *LoggingSuite) TearDownTest(c *gc.C) {
...
-
+	logger.SetupLogger(context.Background(), "debug", true)
 	loggo.ResetLogging()
 	// Don't use the default writer for the test logging, which
 	// means we can still get logging output from tests that
 	// replace the default writer.
 	err := loggo.RegisterWriter(loggo.DefaultWriterName, discardWriter{})
 	c.Assert(err, gc.IsNil)
-	err = loggo.RegisterWriter("loggingsuite", zaputil.NewLoggoWriter(logger))
+	err = loggo.RegisterWriter("loggingsuite", zaputil.NewLoggoWriter(zapctx.Default))
```
ale8k: Interesting way of doing it... It's ok for now but in future I think let's have setup logger return a logger that we explicitly use (and in the main app). We don't use logger inheritance, but we should and this will need to change if we ever go that route (again we should)

internal/testutils/loggingsuite/loggingsuite.go
```diff
@@ -68,15 +55,3 @@ type discardWriter struct{}
 
 func (discardWriter) Write(entry loggo.Entry) {
 }
-
-type gocheckZapWriter struct {
```
ale8k: Do gocheck logs still work fine? (You'll see a line like "------------------------------------", is there anything above or below it?)

internal/testutils/loggingsuite/loggingsuite.go
```diff
@@ -1,16 +1,16 @@
 // Copyright 2024 Canonical.
-package jimmtest
+package loggingsuite
```
ale8k: It's not conventional to name a package different from its folder name, why are we doing this?

ale8k: I think it's actually not idiomatic either*

internal/testutils/cmdtest/jimmsuite.go
```diff
@@ -48,10 +50,23 @@ type JimmCmdSuite struct {
...
+
+	loggingsuite.LoggingSuite
+}
+
+func (s *JimmCmdSuite) SetUpSuite(c *gc.C) {
+	s.JujuConnSuite.SetUpSuite(c)
+	s.LoggingSuite.SetUpSuite(c)
+}
+
+func (s *JimmCmdSuite) TearDownSuite(c *gc.C) {
```
ale8k: This wasn't here?... As in the loggingsuite and setup/teardown??

internal/logger/default_logger.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/juju/zaputil/zapctx"
+	"github.com/mattn/go-colorable"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+)
+
+// SetupLogger sets up the default logger.
+// The local logger is a colorized plain text logger.
+// The production logger is a JSON structured logger.
+func SetupLogger(ctx context.Context, logLevel string, devMode bool) {
```
ale8k, line 19: We don't need a dev mode right? If they pass a loglevel of "development", we could just use that and then remove the need for another env var??

ale8k: What I mean to say is, if you're in devmode, you're at least in debug...

internal/logger/http_logger.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+// Write is called when the request handler has finished.
+func (l *httpLogEntry) Write(status, bytes int, header http.Header, elapsed time.Duration, extra interface{}) {
+	fields := make([]zap.Field, 0)
+
+	if status != 0 {
+		fields = append(fields, zap.Int("status", status))
+	}
+	fields = append(fields, zap.String("method", l.request.Method),
```
ale8k: Some fields are on new line and others inline, can we make this look consistent?

internal/logger/http_logger.go
```diff
@@ -0,0 +1,48 @@
...
+type httpLogEntry struct {
+	*HTTPLogFormatter
+	request *http.Request
+}
+
+// Write is called when the request handler has finished.
+func (l *httpLogEntry) Write(status, bytes int, header http.Header, elapsed time.Duration, extra interface{}) {
+	fields := make([]zap.Field, 0)
+
+	if status != 0 {
```
ale8k, line 27: I don't think this would be debug only, might we want it on all the time but only log specific status codes, and then perhaps if its anything other than a erroneous code, then default to debug?

ale8k (commented):
Some comments

internal/testutils/loggingsuite/loggingsuite.go
SimoneDutto: i agree that we should use a logger and not just change the default, but this is a major change in all the project

internal/testutils/loggingsuite/loggingsuite.go
SimoneDutto: the folder is now `loggingsuite`. I needed to move this package out of jimmtest to use it also in the other suites

internal/testutils/cmdtest/jimmsuite.go
SimoneDutto: loggingsuite uses setupSuite.
You can call setupSuite on in setupSuite methods. That's the reason why jimmcmdsuite has now a setupsuite method

internal/testutils/loggingsuite/loggingsuite.go
SimoneDutto: remove test logs change in favor of new pr

internal/testutils/loggingsuite/loggingsuite.go
SimoneDutto: remove test logs change in favor of new pr

internal/testutils/cmdtest/jimmsuite.go
SimoneDutto: remove test logs change in favor of new pr

internal/logger/default_logger.go
SimoneDutto: as discussed in the standup a flag is better suited.

internal/logger/default_logger_test.go
```diff
@@ -0,0 +1,68 @@
...
+
+// TestLogSetup tests log setup with different configurations and checks for panic.
+func TestLogSetup(t *testing.T) {
+	ctx := context.Background()
+	defer func() {
+		r := recover()
+		if r != nil {
+			t.Errorf("The code panic")
+		}
+	}()
```
babakks: This will fail the entire test if even one test case causes panic. To fix this, I think it's better to move this `defer` statement into the `t.Run` in the loop.

internal/logger/default_logger_test.go
```diff
@@ -0,0 +1,68 @@
...
+			devMode:     true,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.description, func(t *testing.T) {
+			logger.SetupLogger(ctx, tt.logLevel, tt.devMode)
+			zapctx.Info(ctx, "test log")
+		})
+
```
babakks: Unnecessary blank line.

babakks (dismissed):

babakks (approved):

ale8k (approved):

PR #1412, opened 2024-10-22
track response time for http endpoints

SimoneDutto:
## Description

Add middleware to track response time for HTTP requests.
This can help us understand our performance bottlenecks and improve our observability.

> with some work we can use the juju calls' metrics to calculate jimm's overhead

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [ ] Covered by unit tests

### DIFF
--- a/internal/middleware/responsetime.go
+++ b/internal/middleware/responsetime.go
@@ -4,21 +4,40 @@ package middleware
 
 import (
 	"net/http"
+	"strconv"
 	"time"
+
+	"github.com/go-chi/chi/v5"
 
 	"github.com/canonical/jimm/v3/internal/servermon"
 )
+
+// statusRecorder to record the status code from the ResponseWriter
+type statusRecorder struct {
+	http.ResponseWriter
+	statusCode int
+}
+
+func (rec *statusRecorder) WriteHeader(statusCode int) {
+	rec.statusCode = statusCode
+	rec.ResponseWriter.WriteHeader(statusCode)
+}
 
 // MeasureResponseTime tracks response time of requests.
 func MeasureResponseTime(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		// Check the upgrade header because we only track http endpoints
-		if r.Header.Get("Upgrade") != "websocket" {
-			start := time.Now()
-			defer func() {
-				servermon.ResponseTimeHistogram.WithLabelValues(r.URL.Path, r.Method).Observe(time.Since(start).Seconds())
-			}()
+		if r.Header.Get("Upgrade") == "websocket" {
+			next.ServeHTTP(w, r)
+			return
 		}
-		next.ServeHTTP(w, r)
+		rec := statusRecorder{w, 200}
+		start := time.Now()
+		defer func() {
+			route := chi.RouteContext(r.Context()).RoutePattern()
+			statusCode := strconv.Itoa(rec.statusCode)
+			servermon.ResponseTimeHistogram.WithLabelValues(route, r.Method, statusCode).Observe(time.Since(start).Seconds())
+		}()
+		next.ServeHTTP(&rec, r)
 	})
 }
--- a/internal/middleware/responsetime.go
+++ b/internal/middleware/responsetime.go
@@ -23,8 +23,9 @@ func (rec *statusRecorder) WriteHeader(statusCode int) {
 	rec.ResponseWriter.WriteHeader(statusCode)
 }
 
-// MeasureResponseTime tracks response time of requests.
-func MeasureResponseTime(next http.Handler) http.Handler {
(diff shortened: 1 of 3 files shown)


internal/middleware/responsetime.go
```diff
@@ -0,0 +1,21 @@
...
+	"time"
+
+	"github.com/canonical/jimm/v3/internal/servermon"
+)
+
+// MeasureResponseTime tracks response time of requests.
+func MeasureResponseTime(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		start := time.Now()
+		next.ServeHTTP(w, r)
```
alesstimec: suggestion:
```
start := time.Now()
defer servermon.ResponseTimeHistogram.WithLabelValues(r.URL.Path, r.Method).Observe(time.Since(start))
next.ServeHTTP(w, r)
```

babakks: Yeah, I like the `defer` idea, but the value of `time.Since(start)` is evaluated immediately so it wouldn't be the actual time duration of the call. Maybe like this:

```go
start := time.Now()
defer func(){
    servermon.ResponseTimeHistogram.WithLabelValues(r.URL.Path, r.Method).Observe(time.Since(start))
}()
next.ServeHTTP(w, r)
```

internal/servermon/monitoring.go
```diff
@@ -131,6 +131,12 @@ var (
 		Name:      "controller",
 		Help:      "The number of controllers managed by JIMM.",
 	})
+	ResponseTimeHistogram = promauto.NewHistogramVec(prometheus.HistogramOpts{
+		Namespace: "jimm",
+		Name:      "http",
+		Help:      "request_duration_seconds",
```
babakks: It should be like this:

```
Subsystem: "http",
Name:      "request_duration_seconds",
Help:      "The duration of handling an HTTP request in seconds.",
```

internal/middleware/responsetime.go
```diff
@@ -0,0 +1,21 @@
...
+		duration := time.Since(start)
+		route := r.URL.Path
+		servermon.ResponseTimeHistogram.WithLabelValues(route, r.Method).Observe(duration.Seconds())
```
babakks: Due to high cardinality (i.e., infinitely many possible argument values like IDs in `/rebac` or `/model` endpoints), the request URL path should not be used as a value for a label. We should find a way to extract the parameterized route and use that as the label value.

babakks: As I checked, you can use `chi.RouteContext(r.Context()).RoutePattern()` to get the route path (e.g., `/foo/{id}/bar`), but as per docs, this has to be done after calling the `next.Serve` method. The reason is there might be nested routers and until all of them have processed the request, the actual route pattern is not fully identified. Check out [this](https://github.com/go-chi/chi/issues/692#issuecomment-1002384162) issue in Chi for an example.

Anyway, I think this probably works but you have to test it to make sure:

```go
start := time.Now()
defer func(){
    route := chi.RouteContext(r.Context()).RoutePattern()
    servermon.ResponseTimeHistogram.WithLabelValues(r.URL.Path, r.Method).Observe(time.Since(start))
}()
next.ServeHTTP(w, r)
```

babakks (changes requested):
Thanks for this. The issue is the route label should be populated with a low-cardinality value. This is a limitation with Prometheus, warned against in [here](https://prometheus.io/docs/practices/naming/). I've explained the workaround in the comments.

babakks (dismissed):

internal/middleware/responsetime.go
```diff
@@ -0,0 +1,43 @@
...
+	statusCode int
+}
+
+func (rec *statusRecorder) WriteHeader(statusCode int) {
+	rec.statusCode = statusCode
+	rec.ResponseWriter.WriteHeader(statusCode)
+}
+
+// MeasureResponseTime tracks response time of requests.
+func MeasureResponseTime(next http.Handler) http.Handler {
```
ale8k: MeasureHTTPResponseTime & explain in godoc it doens't track websockets

ale8k (dismissed):
one small comment but lgtm

internal/middleware/responsetime.go
```diff
@@ -0,0 +1,44 @@
...
+	statusCode int
+}
+
+func (rec *statusRecorder) WriteHeader(statusCode int) {
+	rec.statusCode = statusCode
+	rec.ResponseWriter.WriteHeader(statusCode)
+}
+
+// MeasureHTTPResponseTime tracks response time of HTTP requests.
+// We don't track websocket requests.
```
ale8k, line 27: Thanks

ale8k (approved):
tyty

babakks (approved):

PR #1413, opened 2024-10-23
Updates dependencies

ale8k:
Updates dependencies via go get -u ./...

Explicitly sets k8s libs to 0.29 (juju req), the upgrade did find 0.31 to be ok but a package was missing. 29 appears to be ok.  Juju 3.5.4 is on 0.29.0.

Lastly, replaces gopkg.in/square/go-jose.v2 => gopkg.in/go-jose/go-jose.v2

The square version is archived and the go-jose org maintains it now and has a fix.


ale8k:
Closed, having some issues modules wise...

PR #1415, opened 2024-10-23
fix panic when connection to controller fails

SimoneDutto:
## Description
This pr fixes a bug that causes Jimm to panic when it is issued `juju status` on a not-existing model.

> I think this bug is even bigger because Jimm will crash on a failing controller's connection generally.

## How to reproduce the bug
- set a qa environment with make `qa-microk8s`
- remove from the database the model (ex. `make dev-env-cleanup` and `make dev-env` again)
- `juju status`

Jimm will panic. It can't find the model in the database, so it is not able to create the connection. Therefore, it will later try to read Json from a nil connection.

This fix returns the error and adds a test for it.

## Before the fix
image
and juju status hangs

## After the fix
image
juju status will return `ERROR model not found (not found)`

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [ ] Covered by unit tests

### DIFF
--- a/internal/rpc/apiproxy.go
+++ b/internal/rpc/apiproxy.go
@@ -401,6 +401,7 @@ func (p *clientProxy) makeControllerConnection(ctx context.Context) error {
 		connWithMetadata, err := p.createControllerConn(ctx)
 		if err != nil {
 			createConnErr = errors.E(op, err)
+			return
 		}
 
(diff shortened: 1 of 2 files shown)


babakks (approved):

alesstimec (approved):
good find, thanks!

PR #1416, opened 2024-10-24
Remove juju debug-log from test logs when a test fails

SimoneDutto:
## Description

The problem we face is that our tests are cluttered with many juju logs.
Ex when test fail:
Screenshot from 2024-10-24 17-11-28
The reason is that `corejujutesting.JujuConnSuite` is logging a lot of stuff it is doing in `setupSuite`.

The solution is to set the env variable in the test config to increase the connsuite log level to error.

Ex when test fail now:
Screenshot from 2024-10-24 17-10-34

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [ ] Covered by unit tests

### DIFF
--- a/internal/logger/gocheck_logger.go
+++ b/internal/logger/gocheck_logger.go
@@ -1,0 +1,37 @@
+// Copyright 2024 Canonical.
+package logger
+
+import (
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+	gc "gopkg.in/check.v1"
+)
+
+func NewGoCheckLogger(c *gc.C) *zap.Logger {
+	output := gocheckZapWriter{c}
+
+	devConfig := zap.NewDevelopmentEncoderConfig()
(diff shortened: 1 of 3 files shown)


babakks (dismissed):

internal/logger/gocheck_logger.go
kian99, line 1: Should we put this inside `testutils` rather since it's for gocheck? It can fit in either `logger` or `testutils` but I can't imagine it'll ever get used outside of tests.

kian99 (dismissed):

internal/logger/gocheck_logger.go
SimoneDutto: i'd like to have all loggers in the same place, so we consistently update all loggers when we update the logger.
For example, we had a logger for gorm tests defined in a test file and we forgot it existed.

kian99 (approved):

ale8k (approved):

PR #1422, opened 2024-11-06
Return the group name when listing a user's groups.

### LINKED ISSUE #1418: ReBAC: User's group names are ids
In the response for a user's groups: `/identities/{id}/groups` the name value contains the id:

```json
{
    "_links": {
        "next": {
            "href": ""
        }
    },
    "_meta": {
        "pageToken": "",
        "size": 1
    },
    "data": [
        {
            "id": "a084509c-33c8-4ec7-b46c-d0f74958f892",
            "name": "a084509c-33c8-4ec7-b46c-d0f74958f892"
        }
    ],
    "message": "",
    "status": 200
}
```

The same group from the `/groups` response is:

```
{
    "_links": {
        "next": {
            "href": "/rebac/v1/groups?page=1"
        }
    },
    "_meta": {
        "page": 0,
        "size": 1,
        "total": 1
    },
    "data": [
        {
            "id": "a084509c-33c8-4ec7-b46c-d0f74958f892",
            "name": "group1"
        }
    ],
    "message": "",
    "status": 200
}
```

kian99:
## Description

This PR fixes the response from the rebac admin's "GetIdentityGroups" endpoint. Currently we return a group name in place of the group ID. One issue we face is that the group name may no longer exist in the DB but an OpenFGA may still exist. To handle this case I considered:
- Returning an error
- Dropping the offending result
- Returning the group with a name of "removed"

~~I've opted for option 3. The first would prevent the user from viewing their groups and the second would hide the issue.~~
Changed to option 2 to drop the result since it is not useful to the user.

Fixes #1418.

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [x] Covered by unit tests

### DIFF
--- a/internal/jimmhttp/rebac_admin/identities.go
+++ b/internal/jimmhttp/rebac_admin/identities.go
@@ -129,7 +129,7 @@ func (s *identitiesService) GetIdentityGroups(ctx context.Context, identityId st
 		dbGroup, err := s.jimm.GetGroupByUUID(ctx, user, t.Target.ID)
 		if err != nil {
 			// Handle the case where the group was removed from the DB but a lingering OpenFGA tuple still exists.
-			// Don't return an error as that would prevent a user from viewing their groups, instead show the group as "removed".
+			// Don't return an error as that would prevent a user from viewing their groups, instead drop the group from the result.
 			if errors.ErrorCode(err) == errors.CodeNotFound {
 				continue
 			}
--- a/internal/jimmhttp/rebac_admin/identities.go
+++ b/internal/jimmhttp/rebac_admin/identities.go
@@ -124,18 +124,21 @@ func (s *identitiesService) GetIdentityGroups(ctx context.Context, identityId st
 		return nil, err
 	}
 
-	groups := make([]resources.Group, len(tuples))
-	for i, t := range tuples {
+	groups := make([]resources.Group, 0, len(tuples))
+	for _, t := range tuples {
 		dbGroup, err := s.jimm.GetGroupByUUID(ctx, user, t.Target.ID)
 		if err != nil {
 			// Handle the case where the group was removed from the DB but a lingering OpenFGA tuple still exists.
 			// Don't return an error as that would prevent a user from viewing their groups, instead show the group as "removed".
-			dbGroup.Name = "(removed)"
+			if errors.ErrorCode(err) == errors.CodeNotFound {
+				continue
+			}
+			return nil, err
 		}
-		groups[i] = resources.Group{
+		groups = append(groups, resources.Group{
 			Id:   &t.Target.ID,
 			Name: dbGroup.Name,
-		}
+		})
 	}
 
 	originalToken := filter.Token()
--- a/internal/jimmhttp/rebac_admin/identities.go
+++ b/internal/jimmhttp/rebac_admin/identities.go
@@ -120,17 +120,24 @@ func (s *identitiesService) GetIdentityGroups(ctx context.Context, identityId st
 		Relation:     ofganames.MemberRelation.String(),
 		TargetObject: openfga.GroupType.String(),
 	}, int32(filter.Limit()), filter.Token()) // #nosec G115 accept integer conversion
-
 	if err != nil {
 		return nil, err
 	}
+
 	groups := make([]resources.Group, len(tuples))
 	for i, t := range tuples {
+		dbGroup, err := s.jimm.GetGroupByUUID(ctx, user, t.Target.ID)
+		if err != nil {
+			// Handle the case where the group was removed from the DB but a lingering OpenFGA tuple still exists.
+			// Don't return an error as that would prevent a user from viewing their groups, instead show the group as "removed".
+			dbGroup.Name = "(removed)"
+		}
 		groups[i] = resources.Group{
 			Id:   &t.Target.ID,
-			Name: t.Target.ID,
+			Name: dbGroup.Name,
 		}
 	}
+
 	originalToken := filter.Token()
 	return &resources.PaginatedResponse[resources.Group]{
 		Data: groups,
--- a/internal/jimmhttp/rebac_admin/identities_integration_test.go
+++ b/internal/jimmhttp/rebac_admin/identities_integration_test.go
@@ -103,7 +103,8 @@ func (s *identitiesSuite) TestIdentityGetGroups(c *gc.C) {
 		c.Assert(err, gc.IsNil)
 		token = *groups.Next.PageToken
 		for j := 0; j < len(groups.Data); j++ {
-			c.Assert(groups.Data[j].Name, gc.Equals, groupTags[i+j].Id())
+			c.Assert(groups.Data[j].Name, gc.Equals, fmt.Sprintf("group-test%d", i+j))
+			c.Assert(groupTags[j].Id(), gc.Matches, `\w*-\w*-\w*-\w*-\w*`)
 		}
 		if *groups.Next.PageToken == "" {
 			break
--- a/internal/jimmhttp/rebac_admin/identities_integration_test.go
+++ b/internal/jimmhttp/rebac_admin/identities_integration_test.go
@@ -113,11 +113,10 @@ func (s *identitiesSuite) TestIdentityGetGroups(c *gc.C) {
 	}
 }
 
-// TestIdentityGetGroupsWithMissingDbGroup tests the behaviour
+// TestGetIdentityGroupsWithDeletedDbGroup tests the behaviour
 // of GetIdentityGroups when a tuple lingers in OpenFGA but the group
 // has been removed from the database.
-func (s *identitiesSuite) TestIdentityGetGroupsWithMissingDbGroup(c *gc.C) {
-	// initialization
+func (s *identitiesSuite) TestGetIdentityGroupsWithDeletedDbGroup(c *gc.C) {
 	ctx := context.Background()
 	ctx = rebac_handlers.ContextWithIdentity(ctx, s.AdminUser)
 	identitySvc := rebac_admin.NewidentitiesService(s.JIMM)
--- a/internal/jimmhttp/rebac_admin/identities_integration_test.go
+++ b/internal/jimmhttp/rebac_admin/identities_integration_test.go
@@ -10,6 +10,7 @@ import (
 	"github.com/juju/names/v5"
 	gc "gopkg.in/check.v1"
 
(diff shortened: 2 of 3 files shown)


internal/jimmhttp/rebac_admin/identities.go
```diff
@@ -120,17 +120,24 @@ func (s *identitiesService) GetIdentityGroups(ctx context.Context, identityId st
...
 		TargetObject: openfga.GroupType.String(),
 	}, int32(filter.Limit()), filter.Token()) // #nosec G115 accept integer conversion
-
 	if err != nil {
 		return nil, err
 	}
+
 	groups := make([]resources.Group, len(tuples))
 	for i, t := range tuples {
+		dbGroup, err := s.jimm.GetGroupByUUID(ctx, user, t.Target.ID)
```
SimoneDutto, line 129: i think since we already have all the ids we could create a `GetGroupsByUUIDs`.
Otherwise this name resolution could potentially be a performance hit.
What do you think?

internal/jimmhttp/rebac_admin/identities.go
```diff
@@ -120,17 +120,24 @@ func (s *identitiesService) GetIdentityGroups(ctx context.Context, identityId st
...
+		if err != nil {
+			// Handle the case where the group was removed from the DB but a lingering OpenFGA tuple still exists.
+			// Don't return an error as that would prevent a user from viewing their groups, instead show the group as "removed".
+			dbGroup.Name = "(removed)"
```
SimoneDutto: thinking out loud...
if the group is missing from the db it won't be listed in the `/groups | group/{id}` endpoint, so showing the id in this endpoint doesn't make sense, because it can't be resolved if the user clicks on it.

So I would remove these entries missing from postgres, so the user doesn't see "dangling" uuid.
Just a `continue` without returning any error is fine.

kian99: I agree, dropping the result does seem like the better thing to do. I'll wait for another review to chime in and if we agree, I'll change it.

internal/jimmhttp/rebac_admin/identities.go
kian99: It could but I didn't want to optimise prematurely. Curious to hear other opinions. If it becomes an issue we can fix it later? Makes me wonder if we have monitoring on our Rebac-admin endpoints.

SimoneDutto: We have, i've added the middleware monitoring in this pr: https://github.com/canonical/jimm/pull/1412

internal/jimmhttp/rebac_admin/identities.go
kian99: Changed it.

internal/jimmhttp/rebac_admin/identities.go
```diff
@@ -124,18 +124,21 @@ func (s *identitiesService) GetIdentityGroups(ctx context.Context, identityId st
...
 	}
 
-	groups := make([]resources.Group, len(tuples))
-	for i, t := range tuples {
+	groups := make([]resources.Group, 0, len(tuples))
+	for _, t := range tuples {
 		dbGroup, err := s.jimm.GetGroupByUUID(ctx, user, t.Target.ID)
 		if err != nil {
 			// Handle the case where the group was removed from the DB but a lingering OpenFGA tuple still exists.
 			// Don't return an error as that would prevent a user from viewing their groups, instead show the group as "removed".
```
SimoneDutto: you've missed to update the comment

internal/jimmhttp/rebac_admin/identities_integration_test.go
```diff
@@ -103,14 +104,55 @@ func (s *identitiesSuite) TestIdentityGetGroups(c *gc.C) {
...
 			break
 		}
 	}
 }
 
+// TestIdentityGetGroupsWithMissingDbGroup tests the behaviour
+// of GetIdentityGroups when a tuple lingers in OpenFGA but the group
+// has been removed from the database.
+func (s *identitiesSuite) TestIdentityGetGroupsWithMissingDbGroup(c *gc.C) {
+	// initialization
```
ale8k: Don't need this

internal/jimmhttp/rebac_admin/identities_integration_test.go
```diff
@@ -103,14 +104,55 @@ func (s *identitiesSuite) TestIdentityGetGroups(c *gc.C) {
...
 		if *groups.Next.PageToken == "" {
 			break
 		}
 	}
 }
 
+// TestIdentityGetGroupsWithMissingDbGroup tests the behaviour
+// of GetIdentityGroups when a tuple lingers in OpenFGA but the group
+// has been removed from the database.
+func (s *identitiesSuite) TestIdentityGetGroupsWithMissingDbGroup(c *gc.C) {
```
ale8k: TestGetIdentityGroupsWithDeletedDbGroup would be better (just because the method is named wrong)

internal/jimmhttp/rebac_admin/identities.go
```diff
@@ -120,17 +120,27 @@ func (s *identitiesService) GetIdentityGroups(ctx context.Context, identityId st
...
 	}
-	groups := make([]resources.Group, len(tuples))
-	for i, t := range tuples {
-		groups[i] = resources.Group{
-			Id:   &t.Target.ID,
-			Name: t.Target.ID,
+
+	groups := make([]resources.Group, 0, len(tuples))
+	for _, t := range tuples {
+		dbGroup, err := s.jimm.GetGroupByUUID(ctx, user, t.Target.ID)
```
ale8k, line 129: Would be nicer GetGroupsByUUIDs but it doesn't matter much

internal/jimmhttp/rebac_admin/identities.go
```diff
@@ -120,17 +120,27 @@ func (s *identitiesService) GetIdentityGroups(ctx context.Context, identityId st
...
+		if err != nil {
+			// Handle the case where the group was removed from the DB but a lingering OpenFGA tuple still exists.
+			// Don't return an error as that would prevent a user from viewing their groups, instead show the group as "removed".
```
ale8k: How is the group shown as "removed"? We just call .First

internal/jimmhttp/rebac_admin/identities.go
```diff
@@ -120,17 +120,27 @@ func (s *identitiesService) GetIdentityGroups(ctx context.Context, identityId st
...
+			if errors.ErrorCode(err) == errors.CodeNotFound {
```
ale8k, line 133: Should we be leaving lingering tuples? Or would .GetGroupsByUUIDs be better in the JIMM layer with tuple cleanup? (Really not sure here just opening up for discussion)

ale8k (changes requested):

ale8k (changes requested):
Requesting changes on test name and some comments need answering before I'm comfortable to approve.

internal/jimmhttp/rebac_admin/identities.go
alesstimec: since we already have the GetGroupByUUID, it's ok. agree with Kian here.

internal/jimmhttp/rebac_admin/identities.go
alesstimec: i think cleanup could possibly be a topic in its own right. we leave a lot of tuples in openfga that refer to a deleted entity in jimm. i think a spec on this would be very welcome, if anybody has time.

alesstimec (dismissed):
LGTM with comments addressed

internal/jimmhttp/rebac_admin/identities.go
kian99: Comment was outdated, we just drop the group from the result.

internal/jimmhttp/rebac_admin/identities.go
kian99: We shouldn't be leaving lingering tuples, but that's a separate issue. This is just handling the case that, for some reason, there is a leftover tuple but no group in the database.

internal/jimmhttp/rebac_admin/identities_integration_test.go
```diff
@@ -103,14 +104,54 @@ func (s *identitiesSuite) TestIdentityGetGroups(c *gc.C) {
...
+	group2Access.Target = ofganames.ConvertTag(group2)
+
+	err := s.JIMM.OpenFGAClient.AddRelation(ctx, group1Access, group2Access)
+	c.Assert(err, gc.IsNil)
+
+	groups, err := identitySvc.GetIdentityGroups(ctx, username, &resources.GetIdentitiesItemGroupsParams{})
+	c.Assert(err, gc.IsNil)
+	c.Assert(groups.Data, gc.HasLen, 2)
+
+	groupToDelete := dbmodel.GroupEntry{Name: "group2"}
```
alesstimec, line 144: i don't think you need to get group2 again.. you have its tag on L126 (which contains the id).

alesstimec (approved):

internal/jimmhttp/rebac_admin/identities_integration_test.go
kian99: I thought so too, but the delete method requires the ID or UUID.

internal/jimmhttp/rebac_admin/identities.go
ale8k: Not sure what you mean here but perhaps explain in tomorrows standup or offline in element?

internal/jimmhttp/rebac_admin/identities.go
```diff
@@ -120,17 +120,27 @@ func (s *identitiesService) GetIdentityGroups(ctx context.Context, identityId st
...
-		groups[i] = resources.Group{
-			Id:   &t.Target.ID,
-			Name: t.Target.ID,
+
+	groups := make([]resources.Group, 0, len(tuples))
+	for _, t := range tuples {
+		dbGroup, err := s.jimm.GetGroupByUUID(ctx, user, t.Target.ID)
+		if err != nil {
+			// Handle the case where the group was removed from the DB but a lingering OpenFGA tuple still exists.
+			// Don't return an error as that would prevent a user from viewing their groups, instead drop the group from the result.
```
ale8k, line 132: We need to handle the lingering tuples... Should have a discussion on this

ale8k (approved):

ale8k (approved):
lgtm bar our lingering naughty tuple issues

Issue #1421, opened 2024-11-05
ReBAC: Filter params ignored


huwshimi:
Filter params on ReBAC endpoints (`/groups?filter=test`, `/entitlements?filter=test`, `/identities?filter=test`) appear to be ignored.

kian99:
It indeed seems we didn't implement filtering. Do you know if the filtering is supposed to be a fuzzy match or exact match? I'll need to take a look at the OpenAPI schema for the rebac-admin-ui to understand how it's intended to work.

huwshimi:
@kian99 the way it works in the example app in https://github.com/canonical/rebac-admin-ui-handlers is a fuzzy match from memory.

SimoneDutto:
Should be fixed now

PR #1426, opened 2024-11-07
JUJU-7073: add filter on group name/id for listGroups

SimoneDutto:
## Description

This pr adds filtering for rebac `listGroups` on group name/id.

Address #1421 

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [ ] Covered by unit tests

### DIFF
--- a/internal/db/group.go
+++ b/internal/db/group.go
@@ -83,7 +83,7 @@ func (d *Database) GetGroup(ctx context.Context, group *dbmodel.GroupEntry) (err
 }
 
 // ListGroups returns a paginated list of groups defined by limit and offset.
-// match is used to fuzzy find based on entries' name using the LIKE operator (ex. LIKE %<match>%).
+// match is used to fuzzy find based on entries' name or uuid using the LIKE operator (ex. LIKE %<match>%).
 func (d *Database) ListGroups(ctx context.Context, limit, offset int, match string) (_ []dbmodel.GroupEntry, err error) {
 	const op = errors.Op("db.ListGroups")
 	if err := d.ready(); err != nil {
@@ -96,7 +96,7 @@ func (d *Database) ListGroups(ctx context.Context, limit, offset int, match stri
 
 	db := d.DB.WithContext(ctx)
 	if match != "" {
-		db = db.Where("name LIKE ?", "%"+match+"%")
+		db = db.Where("name LIKE ? OR uuid LIKE ?", "%"+match+"%", "%"+match+"%")
 	}
 	db = db.Order("name asc")
 	db = db.Limit(limit)
--- a/internal/db/group.go
+++ b/internal/db/group.go
@@ -85,7 +85,7 @@ func (d *Database) GetGroup(ctx context.Context, group *dbmodel.GroupEntry) (err
 // ForEachGroup iterates through every group calling the given function
 // for each one. If the given function returns an error the iteration
 // will stop immediately and the error will be returned unmodified.
-// `match` will filter with the LIKE operator on id or name.
+// `match` will filter with the LIKE operator on uuid or name.
 func (d *Database) ForEachGroup(ctx context.Context, limit, offset int, match string, f func(*dbmodel.GroupEntry) error) (err error) {
 	const op = errors.Op("db.ForEachGroup")
 	if err := d.ready(); err != nil {
--- a/internal/db/group.go
+++ b/internal/db/group.go
@@ -82,14 +82,13 @@ func (d *Database) GetGroup(ctx context.Context, group *dbmodel.GroupEntry) (err
 	return nil
 }
 
-// ForEachGroup iterates through every group calling the given function
-// for each one. If the given function returns an error the iteration
-// will stop immediately and the error will be returned unmodified.
+// ListGroups returns a paginated list of groups defined by limit and offset.
+// match is used to filter entries based on name.
 // `match` will filter with the LIKE operator on uuid or name.
-func (d *Database) ForEachGroup(ctx context.Context, limit, offset int, match string, f func(*dbmodel.GroupEntry) error) (err error) {
+func (d *Database) ListGroups(ctx context.Context, limit, offset int, match string) (_ []dbmodel.GroupEntry, err error) {
 	const op = errors.Op("db.ForEachGroup")
 	if err := d.ready(); err != nil {
-		return errors.E(op, err)
+		return nil, errors.E(op, err)
 	}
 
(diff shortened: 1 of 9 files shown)


internal/db/group.go
```diff
@@ -96,6 +97,10 @@ func (d *Database) ForEachGroup(ctx context.Context, limit, offset int, f func(*
 	defer servermon.ErrorCounter(servermon.DBQueryErrorCount, &err, string(op))
 
 	db := d.DB.WithContext(ctx)
+	if match != "" {
+		match = "%" + match + "%"
+		db = db.Where("uuid LIKE ? OR name LIKE ?", match, match)
```
ale8k: Why not:
```go
db = db.Where("uuid = ? OR name = ?", match, match)
```
Because in theory I could just use % and get everything no?

I see Hue said fuzzy but I don't understand why we'd fuzzy match it. Also, names are unique so there's no reason to iterate over the group table when we can select where name = ?

ale8k (changes requested):
Not sure I'm following why we're doing this. Requesting changes and asking for explanation.

internal/db/group.go
SimoneDutto: The reason is that they usually have a text box where they can start writing a name or id, and the list is filtered upon this fuzzy search, from the webui.
If we provide only an exact match it doesn't make sense because they would need to know the full name in advance, so it's not a listing operation anymore, but just a `GET group/{id}`

SimoneDutto: as discussed in a call, it's fine as it is. 
We need to check if they need fuzzy find on uuid as well

internal/db/group.go
```diff
@@ -96,6 +97,10 @@ func (d *Database) ForEachGroup(ctx context.Context, limit, offset int, f func(*
 	defer servermon.ErrorCounter(servermon.DBQueryErrorCount, &err, string(op))
 
 	db := d.DB.WithContext(ctx)
+	if match != "" {
+		match = "%" + match + "%"
```
alesstimec: perhaps assert that the `match` string does not have a `%` prefix or suffix to start with

alesstimec: what is the allowed format of the `match` string?

kian99 (approved):
lgtm

internal/db/group.go
SimoneDutto: it's just a string coming from a route argument

SimoneDutto: i've added a test for match containing `%`

internal/db/group.go
```diff
@@ -82,41 +82,30 @@ func (d *Database) GetGroup(ctx context.Context, group *dbmodel.GroupEntry) (err
 	return nil
 }
 
-// ForEachGroup iterates through every group calling the given function
-// for each one. If the given function returns an error the iteration
-// will stop immediately and the error will be returned unmodified.
-func (d *Database) ForEachGroup(ctx context.Context, limit, offset int, f func(*dbmodel.GroupEntry) error) (err error) {
+// ListGroups returns a paginated list of groups defined by limit and offset.
+// match is used to filter entries based on name.
```
alesstimec: could we please add the expected format of match (is it a regexp?), or perhaps just provide examples

internal/db/group.go
```diff
@@ -82,41 +82,30 @@ func (d *Database) GetGroup(ctx context.Context, group *dbmodel.GroupEntry) (err
...
+func (d *Database) ListGroups(ctx context.Context, limit, offset int, match string) (_ []dbmodel.GroupEntry, err error) {
 	const op = errors.Op("db.ForEachGroup")
```
alesstimec: s/ForEachGroup/ListGroups/

internal/jimm/access.go
```diff
@@ -790,18 +790,15 @@ func (j *JIMM) RemoveGroup(ctx context.Context, user *openfga.User, name string)
 }
 
 // ListGroups returns a list of groups known to JIMM.
-func (j *JIMM) ListGroups(ctx context.Context, user *openfga.User, filter pagination.LimitOffsetPagination) ([]dbmodel.GroupEntry, error) {
+// `match` will filter the list for name or uuid matching it.
+func (j *JIMM) ListGroups(ctx context.Context, user *openfga.User, filter pagination.LimitOffsetPagination, match string) ([]dbmodel.GroupEntry, error) {
```
alesstimec: could we rename `filter` to `pagination`? otherwise i'm likely to confuse `match` and `filter`.. `match` is used for filtering, but `filter` is used for pagination..

alesstimec (changes requested):

kian99 (commented):
Looks good with Ales' comments addressed

internal/db/group.go
SimoneDutto: updated godoc. I think it's clear now with the LIKE operator, if you'd like examples as well I can add them

internal/jimm/access.go
SimoneDutto: yeah, i've address this in the last commit to make it easier to review

ale8k (approved):

internal/jimmhttp/rebac_admin/groups.go
```diff
@@ -41,7 +41,11 @@ func (s *groupsService) ListGroups(ctx context.Context, params *resources.GetGro
 		return nil, err
 	}
 	page, nextPage, pagination := pagination.CreatePagination(params.Size, params.Page, count)
-	groups, err := s.jimm.ListGroups(ctx, user, pagination)
+	match := ""
```
alesstimec, line 44: maybe s/match/filter/

internal/testutils/jimmtest/mocks/jimm_group_mock.go
```diff
@@ -55,11 +55,11 @@ func (j *GroupService) GetGroupByName(ctx context.Context, user *openfga.User, n
 	return j.GetGroupByName_(ctx, user, name)
 }
 
-func (j *GroupService) ListGroups(ctx context.Context, user *openfga.User, filters pagination.LimitOffsetPagination) ([]dbmodel.GroupEntry, error) {
+func (j *GroupService) ListGroups(ctx context.Context, user *openfga.User, pagination pagination.LimitOffsetPagination, match string) ([]dbmodel.GroupEntry, error) {
```
alesstimec, line 58: some places we call it `filter`, other places `match`. we should decide on one of those and use that term everywhere.

alesstimec (approved):
LGTM with a nitpick

PR #1428, opened 2024-11-08
fix flaky test by checking error properly

SimoneDutto:
## Description

We had a flaky test that in some cases where the test runner was under pressure failed.
ex. 
- https://github.com/canonical/jimm/actions/runs/11509775185/job/32040489299
- https://github.com/canonical/jimm/actions/runs/11570348073/job/32205943025

The reason is that:
we were using `websocket.IsCloseError(err)` wrong and we were returning `internal error` for `close errors`.
Here is the godoc:
```go
// IsCloseError returns boolean indicating whether the error is a *CloseError
// with one of the specified codes.
func IsCloseError(err error, codes ...int) bool
```

## QA steps
To reproduce the bug locally just run the same test many times
image

After the fix:
image

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [ ] Covered by unit tests

### DIFF
--- a/internal/rpc/client_test.go
+++ b/internal/rpc/client_test.go
@@ -518,12 +518,12 @@ func handleWS(f func(*websocket.Conn) error) http.Handler {
 		defer c.Close()
 		err = f(c)
 		var cm []byte
+		closeError, isCloseError := err.(*websocket.CloseError)
 		switch {
 		case err == nil:
 			cm = websocket.FormatCloseMessage(websocket.CloseNormalClosure, "")
-		case websocket.IsCloseError(err):
-			ce := err.(*websocket.CloseError)
-			cm = websocket.FormatCloseMessage(ce.Code, ce.Text)
+		case isCloseError:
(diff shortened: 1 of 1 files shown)


ale8k (approved):
nice, thanks for this

ale8k (approved):

internal/rpc/client_test.go
```diff
@@ -518,12 +518,12 @@ func handleWS(f func(*websocket.Conn) error) http.Handler {
 		defer c.Close()
 		err = f(c)
 		var cm []byte
+		closeError, isCloseError := err.(*websocket.CloseError)
```
alesstimec, line 521: maybe instead of this you can user `errors.Is`?

SimoneDutto: I need to cast the type in anycase to return the format the close message in `websocket.FormatCloseMessage(closeError.Code, closeError.Text)`

alesstimec (approved):

PR #1429, opened 2024-11-08
Support update cloud

ale8k:
I noticed our UpdateCloud wasn't implemented but we had the service layer implementation.

I've plugged it in, and I'm not sure why Martin thought this wouldn't work but it appears fine? Is there any issues plugging this in?

## Description

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [x] Covered by unit tests

### DIFF
--- a/internal/jimm/cloud.go
+++ b/internal/jimm/cloud.go
@@ -600,24 +600,27 @@ func (j *JIMM) UpdateCloud(ctx context.Context, user *openfga.User, ct names.Clo
 	// do this in a transaction so that the local view cannot finish in
 	// an inconsistent state.
 	err = j.Database.Transaction(func(db *db.Database) error {
+		var transactionCloud dbmodel.Cloud
+		transactionCloud.SetTag(ct)
 
-		var c dbmodel.Cloud
-		c.SetTag(ct)
-		if err := db.GetCloud(ctx, &c); err != nil {
+		if err := db.GetCloud(ctx, &transactionCloud); err != nil {
 			return err
 		}
-		c.FromJujuCloud(cloud)
-		for i := range c.Regions {
-			if len(c.Regions[i].Controllers) == 0 {
-				for _, ctl := range controllers {
-					c.Regions[i].Controllers = append(c.Regions[i].Controllers, dbmodel.CloudRegionControllerPriority{
-						Controller: ctl,
-						Priority:   dbmodel.CloudRegionControllerPrioritySupported,
-					})
-				}
-			}
+		transactionCloud.FromJujuCloud(cloud)
+
+		// Set the region controller priorities again for this transactions dbmodel.Cloud.
+		for i := range transactionCloud.Regions {
+			transactionCloud.Regions[i].Controllers = c.Regions[i].Controllers
+			// if len(transactionCloud.Regions[i].Controllers) == 0 {
+			// 	for _, ctl := range controllers {
+			// 		transactionCloud.Regions[i].Controllers = append(transactionCloud.Regions[i].Controllers, dbmodel.CloudRegionControllerPriority{
+			// 			Controller: ctl,
+			// 			Priority:   dbmodel.CloudRegionControllerPrioritySupported,
+			// 		})
+			// 	}
+			// }
 		}
-		return db.UpdateCloud(ctx, &c)
+		return db.UpdateCloud(ctx, &transactionCloud)
 	})
 
 	if err != nil {
--- a/internal/jujuapi/cloud.go
+++ b/internal/jujuapi/cloud.go
@@ -483,11 +483,13 @@ func (r *controllerRoot) UpdateCloud(ctx context.Context, args jujuparams.Update
 	}
 
 	for i, cloud := range args.Clouds {
-		ct, err := names.ParseCloudTag("cloud-" + cloud.Name)
-		if err != nil {
-			results.Results[i].Error = mapError(err)
+		if !names.IsValidCloud(cloud.Name) {
+			results.Results[i].Error = mapError(errors.E(errors.CodeBadRequest, "invalid cloud name"))
+			continue
 		}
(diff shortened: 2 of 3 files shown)


internal/jimm/cloud.go
```diff
@@ -600,24 +600,27 @@ func (j *JIMM) UpdateCloud(ctx context.Context, user *openfga.User, ct names.Clo
...
-						Priority:   dbmodel.CloudRegionControllerPrioritySupported,
-					})
-				}
-			}
+		transactionCloud.FromJujuCloud(cloud)
+
+		// Set the region controller priorities again for this transactions dbmodel.Cloud.
+		for i := range transactionCloud.Regions {
+			transactionCloud.Regions[i].Controllers = c.Regions[i].Controllers
+			// if len(transactionCloud.Regions[i].Controllers) == 0 {
```
ale8k, line 614: I've switched to this to simplify this code, but left it commented for reviews. I think this is OK and not sure why it was looping over again to set the region priority when they could've just used the initially retrieved priority. Any idea why? Or perhaps it was an oversight?

internal/jujuapi/cloud.go
```diff
@@ -481,21 +481,20 @@ func (r *controllerRoot) UpdateCloud(ctx context.Context, args jujuparams.Update
 	results := jujuparams.ErrorResults{
 		Results: make([]jujuparams.ErrorResult, len(args.Clouds)),
 	}
-	for i := range args.Clouds {
-		err := r.updateCloud()
+
+	for i, cloud := range args.Clouds {
+		ct, err := names.ParseCloudTag("cloud-" + cloud.Name)
```
ale8k: I always forget the best way of creating tags, is this it? Using "cloud-" isn't amazing but I can't see an alternate way.

internal/jimm/cloud.go
```diff
@@ -600,24 +600,27 @@ func (j *JIMM) UpdateCloud(ctx context.Context, user *openfga.User, ct names.Clo
...
-					c.Regions[i].Controllers = append(c.Regions[i].Controllers, dbmodel.CloudRegionControllerPriority{
-						Controller: ctl,
-						Priority:   dbmodel.CloudRegionControllerPrioritySupported,
-					})
-				}
-			}
+		transactionCloud.FromJujuCloud(cloud)
+
+		// Set the region controller priorities again for this transactions dbmodel.Cloud.
+		for i := range transactionCloud.Regions {
```
SimoneDutto, line 612: it seems to me that the new code is doing a different thing.
The code before was checking the region was empty and add to that region all the controllers from the previous controller definition.
Now your code is always overriding the controllers, not sure if this is a issue or not.

SimoneDutto: Not sure you need a transaction for this operation though, i think you can do that in a single "replace regions"

internal/jujuapi/cloud.go
SimoneDutto: it should be `NewCloudTag`

internal/jujuapi/cloud_test.go
```diff
@@ -872,18 +872,53 @@ func (s *cloudSuite) TestModifyCloudAccessUnauthorized(c *gc.C) {
...
+	})
+	c.Assert(err, gc.IsNil)
+
+	info, err := client.CloudInfo([]names.CloudTag{names.NewCloudTag("test-cloud")})
+	c.Assert(err, gc.IsNil)
+
+	c.Assert(info, gc.HasLen, 1)
+	c.Assert(info[0].Cloud.StorageEndpoint, gc.Equals, newStorageEndpointValue)
+
+	// Now attempt to update the name erroneously
```
SimoneDutto, line 911: if there aren't any others you should create some tests for `updateCloud`, not just the error case

internal/jimm/cloud.go
ale8k: Yeah I'm not sure either really, it's not clear why the == 0 check was there. And on the transaction, I think it's because we're retrieving the cloud first and in theory the cloud could change between the time you first retrieve it and then set it. So I think a transaction makes sense here.

internal/jujuapi/cloud.go
ale8k: NewCloudTag panics, Parse calls new underneath. We **generally** don't call new ever.

internal/jujuapi/cloud_test.go
ale8k: What do you mean? There is a test above?

SimoneDutto: I could be wrong but with these tests you are not hitting the for loop with the controllers update

ale8k: We are, it's using jujuconnsuite so just the one controller, not many.

SimoneDutto: i'd add a test to make it clear we've changed the behaviour now to always override, so if someone (like us now) is confused and change the code, the test will break

internal/jujuapi/cloud.go
SimoneDutto: what about `cloud.Tag()`?

internal/jimm/cloud.go
SimoneDutto: the thing is that you are not using the `transactionCloud`, you are just overriding the regions field, so you can do that w/o using a transaction. Right?

internal/jujuapi/cloud_test.go
ale8k: That's the thing I'm not really sure why he was doing that, and can't see why?

internal/jimm/cloud.go
ale8k: Ahhh, I see what you mean. Perhaps it should be getting the cloud again, good catch.

ale8k:
Closing PR until UX is resolved.

PR #1432, opened 2024-11-12
import offers during model import

kian99:
## Description

This PR does two things:
1. Refactors the ImportModel method. I started getting linting errors on cognitive complexity when trying to add functionality so I took the time to refactor the method.
2. Adds functionality to the ImportModel method to import a model's application-offers.

**Please Note:** The first commit contains only the first change and the second commit contains the second change.

One issue that remains is how to handle offer URLs that already exist. This will be tackled in a separate PR.

Partially fixes [JUJU-7011](https://warthogs.atlassian.net/browse/JUJU-7011)

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [x] Covered by unit tests

[JUJU-7011]: https://warthogs.atlassian.net/browse/JUJU-7011?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/dbmodel/model.go
+++ b/internal/dbmodel/model.go
@@ -103,7 +103,7 @@ func (m *Model) SetTag(t names.ModelTag) {
 	m.UUID.Valid = true
 }
 
-// FromModelUpdate updates the model from the given ModelUpdate.
+// SwitchOwner updates the model owner.
 func (m *Model) SwitchOwner(u *Identity) {
 	m.OwnerIdentityName = u.Name
 	m.Owner = *u
--- a/internal/dbmodel/model.go
+++ b/internal/dbmodel/model.go
@@ -103,8 +103,8 @@ func (m *Model) SetTag(t names.ModelTag) {
 	m.UUID.Valid = true
 }
 
-// SwitchOwner updates the model owner.
-func (m *Model) SwitchOwner(u *Identity) {
+// SetOwner updates the model owner.
+func (m *Model) SetOwner(u *Identity) {
 	m.OwnerIdentityName = u.Name
 	m.Owner = *u
 }
--- a/internal/jimm/controller.go
+++ b/internal/jimm/controller.go
@@ -398,6 +398,7 @@ type modelImporter struct {
 	model         dbmodel.Model
 	modelInfo     jujuparams.ModelInfo
 	originalOwner names.UserTag
+	offersToAdd   []jujuparams.ApplicationOfferAdminDetailsV5
 }
 
 func newModelImporter(jimm *JIMM) modelImporter {
@@ -429,6 +430,16 @@ func (m *modelImporter) fetchModelInfo(ctx context.Context, controllerName strin
 	m.originalOwner, err = names.ParseUserTag(m.modelInfo.OwnerTag)
 	if err != nil {
 		return errors.E(fmt.Sprintf("invalid username %s from original model owner", m.modelInfo.OwnerTag))
+	}
+
+	m.offersToAdd, err = api.ListApplicationOffers(ctx, []jujuparams.OfferFilter{
+		{
+			OwnerName: m.originalOwner.Id(),
+			ModelName: m.modelInfo.Name,
+		},
+	})
+	if err != nil {
+		return err
 	}
 
 	// fill in data from model info
@@ -476,6 +487,13 @@ func (m *modelImporter) addPermissions(ctx context.Context) error {
 
 	if err := m.jimm.addModelPermissions(ctx, ofgaUser, m.model.ResourceTag(), controllerTag); err != nil {
 		return err
+	}
+
+	for _, offer := range m.offersToAdd {
+		err := m.jimm.OpenFGAClient.AddModelApplicationOffer(ctx, m.model.ResourceTag(), names.NewApplicationOfferTag(offer.OfferUUID))
+		if err != nil {
+			return err
+		}
 	}
 
 	return nil
@@ -540,6 +558,17 @@ func (m *modelImporter) save(ctx context.Context) error {
 				return fmt.Errorf("model (%s) already exists", m.model.Name)
 			}
 			return err
+		}
+		for _, offer := range m.offersToAdd {
+			var dbOffer dbmodel.ApplicationOffer
+			dbOffer.FromJujuApplicationOfferAdminDetailsV5(offer)
+			dbOffer.ModelID = m.model.ID
+			if err := m.jimm.Database.AddApplicationOffer(ctx, &dbOffer); err != nil {
+				if errors.ErrorCode(err) == errors.CodeAlreadyExists {
+					return fmt.Errorf("offer with URL %s already exists", offer.OfferURL)
+				}
+				return err
+			}
 		}
 		return nil
 	})
--- a/internal/jimm/controller.go
+++ b/internal/jimm/controller.go
@@ -487,6 +487,8 @@ func (m *modelImporter) setModelOwner(ctx context.Context) error {
 	return nil
 }
 
+// addPermissions grants the model owner with admin access to the model
+// and, in turn, admin access to any offers within the model.
 func (m *modelImporter) addPermissions(ctx context.Context) error {
 	// Note that only the new owner is given access. All previous users that had access according to Juju
 	// are discarded as access must now be governed by JIMM and OpenFGA.
--- a/internal/jimm/controller.go
+++ b/internal/jimm/controller.go
@@ -394,17 +394,28 @@ func (j *JIMM) GetUserControllerAccess(ctx context.Context, user *openfga.User, 
 }
 
 type modelImporter struct {
-	jimm          *JIMM
-	model         dbmodel.Model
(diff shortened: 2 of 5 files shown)


internal/jimm/controller.go
```diff
@@ -505,11 +514,28 @@ func (j *JIMM) ImportModel(ctx context.Context, user *openfga.User, controllerNa
...
+			if err := j.Database.AddApplicationOffer(ctx, &dbOffer); err != nil {
+				if errors.ErrorCode(err) == errors.CodeAlreadyExists {
+					return errors.E(op, err, fmt.Sprintf("offer with URL %s already exists", offer.OfferURL))
+				}
+				return errors.E(op, err)
+			}
 		}
+		return nil
+	})
+	if err != nil {
```
alesstimec, line 597: question: we run handleModelDeltas on L516, which will connect to the controller, create an all watcher and process a single set of deltas.. is that the intent?

internal/jimm/controller.go
```diff
@@ -505,11 +514,28 @@ func (j *JIMM) ImportModel(ctx context.Context, user *openfga.User, controllerNa
...
+	err = j.Database.Transaction(func(d *db.Database) error {
+		err := j.Database.AddModel(ctx, &model)
+		if err != nil {
+			if errors.ErrorCode(err) == errors.CodeAlreadyExists {
+				return errors.E(op, err, "model already exists")
+			}
+			return errors.E(op, err)
+		}
+		for _, offer := range offersToAdd {
+			var dbOffer dbmodel.ApplicationOffer
```
alesstimec, line 573: while this will add offers to JIMM's database, it will not grant anybody access to imported offers. At the very least, the user importing the model should be granted access.

this will require a bit of thought: each offer comes with a set of users that have access to the offer. If those are external users (those that JIMM cares about), they should be granted access to the offer when it is imported. But i guess in most cases these users would be local users - access to the offer will remain granted on the controller, but JIMM will not know/care about those rights.  perhaps we should simply print this info and let the user importing the model handle it. e.g. we imported this model and these app offer and these are the users that have access to the app offer - handle it!

alesstimec (changes requested):
We need to discuss how to set access rights to imported application offers.

internal/jimm/controller.go
kian99: I can't say why it's like that, but that does seem intentional and has been that way for some time, the tests also verify the delta functionality.
The tests show that the deltas are used to update some properties of the model like the status fields, etc.

internal/jimm/controller.go
kian99: True, I think we need to create an OpenFGA relation between the offers and the model. That way the model owner will automatically get access. Then they can grant access to other users. The users on the controller should never (at least with the new JIMM) be external users, because nowadays those are never stored on Juju's end (only in JIMM) so I think we can safely discard the users coming from the controller.

But yes printing a message would be helpful.

SimoneDutto (approved):
lgtm

internal/jimm/controller.go
```diff
@@ -393,133 +393,231 @@ func (j *JIMM) GetUserControllerAccess(ctx context.Context, user *openfga.User,
...
+	jimm          *JIMM
+	model         dbmodel.Model
+	modelInfo     jujuparams.ModelInfo
+	originalOwner names.UserTag
+	offersToAdd   []jujuparams.ApplicationOfferAdminDetailsV5
+}
 
-	if err := j.checkJimmAdmin(user); err != nil {
-		return err
+func newModelImporter(jimm *JIMM) modelImporter {
```
alesstimec: pass in the new owner name so that you can immediately return an error if we fail to parse the owner tag or owner is a local user
```
func newModelImporter(jimm *JIMM, ownerTag names.UserTag) (error, *modelImporter)

File: internal/dbmodel/model.go [hunk 4]
```diff
@@ -103,7 +103,7 @@ func (m *Model) SetTag(t names.ModelTag) {
 	m.UUID.Valid = true
 }
 
-// FromModelUpdate updates the model from the given ModelUpdate.
+// SwitchOwner updates the model owner.
```
alesstimec: nitpick: SetOwner seems a more appropriate name

internal/jimm/controller.go
```diff
@@ -393,133 +393,231 @@ func (j *JIMM) GetUserControllerAccess(ctx context.Context, user *openfga.User,
...
+		return errors.E(op, err)
+	}
+
+	if err := importer.addPermissions(ctx); err != nil {
+		return errors.E(op, err)
+	}
+
+	// TODO(CSS-5458): Remove the below section on cloud credentials once we no longer persist the relation between
+	// cloud credentials and models.
+	// Update: We need to investigate this further, if a user updates their cloud-credential it will update the credential
```
alesstimec, line 615: hm. this might be a bit confusing. if a user updates their cloud-credential, sure, jimm will try and update the cloud credential on the controller responsible for this model, but the actual credential used by the model will not be affected. the user would need to use the `juju set-credential` command to set credentials first and then update them to get the actual model cloud credentials updated.

internal/jimm/controller.go
ale8k: +1, relation makes sense to me.

internal/jimm/controller.go
```diff
@@ -393,133 +393,231 @@ func (j *JIMM) GetUserControllerAccess(ctx context.Context, user *openfga.User,
...
+					return fmt.Errorf("offer with URL %s already exists", offer.OfferURL)
+				}
+				return err
+			}
+		}
+		return nil
+	})
+}
+
+// ImportModel imports model with the specified UUID from the controller.
```
ale8k: imports a model and the currently existing offers into JIMM. A new owner must be set to represent the external user who will own this model.

Something like this for the comment?

ale8k (approved):
lgtm

internal/jimm/controller.go
ale8k: Can do in a follow up PR @alesstimec

internal/jimm/controller.go
kian99: Yes this will need fixing I think. For a follow up PR.

internal/jimm/controller.go
```diff
@@ -393,133 +393,244 @@ func (j *JIMM) GetUserControllerAccess(ctx context.Context, user *openfga.User,
...
+	err = api.ModelInfo(ctx, &m.modelInfo)
 	if err != nil {
-		return errors.E(op, err)
+		return err
 	}
-	model := dbmodel.Model{}
+
+	m.originalOwner, err = names.ParseUserTag(m.modelInfo.OwnerTag)
+	if err != nil {
+		return errors.E(fmt.Sprintf("invalid username %s from original model owner", m.modelInfo.OwnerTag))
```
alesstimec, line 443: nothing to fix, just to say that it would be very unusual to face this error

internal/jimm/controller.go
```diff
@@ -393,133 +393,244 @@ func (j *JIMM) GetUserControllerAccess(ctx context.Context, user *openfga.User,
...
+	modelImporter := modelImporter{
+		jimm: jimm,
+	}
+	if newOwner == "" {
+		return modelImporter, nil
+	}
+	if !names.IsValidUser(newOwner) {
+		return modelImporter, errors.E(errors.CodeBadRequest, "invalid new username for new model owner")
 	}
+	newOwnerTag := names.NewUserTag(newOwner)
```
alesstimec, line 416: can we also check if this is a local user here?

internal/jimm/controller.go
```diff
@@ -393,133 +393,244 @@ func (j *JIMM) GetUserControllerAccess(ctx context.Context, user *openfga.User,
...
-	if err := j.addModelPermissions(ctx, ofgaUser, modelTag, controllerTag); err != nil {
-		return errors.E(op, err)
+	if err := m.jimm.addModelPermissions(ctx, ofgaUser, m.model.ResourceTag(), controllerTag); err != nil {
+		return err
 	}
 
-	// TODO(CSS-5458): Remove the below section on cloud credentials once we no longer persist the relation between
-	// cloud credentials and models
+	for _, offer := range m.offersToAdd {
+		err := m.jimm.OpenFGAClient.AddModelApplicationOffer(ctx, m.model.ResourceTag(), names.NewApplicationOfferTag(offer.OfferUUID))
```
alesstimec, line 503: maybe just a short godoc explaining that by associating offer with a model we are giving the model owner admin access to the offer

alesstimec (approved):
LGTM with a few suggestions

internal/jimm/controller.go
kian99: Yeah...

internal/jimm/controller.go
kian99: We already do the check when we switch the owner, so I think it's okay like this.

Issue #1419, opened 2024-11-05
ReBAC: Updating user's entitlements fails silently


huwshimi:
When updating a user's entitlements: PATCH `/identities/{id}/entitlements` there is no response body and no errors are returned.

For example the following gets a 200 OK response but does not create the relation:

```json
{
  "patches": [
    {
      "entitlement": {
        "entity_type": "cloud",
        "entitlement": "administrator",
        "entity_id": "localhost"
      },
      "op": "add"
    }
  ]
}
```

But this succeeds:

```
{
  "patches": [
    {
      "entitlement": {
        "entity_type": "controller",
        "entitlement": "administrator",
        "entity_id": "d61369f7-7953-4a99-870a-4d046038fddc"
      },
      "op": "add"
    }
  ]
}
```

kian99:
I've made a test for this but I can't seem to replicate it.
The test is an integration test but calls the handler directly in code (I haven't tried an actual HTTP call). 
But it seems to work, I grant a user access to a "test-cloud" and then verify they have the expected access.

Why do you suggest the first block doesn't actually create the relation? Is it not showing up in one of the other calls?

huwshimi:
Ah interesting, if I try and create that same relation again it says it already exists, so that call is succeeding.

However, the created relation is not included in the entitlements call e.g. `/identities/{id}/entitlements`.

```json
{
    "_links": {
        "next": {
            "href": "/rebac/v1/identities/jimm-test@canonical.com/entitlements?nextToken=eyJraW5kIjoiY2xvdWQiLCJ0b2tlbiI6IiJ9"
        }
    },
    "_meta": {
        "pageToken": "",
        "size": 1
    },
    "data": [
        {
            "entitlement": "administrator",
            "entity_id": "3217dbc9-8ea9-4381-9e97-01eab0b3f6bb",
            "entity_type": "controller"
        }
    ],
    "message": "",
    "status": 200
}
```

kian99:
Digging into this further, I can see why this is happening. In our implementation for pagination of OpenFGA results, we realised it is not trivial to fetch all the relations of a user. What OpenFGA offers is the ability to fetch the relations for a specific user + specific object. So "user" + "cloud" or "user" + "controller".

You'll notice the response from `/identities/{id}/entitlements` includes a next page with a "nextToken", in our implementation, we page over the combinations of "user" + <resource> and so a page can be empty but the next page might not be. 

[...]

PR #1435, opened 2024-11-14
Juju 7058/fix pagination for entitlement

SimoneDutto:
## Description

In this PR we change the way we return pages of entries from OpenFGA.
Before, we returned a page containing only one entity kind, which caused confusion to users because if they asked for a `pageSize` 10 and we returned a `pageSize` 5 with a continuation token, they would assume there weren't more entries.

Now we try to fill the page of `pageSize` entries, independently of entity kind.

> the second-page fetch of `pageSize` is to ensure that if we return a continuation token, there are at least one more entry in openfga.
 

Address #1419 

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [ ] Covered by unit tests

### DIFF
--- a/internal/jimm/relation.go
+++ b/internal/jimm/relation.go
@@ -110,29 +110,58 @@ func (j *JIMM) ListObjectRelations(ctx context.Context, user *openfga.User, obje
 	if !user.JimmAdmin {
 		return nil, e, errors.E(op, errors.CodeUnauthorized, "unauthorized")
 	}
-	continuationToken, kind, err := pagination.DecodeEntitlementToken(entitlementToken)
+	responseTuples, nextToken, err := j.getObjectRelationsPage(ctx, object, pageSize, entitlementToken)
 	if err != nil {
-		return nil, e, err
+		return nil, e, errors.E(op, err)
 	}
+	// Verify next page contain some entries. Otherwise return empty nextToken
+	if len(responseTuples) == int(pageSize) && nextToken.String() != "" {
+		responseTuples, _, err := j.getObjectRelationsPage(ctx, object, 1, nextToken)
+		if err != nil {
+			return nil, e, errors.E(op, "error getting next page to verify it cointains something", err)
+		}
+		if len(responseTuples) == 0 {
+			nextToken = pagination.EntitlementToken{}
+		}
+	}
+	return responseTuples, nextToken, nil
+}
+
+func (j *JIMM) getObjectRelationsPage(ctx context.Context, object string, pageSize int32, entitlementToken pagination.EntitlementToken) ([]openfga.Tuple, pagination.EntitlementToken, error) {
+	var err error
+	var e pagination.EntitlementToken
 	tuple := &openfga.Tuple{}
 	tuple.Object, err = j.parseAndValidateTag(ctx, object)
 	if err != nil {
 		return nil, e, err
 	}
-	tuple.Target, err = j.parseAndValidateTag(ctx, kind.String())
-	if err != nil {
-		return nil, e, err
+	var responseTuples []openfga.Tuple
+	nextToken := entitlementToken
+	for {
+		nextContinuationToken, kind, err := pagination.DecodeEntitlementToken(nextToken)
+		if err != nil {
+			return nil, e, err
+		}
+		tuple.Target, err = j.parseAndValidateTag(ctx, kind.String())
+		if err != nil {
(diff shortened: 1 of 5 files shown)


internal/jimm/relation.go
```diff
@@ -110,29 +110,60 @@ func (j *JIMM) ListObjectRelations(ctx context.Context, user *openfga.User, obje
 	if !user.JimmAdmin {
 		return nil, e, errors.E(op, errors.CodeUnauthorized, "unauthorized")
 	}
-	continuationToken, kind, err := pagination.DecodeEntitlementToken(entitlementToken)
+	responseTuples, nextToken, err := j.getObjectRelationsPage(ctx, object, pageSize, entitlementToken)
 	if err != nil {
-		return nil, e, err
+		return nil, e, errors.E(op, err)
 	}
+	// berify next page contains some entries. Otherwise return empty nextToken.
```
kian99:
```suggestion
	// verify next page contains some entries. Otherwise return empty nextToken.
```

internal/jimm/relation_test.go
```diff
@@ -222,17 +243,29 @@ func TestListObjectRelations(t *testing.T) {
...
 		expectedTuples []ExpectedTuple
 	}{
 		{
-			description:    "test listing all relations",
+			description:    "test listing all relations in single page",
 			object:         user.Tag().String(),
-			expectedLength: 3,
+			pageSize:       10,
+			expectNumPages: 1,
+			expectedLength: 7,
```
kian99: Maybe rename `expectedLength` to `expectedTuples`?

kian99 (approved):
lgtm

internal/jimm/relation.go
```diff
@@ -110,29 +110,60 @@ func (j *JIMM) ListObjectRelations(ctx context.Context, user *openfga.User, obje
...
 		return nil, e, errors.E(op, errors.CodeUnauthorized, "unauthorized")
 	}
-	continuationToken, kind, err := pagination.DecodeEntitlementToken(entitlementToken)
+	responseTuples, nextToken, err := j.getObjectRelationsPage(ctx, object, pageSize, entitlementToken)
 	if err != nil {
-		return nil, e, err
+		return nil, e, errors.E(op, err)
 	}
+	// verify next page contains some entries. Otherwise return empty nextToken.
+	if len(responseTuples) == int(pageSize) && nextToken.String() != "" {
```
alesstimec, line 118: should we care if the next page contains entries? or just keep returning if openfga returns a continuation token.. also i think the correct way would be to 
```
if len(responseTuples) > 0 && nextToken.String() != "" {}
```

internal/jimm/relation.go
```diff
@@ -110,29 +110,60 @@ func (j *JIMM) ListObjectRelations(ctx context.Context, user *openfga.User, obje
...
+		responseTuples, _, err := j.getObjectRelationsPage(ctx, object, 1, nextToken)
+		if err != nil {
+			return nil, e, errors.E(op, "error getting next page to verify it cointains something", err)
+		}
+		if len(responseTuples) == 0 {
```
alesstimec, line 123: should we be making this call? or do we trust openfga to know when to stop, by giving us an empty continuation token?

(Simone: can i edit someone else's comment)

alesstimec (approved):
LGTM with a few questions

internal/jimm/relation.go
SimoneDutto: the problem is:
- we have a pagination token for each entity kind (user, cloud, controller)
- when a next token from openfga is "", that means we have finished the entries for that specific kind, not for all kinds.
- so we implemented a custom token which cycles though entity kinds, which return "" only if we explored all the kinds otherwise it returns a valid token.
Now:
- this check for next page not empty is to avoid an edge case, where we asked for the last 10 entries. So our custom token implementation will return a continuation token which means "ok, you've explored users now explore controllers", the frontend receives a valid token but if it calls for the next page is empty

Hope it clears things up

internal/jimm/relation.go
SimoneDutto: we care because otherwise frontend receives a meaningless token

SimoneDutto: `if len(responseTuples) == int(pageSize)`
this is the right check, because if we have fewer entries from the pagination methods, we are sure the continuation token to return to the frontend will be empty.

PR #1436, opened 2024-11-14
Adds cleanup of orphaned OpenFGA tuples.

alesstimec:
## Description

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [ ] Covered by unit tests


internal/jimm/access.go
```diff
@@ -807,3 +807,58 @@ func (j *JIMM) ListGroups(ctx context.Context, user *openfga.User, filter pagina
...
+		tuples, continuationToken, err := j.OpenFGAClient.ReadRelatedObjects(ctx, openfga.Tuple{}, 10, "")
+		if err != nil {
+			zapctx.Error(ctx, "reading all tuples", zap.Error(err))
+			return err
+		}
+		orphanedTuples := j.orphanedTuples(ctx, tuples...)
+		if len(orphanedTuples) == 0 {
+			if continuationToken == "" {
+				zapctx.Error(ctx, "continue cleanup")
+				continue
```
SimoneDutto: if you are getting the last tuples and you have no orphaned tuples you continue and requery the tuples from the start. Right?

in a case of cleaned db you are potentially spinning and requery continuously and not once every 6hours

internal/jimm/access_test.go
```diff
@@ -1110,3 +1110,99 @@ func TestListGroups(t *testing.T) {
...
+		Database: db.Database{
+			DB: jimmtest.PostgresDB(c, func() time.Time { return now }),
+		},
+		OpenFGAClient: ofgaClient,
+	}
+
+	err = j.Database.Migrate(ctx, false)
+	c.Assert(err, qt.IsNil)
+
+	err = j.OpenFGACleanup(ctx)
```
SimoneDutto, line 1203: can you add another `OpenFGACleanup` on a cleaned db, to make sure we are checking the scenario I've described above?

internal/jimm/access.go
```diff
@@ -807,3 +807,58 @@ func (j *JIMM) ListGroups(ctx context.Context, user *openfga.User, filter pagina
...
+func (j *JIMM) OpenFGACleanup(ctx context.Context) error {
+	for {
+		tuples, continuationToken, err := j.OpenFGAClient.ReadRelatedObjects(ctx, openfga.Tuple{}, 10, "")
+		if err != nil {
+			zapctx.Error(ctx, "reading all tuples", zap.Error(err))
+			return err
+		}
+		orphanedTuples := j.orphanedTuples(ctx, tuples...)
+		if len(orphanedTuples) == 0 {
+			if continuationToken == "" {
```
kian99: Shouldn't this be `if continuationToken != ""`?

internal/jimm/access_test.go
```diff
@@ -1110,3 +1110,99 @@ func TestListGroups(t *testing.T) {
...
+	c := qt.New(t)
+	ctx := context.Background()
+
+	ofgaClient, _, _, err := jimmtest.SetupTestOFGAClient(c.Name())
+	c.Assert(err, qt.IsNil)
+
+	type createTagFunction func(int) *ofga.Entity
+
+	var (
+		createStringTag = func(kind openfga.Kind) createTagFunction {
```
kian99, line 1140: Something worthwhile for creating OpenFGA tuples would be to create a similar construct to our database `yaml` parser where we define a yaml string of what will go into our database. We could have the same for OpenFGA tuples, define it in a yaml string above the test and call some test function to create them.

kian99 (dismissed):
2 comments.
Overall I think this is fine, but my main gripe is that it's a bit of a bandaid over a different problem. If we're not cleaning up tuples, we're doing something wrong elsewhere and then creating a need for this routine.

Just a thought but perhaps we should log when we have orphaned tuples, then we can see we're not cleaning up when we should. We can already see this when we run `jimmctl auth relation list`.

internal/jimm/access.go
alesstimec: fixed, thanks

internal/jimm/access.go
alesstimec: yes, fixed

internal/jimm/access_test.go
alesstimec: i could do, but i wanted to create an arbitrary number of tuples so a function seemed to make more sense

alesstimec:
I agree this is a band-aid for a problem we have in JIMM. I've added logging for tuples that are removed so that we can see where the problem is and address it in followups.

internal/jimm/access_test.go
alesstimec: added

SimoneDutto (approved):
lgtm

internal/jimm/access.go
```diff
@@ -804,3 +804,59 @@ func (j *JIMM) ListGroups(ctx context.Context, user *openfga.User, pagination pa
...
+			return nil
+		default:
+		}
+	}
+}
+
+func (j *JIMM) orphanedTuples(ctx context.Context, tuples ...openfga.Tuple) []openfga.Tuple {
+	orphanedTuples := []openfga.Tuple{}
+	for _, tuple := range tuples {
+		_, err := j.ToJAASTag(ctx, tuple.Object, true)
```
ale8k, line 848: I'd godoc this to explain we're using ToJAASTag specifically to check for existence and it just so happens this method works for us. The code isn't clear without knowing this.

internal/jimm/access.go
```diff
@@ -804,3 +804,59 @@ func (j *JIMM) ListGroups(ctx context.Context, user *openfga.User, pagination pa
 	}
 	return groups, nil
 }
+
+// OpenFGACleanup queries OpenFGA for all existing tuples, tries to resolve each tuple and removes those
+// that JIMM cannot resolved - orphaned tuples. JIMM not being able to resolve a tuple means that the
+// corresponding entity has been removed from JIMM's database.
+func (j *JIMM) OpenFGACleanup(ctx context.Context) error {
+	for {
+		tuples, continuationToken, err := j.OpenFGAClient.ReadRelatedObjects(ctx, openfga.Tuple{}, 10, "")
```
ale8k: 10 feels like a bit of a magic number, what is the reason for doing it in 10s?

ale8k (commented):
I want a little time to think about the "removing if we can't find it in db". 

For example, if we had data loss [somehow; just humour me] in the db. And we just so happened to be performing a cleanup, we'll remove access once data is restored.

And I'll bet there's other scenarios we're not thinking of. So before merging this I implore we all have a think about it.

I'm also unsure about the reading in 10s, and want to be make sure it isn't rate limited. 

Again humour me, we have a data loss of 10000 objects, OpenFGA begins to reject requests after the first 5000, data is restored, we have 5000 lingering, data is restored and 5000 tuples are valid leaving our auth in a mess. (Not sure if this could actually happen but I'm throwing ideas to the wind).

I guess what I'm trying to say is let's have a think and be careful before merging.

internal/jimm/access.go
alesstimec: i can change it to 17?

internal/jimm/access.go
```diff
@@ -804,3 +804,59 @@ func (j *JIMM) ListGroups(ctx context.Context, user *openfga.User, pagination pa
...
+		if err != nil {
+			zapctx.Error(ctx, "reading all tuples", zap.Error(err))
+			return err
+		}
+		zapctx.Debug(ctx, "removing orphaned tuples", zap.Any("tuples", tuples))
+		orphanedTuples := j.orphanedTuples(ctx, tuples...)
+		if len(orphanedTuples) == 0 {
+			if continuationToken != "" {
+				zapctx.Error(ctx, "continue cleanup")
```
kian99: This log line seems arbitrary and at the wrong log level.
I'd suggest to either remove it or add some info
`zapctx.Debug(ctx, "continue cleanup",zap.Int("page",page)`

internal/jimm/access.go
```diff
@@ -804,3 +804,59 @@ func (j *JIMM) ListGroups(ctx context.Context, user *openfga.User, pagination pa
 	}
 	return groups, nil
 }
+
+// OpenFGACleanup queries OpenFGA for all existing tuples, tries to resolve each tuple and removes those
+// that JIMM cannot resolved - orphaned tuples. JIMM not being able to resolve a tuple means that the
+// corresponding entity has been removed from JIMM's database.
+func (j *JIMM) OpenFGACleanup(ctx context.Context) error {
```
kian99, line 813: Let's update the godoc to make it very explicit that this cleanup method is intended to be short-lived until tuples are correctly removed from OpenFGA. Something like.
```
//....
// 
// This approach to cleaning up tuples is intended to be temporary while we implement
// a better approach to eventual consistency of JIMM's database objects and OpenFGA tuples.
```

internal/jimm/access.go
kian99: We're not passing the continuation token into the function?

internal/jimm/access.go
```diff
@@ -804,3 +804,59 @@ func (j *JIMM) ListGroups(ctx context.Context, user *openfga.User, pagination pa
...
+	for {
+		tuples, continuationToken, err := j.OpenFGAClient.ReadRelatedObjects(ctx, openfga.Tuple{}, 10, "")
+		if err != nil {
+			zapctx.Error(ctx, "reading all tuples", zap.Error(err))
+			return err
+		}
+		zapctx.Debug(ctx, "removing orphaned tuples", zap.Any("tuples", tuples))
+		orphanedTuples := j.orphanedTuples(ctx, tuples...)
+		if len(orphanedTuples) == 0 {
```
kian99: Seems to be the wrong way to check for termination, just check if the continuation token is empty and if so, return. Which is already being done below.

internal/jimm/access_test.go
```diff
@@ -1110,3 +1110,103 @@ func TestListGroups(t *testing.T) {
...
+		relation:        "administrator",
+		createTargetTag: createStringTag(openfga.CloudType),
+	}, {
+		createObjectTag: createStringTag(openfga.UserType),
+		relation:        "consumer",
+		createTargetTag: createUUIDTag(openfga.ApplicationOfferType),
+	}}
+
+	orphanedTuples := []ofga.Tuple{}
+	for i := 0; i < 10; i++ {
```
kian99: I think we need a test with more than 10 tags to make sure pagination is working correctly.

internal/jimm/access.go
alesstimec: removed.

internal/jimm/access.go
alesstimec: added

internal/jimm/access.go
alesstimec: changed/fixed

internal/jimm/access_test.go
alesstimec: done

internal/jimm/access.go
alesstimec: done

SimoneDutto:
fyi: i was curious about @ale8k comments and I've checked.

In case of a data loss Postgres enters a recovery mode (read-only), using the WAL to replay the latest operations.
So in case of a data loss, I haven't found any documentation that indicates that Postgres can answer your queries w/o errors but with records missing.

internal/jimm/access.go
```diff
@@ -804,3 +804,62 @@ func (j *JIMM) ListGroups(ctx context.Context, user *openfga.User, pagination pa
...
+		err               error
+		tuples            []ofga.Tuple
+	)
+	for {
+		tuples, continuationToken, err = j.OpenFGAClient.ReadRelatedObjects(ctx, openfga.Tuple{}, 20, continuationToken)
+		if err != nil {
+			zapctx.Error(ctx, "reading all tuples", zap.Error(err))
+			return err
+		}
+		zapctx.Debug(ctx, "removing orphaned tuples", zap.Any("tuples", tuples))
```
kian99: This log line is logging **all** tuples, not the ones that are orphaned.

kian99 (commented):
Looks good, will approve after the log comment is addressed.

internal/jimm/access.go
alesstimec: done

kian99 (approved):
A suggestion for a follow-up. Add a metric to measure the time on the orphan relation cleanup method.

PR #1439, opened 2024-11-18
add role to parse tag

SimoneDutto:
## Description

Add RoleTag to `ParseTag`, the validation is c/p to GroupTag validation.

Fixes _JIRA/GitHub issue number_

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [ ] Covered by unit tests
- [ ] Covered by integration tests

PR #1440, opened 2024-11-18
Juju 7120/update db with roles

ale8k:
## Description

Updates db with role entries. Includes this PR: https://github.com/canonical/jimm/pull/1439#pullrequestreview-2442322946

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [ ] Covered by unit tests


internal/dbmodel/role.go
```diff
@@ -0,0 +1,44 @@
...
+	"time"
+
+	"github.com/juju/names/v5"
+
+	jimmnames "github.com/canonical/jimm/v3/pkg/names"
+)
+
+// RoleEntry holds a role entry.
+type RoleEntry struct {
+	// Note this doesn't use the standard gorm.Model to avoid soft-deletes.
```
SimoneDutto, line 15: nice

internal/dbmodel/role.go
```diff
@@ -0,0 +1,44 @@
...
+	ID        uint `gorm:"primarykey"`
+	CreatedAt time.Time
```
SimoneDutto, line 17: does it mean we need to set these fields manually?

Remember to handle it next PR

SimoneDutto (approved):

internal/dbmodel/sql/postgres/1_13.sql
```diff
@@ -0,0 +1,8 @@
+-- 1_13.sql is a migration that introduces the concept of roles to JIMM.
+CREATE TABLE IF NOT EXISTS roles (
+   id BIGSERIAL PRIMARY KEY,
+   created_at TIMESTAMP WITH TIME ZONE,
```
SimoneDutto, line 4: general question: why do we need to store the timestamp with timezone?

ale8k: I think may be a gorm thing, but I'm not 100%.

internal/dbmodel/role.go
ale8k: I think they're set anyway, it just prevents soft delete fields existing?

internal/dbmodel/role.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	"time"
+
+	"github.com/juju/names/v5"
+
+	jimmnames "github.com/canonical/jimm/v3/pkg/names"
+)
+
+// RoleEntry holds a role entry.
+type RoleEntry struct {
```
kian99, line 14: How comes RoleEntry and not just Role?

ale8k: groups is named similarly

kian99: Ah fair

kian99 (approved):

internal/dbmodel/role.go
alesstimec: yes, gorm will check if DeletedAt is present and use soft deletes if it is

internal/dbmodel/sql/postgres/1_13.sql
alesstimec: time is hard.. use timezones.

internal/dbmodel/version.go
```diff
@@ -20,7 +20,7 @@ const (
 	// Minor is the minor version of the model described in the dbmodel
 	// package. It should be incremented for any change made to the
 	// database model from database model in a released JIMM.
-	Minor = 12
+	Minor = 13
```
alesstimec, line 23: this might have to be bumped.. depends on whether #1443 lands first or not, because it introduces migration 13, too

PR #1441, opened 2024-11-18
refactor listing offers

kian99:
## Description

This PR fixes an issue in ListApplicationOffers that was preventing listing offers when a model had changed ownership.
From the Jira card

>When importing a model from a local user, we allow the owner of the model to be changed. In https://github.com/canonical/jimm/pull/1432  we introduced the ability to import offers alongside a model. 
>
>This causes an issue when running `juju offers` to view the offers available in the model.
The Juju CLI requests for all offers in the specified model and JIMM forwards the request to the controller hosting the model. The hosting controller is not aware of the new model owner and returns an error that the model does not exist.

To fix this, we gather the available offers and then query the Juju controller for the details on each (using the offer URL) before the replacing the fields where JIMM is authoritative and then return the result to the client.

In summary:
- Slightly refactored the FindApplicationOffers method to use the new enrichOfferDetails method.
- Refactored listApplicationOffersForModel to avoid the issue mentioned above.
- Added application-offers to the DB test init package to add offers via yaml.

Fixes [JUJU-7173](https://warthogs.atlassian.net/browse/JUJU-7173)

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [x] Covered by unit tests

## Notes for code reviewers
There's some more cleanup we can do in this area related to the slim-jimm efforts. It seems some of the fields on an offer may have been added to JIMM to handle the filtering at the DB level.

[JUJU-7173]: https://warthogs.atlassian.net/browse/JUJU-7173?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


alesstimec (approved):

kian99:
Turns out this doesn't work because I can't request information about an application offer with just the UUID. The offer URL is required.

kian99:
I've made a change to use the offer URL to query the Juju controller. This implies that the offer URL in JIMM will match the offer URL in Juju.

internal/db/applicationoffer.go
```diff
@@ -148,9 +148,12 @@ func ApplicationOfferFilterByDescription(substring string) ApplicationOfferFilte
 }
 
 // ApplicationOfferFilterByModel filters application offers by model name.
-func ApplicationOfferFilterByModel(modelName string) ApplicationOfferFilter {
+func ApplicationOfferFilterByModel(modelName, modelOwner string) ApplicationOfferFilter {
```
SimoneDutto, line 151: please update name and godoc

internal/jimm/applicationoffer.go
```diff
@@ -612,38 +612,53 @@ func (j *JIMM) FindApplicationOffers(ctx context.Context, user *openfga.User, fi
...
+	if err != nil {
+		return offerDetail, err
+	}
+
+	offerDetail.Users = users
+
+	return offerDetail, nil
 }
 
 func (j *JIMM) applicationOfferFilters(ctx context.Context, jujuFilters ...jujuparams.OfferFilter) ([]db.ApplicationOfferFilter, error) {
```
SimoneDutto, line 656: i know it's not your pr however could you add a godoc to this function?

SimoneDutto (approved):
lgtm with the godoc updated

kian99:
@alesstimec Please take another looks since I switched to using the offer URL to query the controller. I haven't given enough thought to #1443 but I think we can look at landing that after this, thoughts on that?

alesstimec (approved):

PR #1442, opened 2024-11-18
Juju 7122/add methods for role management

ale8k:
## Description

Adds database methods for roles according to the specification. 

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [x] Covered by unit tests


internal/db/role.go
```diff
@@ -0,0 +1,143 @@
...
+	defer servermon.ErrorCounter(servermon.DBQueryErrorCount, &err, string(op))
+
+	if err := d.DB.WithContext(ctx).Delete(role).Error; err != nil {
+		return errors.E(op, dbError(err))
+	}
+	return nil
+}
+
+// ForEachRole iterates through all role entries applying the provided callback function.
+func (d *Database) ForEachRole(ctx context.Context, f func(*dbmodel.RoleEntry) error) (err error) {
```
SimoneDutto: group has ListGroup, which is more suited for our listing use cases. So either replace this forEachRole with ListRoles, or add ListRoles

internal/db/role_test.go
```diff
@@ -0,0 +1,195 @@
...
+		Name: "test-role",
+	}
+	tx := s.Database.DB.First(&re)
+	c.Assert(tx.Error, qt.IsNil)
+	c.Assert(re.ID, qt.Equals, uint(1))
+	c.Assert(re.Name, qt.Equals, "test-role")
+	c.Assert(re.UUID, qt.Equals, uuid)
+}
+
+func (s *dbSuite) TestDatabase_GetRole(c *qt.C) {
```
SimoneDutto: i see you want to bring the underscore in this Go Land

internal/db/role_test.go
```diff
@@ -0,0 +1,195 @@
...
+	ge1.Name = "renamed-role"
+	err = s.Database.UpdateRole(context.Background(), ge1)
+	c.Check(err, qt.IsNil)
+
+	ge2 := &dbmodel.RoleEntry{
+		Name: "renamed-role",
+	}
+	err = s.Database.GetRole(context.Background(), ge2)
+	c.Check(err, qt.IsNil)
+	c.Assert(ge2, qt.DeepEquals, ge1)
```
SimoneDutto: how they are deep equals? At least their `UpdatedAt` should be different, right?

internal/db/role.go
SimoneDutto, line 1: we need the same methods implemented for the groups.
For example, here we are missing `CountRoles` and this is needed for rebac admin pagination

SimoneDutto (changes requested):

internal/db/role.go
ale8k: The spec didn't say ListRoles, but I agree @alesstimec any reason ForEachRole and not ListRoles?

internal/db/role_test.go
ale8k: Aha I do ;)

internal/db/role_test.go
ale8k: Fair point! I'm not actually sure...

SimoneDutto (approved):
lgtm

internal/db/role.go
```diff
@@ -0,0 +1,155 @@
...
+	if err := d.ready(); err != nil {
+		return errors.E(op, err)
+	}
+
+	durationObserver := servermon.DurationObserver(servermon.DBQueryDurationHistogram, string(op))
+	defer durationObserver()
+	defer servermon.ErrorCounter(servermon.DBQueryErrorCount, &err, string(op))
+
+	db := d.DB.WithContext(ctx)
+	if role.ID != 0 {
```
alesstimec, line 51: should we have some validation to make sure at least one of `id`, `uuid` or `name` is non-empty?

internal/db/role.go
```diff
@@ -0,0 +1,155 @@
...
+		db = db.Where("name = ?", role.Name)
+	}
+	if err := db.First(&role).Error; err != nil {
+		return errors.E(op, dbError(err))
+	}
+	return nil
+}
+
+// UpdateRole updates the role identified by its ID or UUID.
+func (d *Database) UpdateRole(ctx context.Context, role *dbmodel.RoleEntry) (err error) {
```
alesstimec: suggestion: UpdateRoleName

internal/db/role.go
```diff
@@ -0,0 +1,155 @@
...
+	const op = errors.Op("db.UpdateRole")
+
+	if role.ID == 0 {
+		return errors.E(errors.CodeNotFound)
+	}
+	if role.UUID == "" {
+		return errors.E("role uuid not specified", errors.CodeNotFound)
+	}
```
alesstimec, line 72: code above contradicts the godoc: both `ID` and `UUID` must be specified. Is this what we want?

internal/db/role.go
```diff
@@ -0,0 +1,155 @@
...
+// RemoveRole removes the role identified by its ID or UUID.
+func (d *Database) RemoveRole(ctx context.Context, role *dbmodel.RoleEntry) (err error) {
+	const op = errors.Op("db.RemoveRole")
+
+	if role.ID == 0 {
+		return errors.E(errors.CodeNotFound)
+	}
+	if role.UUID == "" {
+		return errors.E(errors.CodeNotFound)
+	}
```
alesstimec, line 97: same here: code requires that both be specified.

internal/db/role_test.go
```diff
@@ -0,0 +1,219 @@
...
+		Name: "test-role",
+	}
+	tx := s.Database.DB.First(&re)
+	c.Assert(tx.Error, qt.IsNil)
+	c.Assert(re.ID, qt.Equals, uint(1))
+	c.Assert(re.Name, qt.Equals, "test-role")
+	c.Assert(re.UUID, qt.Equals, uuid)
+}
+
+func (s *dbSuite) TestDatabase_GetRole(c *qt.C) {
```
alesstimec: veto: please no underscores.

internal/db/role_test.go
alesstimec: worth investigating.

alesstimec (changes requested):

internal/db/role_test.go
SimoneDutto: I think gorm is not aware of UpdateAt. 

When we are not using gorm in the models maybe we've also lost this autoupdating mechanism for `CreatedAt, UpdatedAt`. Please investigate this or create a card

internal/db/role.go
ale8k: Is there a need? We have the not-found error, I imagine this is good enough?

internal/db/role.go
ale8k: I'll update groups too (so they look similar)

internal/db/role.go
ale8k: I think so, you'll get the role and it'll have an id and uuid. I guess it's an "extra safe" check.

internal/db/role.go
SimoneDutto: wait, but this method is implicitly used only to update the name. But you can potentially update any field, right? So i think the name is correct.

internal/db/role_test.go
ale8k: Looking now

internal/db/role.go
kian99: I suggest adding a test with no ID, name or UUID specified to verify that behavior.

internal/db/role.go
kian99: We should update the godoc in that case.

kian99 (commented):
Just gave it a quick review so won't approve but looks good.

internal/db/role.go
ale8k: I modified the existing with name set to have nothing.

internal/db/role.go
ale8k: Done

internal/db/role_test.go
ale8k: Created bug.

internal/db/role.go
```diff
@@ -0,0 +1,155 @@
...
+
+// UpdateRoleName updates the role name identified by its ID and UUID.
+func (d *Database) UpdateRoleName(ctx context.Context, role *dbmodel.RoleEntry) (err error) {
+	const op = errors.Op("db.UpdateRole")
+
+	if role.ID == 0 {
+		return errors.E(errors.CodeNotFound)
+	}
+	if role.UUID == "" {
+		return errors.E("role uuid not specified", errors.CodeNotFound)
```
alesstimec: in should be able to do 
```
UpdateRoleName(ctx, &dbmodel.RoleEntry{
  ID: <id>,
  Name: "some-new-name",
}
``` 
or 
```
UpdateRoleName(ctx, &dbmodel.RoleEntry{
  UUID: <uuid>,
  Name: "some-new-name",
}
```
we don't need both to be filled in to update the role name

internal/db/role.go
alesstimec: one i enough to uniquely identify the role we want to remove

internal/db/role_test.go
```diff
@@ -0,0 +1,215 @@
...
+
+	err := s.Database.GetRole(context.Background(), &dbmodel.RoleEntry{})
+	c.Check(errors.ErrorCode(err), qt.Equals, errors.CodeUpgradeInProgress)
+
+	err = s.Database.Migrate(context.Background(), false)
+	c.Assert(err, qt.IsNil)
+
+	role := &dbmodel.RoleEntry{}
+	err = s.Database.GetRole(context.Background(), role)
+	c.Check(errors.ErrorCode(err), qt.Equals, errors.CodeNotFound)
```
alesstimec: hmm.. but if you add a role above this, then we expect to get one role back - whichever is first by some magic order.

internal/db/role.go
alesstimec: I would still require at least one of these fields that (each) uniquely identify a role to be set

alesstimec (changes requested):

internal/db/role.go
ale8k: Done, one must be set now.

internal/db/role_test.go
ale8k: Would you like me to break it into its own test? If I do, it'd probably be worth breaking up all cases.  But many of our tests bundle many cases into one so it'd be a bit of an outlier

internal/db/role.go
ale8k: Updated to be either or

internal/db/role.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+
+	durationObserver := servermon.DurationObserver(servermon.DBQueryDurationHistogram, string(op))
+	defer durationObserver()
+	defer servermon.ErrorCounter(servermon.DBQueryErrorCount, &err, string(op))
+
+	model := d.DB.WithContext(ctx).Model(&dbmodel.RoleEntry{})
+	model.Where("uuid = ?", uuid)
+	if model.Update("name", name).RowsAffected == 0 {
+		return errors.E(op, "role does not exist")
```
alesstimec:
```suggestion
		return errors.E(op, "role not found", errors.CodeNotFound)
```

alesstimec (approved):
LGTM with one suggestion

PR #1443, opened 2024-11-18
Lightweight application offers.

alesstimec:
## Description

    Removes data from ApplicationOffer type that JIMM does not need to store. Instead the source of
    truth for application offers becomes the offering juju controller. Controllers are now queries when
    we need details on application offers.
    
    It also allowed me to remove the update method for application offers and simplify the
    AllWatcher so that we don't have to process deltas for offers.
    
    A drive-by fix: we no longer require the user to have add-model permission to add a k8s cloud.

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [ ] Covered by unit tests

### DIFF
--- a/internal/db/applicationoffer.go
+++ b/internal/db/applicationoffer.go
@@ -5,8 +5,6 @@ package db
 import (
 	"context"
 
-	"gorm.io/gorm"
-
 	"github.com/canonical/jimm/v3/internal/dbmodel"
 	"github.com/canonical/jimm/v3/internal/errors"
 	"github.com/canonical/jimm/v3/internal/servermon"
@@ -33,46 +31,6 @@ func (d *Database) AddApplicationOffer(ctx context.Context, offer *dbmodel.Appli
 	return nil
 }
 
-// UpdateApplicationOffer updates the application offer information.
-func (d *Database) UpdateApplicationOffer(ctx context.Context, offer *dbmodel.ApplicationOffer) (err error) {
-	const op = errors.Op("db.UpdateApplicationOffer")
-
-	if err := d.ready(); err != nil {
-		return errors.E(op, err)
-	}
-
-	durationObserver := servermon.DurationObserver(servermon.DBQueryDurationHistogram, string(op))
-	defer durationObserver()
-	defer servermon.ErrorCounter(servermon.DBQueryErrorCount, &err, string(op))
-
-	db := d.DB.WithContext(ctx)
-	err = db.Transaction(func(tx *gorm.DB) error {
-		tx.Omit("Connections", "Endpoints", "Spaces").Save(offer)
-		err = tx.Model(offer).Association("Connections").Replace(offer.Connections)
-		if err != nil {
-			return err
-		}
-		err = tx.Model(offer).Association("Endpoints").Replace(offer.Endpoints)
-		if err != nil {
-			return err
-		}
-		err = tx.Model(offer).Association("Spaces").Replace(offer.Spaces)
-		if err != nil {
-			return err
-		}
-		return tx.Error
-	})
-	if err != nil {
-		return errors.E(op, dbError(err))
-	}
-
-	if err := d.GetApplicationOffer(ctx, offer); err != nil {
-		return err
-	}
-
-	return nil
-}
-
 // GetApplicationOffer returns application offer information based on the
 // offer UUID or URL.
 func (d *Database) GetApplicationOffer(ctx context.Context, offer *dbmodel.ApplicationOffer) (err error) {
@@ -95,10 +53,8 @@ func (d *Database) GetApplicationOffer(ctx context.Context, offer *dbmodel.Appli
 	default:
 		return errors.E(op, "missing offer UUID or URL")
 	}
-	db = db.Preload("Connections")
-	db = db.Preload("Endpoints")
+
 	db = db.Preload("Model").Preload("Model.Controller")
-	db = db.Preload("Spaces")
 	if err := db.First(&offer).Error; err != nil {
 		err := dbError(err)
 		if errors.ErrorCode(err) == errors.CodeNotFound {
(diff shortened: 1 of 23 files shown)


kian99:
I think we have a solution #1441, let's wait for that to land before landing this please.

internal/jimm/applicationoffer.go
```diff
@@ -599,193 +546,57 @@ func (j *JIMM) FindApplicationOffers(ctx context.Context, user *openfga.User, fi
...
-
-		// non-admin users should not see connections of an application
-		// offer.
-		if accessLevel != "admin" {
-			offerDetails[i].Connections = nil
+			return errors.E(op, err)
 		}
-		users, err := j.listApplicationOfferUsers(ctx, offer.ResourceTag(), user.Identity, accessLevel)
+		defer api.Close()
+		controllerOffers, err := api.ListApplicationOffers(ctx, filters)
```
kian99: I like this change of just querying the Juju controllers but we have some challenges with this approach:
1. It will query all offers on all controllers each time. That could be a large number of offers.
2. It doesn't take into account the fact that model owners have changed when a model was imported into JIMM.
3. It doesn't take into account the difference between `juju find-offers` and `juju offers`. The latter shows you the offers available in a model. The former shows you offers you have access to from all models so that you can use them in your current model.

I think we can solve this pretty easily with some code that already existed. List-offers should just list the offers in the specified model as was already done, no problems there. Find-offers can loop through all controllers/offers but requires the model owner to be treated carefully. If the model owner/name are specified we can just use the approach in list-offers.

Alternatively, we can consider removing the transfer of ownership concept when importing models into JIMM. That would mean two models with the same name/owner on different controllers can't be imported into JIMM but that could possibly be an okay tradeoff.

alesstimec: @kian99 updated as you suggested:
- ListOffers now works as before
- FindApplicationOffers will query all controllers in parallel - not the most efficient, but better than the serial approach
- still does not handle the case of imported models

internal/db/applicationoffer.go
```diff
@@ -207,12 +104,12 @@ func (d *Database) FindApplicationOffers(ctx context.Context, filters ...Applica
 	db := d.DB.WithContext(ctx)
 	db = db.Table("application_offers AS offers")
 
-	for _, filter := range filters {
-		db = filter(db)
-	}
+	db = db.Joins("JOIN models ON models.id = offers.model_id").
+		Where("models.name = ?", modelName).
+		Where("models.owner_identity_name = ?", modelOwner)
```
kian99, line 109: How does this interact with the preload below? Or rather do we need the preload?

internal/jimm/applicationoffer.go
```diff
@@ -599,223 +558,128 @@ func (j *JIMM) FindApplicationOffers(ctx context.Context, user *openfga.User, fi
...
 
-	offerDetails := make([]jujuparams.ApplicationOfferAdminDetailsV5, len(offers))
-	for i, offer := range offers {
-		var offerDetail jujuparams.ApplicationOfferAdminDetailsV5
-		offerDetail.OfferURL = offer.URL
-		err := api.GetApplicationOffer(ctx, &offerDetail)
-		if err != nil {
-			return nil, err
-		}
+				accessLevel, err := j.getUserOfferAccess(ctx, user, names.NewApplicationOfferTag(offer.OfferUUID))
```
kian99: What do you think about using the `enrichOfferDetails` function here and elsewhere where we do this?

kian99 (commented):
Looks good, just 2 comments so far. Also can we please update the PR description

internal/db/applicationoffer.go
```diff
@@ -130,71 +86,12 @@ func (d *Database) DeleteApplicationOffer(ctx context.Context, offer *dbmodel.Ap
...
-		}
-		return db
-	}
-}
-
-// FindApplicationOffers returns application offers matching criteria specified by the filters.
-func (d *Database) FindApplicationOffers(ctx context.Context, filters ...ApplicationOfferFilter) (_ []dbmodel.ApplicationOffer, err error) {
-	const op = errors.Op("db.FindApplicationOffer")
+// FindApplicationOffersByModel returns all application offers in a model specified by model name and owner.
+func (d *Database) FindApplicationOffersByModel(ctx context.Context, modelName, modelOwner string) (_ []dbmodel.ApplicationOffer, err error) {
```
ale8k, line 90: Nice

internal/db/applicationoffer.go
ale8k: Yeah I'm confused here too, will we need to access the model within the offers on this db call?

internal/dbmodel/sql/postgres/1_14.sql
```diff
@@ -0,0 +1,19 @@
+-- 1_13.sql is a migration simplifies application offers
+DROP INDEX IF EXISTS idx_application_offer_connections_deleted_at;
```
ale8k, line 2: Ooft... This definitely felt good

internal/jimm/watcher.go
```diff
@@ -498,6 +493,7 @@ func (w *Watcher) handleDelta(ctx context.Context, modelIDf func(string) *modelS
 			state.changed = true
 			state.units[eid.Id] = true
 		}
+	default:
```
ale8k, line 496: Default and do what?

internal/jujuapi/applicationoffers_test.go
```diff
@@ -281,64 +282,61 @@ func (s *applicationOffersSuite) TestListApplicationOffers(c *gc.C) {
...
+		s.OFGAClient,
+	)
+
+	offer := dbmodel.ApplicationOffer{
+		URL: offerURL,
+	}
+	err = s.JIMM.Database.GetApplicationOffer(context.Background(), &offer)
+	c.Assert(err, gc.Equals, nil)
+
+	c.Assert(testUser.GetApplicationOfferAccess(ctx, offer.ResourceTag()), gc.Equals, ofganames.AdministratorRelation)
```
ale8k: I thought we were not going to in line func calls i.e., from our discussion on c.Assert(mythingreturnserr, gc.IsNil)? This seems like the same thing?

internal/jujuapi/applicationoffers_test.go
```diff
@@ -281,64 +282,61 @@ func (s *applicationOffersSuite) TestListApplicationOffers(c *gc.C) {
...
+
+	err = client.RevokeOffer("test.user@canonical.com", "admin", offerURL)
+	c.Assert(err, jc.ErrorIsNil)
+
+	c.Assert(testUser.GetApplicationOfferAccess(ctx, offer.ResourceTag()), gc.Equals, ofganames.NoRelation)
```
ale8k: Learning question (maybe I've forgotten actually): What is this line? "NoRelation"?

ale8k (approved):
Couple comments, I'm approving but ping me when responded.

internal/jimm/watcher.go
alesstimec: nothing.. ignore deltas we don't care about

internal/jujuapi/applicationoffers_test.go
alesstimec: asserting user does not have access to the application offer.

internal/db/applicationoffer.go
alesstimec: a bit of shameless cargo-culting from cloud defaults. Joins and preloads do not seem to interfere with eachother.

internal/jimm/applicationoffer.go
alesstimec: done.

alesstimec: actually.. done with a little tweak

internal/jujuapi/applicationoffers_test.go
alesstimec: fixed

internal/db/applicationoffer.go
ale8k: Yeah it'll filter before actually loading I imagine

internal/jimm/applicationoffer.go
```diff
@@ -283,19 +288,50 @@ func (j *JIMM) listApplicationOfferUsers(ctx context.Context, offer names.Applic
...
-			// fetch it from the DB if it is REALLY important.
 		})
 	}
 	return userDetails, nil
 }
 
+var noApplicationOfferAccessError = errors.E("no application offer access")
+
+// enrichOfferDetails replaces fields on an application offer's details with information
+// where JIMM is authoritative. It returns a noApplicationOFferAccessError if the user
```
kian99: nit
```suggestion
// where JIMM is authoritative. It returns a noApplicationOfferAccessError if the user
```

internal/jimm/applicationoffer.go
```diff
@@ -599,223 +586,115 @@ func (j *JIMM) FindApplicationOffers(ctx context.Context, user *openfga.User, fi
...
-		return nil, errors.E(op, err)
-	}
-	userOfferUUIDs, err := user.ListApplicationOffers(ctx, ofganames.ReaderRelation)
-	if err != nil {
-		return nil, errors.E(op, err)
-	}
-	offerFilters = append(offerFilters, db.ApplicationOfferFilterByUUID(userOfferUUIDs))
-	offers, err := j.Database.FindApplicationOffers(ctx, offerFilters...)
+	// first we find all application offers user has access to
+	offerUUIDs, err := user.ListApplicationOffers(ctx, ofganames.ReaderRelation)
```
kian99: I wonder if we shouldn't just query all controllers after reading this.
It's either:
1. Query OpenFGA for all offers a user can access.
2. Query DB for each offer to get the controller of the offer.
3. Query the controllers with the user provided filters.
4. Check the user has access to the returned offers.

Versus

1. Query all controllers with the user provided filters.
2. Check the user has access to the returned offers.

internal/jimm/cloud.go
```diff
@@ -224,14 +224,6 @@ func (j *JIMM) AddHostedCloud(ctx context.Context, user *openfga.User, tag names
 		return errors.E(op, err)
 	}
 
-	allowedAddModel, err := user.IsAllowedAddModel(ctx, region.Cloud.ResourceTag())
```
kian99, line 227: Should've been a separate PR.

internal/jimm/applicationoffer.go
kian99: It's worth it to keep `ListApplicationOffers` as is since that is just querying a single model. But I would simplify `FindApplicationOffers` to just query across all the controllers with the provided filters.

internal/jimm/cloud.go
alesstimec: yes, it should, but i (we) keep forgetting about it :)

internal/jimm/applicationoffer.go
alesstimec: done.

internal/jimm/applicationoffer.go
```diff
@@ -599,223 +586,104 @@ func (j *JIMM) FindApplicationOffers(ctx context.Context, user *openfga.User, fi
...
-	if err != nil {
-		return nil, errors.E(op, err)
-	}
-	defer api.Close()
+	for controllerID, _ := range controllerIDs {
+		eg.Go(func() error {
+			ctl := &dbmodel.Controller{
+				ID: controllerID,
+			}
+			err := j.Database.GetController(ctx, ctl)
```
kian99: Minor optimisation for a follow up, but instead of passing in the controller ID you could pass in the controller dbmodel since we fetch it twice.

internal/jimm/applicationoffer.go
```diff
@@ -599,223 +586,104 @@ func (j *JIMM) FindApplicationOffers(ctx context.Context, user *openfga.User, fi
...
+			if err != nil {
+				return errors.E(err)
+			}
+			defer api.Close()
+			controllerOffers, err := query(api)
+			if err != nil {
+				return errors.E(err)
+			}
+			for _, offer := range controllerOffers {
+				offer := offer
```
kian99: You don't need to do this since Go 1.23

internal/jimm/applicationoffer.go
```diff
@@ -283,19 +288,50 @@ func (j *JIMM) listApplicationOfferUsers(ctx context.Context, offer names.Applic
...
 	}
 	return userDetails, nil
 }
 
+var noApplicationOfferAccessError = errors.E("no application offer access")
+
+// enrichOfferDetails replaces fields on an application offer's details with information
+// where JIMM is authoritative. It returns a noApplicationOfferAccessError if the user
+// does not have access to the offer.
+func (j *JIMM) enrichOfferDetails(ctx context.Context, user *openfga.User, dbOffer dbmodel.ApplicationOffer, offerDetail jujuparams.ApplicationOfferAdminDetailsV5) (jujuparams.ApplicationOfferAdminDetailsV5, error) {
```
kian99: The `dbOffer` parameter is no longer needed in this function.

kian99 (approved):
lgtm with some comments, happy for them to get done in a follow up.

PR #1444, opened 2024-11-18
fix(auth model): assignee missing from each type for roles

ale8k:
## Description

Assignee was missing. I've also moved the DSL around to it reads a bit more semantically.

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [ ] Covered by unit tests


alesstimec (approved):
LGTM with an ask: could we please generate the JSON in the build process instead of commiting it every time?

alesstimec:
**Note**: the entitlementsList in internal/jimmhttp/rebac_admin/entitlements.go will need to be fixed in a followup

ale8k:
> LGTM with an ask: could we please generate the JSON in the build process instead of commiting it every time?

Not sure what you mean?

kian99 (approved):

kian99 (commented):
I suggest when making quality-of-life improvements like re-ordering the auth model. Do the actual changes in one commit then the re-ordering in a separate commit. Makes it easier to review.

ale8k:
> I suggest when making quality-of-life improvements like re-ordering the auth model. Do the actual changes in one commit then the re-ordering in a separate commit. Makes it easier to review.

To be fair, I thought this when I did it. It was more of a driveby

SimoneDutto (changes requested):
please update the CI workflow here: https://github.com/canonical/jimm/blob/ab98ccc9a64b1fe6d72d305146b2fe6ad36b7d06/.github/workflows/ci.yaml#L31

The wrong auth model was caused because here we were not actually running the tests

Edit:
we are running the tests but the were lacking the role#assignee case.

openfga/tests.fga.yaml
```diff
@@ -27,6 +27,40 @@ tuples:
...
+    - user: role:ro-role-2#assignee
+      relation: administrator
+      object: controller:ro-controller-1
+    - user: role:ro-role-2#assignee
+      relation: audit_log_viewer
+      object: controller:ro-controller-2
+
+    - user: role:ro-role-2#assignee
+      relation: administrator
+      object: model:ro-model-1
```
alesstimec, line 39: weren't we just complaining about confusing terminology? as i read these test i'm super confused by prefixes such as `user:gr-user-1` or `model:ro-model-1`..  i'd rather see we just used regular names, not names that "indicate" the purpose of the thing being named..

Makefile
```diff
@@ -115,6 +115,8 @@ load-rock:
 	$(eval jimm_version := $(shell cat ./rocks/jimm.yaml | yq ".version"))
 	@sudo /snap/rockcraft/current/bin/skopeo --insecure-policy copy oci-archive:jimm_${jimm_version}_amd64.rock docker-daemon:jimm:latest
 
+transform-auth-model:
```
alesstimec: suggestion: auth-model-json

openfga/README.md
```diff
@@ -21,9 +21,7 @@ go install github.com/openfga/cli/cmd/fga@latest
 ## Adding / modifying [to] the authorsation model
 1. Open the authorisation_model.fga 
 2. Make your modification
-3. Open the Command Pallette using Ctrl+Shift+P (Windows) or Command+Shift+P (OSX)
-4. Select OpenFGA: Transform DSL to JSON
-5. Save the file over the existing authorisation_model.json
+3. Run: `make transform-auth-model`
```
alesstimec, line 24: much better.. but i would still prefer
`make auth-model-json`

alesstimec (approved):
LGTM with a few suggestions

openfga/tests.fga.yaml
ale8k: I agree, but I followed the convention (and I get why piotr did it, to avoid collisions)

Makefile
ale8k: I'll leave this as it's following the verbs in the tool

SimoneDutto (approved):
thank you for updating the tests.

PR #1446, opened 2024-11-19
fix flaky test TestStreamProxyStoppedController

SimoneDutto:
## Description

We have a flaky test: https://github.com/canonical/jimm/actions/runs/11910124487/attempts/1?pr=1444
To reproduce it locally: `go test -run ^TestStreamProxyStoppedController$ github.com/canonical/jimm/v3/internal/rpc -count 100000 -failfast`

This should fix it.

## Possible explanation 
In some cases, a race condition prevents the stopped variable from being updated before the echo server enters the ReadJSON. 
This adds the check for `stopped` after the readJSON, to avoid race conditions.

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [ ] Covered by unit tests


kian99 (commented):
While I agree the use of `stopped` here doesn't seem quite right, I'm not convinced this fixes the flaky test because in the link in the description the error is 
```
 error:
          got nil error but want non-nil
        got error:
          nil
        regexp:
          ".*abnormal closure.*"
```
This error couldn't have come from this block of code you've changed because we check `if err != nil` before returning.

kian99:
Ah but trying it locally, I see if I comment out the line `stopped = true` then I get the same error as in the Github Actions run, so it must be something I'm not seeing.
(edit): Yeah my comment above was looking at the wrong error.

qa-microk8s-controller.yaml
kian99, line 1: I think this was accidentally commited.

internal/rpc/streamproxy_test.go
```diff
@@ -14,19 +15,18 @@ import (
...
-		if *stopped {
-			return errors.New("stopped")
-		}
-		if err := c.ReadJSON(&msg); err != nil {
-			return err
-		}
-		if err := c.WriteJSON(msg); err != nil {
-			return err
-		}
+func streamEcho(c *websocket.Conn, stopped *atomic.Bool) error {
```
kian99: Now this will only send one message

internal/rpc/streamproxy_test.go
```diff
@@ -84,7 +84,32 @@ func TestStreamProxyStoppedController(t *testing.T) {
...
+	c.Assert(err, qt.IsNil)
+	defer ws.Close()
+
+	verifyEcho(c, ws, ".*abnormal closure.*")
+
+	ws.Close()
+	<-doneChan // Ensure go routines are cleaned up
+}
+
+func TestStreamProxyStoppedMidwayController(t *testing.T) {
```
kian99, line 102: I think this test is okay, but it doesn't need the `stopped` variable. I tested it locally and commented out the line `stopped.Store(true)` and it still passes because the streamEcho returns after echoing 1 message.

SimoneDutto: even simpler

kian99 (approved):

internal/rpc/streamproxy_test.go
```diff
@@ -14,19 +14,15 @@ import (
...
-		if *stopped {
-			return errors.New("stopped")
-		}
-		if err := c.ReadJSON(&msg); err != nil {
-			return err
-		}
-		if err := c.WriteJSON(msg); err != nil {
-			return err
-		}
+func streamEcho(c *websocket.Conn) error {
```
alesstimec: so, this will echo a single message.. the function name seems wrong, that's my comment :)

alesstimec (approved):
LGTM with a comment

internal/rpc/streamproxy_test.go
SimoneDutto: updated the name to a more explicit

PR #1450, opened 2024-11-19
add facade and entity to logs

SimoneDutto:
## Description

We now inject facade and entity to zapctx logs, so you can grep for specific entity or facade logs.

> drive by add logs for jimm methods entrypoint

I am using the OP as an identifier to be consistent with errors, etc, etc

Feel free to suggest better name than `facade` and `entity`, keep the brief otherwise the logs are ugly to read locally-

Example:
image

------------------------

After a quick call with @kian99 we've settled to inject into the context the fields at the RPC method calling level, to avoid boilerplate.

New example:
image

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [ ] Covered by unit tests

### DIFF
--- a/internal/jimm/model.go
+++ b/internal/jimm/model.go
@@ -553,6 +553,7 @@ func (b *modelBuilder) JujuModelInfo() *jujuparams.ModelInfo {
 // AddModel adds the specified model to JIMM.
 func (j *JIMM) AddModel(ctx context.Context, user *openfga.User, args *ModelCreateArgs) (_ *jujuparams.ModelInfo, err error) {
 	const op = errors.Op("jimm.AddModel")
+	zapctx.Info(ctx, string(op))
 
 	owner, err := dbmodel.NewIdentity(args.Owner.Id())
 	if err != nil {
@@ -719,6 +720,7 @@ func (j *JIMM) addModelPermissions(ctx context.Context, owner *openfga.User, mt 
 // CodeUnauthorized.
 func (j *JIMM) ModelInfo(ctx context.Context, user *openfga.User, mt names.ModelTag) (*jujuparams.ModelInfo, error) {
 	const op = errors.Op("jimm.ModelInfo")
+	zapctx.Info(ctx, string(op))
 
 	var m dbmodel.Model
 	m.SetTag(mt)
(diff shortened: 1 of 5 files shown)


kian99 (approved):
I think you could reduce the boilerplate and make this happen "magically" by looking at how we dispatch RPC methods in `internal/jujuapi/rpc/`.
This is a good start too though.

SimoneDutto:
> I think you could reduce the boilerplate and make this happen "magically" by looking at how we dispatch RPC methods in `internal/jujuapi/rpc/`. This is a good start too though.

I've tried it, but you need to manipulate the context passed down to make it work.
It should be `func (c rootMethodCaller) Call` here. However, it seems more hacky than just adding the explicit one after the op definition.

https://github.com/canonical/jimm/blob/92b1465a11ee8bf4a167238426a97fca19a5bf4e/internal/jujuapi/rpc/root.go#L105-L109

kian99:
>I've tried it, but you need to manipulate the context passed down to make it work.
It should be func (c rootMethodCaller) Call here. However, it seems more hacky than just adding the explicit one after the op definition.

@SimoneDutto How come you found it more hacky? It's more obscure where the information is getting put on the context for sure. But perhaps we just need to build the right scaffolding around it.

SimoneDutto:
> > I've tried it, but you need to manipulate the context passed down to make it work.
> > It should be func (c rootMethodCaller) Call here. However, it seems more hacky than just adding the explicit one after the op definition.
> 
> @SimoneDutto How come you found it more hacky? It's more obscure where the information is getting put on the context for sure. But perhaps we just need to build the right scaffolding around it.

there isn't one place where we have both "methodName" and "ctx" to inject the zap field

internal/jimm/model.go
```diff
@@ -553,6 +553,7 @@ func (b *modelBuilder) JujuModelInfo() *jujuparams.ModelInfo {
 // AddModel adds the specified model to JIMM.
 func (j *JIMM) AddModel(ctx context.Context, user *openfga.User, args *ModelCreateArgs) (_ *jujuparams.ModelInfo, err error) {
 	const op = errors.Op("jimm.AddModel")
+	zapctx.Info(ctx, string(op))
```
kian99, line 556: With the changes we made yesterday, logging the op will duplicate the facade method

SimoneDutto: not really, this is the log for the jimm's method entrypoint. 
It can be decoupled from the facade methods, you can potentially call multiple jimm methods as well from a single facade, and also call jimm's methods from rebac-admin.

ex. `CreateModel` facade can call multiple times `jimm.AddModel`, each with the same facade method injected but different entity tag.

kian99: Fair enough

ale8k (approved):

PR #1451, opened 2024-11-19
fix(internal/db/group.go): fixes CRUD methods for groups

alesstimec:
## Description

Fixes group handling methods in internal/db package.

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [ ] Covered by unit tests

### DIFF
--- a/internal/db/group.go
+++ b/internal/db/group.go
@@ -62,6 +62,10 @@ func (d *Database) GetGroup(ctx context.Context, group *dbmodel.GroupEntry) (err
 		return errors.E(op, err)
 	}
 
+	if group.UUID == "" && group.Name == "" {
+		return errors.E(op, "must specify uuid or name")
+	}
+
 	durationObserver := servermon.DurationObserver(servermon.DBQueryDurationHistogram, string(op))
 	defer durationObserver()
 	defer servermon.ErrorCounter(servermon.DBQueryErrorCount, &err, string(op))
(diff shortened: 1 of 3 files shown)


internal/db/group.go
```diff
@@ -109,14 +113,11 @@ func (d *Database) ListGroups(ctx context.Context, limit, offset int, match stri
 }
 
 // UpdateGroup updates the group identified by its ID.
-func (d *Database) UpdateGroup(ctx context.Context, group *dbmodel.GroupEntry) (err error) {
+func (d *Database) UpdateGroupName(ctx context.Context, uuid, name string) (err error) {
```
SimoneDutto, line 116: godoc

SimoneDutto (approved):
lgtm with godoc updated

internal/db/group_test.go
```diff
@@ -143,15 +141,15 @@ func (s *dbSuite) TestUpdateGroup(c *qt.C) {
...
-	err = s.Database.UpdateGroup(context.Background(), ge1)
+	err = s.Database.UpdateGroupName(context.Background(), ge1.UUID, ge1.Name)
 	c.Check(err, qt.IsNil)
 
 	ge2 := &dbmodel.GroupEntry{
 		Name: "renamed-group",
 	}
 	err = s.Database.GetGroup(context.Background(), ge2)
 	c.Check(err, qt.IsNil)
-	c.Assert(ge2, qt.DeepEquals, ge1)
```
kian99, line 154: This wasn't working before or you were just worried it could be flaky?

internal/db/group.go
kian99, line 116: It's a liiittle bit strange this method accepts a uuid rather than a dbmodel.GroupEntry. Based on the other methods I'd expect a dbmodel.GroupEntry and the new name as parameters. Otherwise for the remove group I'd expect a modelUUID and ID parameter. But I might be overthinking it.

kian99 (approved):

internal/db/group_test.go
SimoneDutto: I think it's because if you use `Model` you correctly update `UpdateAt` each time you change something in the Database Model, so ge2 and ge1 shouldn't be `DeepEquals`

internal/db/group.go
alesstimec: I was just thinking of intended usage: the only use case is renaming a group. UUID uniquely identifies a group. So if we get a group tag (which contains the UUID) or fetch the GroupEntry from db, we always have the UUID.

alesstimec: done

PR #1452, opened 2024-11-19
JUJU-7125/add-jimm-facade-for-role

SimoneDutto:
## Description

Add facade methods for roles.
The actual implementations are empty.

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [ ] Covered by unit tests

### DIFF
--- a/internal/jimm/role.go
+++ b/internal/jimm/role.go
@@ -1,0 +1,51 @@
+// Copyright 2024 Canonical.
+package jimm
+
+import (
+	"context"
+
+	"github.com/canonical/jimm/v3/internal/common/pagination"
+	"github.com/canonical/jimm/v3/internal/dbmodel"
+	"github.com/canonical/jimm/v3/internal/openfga"
+)
+
+// AddRole creates a role within JIMMs DB for reference by OpenFGA.
+func (j *JIMM) AddRole(ctx context.Context, user *openfga.User, name string) (*dbmodel.RoleEntry, error) {
+	return &dbmodel.RoleEntry{}, nil
+}
+
+// CountRoles returns the number of roles that exist.
+func (j *JIMM) CountRoles(ctx context.Context, user *openfga.User) (int, error) {
+	return 0, nil
+}
+
+// getRole returns a role based on the provided UUID or name.
+func (j *JIMM) getRole(_ context.Context, _ *openfga.User, _ *dbmodel.RoleEntry) (*dbmodel.RoleEntry, error) {
+	return &dbmodel.RoleEntry{}, nil
+}
+
+// GetRoleByUUID returns a role based on the provided UUID.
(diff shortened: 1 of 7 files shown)


internal/jujuapi/role.go
```diff
@@ -0,0 +1,129 @@
...
+	const op = errors.Op("jujuapi.RemoveRole")
+
+	if err := r.jimm.RemoveRole(ctx, r.user, req.Name); err != nil {
+		zapctx.Error(ctx, "failed to remove role", zaputil.Error(err))
+		return errors.E(op, err)
+	}
+	return nil
+}
+
+// ListRole lists relational access control roles within JIMMs DB.
```
kian99: Godoc is a bit weird, why not just,
```suggestion
// ListRole lists access control roles within JIMMs DB.
```

pkg/api/params/params.go
kian99, line 1: All the docstrings in this file still mentioned `group` at the end of the sentence.

internal/jimm/role.go
```diff
@@ -0,0 +1,51 @@
...
+
+	"github.com/canonical/jimm/v3/internal/common/pagination"
+	"github.com/canonical/jimm/v3/internal/dbmodel"
+	"github.com/canonical/jimm/v3/internal/openfga"
+)
+
+// AddRole creates a role within JIMMs DB for reference by OpenFGA.
+func (j *JIMM) AddRole(ctx context.Context, user *openfga.User, name string) (*dbmodel.RoleEntry, error) {
+	return &dbmodel.RoleEntry{}, nil
+}
```
kian99, line 15: I guess you still need to hook all of these up to the DB methods. Will that be for this PR or a follow-up?

SimoneDutto: yes, the jimm implementation is for a follow-up.
So this 300lines skeleton pr is not distracting from the actual implementation.

kian99 (approved):
Looks fine but you might want to start from the bottom up, adding the `internal/jimm` logic first with tests then you can add the facade layer with tests. Because as it stands this PR doesn't have tests so I'm not sure if it should go in even though it's just skeleton.
I'll approve as I think it's fine this time, but worth thinking about.

SimoneDutto:
> Looks fine but you might want to start from the bottom up, adding the `internal/jimm` logic first with tests then you can add the facade layer with tests. Because as it stands this PR doesn't have tests so I'm not sure if it should go in even though it's just skeleton. I'll approve as I think it's fine this time, but worth thinking about.

yes, i did it just so because I was waiting on other prs to land before I could work on jimm's implementations.
Maybe an unnecessary optimization

SimoneDutto:
closing because @kian99 is right, it was an unnecessary PR. It is better to wait or doing the other way around it

PR #1453, opened 2024-11-20
refactor(internal/db): updates to GORM 2.0

alesstimec:
## Description

Updates to GORM 2.0.

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [ ] Covered by unit tests


go.mod
```diff
@@ -339,6 +339,10 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.4.1 // indirect
 )
 
+require github.com/jackc/pgx/v5 v5.7.1
```
ale8k: This should be in requires block and line 344 in indirect block

internal/db/audit.go
```diff
@@ -98,8 +98,12 @@ func (d *Database) ForEachAuditLogEntry(ctx context.Context, filter AuditLogFilt
 	if filter.SortTime {
 		db = db.Order("time DESC")
 	}
-	db = db.Limit(filter.Limit)
-	db = db.Offset(filter.Offset)
+	if filter.Limit > 0 {
```
SimoneDutto, line 101: in this way if somebody set it to 0 (or default zero-valued) it gets all the records. 
Here, which is the rebac_admin entrypoint, we set guards for max number of elements and we use "-1" for the default page size.
https://github.com/canonical/jimm/blob/ee16e51a426683210a6aa5b25459e03114fffdab/internal/common/pagination/pagination.go#L22-L36

internal/db/group.go
```diff
@@ -103,8 +103,12 @@ func (d *Database) ListGroups(ctx context.Context, limit, offset int, match stri
 		db = db.Where("name LIKE ? OR uuid LIKE ?", "%"+match+"%", "%"+match+"%")
 	}
 	db = db.Order("name asc")
-	db = db.Limit(limit)
```
kian99, line 106: An interesting question:
- I think this would've been okay because the limit and offset should be validated before they get to this function, but nothing wrong about defensive programming. But then should we also set an upper limit here or trust an upper limit was applied before reaching here?

internal/db/audit.go
alesstimec: i don't quite understand. If 0, all records should be returned. If somebody wants to limit the number of records returned they should set a limit that is higher than 0. Is that not correct?

kian99 (approved):
lgtm with a question. Also can we update the PR description to include all the changes.
And you mentioned we were already on Gorm v2 so the PR should reflect that we just bumped the Gorm version.

internal/db/group.go
alesstimec: the way it was written, limit would be set to 0 (the default value) and NO records would be returned. Unless we intentionally set a limit that was higher than 0.

internal/db/audit.go
SimoneDutto: we have two entrypoints for groups:
- jimmctl
- rebac_admin

For the rebac_admin implementation there is no way to return all records, we have a default and max page size.

Here with the jimmctl we are not doing that, we are defaulting to return all records.
Is that the correct approach?

alesstimec: yes

internal/db/group.go
SimoneDutto: i think that's because from the rebac endpoint they set the `pageSize`, and if they set `pageSize` to 0 that' fair to return 0 elements

internal/db/audit.go
SimoneDutto: Shouldn't we guard for a max size? I genuinely don't know what's the best approach to display 1000 elements in a CLI tool.
Pagination doesn't make much sense, but also returning many many records doesn't seem very smart.

SimoneDutto: i am not talking about groups, but more in general for all of our outputs

internal/db/group.go
kian99: Yeah I think returning everything is probably not correct, if we have pagination usually we'd have a max size.

internal/db/audit.go
alesstimec: but then we need to change jimmctl commands, as well.. followup as it exceeds the scope of this PR

internal/db/group.go
alesstimec: not all clients (e.g. jimm) use pagination

go.mod
```diff
@@ -42,7 +42,7 @@ require (
 	github.com/hashicorp/vault/api v1.13.0
 	github.com/hashicorp/vault/api/auth/approle v0.6.0
 	github.com/itchyny/gojq v0.12.12
-	github.com/jackc/pgconn v1.14.3
+	github.com/jackc/pgconn v1.14.3 // indirect
```
ale8k, line 45: This is now indirect, should be moved

internal/db/errors.go
```diff
@@ -13,6 +12,10 @@ import (
 // https://www.postgresql.org/docs/11/errcodes-appendix.html.
 const pgUniqueViolation = "23505"
 
+type pgError interface {
```
ale8k, line 15: Why are we creating this? Is it just so we don't need pgconn?

internal/db/group.go
ale8k: I'd imagine max limits should be per query, not an overall max. No?

ale8k (approved):
Bit worried about the limits, I'd imagine it's per query and not "no limit unless specified". I'll approve for now but can we discuss this?

internal/db/group.go
kian99: One thing, happy to check `> 0` and leave the other validation to the calling function since it makes sense that we need to check whether to apply the limit here at the DB layer. But I checked the `jimmctl auth group list` command and it does support pagination and sends `limit=0` by default and this actually works and lists all the groups.

SimoneDutto: for reference: juju returns everything when you do `Model`. As far as I've seen, there is no pagination.

https://github.com/juju/juju/blob/e3263a5bfcb34ede12f8a1601b9847b4d881a051/state/modeluser.go#L169-L243

alesstimec: @kian99 with this dependency update those jimmctl tests that were sending limit=0 started failing, because nothing was returned. If you want all entries to be returned you should be setting limit=-1

internal/db/errors.go
alesstimec: the updated gorm seems to return two different types of errors now.. this interface is here so that we don't need to care about the concrete type..we only ever call the one method on the two implementing types.

internal/db/group.go
kian99: Ahh those tests started failing, didn't realise. Then this change is perfect imo.

go.mod
alesstimec: the rule to sort direct vs indirect reports is not enforcable, unless we have some tooling that would actually sort imports for us.. we can't expect devs to sort imports if `go mod tidy` does not do it for us.

PR #1454, opened 2024-11-20
Fix websocket dial to ipv6

### LINKED ISSUE #1449: No support for dial via ipv6
I don't know if that's an issue or not, but my local tests it seems we are not supporting dialing via ipv6.
```
11:45:30.240	ERROR	BasicDial failed	{"websocket": true, "facade": "jujuapi.CreateModel", "error": "dial tcp: address fd42:791:fa5e:6834:216:3eff:fea4:81e6:17070: too many colons in address"}
```

SimoneDutto:
## Description

This pr fixes our websocket dial to ipv6.
The issue is that `url.URL` doesn't escape properly ipv6. 
See this golang issue: https://github.com/golang/go/issues/47763

Add some tests for both ipv4 and ipv6 dialing. Fly-by, pass down right context to dial to cancel it appropriately.

> [!NOTE] 
> httptest pkg doesn't have an handy `StartIPv6Sever`, so I've created one.

> [!WARNING] 
> dial via ipv6 inside of our docker-compose doesn't work because we don't have an ipv6 configured network.

Fixes: #1449 

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [ ] Covered by unit tests

### DIFF
--- a/internal/rpc/client_test.go
+++ b/internal/rpc/client_test.go
@@ -7,6 +7,7 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"encoding/json"
+	"net"
 	"net/http"
 	"net/http/httptest"
 	"strings"
@@ -495,6 +496,27 @@ type server struct {
 func newServer(f func(*websocket.Conn) error) *server {
 	var srv server
 	srv.Server = httptest.NewTLSServer(handleWS(f))
+	srv.URL = "ws" + strings.TrimPrefix(srv.Server.URL, "http")
+	cp := x509.NewCertPool()
+	cp.AddCert(srv.Certificate())
+	srv.dialer = &rpc.Dialer{
+		TLSConfig: &tls.Config{
+			RootCAs:    cp,
+			MinVersion: tls.VersionTLS12,
+		},
+	}
+	return &srv
+}
+
+func newIPv6Server(f func(*websocket.Conn) error) *server {
+	var srv server
+	l, _ := net.Listen("tcp", "[::1]:0")
+	server := httptest.Server{
+		Listener: l,
+		Config:   &http.Server{Handler: handleWS(f)}, //nolint:gosec
+	}
+	server.StartTLS()
+	srv.Server = &server
 	srv.URL = "ws" + strings.TrimPrefix(srv.Server.URL, "http")
 	cp := x509.NewCertPool()
 	cp.AddCert(srv.Certificate())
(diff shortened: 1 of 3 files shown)


internal/rpc/client_test.go
```diff
@@ -507,6 +508,27 @@ func newServer(f func(*websocket.Conn) error) *server {
...
+func newIPv6Server(f func(*websocket.Conn) error) *server {
+	var srv server
+	l, _ := net.Listen("tcp", "[::1]:0")
+	server := httptest.Server{
+		Listener: l,
+		Config:   &http.Server{Handler: handleWS(f)}, //nolint:gosec
+	}
+	server.StartTLS()
+	srv.Server = &server
+	srv.URL = "ws" + strings.TrimPrefix(srv.Server.URL, "http")
```
alesstimec, line 520: wss?

internal/rpc/dial _test.go
```diff
@@ -0,0 +1,65 @@
+// Copyright 2024 Canonical.
+package rpc_test
```
alesstimec, line 3: new line between copyright and package name, please

internal/rpc/dial.go
```diff
@@ -89,7 +89,13 @@ func Dial(ctx context.Context, ctl *dbmodel.Controller, modelTag names.ModelTag,
 	for _, hps := range ctl.Addresses {
 		for _, hp := range hps {
 			if maybeReachable(hp.Scope) {
-				urls = append(urls, websocketURL(fmt.Sprintf("%s:%d", hp.Value, hp.Port), modelTag, finalPath))
+				var ip string
+				if hp.Type == string(network.IPv6Address) {
```
alesstimec, line 93: suggestion: 
```
addr := net.UDPAddr{
  IP: net.ParseIP(hp.Value),
  Port: hp.Port,
}
ip := addr.String()
```

SimoneDutto: the issue is that we have 3 types:
- ipv4
- ipv6
- hostname

Would this work for hostname?

internal/rpc/client_test.go
SimoneDutto: `<http>s://google.com` -> replace `http` with `ws` -> `wss://google.com`

this code works both for HTTP and https url

kian99 (approved):
lgtm

ale8k (approved):
Lgtm

PR #1455, opened 2024-11-20
Juju 7124/implement jimm role package

ale8k:
## Description
Implements the RoleManager. I've opted to change this package to whitebox testing as we're testing a single unit of JIMM.

You'll see only the constructor is public and the struct defines interfaces to interact with. Additionally, you'll see within the service package and jimm are interfaces to consume this RoleManager. 

To prevent an import cycle, I've moved some jimmtest files into their own packages. When plugging in the rolemanager, and attempting to whitebox test, we had an import cycle as the test used jimmtest, jimmtest imports service and service uses rolemanager. Breaking these testing utilities up into their own packages solves this, allowing us to whitebox test without requiring ALL of jimmtest and all of it's imported dependencies. I hope I've described why I've done this sufficiently.  If we do like this, I'll do a follow up PR which deletes gorm.go, cmp.go, openfga.go and uses the broken down testpackages. 

Additionally, you will see in each manager method test, we don't rely on other methods provided by the manager and interact with store and authSvc directly - this is intentional such that tests are isolated.

Finally, I've had them run in parallel using qt (just a little QoL).

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [ ] Covered by unit tests


internal/jimm/role/role.go
```diff
@@ -0,0 +1,173 @@
...
+
+	"github.com/canonical/jimm/v3/internal/common/pagination"
+	"github.com/canonical/jimm/v3/internal/dbmodel"
+	"github.com/canonical/jimm/v3/internal/errors"
+	"github.com/canonical/jimm/v3/internal/openfga"
+	jimmnames "github.com/canonical/jimm/v3/pkg/names"
+)
+
+// RoleStore defines a storage interface to persist roles.
+type RoleStore interface {
```
kian99: Do we need a role store? Just thinking about what it purpose it serves, we don't expect to have multiple implementations of a RoleStore and the tests would work fine against a real Postgres store. Need to understand if this is useful beyond just an interface for the sake of making an interface.

On a second thought, this interface limits what the RoleManager can do which is nice.

internal/jimm/role/role.go
```diff
@@ -0,0 +1,173 @@
...
+		Name: roleName,
+	}
+	err := rm.store.GetRole(ctx, re)
+	if err != nil {
+		return errors.E(op, err)
+	}
+
+	// TODO(ale8k):
+	// Would be nice to have a way to create a transaction to get, remove tuples, if successful, delete role
+	// somehow. We could pass a callback and change the db methods?
```
kian99, line 86: This probably deserves a Jira card rather than a todo comment. Especially because it's a larger piece of work related to distributed transactions effort we went through previously.

internal/jimm/role/role_test.go
```diff
@@ -0,0 +1,238 @@
...
+	"github.com/canonical/jimm/v3/internal/dbmodel"
+	"github.com/canonical/jimm/v3/internal/errors"
+	"github.com/canonical/jimm/v3/internal/openfga"
+	ofganames "github.com/canonical/jimm/v3/internal/openfga/names"
+	jimmtestcmp "github.com/canonical/jimm/v3/internal/testutils/jimmtest/cmp"
+	jimmtestgorm "github.com/canonical/jimm/v3/internal/testutils/jimmtest/gorm"
+	jimmtestopenfga "github.com/canonical/jimm/v3/internal/testutils/jimmtest/openfga"
+)
+
+type roleManagerSuite struct {
```
kian99, line 26: Really like the suite

internal/testutils/jimmtest/cmp/cmp.go
kian99, line 1: I wonder if it wouldn't be better to remove the dependency that jimmtest has on service rather than moving these all to separate packages.

internal/jimm/jimm.go
```diff
@@ -86,6 +87,28 @@ type JIMM struct {
 	// OAuthAuthenticator is responsible for handling authentication
 	// via OAuth2.0 AND JWT access tokens to JIMM.
 	OAuthAuthenticator OAuthAuthenticator
+
+	// RoleManager provides a means to manage roles within JIMM.
+	RoleManager RoleManager
```
kian99, line 92: Just wondering if this should go directly on JIMM or on a field like `app` or `business`.

kian99 (commented):
Looks really good, thanks for this. Just some comments

internal/jimm/role/role.go
ale8k: Yeah it's just another example where I write the code and feel a bit helpless.

internal/jimm/jimm.go
ale8k: Uhmm I'm not against it actually

internal/jimm/role/role.go
ale8k: Yup, which is really how we should be using interfaces. To consume structs.

internal/jimm/role/role_test.go
ale8k: I just wish it had a before/after suite and before/after test, appears it just an "Init"? So perhaps qt isn't for us or we talk to Fran?

cmd/jimmsrv/service/service.go
```diff
@@ -398,6 +399,12 @@ func NewService(ctx context.Context, p Params) (*Service, error) {
 		JWTService:                 s.jimm.JWTService,
 	}
 
+	roleManager, err := role.NewRoleManager(&s.jimm.Database, s.jimm.OpenFGAClient)
+	if err != nil {
+		return nil, errors.E(op, err, "failed to create RoleManager")
+	}
+	s.jimm.RoleManager = roleManager
+
```
SimoneDutto, line 407: i don't know if I am sold on this.

This is inconsistent with the common pattern we've used all around the code. It's a start going towards breaking up JIMM, but I would've preferred this in a separate PR

internal/jimm/role/role.go
```diff
@@ -0,0 +1,173 @@
...
+		return nil, errors.E("role store cannot be nil")
+	}
+	if authSvc == nil {
+		return nil, errors.E("role authorisation service cannot be nil")
+	}
+	return &roleManager{store, authSvc}, nil
+}
+
+// AddRole adds a role to JIMM.
+func (rm *roleManager) AddRole(ctx context.Context, user *openfga.User, roleName string) (*dbmodel.RoleEntry, error) {
```
SimoneDutto, line 36: nit: i would prefer `rlM` just because `rm` down the code seems the UNIX `rm`

internal/jimm/role/role_test.go
```diff
@@ -0,0 +1,236 @@
...
+	c.Assert(err, qt.IsNil)
+
+	// Create test identity
+	i, err := dbmodel.NewIdentity("alice")
+	c.Assert(err, qt.IsNil)
+	s.user = openfga.NewUser(i, ofgaClient)
+}
+
+func (s *roleManagerSuite) TestAddRole(c *qt.C) {
+	c.Parallel()
```
SimoneDutto, line 59: what's this?

internal/jimm/role/role_test.go
```diff
@@ -0,0 +1,236 @@
...
+	c.Assert(errors.ErrorCode(err), qt.Equals, errors.CodeUnauthorized)
+
+	// Check happypath
+	s.user.JimmAdmin = true
+
+	// Add to the db and openfga.
+	re, err := s.db.AddRole(ctx, "r1")
+	c.Assert(err, qt.IsNil)
+
+	tupe := openfga.Tuple{
```
SimoneDutto, line 117: -> `tuple`

internal/jimm/role/role.go
```diff
@@ -0,0 +1,173 @@
...
+	// RemoveRole removes a role.
+	RemoveRole(ctx context.Context, role jimmnames.RoleTag) error
+}
+
+// roleManager provides a means to manage roles within JIMM.
+type roleManager struct {
+	store   RoleStore
+	authSvc RoleAuthorisationService
+}
+
```
SimoneDutto: thinking out loud:
this is good, very interfacy, i am just worried the boilerplate we are imposing upon us with this new pattern. 

I would have love to discuss this new pattern in a separate PR

SimoneDutto (commented):
The code is fine.

But I would have love to discuss this new pattern in a separate PR to decide all together the direction.
If we merge now, it's very inconsistent because we only have one "manager" on JIMM, and that's the Role manager. It doesn't make sense to be alone.

cmd/jimmsrv/service/service.go
ale8k: We have jwkservice, oauthauthenticator, etc. They're all the same pattern.

internal/jimm/role/role_test.go
ale8k: Allows the tests to run in parallel :) bit of a premature optimisation.

internal/jimm/role/role_test.go
ale8k: I really love the word "tupe" I can't deny aha

internal/jimm/role/role.go
ale8k: This is generally the best way to write go:
- never return interfaces
- create interfaces on consuming side to determine what methods are accessed within your service

This is why in the jujuclient PoC, I taken issue when I had to return interfaces

internal/jimm/role/role_test.go
SimoneDutto: if it works just like that is cool!

internal/jimm/role/role.go
ale8k: Not sure if this really matters so I'll ignore this one if that's ok?

internal/jimm/role/role_test.go
SimoneDutto: is the `init` run one single time? Or for each test? Do they share the same db?

ale8k: Well I never checked if Fran plugged it in but I'm presuming it's calling underlying TB parallel? Can check?

internal/jimm/role/role.go
SimoneDutto: yy, for sure. Just a thing I had on my mind

internal/jimm/role/role_test.go
SimoneDutto: My question is: can we just use it without worrying about test race conditions?

ale8k: I believe each test but need to trawl down a bit, I just tested this via creating new tests against the same object. Fran's quicktest doesn't have the before/after suite/test methods, just an init... Which I guess is another discussion we need to have.

SimoneDutto: Or do they share the same db instance, so should we be careful?

ale8k: In any case we create db by name, so it's safe - I don't believe they're separate instances if that's what you mean?

SimoneDutto: If the `init` is called before each test, then they don't share the same db. 
But if they use the same `s.db` they share same db and table. 
Right? 
If the latter running tests in parallel is not trivial

internal/jimm/role/role.go
```diff
@@ -0,0 +1,173 @@
+// Copyright 2024 Canonical.
+package role
```
alesstimec, line 3: new line, please

internal/jimm/role/role_test.go
```diff
@@ -0,0 +1,236 @@
+// Copyright 2024 Canonical.
+package role
```
alesstimec: new line, please

internal/jimm/role/role.go
alesstimec: if we are not mocking this interface in our units test, why do we need it exactly?

internal/jimm/role/role_test.go
```diff
@@ -0,0 +1,236 @@
...
+type roleManagerSuite struct {
+	manager    *roleManager
+	user       *openfga.User
+	db         *db.Database
+	ofgaClient *openfga.OFGAClient
+}
+
+func (s *roleManagerSuite) Init(c *qt.C) {
+	// Setup DB
+	db := &db.Database{
```
alesstimec, line 35: since we have the RoleStore inteface, shouldn't we be mocking it in out unit tests?

internal/jimm/jimm.go
alesstimec: i think we should just embed the RoleManager type in JIMM

cmd/jimmsrv/service/service.go
```diff
@@ -398,6 +399,12 @@ func NewService(ctx context.Context, p Params) (*Service, error) {
 		JWTService:                 s.jimm.JWTService,
 	}
 
+	roleManager, err := role.NewRoleManager(&s.jimm.Database, s.jimm.OpenFGAClient)
```
alesstimec, line 402: i propose we have a constructor of JIMM, so that i can do `jimm.New(params jimm.Params)`.
We often create an instance of JIMM in our tests, which would now be incomplete without us manually creating the roleManager instance and then setting it on JIMM.

internal/jimm/role/role_test.go
```diff
@@ -0,0 +1,236 @@
...
+	"github.com/canonical/jimm/v3/internal/db"
+	"github.com/canonical/jimm/v3/internal/dbmodel"
+	"github.com/canonical/jimm/v3/internal/errors"
+	"github.com/canonical/jimm/v3/internal/openfga"
+	ofganames "github.com/canonical/jimm/v3/internal/openfga/names"
+	jimmtest "github.com/canonical/jimm/v3/internal/testutils/jimmtest"
+)
+
+type roleManagerSuite struct {
+	manager    *roleManager
```
alesstimec: also we don't necessarily need this field on the suite. Calling the NewRoleManager constructor takes one line in each test and we're not using any of the role package internals really so i don't think this is a compelling reason for whitebox testing here..

internal/jimm/role/role.go
kian99: I'm not sold on using an interface just to limit the scope of a struct. Say we passed in a DB struct instead of a RoleStore. Then the RoleManager _could_ do more things. 
Is it worth it to have an interface to clearly describe what DB operations the RoleManager has access to?
I lean towards yes, but it also feels okay to provide the role manager with raw DB access.
This interface basically feels like a form of documentation (and potentially lets you test with a fake DB if desired).

internal/jimm/role/role_test.go
kian99: before test would just be stuff in the suite? And after test you can add a test.Cleanup() in the suite I believe.
Before/after suite can be achieved with a TestMain() if you need.

internal/jimm/jimm.go
ale8k: The way it is now?...

internal/jimm/role/role.go
ale8k: It limits the methods exposed to the RoleManager

internal/jimm/role/role_test.go
ale8k: Yes! But I haven't for this as to not stray away massively from what we do currently. I want us to reach a consensus on mocking and how small services like this should look.

cmd/jimmsrv/service/service.go
kian99: I agree that it's unfortunate it's so inconsistent with the rest of the business logic. I think Simone is also thinking about the other jimm methods that form all the business logic and how this is very different from that.
Though of course we can get those to look more like this going forward.

internal/jimm/role/role_test.go
ale8k: This is a simple example, but some services may benefit from testing private functions for edge behaviour or perhaps there are variables that can be tweaked. Things like this.

internal/jimm/role/role_test.go
ale8k: From the doc:
```
// If suite has a method of the form
//
//	func (T) Init(*quicktest.C)
//
// this method will be invoked before each test run.
```

SimoneDutto: thank you!

SimoneDutto: so the c.parallel should be kind of a "free lunch"

kian99: Tests from different packages are already run in parallel from what I understand btw. For more info
https://stackoverflow.com/a/72354980

internal/jimm/role/role_test.go
kian99: I think better to do whitebox testing on a case by case basis where needed. Here, as in most of the rest of the code, my suggestion would be to keep it blackbox and using a role_testing package. I liked the idea of being deliberate with how you structure these black vs white tests because it prevents you from accidentally using something that is not exposed and feels like better practice.

cmd/jimmsrv/service/service.go
ale8k: I think a different PR for this - and totally agree!

alesstimec: can do in a follow-up.. please create a jira ticket so we don’t forget.. thanks

ale8k: Created card

cmd/jimmsrv/service/service.go
alesstimec: i would love to split up the jimm package into smaller services that are easier to test

alesstimec (approved):
LGTM

internal/jimm/role/role.go
```diff
@@ -0,0 +1,161 @@
...
+		return errors.E(op, err)
+	}
+
+	return nil
+}
+
+// RenameRole renames a role in JIMM's DB.
+func (rm *roleManager) RenameRole(ctx context.Context, user *openfga.User, uuid, newName string) error {
+	const op = errors.Op("roleManager.RenameRole")
+	zapctx.Debug(ctx, fmt.Sprintf("Running %s", op))
```
kian99, line 101: I suggest making these info level and including the name of the role that's being changed. I doubt they'd be hit often enough to be too noisy. But open to leaving it at debug too.

kian99 (approved):

PR #1456, opened 2024-11-21
chore: remove dependency from test package

kian99:
## Description

Remove a dependency the testutils/jimmtest package has on github.com/canonical/jimm/v3/cmd/jimmsrv/service

Realised this dependency shouldn't exist after reading #1455.
The function I've removed was a helper to create a `jimmsvc.Params` object. Used in 3 places:
- cmd/jimmsrv/service/service_test.go: Added a copy here, makes sense for this test to start a service.
- internal/jimmjwx/utils_test.go: Added a copy here, this test shouldn't need to start a jimm service, will refactor in a follow-up PR.
- internal/testutils/cmdtest/jimmsuite.go: Added a copy here, this makes sense as the package is creating a test jimm service for cmd tests to run against.


kian99:
@ale8k This change should make #1455 easier to land.

ale8k (approved):
ty!

cmd/jimmsrv/service/service_test.go
```diff
@@ -36,12 +38,33 @@ func TestMain(m *testing.M) {
 	os.Exit(code)
 }
 
+// newTestJimmParams returns a set of JIMM params with sensible defaults
+// for tests. A test can override any parameter that it needs.
+// Note that newTestJimmParams will create an empty test database.
+func newTestJimmParams(t jimmtest.Tester) jimmsvc.Params {
```
SimoneDutto, line 44: why not exporting this function and using it all around? As service_test configuration?

SimoneDutto: ignore my comment

SimoneDutto (approved):

PR #1457, opened 2024-11-21
chore: simplify jwx tests

kian99:
## Description

In #1148 I refactored the JWKS cache, previously the cache worked by making an HTTP to call to jimm from jimm which is why this test creates a server. That was changed to remove the circular dependency but the tests weren't updated, this finally does that.

This PR builds on #1456. View the 2nd commit to see only the changes here.

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [x] Covered by unit tests

### DIFF
--- a/internal/jimmjwx/jwt_test.go
+++ b/internal/jimmjwx/jwt_test.go
@@ -3,7 +3,6 @@ package jimmjwx_test
 
 import (
 	"context"
-	"net/url"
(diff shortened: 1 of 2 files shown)


ale8k (approved):

SimoneDutto (approved):
lgtm

PR #1458, opened 2024-11-22
Juju 7209/remove non essentials model

SimoneDutto:
## Description

In this PR we are slimming JIMM model dbmodel.
So:
- remove the field from the model struct all around code and test
- migration sql
- refactor `ListModelSummaries` to gather info from the controllers instead of jimm db

What's missing:
- removing controller watcher
- handle `Life` with a model watcher

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [ ] Covered by unit tests


internal/jimm/watcher_test._go
SimoneDutto, line 1: i've renamed this file, because i'll handle watcher changes in a separate pr. And it wouldn't make sense to work to fix the tests to remove them the PR after it

internal/jimm/model.go
```diff
@@ -748,16 +749,103 @@ func (j *JIMM) ModelInfo(ctx context.Context, user *openfga.User, mt names.Model
...
+	for _, m := range models {
+		modelSummaryFromController, ok := modelSummariesSafeMap.modelSummaries[m.model.UUID.String]
+		modelSummaryResult := m.model.MergeModelSummaryFromController(modelSummaryFromController.Result, maskingControllerUUID, m.userAccess)
+		if modelSummaryFromController.Error != nil {
+			modelSummaryResults = append(modelSummaryResults, jujuparams.ModelSummaryResult{
+				Result: &modelSummaryResult,
+				Error:  modelSummaryFromController.Error,
+			})
+			continue
+		}
```
SimoneDutto, line 824: when you add an error the error message (customizable) is displayed on top of the models like so:
image

internal/jimm/model.go
```diff
@@ -748,16 +749,103 @@ func (j *JIMM) ModelInfo(ctx context.Context, user *openfga.User, mt names.Model
...
+		if !ok {
+			// if model was not found in any controller we mark it as anavailable
+			modelSummaryResult.Status.Status = "unavailable"
```
SimoneDutto, line 827: if you change the status, it is displayed like this:
image

internal/dbmodel/model.go
```diff
@@ -115,10 +84,7 @@ func (m *Model) SetOwner(u *Identity) {
 // will need to be filled in manually by the caller of this function.
 func (m *Model) FromJujuModelInfo(info jujuparams.ModelInfo) error {
```
alesstimec, line 85: do we still need this method?

alesstimec: and other methods that transform to/from juju types..

internal/dbmodel/sql/postgres/1_14.sql
```diff
@@ -1,4 +1,4 @@
--- 1_13.sql is a migration simplifies application offers
+-- 1_14.sql is a migration simplifies application offers
```
alesstimec, line 1: thanks :)

internal/dbmodel/sql/postgres/1_15.sql
```diff
@@ -0,0 +1,7 @@
+-- 1_15.sql remove non essential fields from model.
+ALTER TABLE models DROP COLUMN default_series, DROP COLUMN migration_controller_id, DROP COLUMN is_controller, DROP COLUMN cores,
```
alesstimec, line 2: a good discussion point: since we're removing field from our dbmodel, it wouldn't hurt if we left these columns in the db for the initial release. If anything goes wrong during the update of an existing JIMM, leaving data in the db would mean we can still roll back. If we remove the data in the same step, we run the risk of not being able to roll back. I did the same in the app offers PR.. but i think this is something we need to be aware of - that we're burning the bridges behind us.

internal/jimm/model.go
```diff
@@ -748,16 +749,103 @@ func (j *JIMM) ModelInfo(ctx context.Context, user *openfga.User, mt names.Model
...
+		model      *dbmodel.Model
+		userAccess jujuparams.UserAccessPermission
+	}
+	// we collect models belonging to the user.
+	err := j.ForEachUserModel(ctx, user, func(m *dbmodel.Model, uap jujuparams.UserAccessPermission) error {
+		models = append(models, struct {
+			model      *dbmodel.Model
+			userAccess jujuparams.UserAccessPermission
+		}{model: m, userAccess: uap})
+		return nil
```
alesstimec, line 792: you could assemble the list of unique controllers in this loop.. no need to have a separate for loop that ranges over models

internal/jimm/model.go
```diff
@@ -748,16 +749,103 @@ func (j *JIMM) ModelInfo(ctx context.Context, user *openfga.User, mt names.Model
...
+	})
+	if err != nil {
+		// we log the error and continue, because even if one controller is not reachable we are still able to fill the response.
+		zapctx.Error(ctx, "Error querying the controllers for model summaries", zap.Error(err))
+	}
+
+	// we map models to modelsummaries
+	for _, m := range models {
+		modelSummaryFromController, ok := modelSummariesSafeMap.modelSummaries[m.model.UUID.String]
```
alesstimec, line 816: what `if !ok`.. if the model UUID is not in the `modelSummaries` map?

internal/jimm/model.go
```diff
@@ -748,16 +749,103 @@ func (j *JIMM) ModelInfo(ctx context.Context, user *openfga.User, mt names.Model
...
+		modelSummaryResult := m.model.MergeModelSummaryFromController(modelSummaryFromController.Result, maskingControllerUUID, m.userAccess)
```
alesstimec, line 817: why do we need to merge the data? shouldn't each controller be the authoritative source of information about model summaries?

internal/jimm/model.go
```diff
@@ -748,16 +749,103 @@ func (j *JIMM) ModelInfo(ctx context.Context, user *openfga.User, mt names.Model
...
+		if modelSummaryFromController.Error != nil {
+			modelSummaryResults = append(modelSummaryResults, jujuparams.ModelSummaryResult{
+				Result: &modelSummaryResult,
+				Error:  modelSummaryFromController.Error,
+			})
+			continue
+		}
+		if !ok {
```
alesstimec, line 825: should we not check sooner? before we try to merge data?

internal/jimm/watcher.go
```diff
@@ -343,9 +343,9 @@ func (w *Watcher) watchController(ctx context.Context, ctl *dbmodel.Controller)
 						machines++
 						cores += n
 					}
-					m.Cores = cores
-					m.Machines = machines
-					m.Units = int64(len(v.units))
+					// m.Cores = cores
```
alesstimec, line 346: remove?

internal/jujuclient/modelmanager.go
```diff
@@ -208,6 +208,22 @@ func (c Connection) ControllerModelSummary(ctx context.Context, ms *jujuparams.M
 	return errors.E(op, "controller model not found", errors.CodeNotFound)
 }
 
+// ListModelSummaries retrieves the list of model summaries from the controler
+func (c Connection) ListModelSummaries(ctx context.Context, ms jujuparams.ModelSummariesRequest) (jujuparams.ModelSummaryResults, error) {
+	const op = errors.Op("jujuclient.ControllerModelSummary")
+	args := jujuparams.ModelSummariesRequest{
+		UserTag: c.userTag,
```
alesstimec, line 215: is this correct? shouldn't we just forward the received request?

internal/dbmodel/model.go
SimoneDutto: we still need them because we are still merging info coming from the controller with the info that we have.
For example:
we take info coming from controllers, we change the OwnerIdentity, CloudCredentials. 
And then we return this to the client.

internal/dbmodel/sql/postgres/1_15.sql
SimoneDutto: great comment, it's difficult to solve.

An idea could be to migrate always in <release+1>, so you release in the wild, everything is fine and then you create a pr to "burn the bridge".

internal/jimm/model.go
SimoneDutto: the controller is the source, but there is still info from our db that we want to propagate.
Ex. OwnerIdentity, CloudRegion, CloudCredentials

internal/jimm/model.go
SimoneDutto: it depends. The two screenshots I've posted are alternatives.

the first alternative (the one implemented in the code here), is to return the info we have from our db (name, uuid, etc), and fill the summaries table with status `unavailable`.
https://github.com/canonical/jimm/pull/1458#discussion_r1860240670

the second is to put this model missing as an `Error`, and you can have your custom message (ex. `model-1 is not found` on top of the table entries.
https://github.com/canonical/jimm/pull/1458#discussion_r1860238548

internal/jimm/watcher.go
SimoneDutto: i've done this on purpose, because i didn't remove the watcher yet, but I don't want them to update these fields to avoid tests being green because the watcher is still doing its job.

internal/jujuclient/modelmanager.go
SimoneDutto: with the `all` flag set to true, you are asking for all model summaries but you need to be admin. And the controller user is admin for sure.
Right?

internal/dbmodel/sql/postgres/1_15.sql
ale8k: Well we'd be reverting back to a poorly tested versions nonetheless which could have its own set of issues in the wild right?

internal/jimm/model.go
```diff
@@ -748,16 +749,99 @@ func (j *JIMM) ModelInfo(ctx context.Context, user *openfga.User, mt names.Model
 	return j.mergeModelInfo(ctx, user, mi, m)
 }
 
+type modelSummariesSafeMap struct {
```
kian99: I suggest dropping `safe` in the name. Also add a docstring describing what they key represents or uses a better type than string if possible.
```suggestion
type modelSummariesMap struct {
```

kian99 (approved):
Just one minor comment, otherwise I think this looks good. I'm happy with the approach to return status "unvailable" for now.

internal/dbmodel/sql/postgres/1_15.sql
alesstimec: @ale8k that is not a valid point: if the current version is poorly tested that is our own fault. I'm just saying that we need to be mindful of potential future migrations and give our users the ability to roll back.

internal/jimm/model.go
alesstimec: no?

internal/jimm/model.go
alesstimec: then perhaps the method to merge data should live in this package instead of dbmodel.

internal/jujuclient/modelmanager.go
alesstimec: is `all` set to true or to `ms.All` which might not be set to true?

internal/jimm/model.go
SimoneDutto: i did it, just forgot to resolve the comment. Sorry

internal/jimm/model.go
SimoneDutto: if `!ok` the model source from controller is set to nil, and the info we will send the client will be populated only from our db.

internal/jimm/model.go
SimoneDutto: i prefer the idea of having the data model transformation in the dbmodel pkg. 
But i can move them if you prefer to have it elsewhere!

internal/jujuclient/modelmanager.go
SimoneDutto: from the JIMM pkg is called with the flag set to true. Are you suggesting to ignore `ms.All` and set it to true here?

alesstimec (approved):

PR #1459, opened 2024-11-25
Juju 7125/add jimm facade for role

SimoneDutto:
## Description

Add facade methods for roles and some tests.

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [ ] Covered by unit tests


internal/db/role.go
```diff
@@ -80,7 +80,7 @@ func (d *Database) UpdateRoleName(ctx context.Context, uuid, name string) (err e
 	defer servermon.ErrorCounter(servermon.DBQueryErrorCount, &err, string(op))
 
 	model := d.DB.WithContext(ctx).Model(&dbmodel.RoleEntry{})
-	model.Where("uuid = ?", uuid)
+	model.Where("name = ?", oldname)
```
kian99: I don't see a test for this updated.

internal/jujuapi/controllerroot.go
```diff
@@ -51,6 +51,7 @@ type JIMM interface {
 	GetCloudCredential(ctx context.Context, user *openfga.User, tag names.CloudCredentialTag) (*dbmodel.CloudCredential, error)
 	GetCloudCredentialAttributes(ctx context.Context, u *openfga.User, cred *dbmodel.CloudCredential, hidden bool) (attrs map[string]string, redacted []string, err error)
 	GetCredentialStore() credentials.CredentialStore
+	GetRoleManager() jimm.RoleManager
```
kian99, line 54: Ah now we've walked into an interesting issue here. I don't think this is what we want, but I guess you can't add the methods here because the JIMM struct doesn't implement any role related methods. Maybe embed the RoleManager on JIMM then you can put actual methods here instead.

internal/jujuapi/role.go
```diff
@@ -0,0 +1,129 @@
...
+		zapctx.Error(ctx, "failed to rename role", zaputil.Error(err))
+		return errors.E(op, err)
+	}
+	return nil
+}
+
+// RemoveRole removes a role within JIMMs DB for reference by OpenFGA.
+func (r *controllerRoot) RemoveRole(ctx context.Context, req apiparams.RemoveRoleRequest) error {
+	const op = errors.Op("jujuapi.RemoveRole")
+
```
kian99, line 92: I notice we check the role name is valid in rename role but we don't here. And I get that's because you're checking the new name is valid, but just wondering if we should check the requested role name to be removed to ensure it's valid.

Also applies to GetRole.

pkg/api/params/params.go
```diff
@@ -379,6 +379,64 @@ type ListRelationshipTuplesResponse struct {
...
+
+// Role holds the details of a role currently residing in JIMM.
+type Role struct {
+	UUID      string `json:"uuid" yaml:"uuid"`
+	Name      string `json:"name" yaml:"name"`
+	CreatedAt string `json:"created_at" yaml:"created_at"`
+	UpdatedAt string `json:"updated_at" yaml:"updated_at"`
+}
+
+// ListRoleResponse returns the role tuples currently residing within OpenFGA.
```
kian99: I wouldn't put implementation details in this godoc. Also it would be roles in the DB no?
Just `ListRoleResponse contains a list of roles`.

pkg/api/params/params.go
```diff
@@ -379,6 +379,64 @@ type ListRelationshipTuplesResponse struct {
...
+	// Name holds the name of the role.
+	Name string `json:"name"`
+}
+
+type ListRolesRequest struct {
+	Limit  int `json:"limit"`
+	Offset int `json:"offset"`
+}
+
+// Role holds the details of a role currently residing in JIMM.
```
kian99:
```suggestion
// Role holds the details of a role currently.
```

pkg/api/params/params.go
```diff
@@ -379,6 +379,64 @@ type ListRelationshipTuplesResponse struct {
...
+	NewName string `json:"new-name"`
+}
+
+// RemoveRoleRequest holds a request to remove a role.
+type RemoveRoleRequest struct {
+	// Name holds the name of the role.
+	Name string `json:"name"`
+}
+
+type ListRolesRequest struct {
```
kian99, line 423: missing godoc

internal/jujuapi/controllerroot.go
SimoneDutto: For auth service, credentialstore we've resorted to this `GetThing()` method, so I've c/ped.

@ale8k what do you think?

ale8k: Yeah.. Not keen on this either really but we use a JIMM interface - I think it's a step up than calling methods directly though and the contract is still honoured. I'd say do this for now then let's discuss the humongous JIMM interface?

internal/db/role.go
```diff
@@ -64,10 +64,10 @@ func (d *Database) GetRole(ctx context.Context, role *dbmodel.RoleEntry) (err er
 }
 
 // UpdateRoleName updates the name of a role identified by UUID.
-func (d *Database) UpdateRoleName(ctx context.Context, uuid, name string) (err error) {
+func (d *Database) UpdateRoleName(ctx context.Context, oldname, name string) (err error) {
```
alesstimec:
```suggestion
func (d *Database) UpdateRoleName(ctx context.Context, oldName, name string) (err error) {
```

internal/db/role.go
```diff
@@ -64,10 +64,10 @@ func (d *Database) GetRole(ctx context.Context, role *dbmodel.RoleEntry) (err er
...
 	const op = errors.Op("db.UpdateRole")
 
-	if uuid == "" {
+	if oldname == "" {
 		return errors.E(op, "uuid must be specified")
```
alesstimec: if we are changing this from uuid to name, this error message should also be replaced

internal/jimm/jimm.go
```diff
@@ -131,6 +131,10 @@ func (j *JIMM) AuthorizationClient() *openfga.OFGAClient {
 	return j.OpenFGAClient
 }
 
+func (j *JIMM) GetRoleManager() RoleManager {
+	return j.RoleManager
```
alesstimec, line 135: does RoleManager still need to be exported now that we have a method to get it?

alesstimec: also.. if `j.RoleManager == nil`, what do we do? return nil here, or an error?

internal/jujuapi/role.go
```diff
@@ -0,0 +1,135 @@
...
+	"github.com/canonical/jimm/v3/internal/common/pagination"
+	"github.com/canonical/jimm/v3/internal/dbmodel"
+	"github.com/canonical/jimm/v3/internal/errors"
+	"github.com/canonical/jimm/v3/internal/openfga"
+	apiparams "github.com/canonical/jimm/v3/pkg/api/params"
+	jimmnames "github.com/canonical/jimm/v3/pkg/names"
+)
+
+type RoleManager interface {
```
alesstimec: godoc, please

alesstimec: also.. i don't see this interface being used at all.. are you sure we need it?

internal/jujuapi/role_test.go
```diff
@@ -0,0 +1,202 @@
...
+	_, err := db.AddRole(ctx, "test-role2")
+	c.Assert(err, gc.IsNil)
+
+	role := &dbmodel.RoleEntry{
+		Name: "test-role2",
+	}
+	err = db.GetRole(ctx, role)
+	c.Assert(err, gc.IsNil)
+
+	tuples := []openfga.Tuple{
```
alesstimec: reduces indentation
```
	tuples := []openfga.Tuple{{
		Object:   ofganames.ConvertTag(user.ResourceTag()),
		Relation: ofganames.AssigneeRelation,
		Target:   ofganames.ConvertTag(role.ResourceTag()),
	},{
		Object:   ofganames.ConvertTagWithRelation(role.ResourceTag(), ofganames.AssigneeRelation),
		Relation: "administrator",
		Target:   ofganames.ConvertTag(controller.ResourceTag()),
	},{
		Object:   ofganames.ConvertTagWithRelation(role.ResourceTag(), ofganames.AssigneeRelation),
		Relation: "writer",
		Target:   ofganames.ConvertTag(model.ResourceTag()),
	}}
```

internal/jujuapi/role_test.go
```diff
@@ -0,0 +1,202 @@
...
+	client := api.NewClient(conn)
+	res, err := client.AddRole(&apiparams.AddRoleRequest{Name: "test-role"})
+	c.Assert(err, jc.ErrorIsNil)
+	c.Assert(res.UUID, gc.Not(gc.Equals), "")
+
+	_, err = client.AddRole(&apiparams.AddRoleRequest{Name: "test-role"})
+	c.Assert(err, gc.ErrorMatches, ".*already exists.*")
+}
+
+func (s *accessControlSuite) TestGetRole(c *gc.C) {
```
alesstimec, line 35: add tests that assert that non-admin users cannot manage roles - this is enforced by the RoleManager, i know, at JIMM level, but since these are integration tests...

alesstimec (changes requested):

internal/jujuapi/controllerroot.go
kian99: It's easily resolved if we just embed the RoleManger on JIMM. I'll make a PR for that to illustrate then this PR will be easier to land.

kian99: On second thought, what if we don't export RoleManager and change `GetRoleManager` to just `RoleManager` then the code reads a bit more cleanly as `r.jimm.RoleManager().RemoveRole`. Probably easiest for now.

In fact, if everything was inside a manager, we wouldn't need a large JIMM interface. The interface is done at the level of the manager. I.e. you can still unit test the api layer, you just need a mock manager. I think going all in on the managers will be helpful.

internal/jimm/jimm.go
SimoneDutto: the same goes for OpenFGAClient, Pubsubhub, database, etc etc.

I think you suggested the newJIMM constructor and i fully agree.
Otherwise we need to add some "panic" in these getters.
I would leave as it is now

internal/jujuapi/role.go
SimoneDutto: yeah, it was not used. I've removed it. In the JIMM interface we use the jimm.RoleManager interface

internal/jujuapi/role_test.go
SimoneDutto: added

internal/jujuapi/controllerroot.go
SimoneDutto: we can't use a field and a method with the same name.
we already have `RoleManager` taken, do you have a better name than `GetRoleManager()`?

SimoneDutto: One way is to have `roleManager` private and then having `RoleManager` as the getter and `SetManager` as the setter, but i don't know if that's any better.

internal/jimm/jimm.go
alesstimec: ok. but we need a card to address this issue in a followup

.github/workflows/ci.yaml
```diff
@@ -34,4 +34,4 @@ jobs:
           store-file-path: ./openfga/tests.fga.yaml
 
       - name: Build and Test
-        run: go test -mod readonly ./... -timeout 1h -cover
+        run: TEST_LOGGING_CONFIG=ERROR go test -mod readonly ./... -timeout 1h -cover
```
alesstimec, line 37: intentional?

internal/jimm/jimm.go
SimoneDutto: card created.

alesstimec (approved):

internal/jujuapi/controllerroot.go
ale8k: We can discuss this in standup.

internal/jujuapi/role.go
```diff
@@ -0,0 +1,124 @@
...
+
+// RemoveRole removes a role within JIMMs DB for reference by OpenFGA.
+func (r *controllerRoot) RemoveRole(ctx context.Context, req apiparams.RemoveRoleRequest) error {
+	const op = errors.Op("jujuapi.RemoveRole")
+
+	if !jimmnames.IsValidRoleName(req.Name) {
+		return errors.E(op, errors.CodeBadRequest, "invalid role name")
+	}
+
+	if err := r.jimm.GetRoleManager().RemoveRole(ctx, r.user, req.Name); err != nil {
```
ale8k, line 97: Also if we do go with Getter, in go you don't prefix getters with Get - think essential go or something else states this

internal/testutils/jimmtest/mocks/jimm_role_mock.go
```diff
@@ -0,0 +1,72 @@
...
+	"context"
+
+	"github.com/canonical/jimm/v3/internal/common/pagination"
+	"github.com/canonical/jimm/v3/internal/dbmodel"
+	"github.com/canonical/jimm/v3/internal/errors"
+	"github.com/canonical/jimm/v3/internal/openfga"
+)
+
+// RoleManager is an implementation of the jujuapi.RoleManager interface.
+type RoleManager struct {
```
ale8k, line 15: We need a discussion on mocking stratgegy, I'd even put mock in the name... Uhm I want us to use gomock let's discuss in standup.

ale8k (approved):
lgtm nice work, some questions to talk about later I think.

PR #1464, opened 2024-11-26
refactor(internal/jimmhttp/httpproxy_handler.go): defines a JIMM interface for the proxy to use.

alesstimec:
The httpproxy_handler was directly using jimm.JIMM, which seems a bit excessive.

Eventually i would like to un-export the jimm.JIMM type and anything that uses it should define its own interface.

## Description

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [ ] Covered by unit tests

### DIFF
--- a/internal/jimmhttp/httpproxy_handler.go
+++ b/internal/jimmhttp/httpproxy_handler.go
@@ -3,23 +3,36 @@
 package jimmhttp
 
 import (
+	"context"
 	"net/http"
 
 	"github.com/go-chi/chi/v5"
 	"github.com/juju/names/v4"
 	"gopkg.in/errgo.v1"
 
-	"github.com/canonical/jimm/v3/internal/jimm"
+	"github.com/canonical/jimm/v3/internal/dbmodel"
+	"github.com/canonical/jimm/v3/internal/jimm/credentials"
 	"github.com/canonical/jimm/v3/internal/middleware"
+	"github.com/canonical/jimm/v3/internal/openfga"
 	ofganames "github.com/canonical/jimm/v3/internal/openfga/names"
 	"github.com/canonical/jimm/v3/internal/rpc"
 )
+
+// JIMM defines an interface used by the HTTPProxyHandler to get authentication
+// information.
+type JIMM interface {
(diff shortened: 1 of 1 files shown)


internal/jimmhttp/httpproxy_handler.go
```diff
@@ -3,23 +3,36 @@
...
+	"github.com/canonical/jimm/v3/internal/jimm/credentials"
 	"github.com/canonical/jimm/v3/internal/middleware"
+	"github.com/canonical/jimm/v3/internal/openfga"
 	ofganames "github.com/canonical/jimm/v3/internal/openfga/names"
 	"github.com/canonical/jimm/v3/internal/rpc"
 )
 
+// JIMM defines an interface used by the HTTPProxyHandler to get authentication
+// information.
+type JIMM interface {
```
ale8k, line 23: Let's not do interface name shadowing and come up with a better name

internal/jimmhttp/httpproxy_handler.go
```diff
@@ -3,23 +3,36 @@
...
+	AuthenticateBrowserSession(context.Context, http.ResponseWriter, *http.Request) (context.Context, error)
+	GetCredentialStore() credentials.CredentialStore
```
ale8k, line 25: Doc the methods please.

internal/jimmhttp/httpproxy_handler.go
```diff
@@ -3,23 +3,36 @@
...
+	GetModel(ctx context.Context, uuid string) (dbmodel.Model, error)
+	LoginWithSessionToken(ctx context.Context, sessionToken string) (*openfga.User, error)
+	UserLogin(ctx context.Context, identityName string) (*openfga.User, error)
```
ale8k, line 28: You could make two interfaces here too, one  for authz/n and one called "ModelGetter"

ale8k: Actually yeah, let's do this, three interfaces - Authenticator, Authoriser, ModelGetter - something like this?

SimoneDutto: no please, let's not start this. We are not a 1 billion LOC project

PR #1468, opened 2024-11-26
refactor(internal/jimm/jimm.go): introduces a jimm constructor

alesstimec:
## Description

Introduces a constructor for jimm in internal/jimm. With the lightest touch possible atm.

drive-by fix: the db.Database type holds a `migrated` (uint32) field. however we were passing the database around by value and every time we copied it, whoever tried to use it would get a `CodeUpgradeInProgress` error. In this PR we pass a pointer to the database around, hopefully avoiding further migration voes.

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [ ] Covered by unit tests

### DIFF
--- a/cmd/jimmsrv/service/service.go
+++ b/cmd/jimmsrv/service/service.go
@@ -13,6 +13,7 @@ import (
 	"time"
 
 	"github.com/antonlindstrom/pgstore"
+	service "github.com/canonical/go-service"
 	cofga "github.com/canonical/ofga"
 	"github.com/go-chi/chi/v5"
 	chimiddleware "github.com/go-chi/chi/v5/middleware"
@@ -27,13 +28,12 @@ import (
 	"gorm.io/gorm"
 
 	"github.com/canonical/jimm/v3/internal/auth"
+	"github.com/canonical/jimm/v3/internal/db"
 	"github.com/canonical/jimm/v3/internal/dbmodel"
 	"github.com/canonical/jimm/v3/internal/discharger"
 	"github.com/canonical/jimm/v3/internal/errors"
 	"github.com/canonical/jimm/v3/internal/jimm"
 	jimmcreds "github.com/canonical/jimm/v3/internal/jimm/credentials"
-	"github.com/canonical/jimm/v3/internal/jimm/group"
-	"github.com/canonical/jimm/v3/internal/jimm/role"
 	"github.com/canonical/jimm/v3/internal/jimmhttp"
 	"github.com/canonical/jimm/v3/internal/jimmhttp/rebac_admin"
 	"github.com/canonical/jimm/v3/internal/jimmjwx"
@@ -100,6 +100,9 @@ type Params struct {
 	// ControllerUUID contains the UUID of the JIMM controller, if this
 	// is not set a random UUID will be generated.
 	ControllerUUID string
+
+	// IsLeader indicates that this is the JIMM leader unit.
+	IsLeader bool
 
 	// DSN is the data source name that the JIMM service will use to
 	// connect to its database. If this is empty an in-memory database
@@ -195,14 +198,17 @@ type Params struct {
 
 // A Service is the implementation of a JIMM server.
 type Service struct {
-	jimm jimm.JIMM
+	jimm *jimm.JIMM
+
+	isLeader              bool
+	auditLogCleanupPeriod int
 
 	mux      *chi.Mux
 	cleanups []func() error
 }
 
 func (s *Service) JIMM() *jimm.JIMM {
-	return &s.jimm
+	return s.jimm
 }
 
 // ServeHTTP implements http.Handler.
@@ -236,10 +242,6 @@ func (s *Service) WatchModelSummaries(ctx context.Context) error {
 
 // StartJWKSRotator see internal/jimmjwx/jwks.go for details.
 func (s *Service) StartJWKSRotator(ctx context.Context, checkRotateRequired <-chan time.Time, initialRotateRequiredTime time.Time) error {
-	if s.jimm.JWKService == nil {
-		zapctx.Warn(ctx, "not starting JWKS rotation")
-		return nil
-	}
 	return s.jimm.JWKService.StartJWKSRotator(ctx, checkRotateRequired, initialRotateRequiredTime)
 }
 
@@ -290,66 +292,52 @@ func (s *Service) AddCleanup(f func() error) {
 }
 
 // NewService creates a new Service using the given params.
+//
+//nolint:gocognit // NewService function to be ignored.
 func NewService(ctx context.Context, p Params) (*Service, error) {
 	const op = errors.Op("NewService")
 
 	s := new(Service)
-	s.mux = chi.NewRouter()
 
-	s.mux.Use(chimiddleware.RequestLogger(&logger.HTTPLogFormatter{}))
-	s.mux.Use(middleware.MeasureHTTPResponseTime)
-
-	// Setup all dependency services
-
-	if p.ControllerUUID == "" {
-		controllerUUID, err := uuid.NewRandom()
-		if err != nil {
-			return nil, errors.E(op, err)
-		}
-		p.ControllerUUID = controllerUUID.String()
+	jimmParameters := jimm.Parameters{
+		UUID:   p.ControllerUUID,
+		Pubsub: &pubsub.Hub{MaxConcurrency: 50},
 	}
-	s.jimm.UUID = p.ControllerUUID
-	s.jimm.Pubsub = &pubsub.Hub{MaxConcurrency: 50}
+	// Setup all dependency services
+	if jimmParameters.UUID == "" {
+		jimmParameters.UUID = uuid.NewString()
+	}
 
 	if p.DSN == "" {
 		return nil, errors.E(op, "missing DSN")
 	}
 
-	var err error
-	s.jimm.Database.DB, err = openDB(ctx, p.DSN, p.LogSQL)
+	database, err := openDB(ctx, p.DSN, p.LogSQL)
 	if err != nil {
 		return nil, errors.E(op, err)
 	}
-	if err := s.jimm.Database.Migrate(ctx, false); err != nil {
-		return nil, errors.E(op, err)
+	db := &db.Database{
+		DB: database,
 	}
-
-	if p.AuditLogRetentionPeriodInDays != "" {
-		period, err := strconv.Atoi(p.AuditLogRetentionPeriodInDays)
-		if err != nil {
-			return nil, errors.E(op, "failed to parse audit log retention period")
-		}
-		if period < 0 {
-			return nil, errors.E(op, "retention period cannot be less than 0")
-		}
-		if period != 0 {
-			jimm.NewAuditLogCleanupService(s.jimm.Database, period).Start(ctx)
-		}
-	}
+	jimmParameters.Database = db
 
 	openFGAclient, err := newOpenFGAClient(ctx, p.OpenFGAParams)
 	if err != nil {
 		return nil, errors.E(op, err)
 	}
-	s.jimm.OpenFGAClient = openFGAclient
+	jimmParameters.OpenFGAClient = openFGAclient
+
 	if err := ensureControllerAdministrators(ctx, openFGAclient, p.ControllerUUID, p.ControllerAdmins); err != nil {
 		return nil, errors.E(op, err, "failed to ensure controller admins")
 	}
-	if err := s.setupCredentialStore(ctx, p); err != nil {
+
+	credentialStore, err := s.setupCredentialStore(ctx, p, db)
+	if err != nil {
 		return nil, errors.E(op, err)
 	}
+	jimmParameters.CredentialStore = credentialStore
 
-	sessionStore, err := s.setupSessionStore(ctx, p.CookieSessionKey)
+	sessionStore, err := s.setupSessionStore(ctx, p.CookieSessionKey, db)
 	if err != nil {
 		return nil, errors.E(op, err)
 	}
@@ -374,12 +362,12 @@ func NewService(ctx context.Context, p Params) (*Service, error) {
 			SessionCookieMaxAge: p.OAuthAuthenticatorParams.SessionCookieMaxAge,
 			JWTSessionKey:       p.OAuthAuthenticatorParams.JWTSessionKey,
 			SecureCookies:       p.OAuthAuthenticatorParams.SecureSessionCookies,
-			Store:               &s.jimm.Database,
+			Store:               db,
 			SessionStore:        sessionStore,
 			RedirectURL:         redirectUrl,
 		},
 	)
-	s.jimm.OAuthAuthenticator = authSvc
+	jimmParameters.OAuthAuthenticator = authSvc
 	if err != nil {
 		zapctx.Error(ctx, "failed to setup authentication service", zap.Error(err))
 		return nil, errors.E(op, err, "failed to setup authentication service")
@@ -389,41 +377,35 @@ func NewService(ctx context.Context, p Params) (*Service, error) {
 		p.JWTExpiryDuration = 24 * time.Hour
 	}
 
-	s.jimm.JWKService = jimmjwx.NewJWKSService(s.jimm.CredentialStore)
-	s.jimm.JWTService = jimmjwx.NewJWTService(jimmjwx.JWTServiceParams{
+	jimmParameters.JWKService = jimmjwx.NewJWKSService(credentialStore)
+	jimmParameters.JWTService = jimmjwx.NewJWTService(jimmjwx.JWTServiceParams{
 		Host:   p.PublicDNSName,
-		Store:  s.jimm.CredentialStore,
+		Store:  credentialStore,
 		Expiry: p.JWTExpiryDuration,
 	})
-	s.jimm.Dialer = &jujuclient.Dialer{
-		ControllerCredentialsStore: s.jimm.CredentialStore,
-		JWTService:                 s.jimm.JWTService,
+	jimmParameters.Dialer = &jujuclient.Dialer{
+		ControllerCredentialsStore: credentialStore,
+		JWTService:                 jimmParameters.JWTService,
 	}
-
-	roleManager, err := role.NewRoleManager(&s.jimm.Database, s.jimm.OpenFGAClient)
-	if err != nil {
-		return nil, errors.E(op, err, "failed to create RoleManager")
-	}
-	s.jimm.RoleManager = roleManager
-
-	groupManager, err := group.NewGroupManager(&s.jimm.Database, s.jimm.OpenFGAClient)
-	if err != nil {
-		return nil, errors.E(op, err, "failed to create GroupManager")
-	}
-	s.jimm.GroupManager = groupManager
 
 	if !p.DisableConnectionCache {
-		s.jimm.Dialer = jimm.CacheDialer(s.jimm.Dialer)
+		jimmParameters.Dialer = jimm.CacheDialer(jimmParameters.Dialer)
 	}
 
 	if _, err := url.Parse(p.DashboardFinalRedirectURL); err != nil {
 		return nil, errors.E(op, err, "failed to parse final redirect url for the dashboard")
 	}
 
-	rebacBackend, err := rebac_admin.SetupBackend(ctx, &s.jimm)
+	// instantiate jimm
+	s.jimm, err = jimm.New(jimmParameters)
 	if err != nil {
 		return nil, errors.E(op, err)
 	}
+
+	s.mux = chi.NewRouter()
+
+	s.mux.Use(chimiddleware.RequestLogger(&logger.HTTPLogFormatter{}))
+	s.mux.Use(middleware.MeasureHTTPResponseTime)
 
 	// Setup CORS middleware
 	corsOpts := cors.New(cors.Options{
@@ -440,7 +422,12 @@ func NewService(ctx context.Context, p Params) (*Service, error) {
 
 	s.mux.Mount("/metrics", promhttp.Handler())
 
-	s.mux.Mount("/rebac", middleware.AuthenticateRebac("/rebac", rebacBackend.Handler(""), &s.jimm))
+	rebacBackend, err := rebac_admin.SetupBackend(ctx, s.jimm)
+	if err != nil {
+		return nil, errors.E(op, err)
+	}
+
+	s.mux.Mount("/rebac", middleware.AuthenticateRebac("/rebac", rebacBackend.Handler(""), s.jimm))
 
 	mountHandler(
 		"/debug",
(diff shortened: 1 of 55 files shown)


cmd/jimmsrv/service/service.go
```diff
@@ -388,32 +405,33 @@ func NewService(ctx context.Context, p Params) (*Service, error) {
...
-		s.jimm.Dialer = jimm.CacheDialer(s.jimm.Dialer)
+		s.jimmParameters.Dialer = jimm.CacheDialer(s.jimmParameters.Dialer)
 	}
 
 	if _, err := url.Parse(p.DashboardFinalRedirectURL); err != nil {
 		return nil, errors.E(op, err, "failed to parse final redirect url for the dashboard")
 	}
 
-	rebacBackend, err := rebac_admin.SetupBackend(ctx, &s.jimm)
+	// instanciate jimm
```
kian99:
```suggestion
	// instantiate jimm
```

cmd/jimmsrv/service/service.go
```diff
@@ -194,14 +194,15 @@ type Params struct {
 
 // A Service is the implementation of a JIMM server.
 type Service struct {
-	jimm jimm.JIMM
+	jimmParameters jimm.Parameters
```
kian99: Why do the params need to go on the struct? I assume you're doing it to pass around to some functions that were looking at s.jimm but we could turn those methods into functions instead and pass in just the bits they need?

cmd/jimmsrv/service/service.go
```diff
@@ -388,32 +405,33 @@ func NewService(ctx context.Context, p Params) (*Service, error) {
...
+		s.jimmParameters.Dialer = jimm.CacheDialer(s.jimmParameters.Dialer)
 	}
 
 	if _, err := url.Parse(p.DashboardFinalRedirectURL); err != nil {
 		return nil, errors.E(op, err, "failed to parse final redirect url for the dashboard")
 	}
 
-	rebacBackend, err := rebac_admin.SetupBackend(ctx, &s.jimm)
+	// instanciate jimm
+	j, err := jimm.New(s.jimmParameters)
```
kian99:
```suggestion
	s.jimm, err = jimm.New(s.jimmParameters)
```

cmd/jimmsrv/service/service.go
```diff
@@ -509,11 +527,11 @@ func (s *Service) setupDischarger(p Params) (*discharger.MacaroonDischarger, err
 func (s *Service) setupSessionStore(ctx context.Context, sessionSecret []byte) (*pgstore.PGStore, error) {
```
kian99: Should we change this to a function that accepts jimmParams? Then at least we don't need jimmParams to stick around on the service struct. Or, at the risk of overcomplicating things, we could have a separate struct called `serviceBuilder`. Just some ideas to avoid keeping jimmparams around longer than needed.

cmd/jimmsrv/service/service.go
```diff
@@ -258,6 +275,10 @@ func (s *Service) MonitorResources(ctx context.Context) {
 
 // OpenFGACleanup starts a goroutine that cleans up any orphaned tuples from OpenFGA.
 func (s *Service) OpenFGACleanup(ctx context.Context, trigger <-chan time.Time) error {
+	if s.jimm == nil {
+		zapctx.Error(ctx, "jimm not created")
+		return errors.E("jimm not created")
+	}
```
SimoneDutto: Aren't these s.jimm == nil a good place where we can `panic`. 
I feel like not having configured jimm before starting these services is not a recoverable state

internal/jimm/jimm.go
```diff
@@ -41,16 +42,12 @@ var (
...
-// connections therefore the JIMM object itself does not contain any per-
-// request state.
-type JIMM struct {
+type Parameters struct {
 	// Database is the database used by JIMM, this provides direct access
 	// to the data store. Any client accessing the database directly is
 	// responsible for ensuring that the authenticated user has access to
 	// the data.
-	Database db.Database
+	Database *db.Database
```
SimoneDutto: instead of changing this which required a lot of manual work to update things all around.
Couldn't you change this https://github.com/canonical/jimm/blob/436ba45558dd987bb28d93bb7d1e49bb6dea1eeb/internal/db/db.go#L30 to a pointer?

internal/jimm/jimm.go
```diff
@@ -41,16 +42,12 @@ var (
 	}
 )
 
-// A JIMM provides the business logic for managing resources in the JAAS
-// system. A single JIMM instance is shared by all concurrent API
-// connections therefore the JIMM object itself does not contain any per-
-// request state.
-type JIMM struct {
+type Parameters struct {
```
SimoneDutto: godoc please

internal/jimm/jimm.go
```diff
@@ -193,6 +144,92 @@ type OAuthAuthenticator interface {
...
+
+	return j, nil
+}
+
+// A JIMM provides the business logic for managing resources in the JAAS
+// system. A single JIMM instance is shared by all concurrent API
+// connections therefore the JIMM object itself does not contain any per-
+// request state.
+type JIMM struct {
+	Parameters
```
SimoneDutto, line 265: since we are embedding the struct, it's not `Parameters` but more `Services`. Because we don't start the service in the constructor, we rather take the service reference and store it in the JIMM struct.

internal/jimm/jimm.go
```diff
@@ -193,6 +144,92 @@ type OAuthAuthenticator interface {
...
+	if err != nil {
+		return nil, err
+	}
+	j.RoleManager = roleManager
+	for _, option := range options {
+		option(j)
+	}
+
+	return j, nil
+}
```
SimoneDutto: i thought this method would have been a method where you receive some parameters (ex. db connection string), and you create the `Database, Openfgaclient`.
So we'll end up with a central point where we validate jimm's config, start the jimm services and return the well defined jimm struct.

internal/jimm/admin_test.go
```diff
@@ -38,10 +45,19 @@ func TestLoginDevice(t *testing.T) {
 func TestGetDeviceSessionToken(t *testing.T) {
 	c := qt.New(t)
 	pollingChan := make(chan string, 1)
+	ofgaClient, _, _, err := jimmtest.SetupTestOFGAClient(c.Name())
```
kian99: Why did this test get an OpenFGA client?

internal/jimm/cloud.go
```diff
@@ -160,7 +160,7 @@ func (j *JIMM) AddCloudToController(ctx context.Context, user *openfga.User, con
 		return errors.E(op, err)
 	}
 
-	if err := validateCloudRegion(ctx, &j.Database, user, cloud, controllerName); err != nil {
+	if err := validateCloudRegion(ctx, j.Database, user, cloud, controllerName); err != nil {
```
kian99, line 163: It would've been good to keep the mechanical changes to a separate commit/PR if possible.

internal/jimm/jimm.go
kian99: Missing godoc

kian99 (commented):
Looks good with some comments

internal/jimm/jimm.go
kian99: I think it's correct like this, the JIMM method sets up the business logic, so when we have more "managers", the JIMM struct won't need any of the fields in Params, it will just take the pieces of params like the DB, OpenFGA client, pubSub, etc and pass them to the relevant constructors for the managers. Whoever calls jimm.New() decides what they want to use as a DB, as an auth service, etc (even if you can't mock the DB or auth service today).

internal/jimm/jimm.go
alesstimec: but then we'd need a constructor for db.Database.

cmd/jimmsrv/service/service.go
alesstimec: isn't returning an error better than panicking?

internal/jimm/jimm.go
alesstimec: constructors have parameters. and the constructor will then instantiate all the managers/service provided by JIMM.

cmd/jimmsrv/service/service.go
alesstimec: because we have many methods that just create "stuff" (e.g. oauth authenticator) needed by jimm. and all that "stuff" gets put into the parameters and then passed on to the constructor

internal/jimm/admin_test.go
alesstimec: jimm constructor needs an openfga client, otherwise it cannot start the role manager

internal/jimm/cloud.go
alesstimec: absolutely.. my bad

cmd/jimmsrv/service/service.go
alesstimec: that is a very good point.. changing

cmd/jimmsrv/service/service.go
```diff
@@ -213,6 +217,10 @@ func (s *Service) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 // monitoring all changes to models. WatchControllers finishes when the
 // given context is canceled, or there is a fatal error watching models.
 func (s *Service) WatchControllers(ctx context.Context) error {
+	if s.jimm == nil {
```
kian99: I wouldn't add all these checks. If jimm is nil and this panics we should see that from our tests. If it's happening in production I'd rather have the stack trace of where we tried to use a nil pointer, because this is an unrecoverable error. With this change all we get is an error messsage, "jimm not created" which is arguably worse because we can't see where it happened.

cmd/jimmsrv/service/service.go
```diff
@@ -499,21 +562,17 @@ func (s *Service) setupDischarger(p Params) (*discharger.MacaroonDischarger, err
...
-func (s *Service) setupSessionStore(ctx context.Context, sessionSecret []byte) (*pgstore.PGStore, error) {
+func (s *Service) setupSessionStore(ctx context.Context, sessionSecret []byte, db *db.Database) (*pgstore.PGStore, error) {
 	const op = errors.Op("setupSessionStore")
 
-	if s.jimm.CredentialStore == nil {
-		return nil, errors.E(op, "credential store is not configured")
-	}
-
-	sqlDb, err := s.jimm.Database.DB.DB()
+	sqlDb, err := db.DB.DB()
```
kian99, line 559: Can we improve this at the moment?

cmd/jimmsrv/service/service_test.go
```diff
@@ -82,24 +86,32 @@ func TestDefaultService(t *testing.T) {
...
 	c.Assert(err, qt.ErrorMatches, "jimm cannot start without a credential store")
 }
 
 func TestAuthenticator(t *testing.T) {
 	c := qt.New(t)
 
+	ctx, s := gosvc.NewService(context.Background(), os.Interrupt, syscall.SIGTERM)
+
+	var cancel func()
+	ctx, cancel = context.WithCancel(ctx)
```
kian99: Why are we creating a cancel function here but not in all the other tests?

internal/jimm/jimm.go
```diff
@@ -41,16 +42,13 @@ var (
 	}
 )
 
-// A JIMM provides the business logic for managing resources in the JAAS
-// system. A single JIMM instance is shared by all concurrent API
-// connections therefore the JIMM object itself does not contain any per-
-// request state.
-type JIMM struct {
+// Parameters holds parameters for the JIMM type.
```
kian99: Let's better explain this. 
```suggestion
// Parameters holds the services and static fields passed to the jimm.New() constructor.
// You can provide mock implementations of certain services where necessary for dependency injection.
```

internal/jimm/jimm.go
```diff
@@ -193,6 +145,92 @@ type OAuthAuthenticator interface {
...
+	// RenameRole renames a role in JIMM's DB.
+	RenameRole(ctx context.Context, user *openfga.User, uuid, newName string) error
+	// ListRoles returns a list of roles known to JIMM.
+	// `match` will filter the list fuzzy matching role's name or uuid.
+	ListRoles(ctx context.Context, user *openfga.User, pagination pagination.LimitOffsetPagination, match string) ([]dbmodel.RoleEntry, error)
+	// CountRoles returns the number of roles that exist.
+	CountRoles(ctx context.Context, user *openfga.User) (int, error)
+}
+
+// Option modifies JIMM returned by the constructor.
```
kian99: Again I think it would be good to explain why this is useful.
```suggestion
// Option modifies the JIMM returned by the constructor.
// This is useful to create mock domain objects, e.g. replace the real RoleManager with a mock 
// object in tests.
```

internal/jimm/jimm.go
```diff
@@ -193,6 +145,92 @@ type OAuthAuthenticator interface {
...
+type Option func(*JIMM)
+
+// RoleManagerOption replaces the role manager in the returned JIMM with the specified one.
+var RoleManagerOption = func(rm RoleManager) Option {
+	return func(j *JIMM) {
+		j.RoleManager = rm
+	}
+}
+
+// New returns a new instance of JIMM based on the provided parameters and options.
```
kian99:
```suggestion
// New returns a new instance of JIMM.
// See [Option] and [Parameters] to better understand how to perform dependency injection.
// Primitives like the dialer or authentication service can be mocked at a low level, alternatively
// top business layer objects like the RoleManager can be mocked instead.
```

cmd/jimmsrv/service/service.go
```diff
@@ -289,66 +313,50 @@ func (s *Service) AddCleanup(f func() error) {
 }
 
 // NewService creates a new Service using the given params.
-func NewService(ctx context.Context, p Params) (*Service, error) {
+func NewService(ctx context.Context, p Params, svc *service.Service) (*Service, error) {
```
kian99: Can we refactor this so that we don't need to pass in a service.Service here.

cmd/jimmsrv/service/service.go
SimoneDutto: i agree, i'd rather panic when creating the service and avoid these checks.

cmd/jimmsrv/service/service.go
```diff
@@ -289,66 +313,50 @@ func (s *Service) AddCleanup(f func() error) {
...
 	const op = errors.Op("NewService")
 
 	s := new(Service)
-	s.mux = chi.NewRouter()
-
-	s.mux.Use(chimiddleware.RequestLogger(&logger.HTTPLogFormatter{}))
-	s.mux.Use(middleware.MeasureHTTPResponseTime)
 
+	jimmParameters := jimm.Parameters{
```
SimoneDutto, line 302: i would still call the services. But I won't die on this hill

internal/jimm/jimm.go
```diff
@@ -193,6 +145,92 @@ type OAuthAuthenticator interface {
...
+func New(p Parameters, options ...Option) (*JIMM, error) {
```
SimoneDutto: does it make sense to validate here that our parameters(services) are not nil? So we can remove the checks in the getters

SimoneDutto (commented):
few comments then LGTM

internal/jimm/jimm.go
alesstimec: we probably could.. but then we'd need to provide a "valid" set of parameters every time constructor was called

internal/jimm/jimm.go
```diff
@@ -41,16 +42,103 @@ var (
...
+	CountRoles(ctx context.Context, user *openfga.User) (int, error)
+}
+
+// Option modifies the JIMM returned by the constructor.
+// This is useful to create mock domain objects, e.g. replace the real RoleManager with a mock
+// object in tests.
+type Option func(*JIMM)
+
+// RoleManagerOption replaces the role manager in the returned JIMM with the specified one.
+var RoleManagerOption = func(rm RoleManager) Option {
```
ale8k: Just make a func called WithRoleManager?

internal/jimm/jimm.go
```diff
@@ -87,28 +175,86 @@ type JIMM struct {
...
+
+	if p.OpenFGAClient == nil {
+		return errors.E("missing openfga client")
+	}
+
+	if p.JWKService == nil {
+		return errors.E("missing jwks service")
+	}
+
+	if p.JWTService == nil {
```
ale8k, line 217: I imagine this should only contain primitives? Not Serices?

internal/jimm/jimm.go
ale8k: I wouldn't embed it, params you pass into the func, and on JIMM it filles in another struct. Don't couple them please?

internal/jimm/jimm.go
alesstimec: done

internal/jimm/jimm.go
ale8k: Also uhm, might be a good idea to have a file just called options? Be easier to track these options as time goes on.

alesstimec: removed

cmd/jimmsrv/service/service.go
alesstimec: improve how?

internal/jimm/access_test.go
```diff
@@ -428,20 +415,7 @@ func TestParseAndValidateTag(t *testing.T) {
...
-		UUID: uuid.NewString(),
-		Database: db.Database{
-			DB: jimmtest.PostgresDB(c, func() time.Time { return now }),
-		},
-		OpenFGAClient: ofgaClient,
-	}
-
-	err = j.Database.Migrate(ctx, false)
-	c.Assert(err, qt.IsNil)
+	j := jimmtest.NewJIMM(c, nil)
```
SimoneDutto, line 418: this is very nice!

internal/testutils/jimmtest/jimm.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+	if p.CredentialStore == nil {
+		p.CredentialStore = p.Database
+	}
+	if p.OpenFGAClient == nil {
+		ofgaClient, _, _, err := SetupTestOFGAClient(t.Name())
+		if err != nil {
+			t.Fatalf("setting up openfga client: %v", err)
+		}
+		p.OpenFGAClient = ofgaClient
```
SimoneDutto, line 85: so, this is the default config for jimmtest now. Update the godoc please

internal/testutils/jimmtest/jimm.go
```diff
@@ -0,0 +1,98 @@
...
+
+type Option func(j *jimm.JIMM)
+
+var (
+	UnsetCredentialStore Option = func(j *jimm.JIMM) {
+		j.CredentialStore = nil
+	}
+)
+
+func NewJIMM(t Tester, additionalParameters *jimm.Parameters, options ...Option) *jimm.JIMM {
```
SimoneDutto, line 26: godoc

internal/testutils/jimmtest/jimm.go
```diff
@@ -0,0 +1,98 @@
+// Copyright 2024 Canonical.
+
```
SimoneDutto, line 2: i was hoping for my first "new line please". I'll get you next time

SimoneDutto, line 2: i was hoping for my first "new line please". I'll get you next time

internal/testutils/jimmtest/jimm.go
```diff
@@ -0,0 +1,98 @@
...
+var now = (time.Time{}).UTC().Round(time.Millisecond)
+
+type Option func(j *jimm.JIMM)
+
+var (
+	UnsetCredentialStore Option = func(j *jimm.JIMM) {
+		j.CredentialStore = nil
+	}
+)
```
SimoneDutto, line 24: if we give default options to choose from, do we need to have the `Option` exported?

internal/testutils/jimmtest/suite.go
```diff
@@ -47,7 +46,11 @@ func (t GocheckTester) Name() string {
 }
 
 func (t GocheckTester) Cleanup(f func()) {
-	t.C.Logf("warning: gocheck does not support Cleanup functions; make sure you're using suite's tear-down method")
+	if t.AddCleanup != nil {
+		t.AddCleanup(f)
```
SimoneDutto, line 50: why bother with adding the cleanup? Since we don't have any cleanup with all the `ŧesting.t` tests.

I think it's just enough to increase the conn limit on postgres

SimoneDutto: btw, where did the change go? Or now it's fine because we are not creating that many connections?

SimoneDutto (approved):
lgtm, but please new time let's split this in multiple pr, with a feature branch. 
Too many mechanical changes in the same pr

internal/testutils/jimmtest/jimm.go
alesstimec: yes. we have a slight bug around cloud credentials. and there is a test that relies on the fact that the credential store is unset.. until we fix that bug, we need this option

internal/testutils/jimmtest/suite.go
alesstimec: https://github.com/canonical/jimm/pull/1468/files#diff-3fde9d1a396e140fefc7676e1bd237d67b6864552b6f45af1ebcc27bcd0bb6e9R73

cmd/jimmsrv/service/service.go
```diff
@@ -389,42 +377,36 @@ func NewService(ctx context.Context, p Params) (*Service, error) {
...
+		jimmParameters.Dialer = jimm.CacheDialer(jimmParameters.Dialer)
 	}
 
 	if _, err := url.Parse(p.DashboardFinalRedirectURL); err != nil {
 		return nil, errors.E(op, err, "failed to parse final redirect url for the dashboard")
 	}
 
-	rebacBackend, err := rebac_admin.SetupBackend(ctx, &s.jimm)
+	// instantiate jimm
+	s.jimm, err = jimm.New(jimmParameters)
```
ale8k, line 400: Happy with this as is, as we discussed, but we should definitely break this down further into primitives and application domains later on. I'd love to see primitive db, openfga, authsvc (as params) -> and within the constructor: rolemanager, identityservice, etc. all constructed returning a nicely complete jimm. And, if we want to override any application domain we pass an option.

cmd/jimmsrv/service/service.go
```diff
@@ -486,16 +473,69 @@ func NewService(ctx context.Context, p Params) (*Service, error) {
...
-	s.mux.Handle("/model/*", websocketCors.Handler(http.StripPrefix("/model", jujuapi.ModelHandler(ctx, &s.jimm, params))))
+	s.mux.Handle("/api", websocketCors.Handler(jujuapi.APIHandler(ctx, s.jimm, params)))
+	s.mux.Handle("/model/*", websocketCors.Handler(http.StripPrefix("/model", jujuapi.ModelHandler(ctx, s.jimm, params))))
 	mountHandler(
 		"/model/{uuid}/{type:charms|applications}",
-		jimmhttp.NewHTTPProxyHandler(&s.jimm),
+		jimmhttp.NewHTTPProxyHandler(s.jimm),
 	)
 
+	if p.AuditLogRetentionPeriodInDays != "" {
```
ale8k, line 483: I feel like we should have service params honestly, it's confusing / bit odd, to see params which re for both jimm and the service and they take effect when you start the service. @kian99 I think you mentioned this right?

cmd/jimmsrv/service/service.go
```diff
@@ -486,16 +473,69 @@ func NewService(ctx context.Context, p Params) (*Service, error) {
...
+			return nil, errors.E(op, "retention period cannot be less than 0")
+		}
+	}
+	s.isLeader = p.IsLeader
+
 	return s, nil
 }
 
+func (s *Service) StartServices(ctx context.Context, svc *service.Service) {
+	// on the leader unit we start additional routines
```
ale8k, line 499: I do like how we start all services here, but in main.go we're running this and then we start the jimm service. Would be nice if jimm was also started here? Could even setup the on shutdown etc all here.

internal/db/cloud_test.go
```diff
@@ -244,7 +244,7 @@ controllers:
     region: test-region
     priority: 1
 `)
-	env.PopulateDB(c, *s.Database)
+	env.PopulateDB(c, s.Database)
```
ale8k, line 247: Interesting we could handle ** here. Good spot.

cmd/jimmsrv/service/service.go
kian99: For a separate PR, and not for you per se, but I was just remarking that the naming isn't great. database.database.database() is crazy.

docker-compose.yaml
```diff
@@ -70,7 +70,7 @@ services:
       POSTGRES_PASSWORD: jimm
     # Since it's mainly used for testing purposes, it's okay to set fsync=off for
     # improved performance.
-    command: -c fsync=off -c full_page_writes=off
+    command: -c fsync=off -c full_page_writes=off -c max_connections=200
```
kian99, line 73: This is a bit concerning. I wonder if we're going over the default 100 connections because we're leaking connections or because of tests running in parallel. Perhaps one way to check would be to add a global test cleanup methods that checks the number of open connections to Postgres. Or run tests with -p 1 and print the number of connections to Postgres between each test incovation.

internal/jimm/jimm.go
kian99: They're only embedded for the time being. Once we have the domain objects, we can delete this.

internal/testutils/jimmtest/jimm.go
```diff
@@ -0,0 +1,98 @@
...
+			p.JWTService = additionalParameters.JWTService
+		}
+		if additionalParameters.OAuthAuthenticator != nil {
+			p.OAuthAuthenticator = additionalParameters.OAuthAuthenticator
+		}
+	}
+
+	if p.Database == nil {
+		p.Database = &db.Database{
+			DB: PostgresDB(t, func() time.Time { return now }),
```
kian99, line 74: I'm currently not in favour of this approach where every test that calls `jimmtest.NewJIMM` gets a whole database and OpenFGA store. Some tests need neither, some tests need one. That was nicely captured in the tests and when we move to managers we can use suites so that there is less duplication of that boilerplate test setup.

I think the default should be all mocks or all nil fields unless the test requests them.

internal/testutils/jimmtest/suite.go
```diff
@@ -172,14 +190,14 @@ func (s *JIMMSuite) TearDownTest(c *gc.C) {
 	}
 }
 
-func (s *JIMMSuite) setupMacaroonDischarger(c *gc.C) *discharger.MacaroonDischarger {
+func setupMacaroonDischarger(c *gc.C, uuid string, db *db.Database, ofgaClient *openfga.OFGAClient) *discharger.MacaroonDischarger {
```
kian99, line 193: Why was this changed to a function?

kian99 (commented):
I think it's almost okay. My biggest gripe is with all tests getting a DB and OpenFGA client.

internal/testutils/jimmtest/suite.go
kian99: I think this link stops working when you force push btw.

internal/testutils/jimmtest/jimm.go
```diff
@@ -0,0 +1,98 @@
...
+
+	"github.com/canonical/jimm/v3/internal/db"
+	"github.com/canonical/jimm/v3/internal/jimm"
+	"github.com/canonical/jimm/v3/internal/jimmjwx"
+	"github.com/canonical/jimm/v3/internal/pubsub"
+)
+
+var now = (time.Time{}).UTC().Round(time.Millisecond)
+
+type Option func(j *jimm.JIMM)
```
ale8k, line 18: I see, we use option pattern only in the context of tests... I think... It's more useful the have the pattern on the JIMM constructor for when we come to (or not to) initialise services. Kinda like juju's manifolds where they configure the binary in different ways - if we ever need to go down that path we can. It's also no harm I think rather than having the option pattern only apply to the jimmtest constructor, I think it's a lot clearer if we just use jimm.New() and pass test options.

ale8k (approved):

ale8k (approved):
lgtm, I've left some comments. Nothing too serious, but more about options and service. I'd much prefer we just use a single constructor for initialisng jimm in tests and in main.

internal/testutils/jimmtest/jimm.go
alesstimec: nope.. in  production we never expect to use options.. only for tests, so why pollute production code with options?

PR #1469, opened 2024-11-27
feat: add rebac-admin handlers for roles

kian99:
## Description

This PR adds the rebac-admin handlers for roles, specifically:
- ListRoles
- CreateRole
- GetRole
- UpdateRole
- DeleteRole
- GetRoleEntitlements
- PatchRoleEntitlements
- GetIdentityRoles
- PatchIdentityRoles

It's a big PR so please let me know if I should rather split it up. But the implementation and tests is all identical to those for groups.

Fixes [JUJU-7127](https://warthogs.atlassian.net/browse/JUJU-7127)

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [x] Covered by unit tests

[JUJU-7127]: https://warthogs.atlassian.net/browse/JUJU-7127?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/jimmhttp/rebac_admin/identities.go
```diff
@@ -146,16 +232,17 @@ func (s *identitiesService) GetIdentityGroups(ctx context.Context, identityId st
...
 		Data: groups,
 		Meta: resources.ResponseMeta{
 			Size:      len(groups),
 			PageToken: &originalToken,
 		},
-		Next: resources.Next{
-			PageToken: &cNextToken,
-		},
-	}, nil
+	}
```
SimoneDutto, line 241: before we were returning the empty string instead of the nil pointer for the last token? strange we didn't catch that before!

SimoneDutto (approved):
Since now we support `roles` we should update also `rebac_admin/entitlements.go|capabilities.go`

If you want to do in a followup, it's fine.

internal/jimmhttp/rebac_admin/identities.go
```diff
@@ -93,19 +93,105 @@ func (s *identitiesService) UpdateIdentity(ctx context.Context, identity *resour
...
+	if len(additions) > 0 {
+		err = s.jimm.AddRelation(ctx, user, additions)
+		if err != nil {
+			zapctx.Error(context.Background(), "cannot add relations", zap.Error(err))
+			return false, v1.NewUnknownError(err.Error())
+		}
+	}
+	if len(deletions) > 0 {
+		err = s.jimm.RemoveRelation(ctx, user, deletions)
```
alesstimec, line 188: ah, ouch, this one could be tricky. what if user is member of a group that is an asignee of a role.. here we attempt to remove only direct relations, but `GetIdentityRoles` returnes indirect relations, too. 

in this case, should we verify if access was actually removed - identity no longer has asignee relation to the role - and report an error/suggestion that this might be a case of indirect relations?

internal/jimmhttp/rebac_admin/roles.go
```diff
@@ -0,0 +1,230 @@
...
+	"github.com/canonical/jimm/v3/internal/jimmhttp/rebac_admin/utils"
+	"github.com/canonical/jimm/v3/internal/jujuapi"
+	ofganames "github.com/canonical/jimm/v3/internal/openfga/names"
+	apiparams "github.com/canonical/jimm/v3/pkg/api/params"
+	jimmnames "github.com/canonical/jimm/v3/pkg/names"
+)
+
+// rolesService implements the `RolesService` interface.
+type rolesService struct {
+	jimm jujuapi.JIMM
```
alesstimec, line 23: hmm.. should we instead require 
```
type RoleManager interface {
 ...
}

type RoleManagerGetter interface {
  GetRoleManager() RoleManager
}
```
then this would become 
```
type rolesService struct {
  jimm RoleManagerGetter
}
```

naming, obviously, is hard

internal/jimmhttp/rebac_admin/roles.go
```diff
@@ -0,0 +1,230 @@
...
+		return false, err
+	}
+	if toAdd != nil {
+		err := s.jimm.AddRelation(ctx, user, toAdd)
+		if err != nil {
+			return false, err
+		}
+	}
+	if toRemove != nil {
+		err := s.jimm.RemoveRelation(ctx, user, toRemove)
```
alesstimec, line 224: same thing applies.. we can only do this in case of direct relations

internal/jimmhttp/rebac_admin/identities.go
kian99: I think direct access is the right "view" for this handler. Indirect relation checks should be handled separately.

kian99: There are separate handlers for assigning/removing groups from roles that I will make in a follow-up PR.

internal/jimmhttp/rebac_admin/roles.go
kian99: I'd prefer to keep it as is for consistency and then switch them over once we no longer have the `RoleManagerGetter`.

ale8k (approved):

PR #1470, opened 2024-11-27
fix sdk release

SimoneDutto:
## Description

This workflow https://github.com/canonical/jimm/actions/runs/12049367128 is broken.
After some googling I think it's because the shell don't allow for pattern matching with `(` unless you set the option.

See https://unix.stackexchange.com/questions/504590/what-is-the-meaning-of-the-following-shell-option-shopt-s-nullglob-extglob

> fly-by remove go checkout because i don't see any go commands.

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [ ] Covered by unit tests

### DIFF
--- a/.github/workflows/update-sdk.yaml
+++ b/.github/workflows/update-sdk.yaml
@@ -32,11 +32,6 @@ jobs:
         ref: ${{ github.event.inputs.sdk-version }}
         path: ./sdk
 
-    - name: Setup Go
-      uses: actions/setup-go@v5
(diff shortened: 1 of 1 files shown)


kian99 (approved):

kian99:
You can also try running the workflow in your fork and debugging with the https://github.com/mxschmitt/action-tmate action

alesstimec (approved):

PR #1474, opened 2024-11-29
fix: flaky test in TestProxySocketsAdminFacade

kian99:
## Description

This PR fixes a flaky test case in `TestProxySocketsAdminFacade`.

The test was incorrectly cancelling the context. When the proxySockets function encounters an error like a failure to connect to a Juju controller it returns with an error. Otherwise it returns a nil error when the context is cancelled. Here we were expecting the former for one of the tests cases but sometimes encountering the latter in a race condition.

The test failed often when running `go test -timeout 30s -run ^TestProxySocketsAdminFacade$ github.com/canonical/jimm/v3/internal/rpc -count=1000`, after the fix I didn't encounter any failures even after increasing count to 10 000.

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [x] Covered by unit tests

### DIFF
--- a/internal/rpc/apiproxy_test.go
+++ b/internal/rpc/apiproxy_test.go
@@ -70,8 +70,7 @@ func TestProxySocketsAdminFacade(t *testing.T) {
 		expectedClientResponse    *message
 		expectedControllerMessage *message
 		oauthAuthenticatorError   error
-		expectProxyError          bool
-		proxyErrorMessage         string
+		expectedProxyError        string
 	}{{
 		about: "login device call - client gets response with both user code and verification uri",
 		messageToSend: message{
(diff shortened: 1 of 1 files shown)


SimoneDutto (approved):
beating the flaky tests allegation

internal/rpc/apiproxy_test.go
```diff
@@ -275,8 +277,8 @@ func TestProxySocketsAdminFacade(t *testing.T) {
 			go func() {
 				defer wg.Done()
 				err = rpc.ProxySockets(ctx, helpers)
-				if test.expectConnectTofail {
-					c.Assert(err, qt.ErrorMatches, "failed to connect to controller: controller connection error")
+				if test.expectProxyError {
```
alesstimec: could we simplify this and do 
```
if test.proxyErrorMessage != ""
```
and lose the bool?

alesstimec (approved):
LGTM with a question/suggestion

internal/rpc/apiproxy_test.go
kian99: Yes, done.

ale8k (approved):

PR #1475, opened 2024-11-29
feat: introduce group manager

kian99:
## Description

This PR follows on from https://github.com/canonical/jimm/pull/1455 and create a group manager.

Moved group methods to group package.
Moved createTestControllerEnvironment function to jimmtest package.

Fixes [JUJU-7245](https://warthogs.atlassian.net/browse/JUJU-7245)

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [x] Covered by unit tests

[JUJU-7245]: https://warthogs.atlassian.net/browse/JUJU-7245?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/jimm/group/group.go
+++ b/internal/jimm/group/group.go
@@ -13,7 +13,7 @@ import (
 	"github.com/canonical/jimm/v3/internal/openfga"
 )
 
-// permissionManager provides a means to manage roles within JIMM.
+// groupManager provides a means to manage groups within JIMM.
 type groupManager struct {
 	store   *db.Database
 	authSvc *openfga.OFGAClient
@@ -23,10 +23,10 @@ type groupManager struct {
 // creation, modification, and removal.
 func NewGroupManager(store *db.Database, authSvc *openfga.OFGAClient) (*groupManager, error) {
 	if store == nil {
-		return nil, errors.E("role store cannot be nil")
+		return nil, errors.E("group store cannot be nil")
 	}
 	if authSvc == nil {
-		return nil, errors.E("role authorisation service cannot be nil")
+		return nil, errors.E("group authorisation service cannot be nil")
 	}
 	return &groupManager{store, authSvc}, nil
 }
--- a/internal/jimm/group/group.go
+++ b/internal/jimm/group/group.go
@@ -1,0 +1,146 @@
+// Copyright 2024 Canonical.
+package group
+
+import (
+	"context"
+
+	"github.com/canonical/jimm/v3/internal/common/pagination"
+	"github.com/canonical/jimm/v3/internal/db"
+	"github.com/canonical/jimm/v3/internal/dbmodel"
+	"github.com/canonical/jimm/v3/internal/errors"
+	"github.com/canonical/jimm/v3/internal/openfga"
+)
+
+// permissionManager provides a means to manage roles within JIMM.
+type groupManager struct {
+	store   *db.Database
+	authSvc *openfga.OFGAClient
+}
+
+// NewGroupManager returns a new group manager that provides group
+// creation, modification, and removal.
+func NewGroupManager(store *db.Database, authSvc *openfga.OFGAClient) (*groupManager, error) {
+	if store == nil {
+		return nil, errors.E("role store cannot be nil")
+	}
+	if authSvc == nil {
+		return nil, errors.E("role authorisation service cannot be nil")
+	}
+	return &groupManager{store, authSvc}, nil
+}
+
+// AddGroup creates a group within JIMMs DB for reference by OpenFGA.
+func (j *groupManager) AddGroup(ctx context.Context, user *openfga.User, name string) (*dbmodel.GroupEntry, error) {
+	const op = errors.Op("jimm.GetGroupManager().AddGroup")
+
+	if !user.JimmAdmin {
+		return nil, errors.E(op, errors.CodeUnauthorized, "unauthorized")
+	}
+
+	ge, err := j.store.AddGroup(ctx, name)
+	if err != nil {
+		return nil, errors.E(op, err)
+	}
+	return ge, nil
+}
+
+// CountGroups returns the number of groups that exist.
+func (j *groupManager) CountGroups(ctx context.Context, user *openfga.User) (int, error) {
+	const op = errors.Op("jimm.CountGroups")
+
+	if !user.JimmAdmin {
+		return 0, errors.E(op, errors.CodeUnauthorized, "unauthorized")
+	}
+	count, err := j.store.CountGroups(ctx)
+	if err != nil {
+		return 0, errors.E(op, err)
+	}
+	return count, nil
+}
+
+// getGroup returns a group based on the provided UUID or name.
+func (j *groupManager) getGroup(ctx context.Context, user *openfga.User, group *dbmodel.GroupEntry) (*dbmodel.GroupEntry, error) {
+	const op = errors.Op("jimm.getGroup")
+
+	if !user.JimmAdmin {
+		return nil, errors.E(op, errors.CodeUnauthorized, "unauthorized")
+	}
+	if err := j.store.GetGroup(ctx, group); err != nil {
+		return nil, errors.E(op, err)
+	}
+	return group, nil
+}
+
+// GetGroupByUUID returns a group based on the provided UUID.
+func (j *groupManager) GetGroupByUUID(ctx context.Context, user *openfga.User, uuid string) (*dbmodel.GroupEntry, error) {
+	return j.getGroup(ctx, user, &dbmodel.GroupEntry{UUID: uuid})
+}
+
+// GetGroupByName returns a group based on the provided name.
+func (j *groupManager) GetGroupByName(ctx context.Context, user *openfga.User, name string) (*dbmodel.GroupEntry, error) {
+	return j.getGroup(ctx, user, &dbmodel.GroupEntry{Name: name})
+}
+
+// RenameGroup renames a group in JIMM's DB.
+func (j *groupManager) RenameGroup(ctx context.Context, user *openfga.User, oldName, newName string) error {
+	const op = errors.Op("jimm.GetGroupManager().RenameGroup")
+
+	if !user.JimmAdmin {
+		return errors.E(op, errors.CodeUnauthorized, "unauthorized")
+	}
+
+	group := &dbmodel.GroupEntry{
+		Name: oldName,
+	}
+	err := j.store.GetGroup(ctx, group)
+	if err != nil {
+		return errors.E(op, err)
+	}
+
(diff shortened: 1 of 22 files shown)


internal/jimm/group/group.go
```diff
@@ -0,0 +1,146 @@
+// Copyright 2024 Canonical.
+package group
```
alesstimec, line 4: new line (+1 for my comment count \o/)

alesstimec: also, as the package needs a godoc (other than copyright)

internal/jimm/group/group.go
```diff
@@ -0,0 +1,146 @@
...
+	const op = errors.Op("jimm.GetGroupManager().RenameGroup")
+
+	if !user.JimmAdmin {
+		return errors.E(op, errors.CodeUnauthorized, "unauthorized")
+	}
+
+	group := &dbmodel.GroupEntry{
+		Name: oldName,
+	}
+	err := j.store.GetGroup(ctx, group)
```
alesstimec: hmm.. slightly worried that this is not done in a transaction. while i don't expect us to have much load on the group manger for this to become a problem i still think it should be done in a transaction.

internal/jimm/group/group.go
```diff
@@ -0,0 +1,146 @@
...
+	const op = errors.Op("jimm.GetGroupManager().RemoveGroup")
+
+	if !user.JimmAdmin {
+		return errors.E(op, errors.CodeUnauthorized, "unauthorized")
+	}
+
+	group := &dbmodel.GroupEntry{
+		Name: name,
+	}
+	err := j.store.GetGroup(ctx, group)
```
alesstimec: same here.. transaction, please.

internal/jimm/jimm.go
```diff
@@ -90,6 +90,9 @@ type JIMM struct {
 
 	// RoleManager provides a means to manage roles within JIMM.
 	RoleManager RoleManager
+
+	// RoleManager provides a means to manage groups within JIMM.
+	GroupManager GroupManager
```
alesstimec, line 95: does the `GroupManager` need to be exported, or could we use the `GroupManager()` getter method instead?

internal/testutils/jimmtest/fixture.go
```diff
@@ -0,0 +1,126 @@
+// Copyright 2024 Canonical.
+package jimmtest
```
alesstimec, line 3: you know..

internal/testutils/jimmtest/fixture.go
```diff
@@ -0,0 +1,126 @@
...
+//   - cloud
+//   - cloud credential
+//   - role
+//
+// Into the test database, returning the dbmodels to be utilised for values within tests.
+//
+// TODO(ale8k): Make this an implicit thing on the JIMM suite per test & refactor the current state.
+// and make the suite argument an interface of the required calls we use here.
+func CreateTestControllerEnvironment(ctx context.Context, c *qt.C, db db.Database) (
+	dbmodel.Identity,
```
alesstimec, line 35: this needs to return a struct type. there are way to many returns

alesstimec (changes requested):

internal/jimm/group/group.go
```diff
@@ -0,0 +1,146 @@
...
+	"context"
+
+	"github.com/canonical/jimm/v3/internal/common/pagination"
+	"github.com/canonical/jimm/v3/internal/db"
+	"github.com/canonical/jimm/v3/internal/dbmodel"
+	"github.com/canonical/jimm/v3/internal/errors"
+	"github.com/canonical/jimm/v3/internal/openfga"
+)
+
+// permissionManager provides a means to manage roles within JIMM.
```
ale8k: s/permission/group

internal/jimm/jimm.go
kian99: It doesn't need to be exported right now, but the plan is to remove the GroupManager() getter as soon as we can. Up to you whether you want me to make it exported or not right now.

internal/testutils/jimmtest/fixture.go
kian99: I wonder if this isn't the linter always adding this copyright header like this.

internal/testutils/jimmtest/fixture.go
kian99: Agreed, but that's a refactoring for a separate PR.

kian99:
@alesstimec Addressed your comments, ptal.

internal/jimm/jimm.go
```diff
@@ -111,6 +114,25 @@ type RoleManager interface {
 	CountRoles(ctx context.Context, user *openfga.User) (int, error)
 }
 
+// GroupManager provides a means to manage groups within JIMM.
+type GroupManager interface {
```
ale8k, line 118: I thought we weren't doing interfaces for now? If we are, can you add one for roles too please.

internal/jimm/jimm.go
```diff
@@ -135,6 +157,10 @@ func (j *JIMM) GetRoleManager() RoleManager {
 	return j.RoleManager
 }
 
+func (j *JIMM) GetGroupManager() GroupManager {
```
ale8k, line 160: Don't start getters with "Get", it is "neither idiomatic nor necessary" according to effective Go:
https://go.dev/doc/effective_go#Getters

internal/jimmhttp/rebac_admin/groups_test.go
```diff
@@ -28,11 +28,14 @@ import (
 func TestCreateGroup(t *testing.T) {
 	c := qt.New(t)
 	var addErr error
+	groupService := mocks.GroupService{
```
ale8k: Mock is still called GroupService, let's rename it so this doesn't end up really confusing down the line?

ale8k (changes requested):
Some bits and bob needs sorting first, then lgtm.

internal/jimm/jimm.go
ale8k: Nvm gotten confused

internal/jimm/jimm.go
kian99: It can't be named GroupManager because there is a field named GroupManager. Once we get rid of the large jimm interface, we can delete these getters.

internal/jimmhttp/rebac_admin/groups_test.go
kian99: Sure, to `groupMock` or `groupManagerMock`?

kian99: I'll go with just GroupManager since it's in the mock package it should be clear.

internal/jimm/group/group.go
```diff
@@ -0,0 +1,162 @@
...
+type groupManager struct {
+	store   *db.Database
+	authSvc *openfga.OFGAClient
+}
+
+// NewGroupManager returns a new group manager that provides group
+// creation, modification, and removal.
+func NewGroupManager(store *db.Database, authSvc *openfga.OFGAClient) (*groupManager, error) {
+	if store == nil {
+		return nil, errors.E("role store cannot be nil")
```
alesstimec: error message is wrong

internal/jimm/group/group.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+	if authSvc == nil {
+		return nil, errors.E("role authorisation service cannot be nil")
```
alesstimec: error message is wrong

alesstimec (approved):
LGTM but the error messages in the constructor need changing

internal/jimm/group/group.go
kian99: Thanks, fixed.

internal/jujuapi/access_control.go
```diff
@@ -44,7 +33,7 @@ func (r *controllerRoot) AddGroup(ctx context.Context, req apiparams.AddGroupReq
 		return resp, errors.E(op, errors.CodeBadRequest, "invalid group name")
 	}
 
-	groupEntry, err := r.jimm.AddGroup(ctx, r.user, req.Name)
+	groupEntry, err := r.jimm.GetGroupManager().AddGroup(ctx, r.user, req.Name)
```
alesstimec, line 36: what if 
```
r.jimm.GetGroupManager()
```
returns nil?

kian99: I consider that the same as if the `r.jimm.db` is nil or `r.jimm.openFGAClient` is nil. Just let it panic because at that point we messed up. For now, that's more possible to happen because we don't have a jimm constructor but once your PR lands then we have a stronger guarantee this can't be nil.

alesstimec: ok. agreed. these scenarios should be caught by constructors. same way `r.jimm` could be nil, right...

internal/jimm/service_account_test.go
```diff
@@ -240,17 +241,20 @@ func TestGrantServiceAccountAccess(t *testing.T) {
 			}
 			err = pgDb.Migrate(context.Background(), false)
 			c.Assert(err, qt.IsNil)
+			groupManager, err := group.NewGroupManager(&pgDb, ofgaClient)
+			c.Assert(err, qt.IsNil)
 			jimm := &jimm.JIMM{
 				Database:      pgDb,
 				OpenFGAClient: ofgaClient,
+				GroupManager:  groupManager,
```
SimoneDutto, line 249: i don't particularly like this, but i guess it's solved by ales's new constructor

internal/jimmhttp/rebac_admin/groups.go
SimoneDutto, line 1: can we change the groupService constructor to accept the `GroupManager` interface?
So we can get rid of the `jujuapi.JIMM` interface.

SimoneDutto (commented):
one comment

internal/jimm/service_account_test.go
ale8k: Ya, this be solved

internal/testutils/jimmtest/fixture.go
ale8k: ALso this was temporary, wasn't supposed to be a catchall function when I did it but we ended up blasting it everywhere aha. We need a better way of doing this really

ale8k (approved):
lgtm,

internal/jimmhttp/rebac_admin/groups.go
kian99: We can't yet because the group service also does things like `s.jimm.ListRelationshipTuples` and `s.jimm.AddRelation`. At a later stage we can pass in just the managers that are required.

PR #1478, opened 2024-12-03
remove controller watcher

SimoneDutto:
## Description

In this PR we remove the Model Watcher controller from JIMM.
To do so:
- we've handled the modeldestroy with a cleanup routine

Removed:
- jujuclient code
- most of the watcher tests

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [ ] Covered by unit tests


internal/jimm/watcher_test.go
SimoneDutto, line 1: in the previous pr i've renamed the watcher_test.go file to avoid fixing useless tests failing, but now this commit seems to add new test in this watcher_test.go

It is not, I've just removed all tests regarding `ModelWatcher` and readded the  `ModelSummaryWatcher` tests.

internal/jimm/model_cleanup.go
```diff
@@ -0,0 +1,59 @@
...
+
+// CleanupModelsDying loops over dying models, contacting the respective controller.
+// And deleting the model from our database if the error is `NotFound` which means the model was successfully deleted.
+func (j *JIMM) CleanupModelsDying(ctx context.Context) error {
+	const op = errors.Op("jimm.WatchModelsDying")
+
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	err := j.DB().ForEachModel(ctx, func(m *dbmodel.Model) error {
```
kian99, line 23: Instead of for each models, rather list models that are dying.

internal/jimm/model_cleanup.go
```diff
@@ -0,0 +1,59 @@
...
+		if m.Life == state.Dying.String() {
+			// if the model is dying and not found by querying the controller we can assume it is dead.
+			// And safely delete the reference from our db.
+			api, err := j.dialModel(ctx, &m.Controller, m.ResourceTag())
```
kian99: What about following the same pattern as the previous code. Something like:
1. Start a routine for each controller.
2. Open API connection to controller.
3. GetDyingModels(ThisController)
4. Query controller for model status.

If the routine returns with an error, the caller can reschedule after e.g. 30s. It just saves us for making a new connection for each model. Granted this may be a premature optimisation but it feels cleaner and aligns with how the watcher worked before anyway.

kian99 (commented):
Nice stuff! Just 2 comments

internal/jimm/model_cleanup.go
SimoneDutto: Potentially we could do something like this. 
However I though starting a routine for each controller feels like a useless optimization because we run this cleanup every 1 minute, so do we care how long the cleanup is? No user is waiting for this to finish.

Saving the connection for each model is fine, if we move towards listing the dying model.

internal/jimm/model_cleanup.go
SimoneDutto: yep, i was battled because i feel like creating a method in the database called `GetDyingModels()` is not very clever.  I'd rather add a `List(filter)` method in our db methods.
But i remember last time there was some opposition to that, and rather just using "ForEach" to keep it consistent.

cmd/jimmsrv/service/service.go
```diff
@@ -272,6 +261,22 @@ func (s *Service) OpenFGACleanup(ctx context.Context, trigger <-chan time.Time)
 	}
 }
 
+// CleanupModelsDying triggers every `trigger` time and calls the jimm methods to cleanup dying models.
+func (s *Service) CleanupModelsDying(ctx context.Context, trigger <-chan time.Time) error {
```
alesstimec: suggestion: CleanupDyingModels

internal/jimm/model_cleanup.go
```diff
@@ -0,0 +1,59 @@
...
+// CleanupModelsDying loops over dying models, contacting the respective controller.
+// And deleting the model from our database if the error is `NotFound` which means the model was successfully deleted.
+func (j *JIMM) CleanupModelsDying(ctx context.Context) error {
+	const op = errors.Op("jimm.CleanupModelsDying")
+
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	err := j.DB().ForEachModel(ctx, func(m *dbmodel.Model) error {
+		if m.Life == state.Dying.String() {
```
alesstimec: suggestion: revert the condition and exit early -> reduces indentation for the "happy path"
```
if m.Life != state.Dying.String() {
   return nil
}
```

internal/jimm/model_cleanup.go
```diff
@@ -0,0 +1,59 @@
...
+			// if the model is dying and not found by querying the controller we can assume it is dead.
+			// And safely delete the reference from our db.
+			api, err := j.dialModel(ctx, &m.Controller, m.ResourceTag())
+			if err != nil {
```
alesstimec: can you still dial a model if it doesn't exist? wouldn't we expect an error here where we try to dial it?

internal/jimm/model_cleanup.go
```diff
@@ -0,0 +1,59 @@
...
+				zapctx.Error(ctx, fmt.Sprintf("Cannot dial model %s: %s\n", m.UUID.String, err))
+				return nil
+			}
+			if err := api.ModelInfo(ctx, &jujuparams.ModelInfo{UUID: m.UUID.String}); err != nil {
+				// Some versions of juju return unauthorized for models that cannot be found.
```
alesstimec: can you please explain which versions return which error?

internal/jimm/model_cleanup.go
```diff
@@ -0,0 +1,59 @@
...
+					zapctx.Error(ctx, fmt.Sprintf("Cannot get ModelInfo for model %s: %s\n", m.UUID.String, err))
+					return nil
+				}
+			}
+		}
+		return nil
+	})
+	if err != nil {
+		// Ignore temporary database errors.
+		if errors.ErrorCode(err) != errors.CodeDatabaseLocked {
```
alesstimec: we shouldn't need to ignore these errors.. the cleanup worker will simply log the error and retry next time time triggers.

alesstimec (commented):
this looks nice, but i have some questions/suggestions.

internal/jimm/model_cleanup.go
SimoneDutto: i copy/pasted this from the watcher code. I honestly have no idea.

internal/jimm/model_cleanup.go
SimoneDutto: yeah, fair enough

internal/jimm/model_cleanup.go
SimoneDutto: yeah, i think it too. I switched to dialing the controller.
Now I am thinking, if a controller is deleted without being removed from jimm, the dangling models will be there forever.
Shouldn't be an issue right?

internal/jimm/model_cleanup.go
ale8k: I think foreach is fine honestly and dialing the controllers makes a lot of sense to me. Why would we have a db method?

internal/jimm/model_cleanup.go
alesstimec: hmm.. we should deal with controller removal in a separate pr

internal/jimm/model_cleanup.go
alesstimec: it would be nice to verify

internal/jimm/model_cleanup.go
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/canonical/jimm/v3/internal/errors"
+)
+
+// CleanupDyingModels loops over dying models, contacting the respective controller.
+// And deleting the model from our database if the error is `NotFound` which means the model was successfully deleted.
+func (j *JIMM) CleanupDyingModels(ctx context.Context) error {
+	const op = errors.Op("jimm.CleanupDyingModels")
+	zapctx.Info(ctx, string(op))
+
+	ctx, cancel := context.WithCancel(ctx)
```
alesstimec: why do we need to cancel the context?

internal/jimm/model_cleanup.go
```diff
@@ -0,0 +1,56 @@
...
+	defer cancel()
+
+	err := j.DB().ForEachModel(ctx, func(m *dbmodel.Model) error {
```
alesstimec, line 23: ok, we're dialing controllers now. could we get the list of models and unique controllers first, so that we only dial each controller once and check for all dying models on that controller?

internal/jimm/model_cleanup.go
SimoneDutto: https://github.com/canonical/jimm/commit/2d232e156c9688feedeacc202bb4b6f060b4b195
it's a 3 years old comment from Martin

internal/jimm/model_cleanup.go
SimoneDutto: we could, but since the dying models would always few and 10-20 maximum, this shouldn't be an issue right?
But it's up to you, i can make the change

internal/jimm/model_cleanup.go
SimoneDutto: yes, i can do that

internal/jimm/model_cleanup.go
SimoneDutto: let me reprhase it.
I don't expect to have many dying models per controller each time. So the "wasted" connection shouldn't be an issue. It's a single connection created and then dropped after it.
But maybe i am wrong on this!

cmd/jimmsrv/main.go
```diff
@@ -206,6 +203,9 @@ func start(ctx context.Context, s *service.Service) error {
 		s.Go(func() error {
 			return jimmsvc.OpenFGACleanup(ctx, time.NewTicker(6*time.Hour).C)
 		})
+		s.Go(func() error {
```
alesstimec, line 206: interesting.. i thought i move code that starts services to service/service.go

internal/jimm/model.go
```diff
@@ -1165,15 +1165,21 @@ func (j *JIMM) DestroyModel(ctx context.Context, user *openfga.User, mt names.Mo
...
-			// If the database fails to update don't worry too much the
-			// monitor should catch it.
 			zapctx.Error(ctx, "failed to store model change", zaputil.Error(err))
+			return err
 		}
+		if err := api.DestroyModel(ctx, mt, destroyStorage, force, maxWait, timeout); err != nil {
+			zapctx.Error(ctx, "failed to call destroy juju api", zaputil.Error(err))
+			// this is a manual way of restoring the life state to alive if the JUJU api fails.
+			m.Life = state.Alive.String()
+			if err := j.Database.UpdateModel(ctx, m); err != nil {
```
alesstimec: to avoid reader confusion about which error is returned on L1180..

```suggestion
			if uerr := j.Database.UpdateModel(ctx, m); uerr != nil {
```

internal/jimm/model.go
```diff
@@ -1165,15 +1165,21 @@ func (j *JIMM) DestroyModel(ctx context.Context, user *openfga.User, mt names.Mo
...
+				zapctx.Error(ctx, "failed to store model change", zaputil.Error(err))
```
alesstimec: and to follow my comment above, we should log `uerr`
```suggestion
				zapctx.Error(ctx, "failed to store model change", zaputil.Error(uerr))
```

internal/jimm/model_cleanup.go
alesstimec: ok. let's leave it as-is for the first pass, but please add metrics to measure how many dying models we're cleaning up and how many different controllers we're dialing each time.

alesstimec (approved):
LGTM with a few nitpicks/suggestions

cmd/jimmsrv/main.go
SimoneDutto: the target branch here is not v3. I'll rebase after this second slim pr is merged and propose the merge in v3

internal/jimm/model_cleanup.go
SimoneDutto: will do after the merge in v3, i don't want this branch to diverge too much

SimoneDutto: https://warthogs.atlassian.net/browse/JUJU-7288

internal/jimm/model.go
```diff
@@ -1165,15 +1165,21 @@ func (j *JIMM) DestroyModel(ctx context.Context, user *openfga.User, mt names.Mo
...
-		}
 		m.Life = state.Dying.String()
 		if err := j.Database.UpdateModel(ctx, m); err != nil {
-			// If the database fails to update don't worry too much the
-			// monitor should catch it.
 			zapctx.Error(ctx, "failed to store model change", zaputil.Error(err))
+			return err
 		}
+		if err := api.DestroyModel(ctx, mt, destroyStorage, force, maxWait, timeout); err != nil {
+			zapctx.Error(ctx, "failed to call destroy juju api", zaputil.Error(err))
```
ale8k: failed to call destroymodel?

internal/jimm/model_cleanup.go
```diff
@@ -0,0 +1,53 @@
...
+		if err := api.ModelInfo(ctx, &jujuparams.ModelInfo{UUID: m.UUID.String}); err != nil {
+			// Some versions of juju return unauthorized for models that cannot be found.
+			if errors.ErrorCode(err) == errors.CodeNotFound || errors.ErrorCode(err) == errors.CodeUnauthorized {
+				if err := j.DB().DeleteModel(ctx, m); err != nil {
+					zapctx.Error(ctx, fmt.Sprintf("Cannot delete model %s: %s\n", m.UUID.String, err))
+				} else {
+					return nil
+				}
+			} else {
+				zapctx.Error(ctx, fmt.Sprintf("Cannot get ModelInfo for model %s: %s\n", m.UUID.String, err))
```
ale8k: errors shouldn't be capitalised, please adjust all errors

internal/jimm/model_cleanup_test.go
```diff
@@ -0,0 +1,195 @@
...
+			},
+			DestroyModel_: func(ctx context.Context, mt names.ModelTag, b1, b2 *bool, d1, d2 *time.Duration) error {
+				return nil
+			},
+		},
+	}
+	err = j.DestroyModel(ctx, jimmAdmin, names.NewModelTag(env.Models[0].UUID), nil, nil, nil, nil)
+	c.Assert(err, qt.IsNil)
+
+	// test
```
ale8k, line 176: Can just put the rest in a setup function and remove this comment? Yeah actually, make a suite please it's much cleaner. See role for example.

internal/jimm/watcher_test.go
```diff
@@ -0,0 +1,361 @@
...
+  uuid: 00000002-0000-0000-0000-000000000003
+  controller: controller-1
+  cloud: test-cloud
+  region: test-cloud-region
+  cloud-credential: cred-1
+  owner: alice@canonical.com
+  life: dead
+`
+
+var modelSummaryWatcherTests = []struct {
```
ale8k, line 74: Has this file just been renamed?

ale8k (approved):
lgtm, just some minor comments. Please do them :) good work!

internal/jimm/watcher_test.go
SimoneDutto: https://github.com/canonical/jimm/pull/1478#discussion_r1869785805

internal/jimm/model_cleanup_test.go
SimoneDutto: done it

PR #1484, opened 2024-12-06
refactor[internal/openfga] simplifies some model access checks

alesstimec:
## Description

- adds a method to internal/openfga to check if user has specific relation to a model
- removes the need for allowedModelAccess map, which was horrible
- makes doModel method slightly faster because now it does a single openfga check instead of 3
- simplifies userModelAccess in jujuapi
- 
## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [ ] Covered by unit tests


internal/jimm/model.go
```diff
@@ -1194,11 +1194,12 @@ func (j *JIMM) doModel(ctx context.Context, user *openfga.User, mt names.ModelTa
...
 	}
 
-	accessLevel, err := j.GetUserModelAccess(ctx, user, mt)
+	hasAccess, err := user.HasModelRelation(ctx, mt, requireRelation)
 	if err != nil {
 		return errors.E(op, err)
 	}
-	if !allowedModelAccess[access][accessLevel] {
+
+	if !hasAccess {
```
SimoneDutto, line 1288: basically before we were handling hierarchy between roles manually? And now we rely on our authorization model, good

SimoneDutto (approved):
LGTM.

internal/jujuapi/cloud.go
```diff
@@ -276,30 +277,18 @@ func collapseUpdateCredentialResults(args jujuparams.TaggedCredentials, updateRe
...
-	if hasWriteAccess {
-		return "write", nil
-	}
-	hasReadAccess, err := user.IsModelReader(ctx, model)
-	if err != nil {
-		return "", errors.E(err)
-	}
-	if hasReadAccess {
-		return "read", nil
+func userModelAccess(ctx context.Context, user *openfga.User, model names.ModelTag) string {
```
SimoneDutto, line 280: godoc please

internal/jujuapi/cloud.go
```diff
@@ -276,30 +277,18 @@ func collapseUpdateCredentialResults(args jujuparams.TaggedCredentials, updateRe
...
+	userRelation := user.GetModelAccess(ctx, model)
+	switch userRelation {
+	case ofganames.AdministratorRelation:
```
ale8k, line 283: I know we have ofganames now, but I really think permission package with access constants would've been better from juju.

ale8k (approved):
lgtm ty

PR #1485, opened 2024-12-09
JUJU-7289 remove identity defaults

SimoneDutto:
## Description

This PR removes IdentityModelDefaults.

IdentityModelDefaults were used when building a model to set up config together with CloudDefault..

But the `Set` method was never used for IdentityModelDefaults, so after an investigation, it turns out it is not used because to set config per Model per User, we can just use `CloudDefaults`.

`juju model-defaults` automatically fills the cloud arg in case of a single cloud, in case of multiple clouds you need to manually specify one.

```
# from juju-cli model-defaults
Model default configuration settings are specific to the cloud on which the
model is deployed. If the controller hosts more than one cloud, the cloud
(and optionally region) must be specified using the --cloud flag. This flag
accepts arguments in the following forms:
```

image

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [ ] Covered by unit tests


ale8k (approved):
lgtm, still a little lost on their purpose

internal/dbmodel/sql/postgres/1_16.sql
```diff
@@ -0,0 +1,5 @@
+-- 1_16.sql is a migration to delete identitymodel defaults
+DROP TABLE identity_model_defaults;
```
alesstimec, line 2: it ok for this PR, but (and i'll keep complaining about this) we will need to be more careful in the future as there will be JIMMs running in the wild and at some point (hard as it might be to believe) some update will fail and customers will need to roll back the update..

so i propose we start dropping tables/columns a release after the code change - a more elaborate migration strategy needs to be discussed, of course

internal/jimm/model.go
```diff
@@ -577,14 +577,8 @@ func (j *JIMM) AddModel(ctx context.Context, user *openfga.User, args *ModelCrea
...
 
-	// fetch user model defaults
-	userConfig, err := j.IdentityModelDefaults(ctx, user.Identity)
-	if err != nil && errors.ErrorCode(err) != errors.CodeNotFound {
-		return nil, errors.E(op, "failed to fetch cloud defaults")
-	}
-	builder = builder.WithConfig(userConfig)
-
 	// fetch cloud defaults
+	// TODO(SimoneDutto): we should get the implicit cloud and then get the defaults.
```
alesstimec, line 581: yes, if there is only one cloud known to JIMM. otherwise there is no such thing as a default cloud (i think)

alesstimec (approved):
LGTM

internal/jimm/model.go
SimoneDutto: i am working on this

internal/dbmodel/sql/postgres/1_16.sql
SimoneDutto: as soon as kian is back we could have a meeting

ale8k: Well I think in future you'd have an up/down step in migrations, this will solve that. Once we're on golang-migrate, we can write a bunch of up/down @alesstimec so don't worry too much. For example, an up (which drops a table) could simply just move data into a jsonb table for situations like this.

PR #1487, opened 2024-12-10
fix model default order

SimoneDutto:
## Description

Before we get the model defaults before getting the `implicit` cloud and cloud region, this could result in having no default.
Now we first get the implicit cloud and then we get the defaults.

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [ ] Covered by unit tests


ale8k (approved):
Not 100% what implicit cloud is, rember us discussing it on call but sounds right... approving.

internal/jimm/model.go
```diff
@@ -245,6 +245,7 @@ func (b *modelBuilder) WithCloudRegion(region string) *modelBuilder {
 				continue
 			}
 			region = r.Name
+			break
```
SimoneDutto, line 250: i've added this break, because before it seemed to me it was wrong to continue the loop if you've already found a region.
Let me know if we were relying on this behaviour for some reason

alesstimec (approved):

PR #1488, opened 2024-12-10
Slim jimm model

SimoneDutto:
## Description

feature-brach into v3, with merge conflicts

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [ ] Covered by unit tests


alesstimec (approved):

internal/jimm/model.go
```diff
@@ -742,16 +743,101 @@ func (j *JIMM) ModelInfo(ctx context.Context, user *openfga.User, mt names.Model
 	return j.mergeModelInfo(ctx, user, mi, m)
 }
 
+// modelSummariesMap is a safe map to add records concurrently because the access is guarded by a Mutex.
+// The read operations are not guarded because only inserts are done concurrently.
+type modelSummariesMap struct {
+	mu             sync.Mutex
```
ale8k, line 749: Can just use sync.Map, no need for mutex @SimoneDutto

internal/jimm/watcher.go
```diff
@@ -343,9 +343,9 @@ func (w *Watcher) watchController(ctx context.Context, ctl *dbmodel.Controller)
 						machines++
 						cores += n
 					}
-					m.Cores = cores
-					m.Machines = machines
-					m.Units = int64(len(v.units))
+					// m.Cores = cores
+					// m.Machines = machines
+					// m.Units = int64(len(v.units))
```
ale8k: This is left commented out?

internal/jujuapi/modelmanager.go
```diff
@@ -71,6 +71,7 @@ type ModelManager interface {
 	ImportModel(ctx context.Context, user *openfga.User, controllerName string, modelTag names.ModelTag, newOwner string) error
 	ModelDefaultsForCloud(ctx context.Context, user *dbmodel.Identity, cloudTag names.CloudTag) (jujuparams.ModelDefaultsResult, error)
 	ModelInfo(ctx context.Context, u *openfga.User, mt names.ModelTag) (*jujuparams.ModelInfo, error)
+	ModelSummaries(ctx context.Context, user *openfga.User, maskingControllerUUID string) (jujuparams.ModelSummaryResults, error)
```
ale8k, line 74: This interface is confusing lol (not you)

internal/jimm/model.go
```diff
@@ -742,16 +743,101 @@ func (j *JIMM) ModelInfo(ctx context.Context, user *openfga.User, mt names.Model
...
+	defer m.mu.Unlock()
+	if m.modelSummaries == nil {
+		m.modelSummaries = make(map[string]jujuparams.ModelSummaryResult)
+	}
+	m.modelSummaries[summary.Result.UUID] = summary
+}
+
+// ModelSummaries returns the list of modelsummary the user has access to.
+// It queries the controllers and then merge the info from the JIMM db.
+func (j *JIMM) ModelSummaries(ctx context.Context, user *openfga.User, maskingControllerUUID string) (jujuparams.ModelSummaryResults, error) {
```
ale8k, line 764: Probavbly better to just name it ListModelSummaries right?

ale8k (approved):
some comments... Feel like we can do them in this branch

internal/jimm/watcher.go
SimoneDutto: this was then removed all togethere in the commit after, so no worries

internal/jimm/model.go
SimoneDutto: https://pkg.go.dev/sync#Map

from the go doc, you use sync.Map for very specific use case. Usually they advice you to use a separate mutex.

ale8k: Ah I see, so if routine A handles keys 1-10, and routine B handled keys 11-20, then sync.Map is preferred. Never noticed the second part really.

PR #1491, opened 2024-12-11
Juju 7288/add prom metric for jimm methods

SimoneDutto:
## Description

Add Prometheus for jimm's methods.
In particular in this pr, I've added the tracking for Cleanup functions.

> fly-by update name of one jimm's method (second commit)

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [ ] Covered by unit tests

### DIFF
--- a/internal/jimm/access.go
+++ b/internal/jimm/access.go
@@ -23,6 +23,7 @@ import (
 	"github.com/canonical/jimm/v3/internal/jimmjwx"
 	"github.com/canonical/jimm/v3/internal/openfga"
 	ofganames "github.com/canonical/jimm/v3/internal/openfga/names"
+	"github.com/canonical/jimm/v3/internal/servermon"
 	jimmnames "github.com/canonical/jimm/v3/pkg/names"
 )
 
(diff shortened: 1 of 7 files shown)


ale8k (approved):

internal/jimm/access.go
```diff
@@ -723,10 +724,14 @@ func (j *JIMM) parseAndValidateTag(ctx context.Context, key string) (*ofganames.
 //
 // This approach to cleaning up tuples is intended to be temporary while we implement
 // a better approach to eventual consistency of JIMM's database objects and OpenFGA tuples.
-func (j *JIMM) OpenFGACleanup(ctx context.Context) error {
+func (j *JIMM) OpenFGACleanup(ctx context.Context) (err error) {
+	const op = errors.Op("jimm.CleanupDyingModels")
+	zapctx.Info(ctx, string(op))
+	durationObserver := servermon.DurationObserver(servermon.JimmMethodsDurationHistogram, string(op))
+	defer durationObserver()
+	defer servermon.ErrorCounter(servermon.DBQueryErrorCount, &err, string(op))
```
alesstimec: why are we using the `DBQueryErrorCount` here? i don't think this is what the metric was intended to measure.

internal/jimm/model_cleanup.go
```diff
@@ -12,15 +12,19 @@ import (
...
 
 // CleanupDyingModels loops over dying models, contacting the respective controller.
 // And deleting the model from our database if the error is `NotFound` which means the model was successfully deleted.
-func (j *JIMM) CleanupDyingModels(ctx context.Context) error {
+func (j *JIMM) CleanupDyingModels(ctx context.Context) (err error) {
 	const op = errors.Op("jimm.CleanupDyingModels")
 	zapctx.Info(ctx, string(op))
+	durationObserver := servermon.DurationObserver(servermon.JimmMethodsDurationHistogram, string(op))
+	defer durationObserver()
+	defer servermon.ErrorCounter(servermon.DBQueryErrorCount, &err, string(op))
```
alesstimec: again, this is not the right metric to use here. `DBQueryErroCount` is intended for database query errors only.

alesstimec (changes requested):

internal/jimm/access.go
SimoneDutto: you are right. removed it

internal/jimm/model_cleanup.go
SimoneDutto: you are right. removed it

alesstimec (approved):

PR #1493, opened 2024-12-11
feat(modelmanager): listmodels

ale8k:
ListModels now correctly calls ListModels. Dialing the controllers themselves and returning models only those users can view. This is used upon Login and Register commands to persist model details, and whether or not it is a CAAS model.

We don't return controller models as with JIMM this does not make sense - we intentionally choose not to expose these details to the user.

Driveby: The api was trying to call /version, but our mock didn't implement it.

## Description

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [x] Covered by unit tests


internal/jujuclient/modelmanager.go
```diff
@@ -325,3 +327,14 @@ func (c Connection) ChangeModelCredential(ctx context.Context, model names.Model
...
+
+// ListModels returns the models that the specified user
+// has access to in the current server.  Only that controller owner
+// can list models for any user (at this stage).  Other users
+// can only ask about their own models.
+//
+// In our wrapper, we ask for the owner. So expect ALL models from
+// the controller.
+func (c Connection) ListModels(ctx context.Context) ([]base.UserModel, error) {
+	return modelmanager.NewClient(&c).ListModels("admin")
```
SimoneDutto, line 337: learning question:
if a local user creates a model, and then the controller is migrated to jaas.
What would be the owneridentity of that model under the juju controller? And under JIMM?

internal/jujuapi/modelmanager_test.go
```diff
@@ -221,23 +221,35 @@ func (s *modelManagerSuite) TestListModelSummariesWithoutControllerUUIDMasking(c
 }
 
 func (s *modelManagerSuite) TestListModels(c *gc.C) {
-	conn := s.open(c, nil, "bob")
+	conn := s.open(c, nil, "charlie@canonical.com")
 	defer conn.Close()
 
 	client := modelmanager.NewClient(conn)
-	models, err := client.ListModels("bob")
+	models, err := client.ListModels("charlie@canonical.com")
```
SimoneDutto, line 228: since, we are not really filtering based of this email when using the juju client. Can we use another more expressive email?
Like: anyuser@canonical.com

internal/jimm/model.go
```diff
@@ -1347,3 +1351,67 @@ func (j *JIMM) ChangeModelCredential(ctx context.Context, user *openfga.User, mo
...
+
+		// Filter the models returned according to the uuids
+		// returned from OpenFGA for read access.
+		//
+		// NOTE: We skip controller models as ListModels is used for login and register.
+		// The models returned are stored locally and used for reference. In the case of JIMM,
+		// we do not want to show the controller models.
+		for _, m := range ums {
+			if slices.Contains(uuids, m.UUID) {
+				userModels = append(userModels, m)
```
SimoneDutto: i dont think you can return it right away from the controller, but rather override the `OwnerTag` from our db

SimoneDutto (changes requested):

internal/jujuclient/modelmanager.go
ale8k: We have a flag on import model to change the owner from a local user to an external user. This call specifically though is getting all models, so we use our auth system to filter and not juju.

internal/jujuapi/modelmanager_test.go
ale8k: We are filtering on this user? Whatever models this user has access to we filter upon.

internal/jimm/model.go
ale8k: I'm not sure what you mean here?

internal/jujuclient/modelmanager.go
alesstimec: jimm does not care about models created by local users.. unless we intentionally import the model.

internal/jimm/jimm.go
```diff
@@ -494,6 +494,15 @@ type API interface {
 
 	// ListStorageDetails lists all storage.
 	ListStorageDetails(ctx context.Context) ([]jujuparams.StorageDetails, error)
+
+	// ListModels returns the models that the specified user
```
alesstimec: s/specified/logged in/

internal/jimm/jimm.go
```diff
@@ -494,6 +494,15 @@ type API interface {
...
+	// has access to in the current server.  Only that controller owner
```
alesstimec: "current server"? and only the jimm administrator or controller superuser  can list models for any user?

internal/jimm/jimm.go
```diff
@@ -494,6 +494,15 @@ type API interface {
...
+	// can list models for any user (at this stage).  Other users
+	// can only ask about their own models.
+	//
+	// In our wrapper, we ask for the owner. So expect ALL models from
```
alesstimec: we "ask" as controller admin, i think..

internal/jimm/jimm.go
```diff
@@ -494,6 +494,15 @@ type API interface {
...
+	// the controller.
+	ListModels(ctx context.Context) ([]base.UserModel, error)
```
alesstimec, line 499: also, i think implementation detail belong on the modelmanager implementation rather than the interface.

internal/jimm/model.go
```diff
@@ -1347,3 +1351,67 @@ func (j *JIMM) ChangeModelCredential(ctx context.Context, user *openfga.User, mo
...
+	const op = errors.Op("jimm.ListModels")
+	zapctx.Info(ctx, string(op))
+
+	// Get models uuids user has access to
+	uuids, err := user.ListModels(ctx, ofganames.ReaderRelation)
+	if err != nil {
+		return nil, errors.E(op, err, "failed to list models")
+	}
+
+	// Get the models themselves
```
alesstimec: get models from the database?

internal/jimm/model.go
```diff
@@ -1347,3 +1351,67 @@ func (j *JIMM) ChangeModelCredential(ctx context.Context, user *openfga.User, mo
...
 	return nil
 }
+
+// ListModels list the models that the user has access to. It intentionally excludes the
+// controller model as this call is used within the context of login and register commands.
+func (j *JIMM) ListModels(ctx context.Context, user *openfga.User) ([]base.UserModel, error) {
+	const op = errors.Op("jimm.ListModels")
+	zapctx.Info(ctx, string(op))
+
+	// Get models uuids user has access to
```
alesstimec:
```suggestion
	// Get uuids of models user has access to
```

internal/jimm/model.go
```diff
@@ -1347,3 +1351,67 @@ func (j *JIMM) ChangeModelCredential(ctx context.Context, user *openfga.User, mo
...
+	uuids, err := user.ListModels(ctx, ofganames.ReaderRelation)
+	if err != nil {
+		return nil, errors.E(op, err, "failed to list models")
```
alesstimec: i'm thinking about the conversation the juju team had about error messages. perhaps we too should try to have more informative error messages.. perhaps something like
```
return nil, errors.E(op, "failed to list user models", err)
```
that way the error states the intent of the operation that failed..

internal/jimm/model.go
```diff
@@ -1347,3 +1351,67 @@ func (j *JIMM) ChangeModelCredential(ctx context.Context, user *openfga.User, mo
...
+	}
+
+	// Get the models themselves
+	models, err := j.DB().GetModelsByUUID(ctx, uuids)
```
alesstimec, line 1367: we need to a bit careful here. .there might be some leftover tuples in openfga that would state users still have access to models that have since been deleted.. in this case we would be unable to fetch model info from the database, since those no longer exists, but that is not an error.. just a temporary inconsistency..

internal/jimm/model.go
```diff
@@ -1347,3 +1351,67 @@ func (j *JIMM) ChangeModelCredential(ctx context.Context, user *openfga.User, mo
...
+	if err != nil {
+		return nil, errors.E(op, err, "failed to get models by uuid")
+	}
+
+	// Find the controllers these models reside on and remove duplicates
+	var controllers []dbmodel.Controller
```
alesstimec, line 1375: i wonder if there's a way we could write this bit with generics or something, because we seem to be looking for unique lists in a few places in jimm code. also not for this pr :)

internal/jimm/model.go
```diff
@@ -1347,3 +1351,67 @@ func (j *JIMM) ChangeModelCredential(ctx context.Context, user *openfga.User, mo
...
+		if err != nil {
+			return err
+		}
+		mutex.Lock()
+		defer mutex.Unlock()
+
+		// Filter the models returned according to the uuids
+		// returned from OpenFGA for read access.
+		//
+		// NOTE: We skip controller models as ListModels is used for login and register.
```
alesstimec: i don't see where we skip controller models..

internal/jimm/model_test.go
```diff
@@ -3804,6 +3805,199 @@ controllers:
...
+  controller-access: superuser
+`
+
+var modelListTests = []struct {
+	name                           string
+	env                            string
+	username                       string
+	expectedUserModels             []base.UserModel
+	expectedError                  string
+	listModelsMockByControllerName map[string]func(context.Context) ([]base.UserModel, error)
```
alesstimec, line 3897: would it not be better to just have a jimmtest.DialerMap here? otherwise you force each test case to test with an equal number of controllers.

internal/jimm/model.go
```diff
@@ -1347,3 +1351,67 @@ func (j *JIMM) ChangeModelCredential(ctx context.Context, user *openfga.User, mo
...
+
+	// Call controllers for their models. We always call as admin, and we're
+	// filtering ourselves. We do this rather than send the user to be 100%
+	// certain that the models do belong to user according to OpenFGA. We could
+	// in theory rely on Juju correctly returning the models (by owner), but this
+	// is more reliable.
+	var userModels []base.UserModel
+	var mutex sync.Mutex
+	err = j.forEachController(ctx, controllers, func(_ *dbmodel.Controller, api API) error {
+		ums, err := api.ListModels(ctx)
```
alesstimec, line 1394: here we list all models as superusers. ideally we would skip all the models owned by local users, but i think it should be fine as we only return models owned by the user that made the call..

alesstimec (changes requested):

internal/jimm/model.go
```diff
@@ -1347,3 +1351,67 @@ func (j *JIMM) ChangeModelCredential(ctx context.Context, user *openfga.User, mo
 
 	return nil
 }
+
+// ListModels list the models that the user has access to. It intentionally excludes the
+// controller model as this call is used within the context of login and register commands.
+func (j *JIMM) ListModels(ctx context.Context, user *openfga.User) ([]base.UserModel, error) {
```
SimoneDutto, line 1356: thinking about this implementation a little more, i think it's different from `ListModelSummaries` in an important way.

`ListModelSummaries` uses the models in the db as the source and then uses the controller as another source to merge info before returning the facade. And that's the reason why you don't need to check for uuid to be present in the list of user uuids.
Since our source of truth is always the database and not the openfga tuples, i think you should switch to a similar implementation to `ListModelSummaries`.

internal/jimm/model.go
SimoneDutto: In your implementation we return the `OwnerTag` coming from the controllers, in most of the other model's method we override the `OwnerTag` with the one in our database.
Are they always the same? https://github.com/canonical/jimm/blob/19a40d7c399638e56dc33bc5a876b0aae7a56375/internal/dbmodel/model.go#L152

SimoneDutto: another example:
https://github.com/canonical/jimm/blob/19a40d7c399638e56dc33bc5a876b0aae7a56375/internal/dbmodel/model.go#L129

internal/jujuapi/modelmanager_test.go
SimoneDutto: i've taken a look at the suite, i didn't realise charlie had access to these two models.
Ignore this comment.

internal/jujuclient/modelmanager.go
SimoneDutto: ty

internal/jimm/model.go
ale8k: I agree, I liked that, changed

internal/jimm/model.go
ale8k: Updated

internal/jimm/model.go
ale8k: I'm really not sure what you mean, this call is returning models the user has access to?

internal/jimm/model.go
ale8k: Mhmmm... You're thinking discover the models from the controllers?

internal/jimm/model.go
ale8k: Already on it :P

internal/jimm/model.go
ale8k: I guess it is a little inefficient in that regard? But I didn't want us to rely on Juju's auth to determine what models we see.

internal/jimm/model.go
ale8k: But checking UUIDs, we never relate users to controller models.

internal/jimm/model.go
ale8k: This is models the user has *access* to, not that they necessarily *own* the model?

internal/jimm/model_test.go
ale8k: I prefer how it is, less bloat in the test case.

internal/jimm/jimm.go
ale8k: This is the calls original godoc. Going to leave it as it's more accurate

internal/jimm/jimm.go
ale8k: This is from the original godoc of juju

internal/jimm/jimm.go
ale8k: Yes sorry, updating

internal/jimm/jimm.go
ale8k: Not sure what you mean here?

internal/jimm/model.go
ale8k: By*

internal/jimm/model.go
alesstimec: no, we just need to fetch models one by one and fail gracefully if a model cannot be found in the db. it might be because of a leftover tuple in openfga.

internal/jimm/model.go
ale8k: As disussed is OK, resolving...

internal/jimm/model.go
ale8k: As discussed, resolving.

internal/jimm/model.go
ale8k: As discussed, resolving.

internal/jimm/jimm.go
alesstimec: Interfaces should not prescribe implementation details.

ale8k: I think I'd literally just say the exact same thing, bar the wrapper part, want me to remove "In our wrapper we ask for the owner." and just leave expected all models?

internal/jimm/model.go
alesstimec: Then amend the godoc to state that rather than mentioning implementation that does not exist, which just confuses the reader.

ale8k: Ah I get you, amended

internal/jimm/jimm.go
ale8k: Updated

internal/jimm/model_test.go
```diff
@@ -3804,6 +3805,199 @@ controllers:
...
+						"controller-1": &jimmtest.Dialer{
+							API: &jimmtest.API{
+								ListModels_: test.listModelsMockByControllerName["controller-1"],
+							},
+						},
+						"controller-2": &jimmtest.Dialer{
+							API: &jimmtest.API{
+								ListModels_: test.listModelsMockByControllerName["controller-2"],
+							},
+						},
```
SimoneDutto, line 3979: nice

internal/jujuapi/modelmanager.go
```diff
@@ -124,7 +124,34 @@ func (r *controllerRoot) ListModelSummaries(ctx context.Context, _ jujuparams.Mo
 // ListModels returns the models that the authenticated user
 // has access to. The user parameter is ignored.
 func (r *controllerRoot) ListModels(ctx context.Context, _ jujuparams.Entity) (jujuparams.UserModelList, error) {
-	return r.allModels(ctx)
```
SimoneDutto, line 127: now, we can either delete this `allModels` and `Model.ToJujuModel()` or you need to refactor the other consumer of this function.

internal/jimm/model.go
```diff
@@ -1347,3 +1351,76 @@ func (j *JIMM) ChangeModelCredential(ctx context.Context, user *openfga.User, mo
...
+	// Get the models from the database
+	models, err := j.DB().GetModelsByUUID(ctx, uuids)
+	if err != nil {
+		return nil, errors.E(op, err, "failed to get models by uuid")
+	}
+
+	// Find the controllers these models reside on and remove duplicates
+	var controllers []dbmodel.Controller
+	seen := make(map[uint]bool)
+	for _, model := range models {
```
SimoneDutto, line 1377: since you are already iterating over models, why don't you build a map with.
map uuid-{model}?

So later down instead of doing: 
```
if slices.Contains(uuids, um.UUID) {
  // Find that model in the db models
  index := slices.IndexFunc(models, func(m dbmodel.Model) bool {
	  return m.UUID.String == um.UUID
  })
```
you can do
if `modelmap[uuid]`.
Which is O(1) instead of O(n^2)

SimoneDutto (changes requested):

internal/jujuapi/modelmanager.go
ale8k: Follow up pr for all models, it's used in AllModels still, don't want to do all in this PR, was going to do it in the next.

internal/jimm/model.go
SimoneDutto: nit, obviously

internal/jujuapi/modelmanager.go
SimoneDutto: okok, perfect

internal/jimm/model.go
ale8k: Agreed, much nicer, changed

SimoneDutto (approved):
good job

internal/jimm/model.go
```diff
@@ -1347,3 +1351,76 @@ func (j *JIMM) ChangeModelCredential(ctx context.Context, user *openfga.User, mo
...
+	models, err := j.DB().GetModelsByUUID(ctx, uuids)
+	if err != nil {
+		return nil, errors.E(op, err, "failed to get models by uuid")
+	}
+
+	// Create map for lookup later
+	modelsMap := make(map[string]dbmodel.Model)
+	for _, m := range models {
+		modelsMap[m.UUID.String] = m
+	}
```
SimoneDutto: move this code at line 1383

internal/jimm/model.go
```diff
@@ -1347,3 +1351,73 @@ func (j *JIMM) ChangeModelCredential(ctx context.Context, user *openfga.User, mo
...
+		if err != nil {
+			return err
+		}
+		mutex.Lock()
+		defer mutex.Unlock()
+
+		// Filter the models returned according to the uuids
+		// returned from OpenFGA for read access.
+		//
+		// NOTE: The controller models are skipped because we do not relate users
```
alesstimec: this godoc is a bit repetitive - keeps explaining that we're not returning controller models

internal/jimm/model.go
```diff
@@ -1347,3 +1351,73 @@ func (j *JIMM) ChangeModelCredential(ctx context.Context, user *openfga.User, mo
...
+		// to controller models, we skip the controller models as ListModels is used for
+		// login and register - the models returned are stored locally and used for reference.
+		// In the case of JIMM, we do not want to show the controller models.
+		for _, um := range ums {
+			// Filter models that match authorised uuids list
+			if slices.Contains(uuids, um.UUID) {
+				um.Owner = modelsMap[um.UUID].OwnerIdentityName
```
alesstimec: this might panic, please make sure um.UUID exists in the modelsMap before getting its OwnerIdentityName

internal/jimm/model.go
```diff
@@ -1347,3 +1351,73 @@ func (j *JIMM) ChangeModelCredential(ctx context.Context, user *openfga.User, mo
...
+		// Filter the models returned according to the uuids
+		// returned from OpenFGA for read access.
+		//
+		// NOTE: The controller models are skipped because we do not relate users
+		// to controller models, we skip the controller models as ListModels is used for
+		// login and register - the models returned are stored locally and used for reference.
+		// In the case of JIMM, we do not want to show the controller models.
+		for _, um := range ums {
+			// Filter models that match authorised uuids list
+			if slices.Contains(uuids, um.UUID) {
```
SimoneDutto: you can remove this check. and replace it with `model, ok = modelsMap[um.UUID]; if ok`.
if it's inside the map it's inside `uuids`.

ale8k: done

alesstimec (approved):

PR #1495, opened 2024-12-12
cve fix

ale8k:
## Description

High CVE fix. Fixes: https://github.com/canonical/jimm/security/dependabot/16 .

## Engineering checklist
*Check only items that apply*

- [ ] Documentation updated
- [ ] Covered by unit tests


SimoneDutto (approved):
can you add a link to the security issue? so in the future we can try why this pr was merged in.

alesstimec (approved):

ale8k:
> can you add a link to the security issue? so in the future we can try why this pr was merged in.

Like copy and paste the link from security tab? I think we probably want a standard for this. Like another PR template specifically for CVEs. What do you think?

SimoneDutto:
> > can you add a link to the security issue? so in the future we can try why this pr was merged in.
> 
> Like copy and paste the link from security tab? I think we probably want a standard for this. Like another PR template specifically for CVEs. What do you think?

yes, a standard would be great. Until we have it, a simple c/p of the link from the security tab is ok

PR #1497, opened 2024-12-16
add test for group deletion

SimoneDutto:
## Description
we were missing a test in the dbmodel pkg for deleted groups.

## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests

### DIFF
--- a/internal/dbmodel/group_test.go
+++ b/internal/dbmodel/group_test.go
@@ -37,7 +37,8 @@ func TestGroupEntry(t *testing.T) {
 	c.Assert(ge3, qt.DeepEquals, ge)
 }
 
-func TestDeleteGroupEntry(t *testing.T) {
+// test we hard delete groups
+func TestHardDeleteGroupEntry(t *testing.T) {
 	c := qt.New(t)
 	db := gormDB(t)
 
--- a/internal/dbmodel/group_test.go
+++ b/internal/dbmodel/group_test.go
@@ -37,7 +37,7 @@ func TestGroupEntry(t *testing.T) {
 	c.Assert(ge3, qt.DeepEquals, ge)
 }
 
-// test we hard delete groups, so when we can recreate the group with the same name after deleting it.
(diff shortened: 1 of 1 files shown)


internal/dbmodel/group_test.go
```diff
@@ -36,3 +36,19 @@ func TestGroupEntry(t *testing.T) {
 	c.Assert(result.Error, qt.IsNil)
 	c.Assert(ge3, qt.DeepEquals, ge)
 }
+
+func TestDeleteGroupEntry(t *testing.T) {
```
ale8k: TestHardDeleteGroupEntry is a better name. Maybe even a comment too explaining why we're doing this test would be good.

ale8k (approved):
approving but please do my comment.

internal/dbmodel/group_test.go
```diff
@@ -36,3 +36,20 @@ func TestGroupEntry(t *testing.T) {
 	c.Assert(result.Error, qt.IsNil)
 	c.Assert(ge3, qt.DeepEquals, ge)
 }
+
+// test we hard delete groups
+func TestHardDeleteGroupEntry(t *testing.T) {
```
alesstimec, line 41: i think this test would've passed with soft deletes as well.. what you need to do is
1. create a group
2. delete the created group
3. try to create another group with the same name

alesstimec (changes requested):

internal/dbmodel/group_test.go
SimoneDutto: did it

internal/dbmodel/group_test.go
```diff
@@ -36,3 +36,25 @@ func TestGroupEntry(t *testing.T) {
 	c.Assert(result.Error, qt.IsNil)
 	c.Assert(ge3, qt.DeepEquals, ge)
 }
+
+// test we hard delete groups, so when we can recreate the group with the same name after deleting it.
```
ale8k: Godocs should start with the func name

alesstimec (approved):

PR #1498, opened 2024-12-16
refactor(addhostedcloud): refactors AddHostedCloud

alesstimec:
## Description

AddHostedCloud has been a bit misleading in how one could specify the host cloud and region.

Previously we would only accept `<cloud-type>/<cloud-region>`.

This PR proposes a change to accept:
- `<cloud-name>/<cloud-region>`
- `<cloud-type>/<cloud-region>`
- `<cloud>`   -> in this case there needs to be only one region for that cloud

This aligns more closely to what juju accepts in the `add-k8s` command, but we'll need to rethink that command in the future: i suspect the `--cloud` and `--region` flags were added specifically for JAAS. But the JIMM admin may want to be more specific as to which controller they want to add the cloud to - future discussions.

## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests


internal/jimm/cloud.go
```diff
@@ -213,15 +213,43 @@ func (j *JIMM) AddHostedCloud(ctx context.Context, user *openfga.User, tag names
...
+		region = &cl.Regions[0]
+	} else {
+		parts := strings.Split(cloud.HostCloudRegion, "/")
+		if len(parts) != 2 || parts[0] == "" {
+			return errors.E(op, errors.CodeBadRequest, fmt.Sprintf("invalid cloud/region format %q", cloud.HostCloudRegion))
+		}
+
+		findRegionFunctions := []func(context.Context, string, string) (*dbmodel.CloudRegion, error){
+			j.Database.FindRegionByCloudType,
+			j.Database.FindRegionByCloudName,
```
kian99: Technically this will cause an issue if you have two clouds, where the name of one cloud is the same as the "type" of the other. I don't know where the cloud types come from through?
E.g. 
Cloud 1: 
Type=k8s
Name=myCloud

Cloud 2:
Type=foo
Name=k8s

Maybe this situation is impossible/unlikely.

alesstimec: juju has a cloud type defined per provider in their code..

kian99: This is likely not a concern for this PR since we don't seem to accept `<cloud-name>/<cloud-region>` currently anyway but what if a user's cloud is named the same as a cloud type, might make things hairy. Anyway, I think the PR looks good, approving.

kian99 (approved):

internal/jimm/cloud.go
```diff
@@ -213,15 +213,43 @@ func (j *JIMM) AddHostedCloud(ctx context.Context, user *openfga.User, tag names
...
+		cl := dbmodel.Cloud{
+			Name: cloud.HostCloudRegion,
+		}
+		if err := j.Database.GetCloud(ctx, &cl); err != nil {
+			return errors.E(op, errors.CodeBadRequest, "unable to find cloud %q", cloud.HostCloudRegion)
+		}
+		if len(cl.Regions) > 1 {
+			return errors.E(op, errors.CodeBadRequest, "unable to determine a unique cloud region for cloud %q", cloud.HostCloudRegion)
+		}
+		region = &cl.Regions[0]
```
SimoneDutto: can `Regions` be empty? above you don't guard against len = 0

internal/jimm/cloud.go
```diff
@@ -213,15 +213,43 @@ func (j *JIMM) AddHostedCloud(ctx context.Context, user *openfga.User, tag names
...
-		return errors.E(op, err)
+
+	var region *dbmodel.CloudRegion
+	var err error
+	if !strings.Contains(cloud.HostCloudRegion, "/") {
+		cl := dbmodel.Cloud{
+			Name: cloud.HostCloudRegion,
+		}
+		if err := j.Database.GetCloud(ctx, &cl); err != nil {
+			return errors.E(op, errors.CodeBadRequest, "unable to find cloud %q", cloud.HostCloudRegion)
```
SimoneDutto: in the godoc of this method there is `If the requested cloud cannot be created on this JAAS system an error with a
 code of CodeIncompatibleClouds will be returned`.
 Maybe there was a reason for it (comply with some juju cli paradigm).
Either change the error code or the godoc.

internal/jimm/cloud.go
```diff
@@ -213,15 +213,43 @@ func (j *JIMM) AddHostedCloud(ctx context.Context, user *openfga.User, tag names
...
+			if err == nil {
+				break
+			}
+		}
+		if err != nil {
+			if errors.ErrorCode(err) == errors.CodeNotFound {
+				return errors.E(op, err, errors.CodeNotFound, fmt.Sprintf("unable to find cloud/region %q", cloud.HostCloudRegion))
+			}
+			return err
+		}
```
SimoneDutto: since all of this code is just to extract the controllers from the region, could we export this logic to a separate function? it would make this already complex function a little easier to read.

SimoneDutto (commented):
looking good, left some comments

SimoneDutto (approved):
lgtm

PR #1500, opened 2024-12-17
feat(jimmctl migrate): enable users to specify model names

### LINKED ISSUE #1371: Improve usability of the migrate command
In the [migrate your model doc](https://canonical-jaas-documentation.readthedocs-hosted.com/en/latest/how-to/migrate_models_internal/#migrate-your-model) it is stated:
>The following command will migrate a model named my-model to the desired controller, in this case called my-controller.
>
>```
>MODEL_NAME=my-model
>MODEL_UUID=$(juju show-model $MODEL_NAME --format yaml | yq .$MODEL_NAME.model-uuid)
>jimmctl migrate my-controller $MODEL_UUID
>```

The user shouldn't need to run a juju command to get the the model UID to use that in a jimmctl command. To improve the usability, `jimmctl` should be able to do `jimmctl migrate my-controller my-model`.

If there's ambiguity in the `my-model` name since there could be models with the same name but in different controllers, jimm should require a unique identifier like <controller-name>.<model-name>.

ale8k:
## Description

Enables users to specify model names when migrating. Fixes [#1371](https://github.com/canonical/jimm/issues/1371)

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests


cmd/jimmctl/cmd/migratemodel.go
```diff
@@ -20,15 +17,24 @@ import (
...
-a model-uuid to migrate and the destination controller name.
+The migrate commands migrates a model, or many models between two controllers
+registered within JIMM. 
+
+You may specify a model name (of the form owner/name) or model UUID.
+
+For example, JIMM is aware of controller-a and controller-b and controller-a has
+models:
+	- alice@canonical.com/model-a
+	- 2cb433a6-04eb-4ec4-9567-90426d20a004
```
kian99: This is confusing, are we saying there is one model or two? And why if there's two why is the one being shown with a name and the other with a UUID?

cmd/jimmctl/cmd/migratemodel.go
```diff
@@ -20,15 +17,24 @@ import (
...
+
+You may run the migrate command like so to migrate the two models
+from controller-a to controller-b:
+	jimmctl migrate controller-b model-a 2cb433a6-04eb-4ec4-9567-90426d20a004
```
kian99: Examples should be left out of the explanation doc and just left for the examples

cmd/jimmctl/cmd/migratemodel.go
```diff
@@ -20,15 +17,24 @@ import (
...
 
-Note that multiple models can be targeted for migration by supplying
-multiple model uuids.
 `
 	migrateModelCommandExample = `
     jimmctl migrate mycontroller 2cb433a6-04eb-4ec4-9567-90426d20a004 
-    jimmctl migrate mycontroller 2cb433a6-04eb-4ec4-9567-90426d20a004 fd469983-27c2-423b-bebf-84f616fb036b ...
+    jimmctl migrate mycontroller model-a model-b fd469983-27c2-423b-bebf-84f616fb036b ...
```
kian99: Let's give separate examples like
```
jimmctl migrate mycontroller model-a model-b
jimmctl migrate mycontroller fd469983-27c2-423b-bebf-84f616fb036b ...
```

kian99: Also, doesn't the example need to be `model-owner@domain/model-name`? The model name alone isn't enough to uniquely identity the model.

pkg/api/params/params.go
```diff
@@ -471,8 +471,8 @@ type PurgeLogsResponse struct {
 // target controller must be specified with both the source model and
 // target controller residing within JIMM.
 type MigrateModelInfo struct {
-	// ModelTag is a tag of the form "model-<UIID>"
-	ModelTag string `json:"model-tag"`
+	// TargetModelNameOrUUID is a tag of the form "model-<UUID>/model-<owner/name>"
```
kian99: It's no longer a tag right? So there is no "model-" prefix.

cmd/jimmctl/cmd/migratemodel.go
ale8k: It's because we want to allow names AND uuids.

cmd/jimmctl/cmd/migratemodel.go
ale8k: I like this example and want to keep it, but putting it in examples might look a bit strange as it will display like:

```
Examples:

You may specify a model name (of the form owner/name) or model UUID.

For example, JIMM is aware of controller-a and controller-b and controller-a has
models:
	- alice@canonical.com/model-a
	- 2cb433a6-04eb-4ec4-9567-90426d20a004

You may run the migrate command like so to migrate the two models
from controller-a to controller-b:
	jimmctl migrate controller-b model-a 2cb433a6-04eb-4ec4-9567-90426d20a004
```

cmd/jimmctl/cmd/migratemodel.go
ale8k: Yes, it does lol

ale8k: Done.

pkg/api/params/params.go
ale8k: Yup, updated

cmd/jimmctl/cmd/migratemodel.go
kian99: I suggest keeping the explanation in the doc and then just move the example.
```
You may specify a model name (of the form owner/name) and/or model UUID.

Examples:
	jimmctl migrate myController myModel myOtherModel
        jimmctl migrate myController 2cb433a6-04eb-4ec4-9567-90426d20a004 34b433a6-04eb-4ec4-9567-90426d20a005
```

cmd/jimmctl/cmd/migratemodel.go
kian99: Yeah but we don't need to show them with mixed usage like this. We can just say you can provide one and/or the other. This is very confusing to read.

cmd/jimmctl/cmd/migratemodel.go
ale8k: Done

cmd/jimmctl/cmd/migratemodel.go
ale8k: Done, removed and left example as suggested lower down

kian99 (approved):
lgtm

internal/jimm/jimm.go
```diff
@@ -730,25 +731,45 @@ func fillMigrationTarget(db *db.Database, credStore credentials.CredentialStore,
...
-		UUID: sql.NullString{
-			String: modelTag.Id(),
+
+	model := dbmodel.Model{}
+	// Check if the user is providing a model UUID or name
+	_, err = uuid.Parse(modelNameOrUUID)
+	if err != nil {
+		s := strings.Split(modelNameOrUUID, "/")
+		if len(s) != 2 {
+			return jujuparams.InitiateMigrationResult{}, errors.E(op, "missing model name")
```
SimoneDutto: this is more "invalid target" right? because we don't know if the model or the user name is wrong

cmd/jimmctl/cmd/migratemodel.go
```diff
@@ -20,15 +17,17 @@ import (
 
 const (
 	migrateModelCommandDoc = `
-The migrate command migrates a model(s) to a new controller. Specify
-a model-uuid to migrate and the destination controller name.
+The migrate commands migrates a model, or many models between two controllers
+registered within JIMM. 
+
+You may specify a model name (of the form owner/name) or model UUID.
```
SimoneDutto, line 23: learning question: do we always use this format `owner/name`? Couldn't we use model name where there is no ambiguity? Otherwise it can be annoying to find the owner of a model every time.

SimoneDutto (approved):
just a single comment

cmd/jimmctl/cmd/migratemodel.go
ale8k: It's probably better to always prefix with name to guarantee it is distinct. We'd have to have a query that ensures only 1 of that model name exists otherwise to do this, so I think specifying owner is easier. Plus, people may very well often create models of the same name part of our model name... Unless we make model names completely distinct in and of themselves I don't think this is worth it.

internal/jimm/jimm.go
ale8k: Yeah true, I'll update it

cmd/jimmctl/cmd/migratemodel.go
kian99: Just to chime in, it's a bug if you leave out the `owner` because Juju doesn't have unique model names, only `owner/model-name` is unique.

PR #1501, opened 2024-12-17
chore: move jwt generator

kian99:
## Description
This PR is mostly a mechanical change to move the jwtGenerator into a separate package. It was already well structured enough to do this. I've also added a package doc and renamed the NewJWTGenerator function to simply `New`.

The package has been renamed to `jujuauth`.

Fixes [JUJU-7272](https://warthogs.atlassian.net/browse/JUJU-7272)

[JUJU-7272]: https://warthogs.atlassian.net/browse/JUJU-7272?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/jimm/jwtgenerator/jwtgenerator.go
```diff
@@ -0,0 +1,180 @@
+// Copyright 2024 Canonical.
+
+// jwtgenerator generates JWT tokens to authenticate
```
alesstimec: package godoc usually starts with `Package <name>`

internal/jimm/jwtgenerator/jwtgenerator.go
```diff
@@ -0,0 +1,180 @@
...
+
+	"github.com/canonical/jimm/v3/internal/dbmodel"
+	"github.com/canonical/jimm/v3/internal/errors"
+	"github.com/canonical/jimm/v3/internal/jimmjwx"
+	"github.com/canonical/jimm/v3/internal/openfga"
+)
+
+// jwtGeneratorDatabase specifies the database interface used by the
+// JWT generator.
+type jwtGeneratorDatabase interface {
```
alesstimec: i think these three interfaces should be exported as they are parameter types for the exported constructor

alesstimec: also interface names do not need the `jwt` prefix now

internal/jimm/jwtgenerator/jwtgenerator_test.go
```diff
@@ -0,0 +1,357 @@
+// Copyright 2024 Canonical.
+package jwtgenerator_test
```
alesstimec: new line, please

alesstimec (commented):
i wonder if we could make this a bit less repetitive and have a `jwt` package with a `NewGenerator` constructor that would return `jwt.Generator`

internal/jimm/jwtgenerator/jwtgenerator.go
kian99: Will remove the jwt prefix, but why do they need to be exported? Maybe for a caller to verify that a struct satisfies the interface with a `_ = jwtGeneratorDatabase` but even that's a stretch because the caller can just call `New()` with whatever struct it has to satisfy the interface.

kian99:
>i wonder if we could make this a bit less repetitive and have a jwt package with a NewGenerator constructor that would return jwt.Generator

@alesstimec What aspect is repetitive?

kian99:
@alesstimec 
I was also thinking we could rename this package to something like `jujuTokens` to highlight that this is an business logic thing that relies on the JWT service primitive.

alesstimec:
@kian99 `jwtgenerator.Generator` sounds a bit repetitive, does it not?

internal/jimm/jwtgenerator/jwtgenerator.go
alesstimec: but the thing is that you don't know what interface you need to satisfy without looking at the code since it is unexported. generally having exported methods with unexported argument types or returns is an anti-pattern

kian99: >but the thing is that you don't know what interface you need to satisfy without looking at the code since it is unexported

But even if it was exported, how would you know how to satisfy the interface without looking at the code?

internal/jimm/jujuauth/jwtgenerator.go
```diff
@@ -0,0 +1,180 @@
+// Copyright 2024 Canonical.
+
+// Package jwtgenerator generates JWT tokens to
```
alesstimec: s/jwtgenerator/jujuauth/

internal/jimm/jujuauth/jwtgenerator.go
```diff
@@ -0,0 +1,180 @@
...
+	mu             sync.Mutex
+	accessMapCache map[string]string
+	mt             names.ModelTag
+	ct             names.ControllerTag
+	user           *openfga.User
+	callCount      int
+}
+
+// New returns a new JWTGenerator.
+func New(database generatorDatabase, accessChecker generatorAccessChecker, jwtService jwtService) TokenGenerator {
```
alesstimec: i still think argument types of exported methods need to be exported

internal/jimm/jujuauth/jwtgenerator.go
```diff
@@ -0,0 +1,180 @@
...
+
+	"github.com/canonical/jimm/v3/internal/dbmodel"
+	"github.com/canonical/jimm/v3/internal/errors"
+	"github.com/canonical/jimm/v3/internal/jimmjwx"
+	"github.com/canonical/jimm/v3/internal/openfga"
+)
+
+// generatorDatabase specifies the database interface used by the
+// JWT generator.
+type generatorDatabase interface {
```
SimoneDutto: do we want to go in this direction of having <package_name>Database? since it is unexported wouldn't be nice just to have database?

internal/jimm/jujuauth/jwtgenerator.go
kian99: I'm trying to think of a reason why and I think I see a good concrete reason. An internal type can be freely changed without impacting the users of a package. In this this case if we were to change the definition of one of these interfaces we would break users of the package so it should be exported to indicate that this is used by external users.

SimoneDutto (approved):
lgtm

internal/jimm/jujuauth/jwtgenerator.go
kian99: I've changed it back to being exported. It could still be database but I don't mind much.

alesstimec (approved):

PR #1502, opened 2024-12-17
chore: remove unneeded getters

kian99:
## Description
Remove getters that allowed accessing the database or authorization client from outside the application layer.
Methods like `DB()` or `AuthorizationClient()` were previously necessary when we had code in `jujuapi` manipulating the database or adding/removing access tuples. Now that that has all been moved to the app layer inside `jimm` we can get rid of these getters.


SimoneDutto (approved):
lgtm

alesstimec (approved):

PR #1503, opened 2024-12-17
chore: remove JWK service from jimm

kian99:
## Description
The JWK service that creates and rotates a JSON web key set doesn't and arguably shouldn't exist on the jimm struct as it goes unused. The service is quite separate from jimm.

Creating this as a draft to see other's opinion on this.


SimoneDutto (approved):
i agree

alesstimec (approved):
LGTM

PR #1504, opened 2024-12-19
feat: use golang migrate

kian99:
## Description
This PR removes our home-grown migration code in favour of [golang-migrate](https://github.com/golang-migrate/migrate/).

Some reservations I had with Golang-migrate over Goose and why I'm okay with it now:
- Migrations, by default, are not run in a transaction.
  - Thanks to Postgres and multi-statements, each migration file is run in a transaction. If we want to move to a different SQL DB or be more explicit we can add BEGIN/COMMIT statements to our transactions.
- Large set of dependencies in the golang-migrate library. 
  - After importing the library and the Postgres driver, we only have 4 new indirect dependencies so it doesn't seem too bad.
- When a migration fails, the DB is marked as "dirty" and requires manual intervention. 
  - Thanks to Postgres/transactions, we don't end up in an inconsistent state and when we detect this we can handle it in code. I've also added a test for this.

I recommend reviewing this PR commit-by-commit to more easily see the changes.

The benefits this gives us aren't huge. One benefit is that we don't need to modify the versions.go file when we add a new migration file, it will just automatically be applied, leaving less room for error.

Fixes [JUJU-7047](https://warthogs.atlassian.net/browse/JUJU-7047)

## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests

## Test instructions
To test the new migrations on a fresh DB, just `docker compose --profile dev up`.

To test migrations on a DB that was using the old-style migrations. Switch to the `v3` branch then ``docker compose --profile dev up` then switch to this branch and observe the jimm logs.

[JUJU-7047]: https://warthogs.atlassian.net/browse/JUJU-7047?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


docker-compose.yaml
```diff
@@ -101,6 +101,7 @@ services:
       OPENFGA_AUTHN_PRESHARED_KEYS: "jimm"
       OPENFGA_DATASTORE_ENGINE: "postgres"
       OPENFGA_DATASTORE_URI: "postgresql://jimm:jimm@db/jimm?sslmode=disable"
+      OPENFGA_LOG_LEVEL: "warn"
```
SimoneDutto, line 104: why this?

internal/db/db.go
```diff
@@ -49,57 +61,116 @@ func (d *Database) Transaction(f func(*Database) error) error {
...
+	if line[len(line)-1] == '\n' {
+		line = line[:len(line)-1]
+	}
+	l.logger.Info(line)
+}
+
+func (l migrationLogger) Verbose() bool {
+	return l.verbose
+}
+
```
SimoneDutto: can you move this to `internal/logger`, so we have them all in a single place

internal/db/db.go
```diff
@@ -49,57 +61,116 @@ func (d *Database) Transaction(f func(*Database) error) error {
...
+			return errors.E(op, fmt.Errorf("failed to get db version: %w", err))
 		}
-		// The major versions are unchanged, the database can be migrated.
-		v.Minor += 1
-		schema, err := dbmodel.SQL.ReadFile(path.Join("sql", db.Name(), fmt.Sprintf("%d_%d.sql", v.Major, v.Minor)))
-		if err != nil {
-			return errors.E(op, err)
+	}
+
+	if dirty {
```
SimoneDutto, line 125: just to make sure I understand. If the migration we've started before fails, it sets the dirty flag in the db, and the next time we call migrate it forces the version to the one before it and reruns the migration.
Right?

internal/db/db.go
```diff
@@ -49,57 +61,116 @@ func (d *Database) Transaction(f func(*Database) error) error {
...
-		})
-		if err != nil {
-			return errors.E(op, dbError(err))
+	if err := m.Up(); err != nil {
+		if !stderr.Is(err, migrate.ErrNoChange) {
+			return errors.E(op, fmt.Errorf("failed to migrate db: %w", err))
 		}
 	}
+
+	atomic.StoreUint32(&d.migrated, 1)
```
SimoneDutto, line 140: why don't we also check this flag at the start of the `Migrate` method?

internal/dbmodel/sql/postgres/005_add_migrating_controller_column.up.sql
```diff
@@ -0,0 +1,3 @@
+-- 1_5.sql is a migration that adds a column necessary for model migrations
```
SimoneDutto: i think you should also edit this comments from 1_5 to 005, or just remove it

SimoneDutto (commented):
nice work, left some comments

docker-compose.yaml
kian99: It reduces the noise after running `docker compose up`, hiding the health check pings that happen every few seconds.

internal/db/db.go
kian99: Correct!

internal/db/db.go
kian99: We could, but we didn't before so I haven't changed that logic.

internal/dbmodel/sql/postgres/005_add_migrating_controller_column.up.sql
kian99: Good point, will do

internal/db/db.go
SimoneDutto: i guess it's uncommon we call the `Migrate` twice in our tests, because we usually create a new db for each test.

SimoneDutto (approved):
lgtm

internal/db/db.go
```diff
@@ -6,15 +6,29 @@ package db
...
 
 	"github.com/canonical/jimm/v3/internal/dbmodel"
 	"github.com/canonical/jimm/v3/internal/errors"
+	"github.com/canonical/jimm/v3/internal/logger"
+)
+
+// Use a custom table name so that we don't run into collisions when OpenFGA or other tools
+// are using the same DB as JIMM in our Docker Compose setup.
+const (
+	MigrationTableName = "jimm_schema_migrations"
```
alesstimec: does this need to be exported?

alesstimec (approved):

internal/db/db.go
kian99: Good catch, it doesn't. Fixed.

PR #1506, opened 2024-12-20
feat: add login manager

kian99:
## Description
This PR introduces a login manager. This PR builds on #1505.

Fixes [JUJU-7253](https://warthogs.atlassian.net/browse/JUJU-7253)

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests
- [x] Covered by integration tests

[JUJU-7253]: https://warthogs.atlassian.net/browse/JUJU-7253?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1507, opened 2025-01-06
feat: add permission manager

kian99:
## Description
This PR builds on #1506.

This PR introduces a permission manager package dedicated to granting/revoking access and checking a user's permissions.
Some notable changes:
- The `jujuauth` now uses a factory pattern since this is a stateful object that needs to be recreated for each websocket connection. After introducing the permission manager package this simplifies how much access the API layer needs to the underlying business layer structs.
-  Refactored some application offer grant/revoke tests. They were not using the standard `yaml` -> DB approach for creating test objects.

The rest of the changes are largely mechanical but unfortunately very long because we had access related logic in many different places.

The changes are easier to review commit-by-commit (or at least the first commit on its own then the rest separately).

Fixes [JUJU-7252](https://warthogs.atlassian.net/browse/JUJU-7252)

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

[JUJU-7252]: https://warthogs.atlassian.net/browse/JUJU-7252?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/jimm/jujuauth/factory.go
```diff
@@ -0,0 +1,25 @@
+// Copyright 2025 Canonical.
+package jujuauth
```
alesstimec, line 3: new line, please

internal/jimm/permissions/permissionmanager.go
```diff
@@ -0,0 +1,33 @@
+// Copyright 2024 Canonical.
```
alesstimec: 2025

internal/jimm/permissions/export_test.go
```diff
@@ -0,0 +1,21 @@
+// Copyright 2024 Canonical.
```
alesstimec: 2025

internal/jimm/permissions/suite_test.go
```diff
@@ -0,0 +1,71 @@
+// Copyright 2024 Canonical.
```
alesstimec: 2025

internal/jimm/permissions/access.go
```diff
@@ -0,0 +1,799 @@
+// Copyright 2024 Canonical.
```
alesstimec: 2025

internal/jimm/permissions/access_test.go
```diff
@@ -0,0 +1,2383 @@
+// Copyright 2024 Canonical.
```
alesstimec: 2025

internal/jimm/permissions/access.go
```diff
@@ -0,0 +1,799 @@
...
+		return ofganames.ConsumerRelation, nil
+	case string(jujuparams.OfferReadAccess):
+		return ofganames.ReaderRelation, nil
+	default:
+		return ofganames.NoRelation, errors.E("unknown application offer access")
+	}
+}
+
+// GetUserControllerAccess returns the user's level of access to the desired controller.
+func (j *permissionManager) GetUserControllerAccess(ctx context.Context, user *openfga.User, controller names.ControllerTag) (string, error) {
```
alesstimec, line 119: suggestion: keep all the permissionManager method implementations in permissionmanager.go

internal/jimm/permissions/permissionmanager.go
```diff
@@ -0,0 +1,33 @@
...
+	store   *db.Database
+	authSvc *openfga.OFGAClient
+	// JIMM's UUID and tag
+	uuid string
+	tag  names.ControllerTag
+}
+
+// NewPermissionManager returns a new permission manager that provides
+// permission handling and resolution of JAAS tags.
+func NewPermissionManager(store *db.Database, authSvc *openfga.OFGAClient, uuid string, tag names.ControllerTag) (*permissionManager, error) {
```
alesstimec: suggestion: rename to NewManager.. that way whoever calls this will do `permissions.NewManager` instead of `permissions.NewPermissionManager` (the word permission is repeated)

alesstimec (commented):
a few comments

internal/jimm/permissions/access.go
kian99: Do you mean just move all the methods in `access.go` to `permissionmanager.go`? What about `relations.go` which contains some more methods but these ones are generic OpenFGA ones.

internal/jimm/permissions/permissionmanager.go
kian99: Done, the other managers need the same change (will do later).

internal/jimm/permissions/relations.go
```diff
@@ -76,7 +76,7 @@ func (j *JIMM) CheckRelation(ctx context.Context, user *openfga.User, tuple apip
 
 // ListRelationshipTuples checks user permission and lists relationship tuples based of tuple struct with pagination.
 // Listing filters can be relaxed: optionally exclude tuple.Relation or tuple.Object or specify only tuple.TargetObject.Kind.
-func (j *JIMM) ListRelationshipTuples(ctx context.Context, user *openfga.User, tuple apiparams.RelationshipTuple, pageSize int32, continuationToken string) ([]openfga.Tuple, string, error) {
+func (j *permissionManager) ListRelationshipTuples(ctx context.Context, user *openfga.User, tuple apiparams.RelationshipTuple, pageSize int32, continuationToken string) ([]openfga.Tuple, string, error) {
```
alesstimec, line 79: remembering our discussion the other day about exposing dbmodel types in manager method, this is doing something similar, but exposing the underlying `openfga.Tuple` type. Eventually we might want to consider decoupling..

internal/jimm/permissions/access.go
alesstimec: yes, i think it would be much cleaner to have all methods of a type in a single file..

alesstimec (approved):
LGTM.. but ptal my comment about exposing types from lower levels (openfga. dbmodel)

internal/jimm/permissions/access.go
kian99: Let's re-organise in a separate PR then, there's a couple of ways of chopping it up.

internal/jimm/permissions/relations.go
kian99: That seems fair.

internal/jimm/jujuauth/factory.go
```diff
@@ -0,0 +1,26 @@
...
+	return &Factory{
+		db:            db,
+		jwtService:    jwtService,
+		accessChecker: accessChecker,
+	}
+}
+
+// New returns a new Juju token generator.
+func (f *Factory) New() TokenGenerator {
+	return new(f.db, f.accessChecker, f.jwtService)
```
SimoneDutto: i don't like the use of  `new`. From the doc is 
```
// The new built-in function allocates memory. The first argument is a type,
// not a value, and the value returned is a pointer to a newly
// allocated zero value of that type.
````
Here we are not returning neither a zero-valued struct or a pointer.
Can we just make it a method like `newJWTGenerator`? What do you think?

File: internal/jimm/permissions/permissionmanager.go [hunk 11]
```diff
@@ -0,0 +1,33 @@
...
+	"github.com/canonical/jimm/v3/internal/errors"
+	"github.com/canonical/jimm/v3/internal/openfga"
+)
+
+// permissionManager provides a means to manage roles within JIMM.
+type permissionManager struct {
+	store   *db.Database
+	authSvc *openfga.OFGAClient
+	// JIMM's UUID and tag
+	uuid string
```
SimoneDutto: what do you think about prepending JIMM to these fields?

internal/jimm/permissions/permissionmanager.go
```diff
@@ -0,0 +1,33 @@
...
+	tag  names.ControllerTag
+}
+
+// NewManager returns a new permission manager that provides
+// permission handling and resolution of JAAS tags.
+func NewManager(store *db.Database, authSvc *openfga.OFGAClient, uuid string, tag names.ControllerTag) (*permissionManager, error) {
```
SimoneDutto, line 24: i see we are going with `store` instead of `db`. Just to keep a consistency with non-manager jimm's methods i'd like to keep db.
Also because store kind of give interface vibes and we are using the concrete type. What do you think?

SimoneDutto: the same goes for openfga client. If we are using the concrete type just call it openfgaClient no?

SimoneDutto (commented):
lgtm, with some questions

internal/jimm/jujuauth/factory.go
kian99: Yeah I'm +1 for that, I'll change it to `newTokenGenerator`

internal/jimm/permissions/permissionmanager.go
kian99: Yeah I think I did that elsewhere, I'll make them `jimmUUID`, etc.

internal/jimm/permissions/permissionmanager.go
kian99: The rest of the manager already have it like this so I'd like to leave this one as is, then we can change it later if we don't think it's fit.

SimoneDutto (approved):

PR #1508, opened 2025-01-07
chore: create auditlog package

kian99:
## Description
~~This PR creates a dedicated `auditlog` package where audit logic lives. Originally this code was inside `internal/jimm` and I initially tried to create a "manager" object for audit logs when I realised that there is no "business logic" related methods for audit logging on the JIMM struct. Instead, audit logging was always initialised separately from JIMM, which makes sense when considering that these logs are a separate service from our business logic. So moving them to a new package entirely in `internal/auditlog` made the most sense to me and introduces very few changes.~~

This PR now creates an `auditLogManager` inside the `jimm` package that contains business logic for persisting/fetching/purging audit logs. The logic around recording audit logs has been moved to a separate file in recorder.go in `jujuapi` that is only concerned with recording RPC messages and then passing them elsewhere for persistence.

The test for cleaning up audit logs periodically was broken and wasn't actually doing anything, so I've also refactored things a bit to make that easier to test.

Fixes [JUJU-7274](https://warthogs.atlassian.net/browse/JUJU-7274)

## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests

[JUJU-7274]: https://warthogs.atlassian.net/browse/JUJU-7274?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/auditlog/auditlog.go
```diff
@@ -19,29 +19,29 @@ import (
 	"github.com/canonical/jimm/v3/internal/utils"
 )
 
-// AuditLoggerBackend defines the interface used by the DbAuditLogger to store
+// LogBackend defines the interface used by the DbAuditLogger to store
```
SimoneDutto:
```suggestion
// LogBackend defines the interface used by the Logger to store
```

internal/jujuapi/controllerroot.go
```diff
@@ -135,8 +135,8 @@ func (r *controllerRoot) setupUUIDGenerator() error {
 	return nil
 }
 
-func (r *controllerRoot) newAuditLogger() jimm.DbAuditLogger {
-	return jimm.NewDbAuditLogger(r.jimm, r.getUser)
+func (r *controllerRoot) newAuditLogger() auditlog.Logger {
+	return auditlog.NewLogger(r.jimm, r.getUser)
```
SimoneDutto: wouldn't be better here to pass the db down to the auditlogger? I don't it needs the whole jimm, right?

SimoneDutto (commented):
lgtm, with a godoc fix and a question

internal/jujuapi/controllerroot.go
kian99: The method to do the logging is "business logic" (and does actually contain some business logic to remove sensitive fields) so it's on JIMM, and we can't/shouldn't extract the DB out of JIMM at this layer because that would leak the database to the API layer.

We can create a manager for doing logging related logic in the future. Then we can pass in just the manager, something like `r.jimm.logger` or whatever is appropriate.

internal/jimm/auditlog/auditlog.go
```diff
@@ -0,0 +1,110 @@
+// Copyright 2025 Canonical.
+
+// The auditlot package provides business logic for handling audit log related methods.
```
alesstimec: s/auditlot/auditlog/

internal/jimm/auditlog/cleanup.go
```diff
@@ -0,0 +1,69 @@
+// Copyright 2025 Canonical.
+package auditlog
```
alesstimec, line 3: new line, please

internal/jimm/auditlog/cleanup.go
```diff
@@ -0,0 +1,69 @@
...
+
+// pollTimeOfDay holds the time hour, minutes and seconds to poll for cleanup.
+type pollTimeOfDay struct {
+	Hours   int
+	Minutes int
+	Seconds int
+}
+
+func (j *auditLogManager) cleanup(ctx context.Context) {
+	retentionDate := time.Now().AddDate(0, 0, -(j.retentionPeriodInDays))
```
alesstimec, line 45: for testing purposes it would be nice if the auditLogManager had a "clock" filed (see https://pkg.go.dev/github.com/juju/clock@v1.1.1#Clock) so in tests you could pass in your own "test" clock, but in production it would just be the WallClock

internal/jimm/auditlog/cleanup.go
```diff
@@ -0,0 +1,69 @@
...
+// auditLogCleanupTime indicates that we poll at 9 AM.
+var auditLogCleanupTime = pollTimeOfDay{
+	Hours: 9,
+}
+
+// StartCleanup loop forever and checks daily for any logs
+// that need to be cleaned up. This method should be run
+// in a separate Go routine to avoid blocking, it will terminate
+// when the provided context is cancelled.
+func (j *auditLogManager) StartCleanup(ctx context.Context) {
```
alesstimec, line 22: also, again, not a fan of splitting methods of a type in multiple files

internal/jimm/jimm.go
```diff
@@ -220,6 +219,17 @@ type PermissionManager interface {
 	ToJAASTag(ctx context.Context, tag *ofganames.Tag, resolveUUIDs bool) (string, error)
 }
 
+type AuditLogManager interface {
```
alesstimec, line 223: godoc, please

internal/jujuapi/recorder.go
```diff
@@ -0,0 +1,126 @@
...
+	"github.com/canonical/jimm/v3/internal/utils"
+)
+
+// LogBackend defines the interface used by the Logger to store
+// audit events.
+type LogBackend interface {
+	AddAuditLogEntry(ale *dbmodel.AuditLogEntry)
+}
+
+type AuditLogger struct {
```
alesstimec: godoc, please

alesstimec (approved):
LGTM with a few suggestions.

internal/jimm/auditlog/cleanup.go
kian99: Yeah it's not ideal but I'm preferring to keep it like this to make things easier to find.

internal/jimm/auditlog/cleanup.go
kian99: Here is the blog post I was talking about that mention's Juju and in Go 1.24 there will also be a neat way of testing time sensitive code.
https://devcenter.upsun.com/posts/go-124/#a-new-testingsynctest-package-for-testing-concurrent-code

kian99: But for now I think this is sufficiently tested.

internal/jujuapi/recorder.go
kian99: Added.

internal/jimm/auditlog/cleanup.go
```diff
@@ -0,0 +1,70 @@
...
```
SimoneDutto, line 22: j -> auditLogManager or am

SimoneDutto (approved):
lgtm

internal/jimm/auditlog/cleanup.go
kian99: Good catch, I'll update it in a follow-up.

PR #1509, opened 2025-01-07
fix(internal/jujuclient/applicationoffers.go): set user when getting offer consume details

alesstimec:
When getting consume details for an application offer the juju controller will, among other things,
also return a macaroon  that is passed to the consuming side. That macaroon contains a
third party caveat addressed to JIMM in which it instructs JIMM to verify that the user
is consumer of the application offer. Previously we were passing in an empty user tag
which meant that Juju added that 3rd party caveat for the authenticated user and since
JIMM always dials the juju controller as user `admin` that caveat was instructing JIMM
to check that the `admin` user has `consume` access to the application offer.

This PR proposes one way to fix is, which is to specify the user when getting consume
details from the controller. The other way would be to dial the controller as the
logged-in user.

## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests


internal/jujuclient/applicationoffers.go
```diff
@@ -209,9 +209,7 @@ func (c Connection) GetApplicationOfferConsumeDetails(ctx context.Context, user
 			OfferURLs:     []string{info.Offer.OfferURL},
 			BakeryVersion: v,
 		},
-		// Do not include a user in the args, Juju will opt to use the user authenticated in the connection.
-		// There is a bug where setting the user tag does not behave as expected.
-		UserTag: "",
+		UserTag: user.String(),
```
kian99, line 212: Can you add a comment about why we do this.

kian99 (commented):
I changed this line in https://github.com/canonical/jimm/pull/1329, mentioning "Fix a bug in GetApplicationOfferConsumeDetails to exclude the user field which currently causes an unknown error.". I wish I'd been more descriptive, but it was causing an error in the Terraform Provider's integration tests. It would be good to run the providers tests against JIMM with this change and see what test fails (if any).

kian99 (approved):

internal/jujuclient/applicationoffers.go
alesstimec: done

SimoneDutto (approved):
lgtm

internal/jujuclient/applicationoffers.go
kian99: I don't see a comment in the code? Did you make a comment elsewhere like in the commit description?

PR #1510, opened 2025-01-07
feat: add service account manager

kian99:
## Description
~~This PR builds on #1507.~~

This PR introduces a service account manager. This manager handles operations related to service accounts and moves more logic out of the `jimm` package and into a separate `serviceaccount` package.

There is one difference between this manager versus those previously introduced. The service account manager method "CopyServiceAccountCredential" calls other jimm methods like `j.GetCloudCredential` which contain business logic related to cloud credentials. This is the first instance where a unit of business logic relies on another unit of business logic.

I addressed this by:
- Created an interface to limit what the service account manager can do with the JIMM struct.
- Moved the logic for copying cloud-credentials from proximity to service accounts and placed it closer to cloud-credential logic.

In the future when we have a cloud-credential manager, we can simply pass that manager into the service account manager.

Fixes [JUJU-7273](https://warthogs.atlassian.net/browse/JUJU-7273)
 
## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

[JUJU-7273]: https://warthogs.atlassian.net/browse/JUJU-7273?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/jimm/cloudcredential.go
```diff
@@ -350,3 +350,38 @@ func (j *JIMM) getCloudCredentialAttributes(ctx context.Context, cred *dbmodel.C
 	}
 	return attr, nil
 }
+
+// CopyCredential copies a cloud credential from one user to another.
+func (j *JIMM) CopyCredential(ctx context.Context, originalUser *openfga.User, newUser *openfga.User, cred names.CloudCredentialTag) (names.CloudCredentialTag, []jujuparams.UpdateCredentialModelResult, error) {
+	op := errors.Op("jimm.CopyCredential")
+
+	credential, err := j.GetCloudCredential(ctx, originalUser, cred)
```
alesstimec, line 358: where do we check authorization.. is the original user allowed to set cloud credentials for the new user -> is the original user administrator of the "new user" service account?

internal/jimm/cloudcredential_test.go
```diff
@@ -1740,3 +1740,110 @@ func (s testCloudCredentialAttributeStore) GetOAuthSecret(ctx context.Context) (
...
+	j := jimmtest.NewJIMM(c, &jimm.Parameters{
+		Dialer: &jimmtest.Dialer{
+			API: api,
+		},
+	})
+
+	svcAccId, err := dbmodel.NewIdentity("39caae91-b914-41ae-83f8-c7b86ca5ad5a@serviceaccount")
+	c.Assert(err, qt.IsNil)
+	c.Assert(j.Database.DB.Create(&svcAccId).Error, qt.IsNil)
+	svcAcc := openfga.NewUser(svcAccId, j.OpenFGAClient)
```
alesstimec, line 1767: nowhere in this test do we grant alice@canonical.com any sort of rights on the svcAcc, which worries me a bit. i think we should only allow svc account admins to update cloud credentials

internal/jimm/serviceaccount/serviceaccount.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/juju/names/v5"
+
+	"github.com/canonical/jimm/v3/internal/db"
+	"github.com/canonical/jimm/v3/internal/errors"
+	"github.com/canonical/jimm/v3/internal/openfga"
+	ofganames "github.com/canonical/jimm/v3/internal/openfga/names"
+	jimmnames "github.com/canonical/jimm/v3/pkg/names"
+)
+
+type CredentialCopier interface {
```
alesstimec, line 20: godoc, please

internal/jimm/serviceaccount/serviceaccount.go
```diff
@@ -0,0 +1,91 @@
...
+	if authSvc == nil {
+		return nil, errors.E("role authorisation service cannot be nil")
+	}
+	if cc == nil {
+		return nil, errors.E("credential copier service cannot be nil")
+	}
+	return &serviceAccountManager{store, authSvc, cc}, nil
+}
+
+// AddServiceAccount checks that no one owns the service account yet
```
alesstimec, line 47: ah, ok. this is a bit confusing and took me 2min to wrap my thoughts around it: when adding a service account, we check that nobody else has administrator relation to it - assert that there is only one administrator relation to it. but once the account is added, the administrator is allowed to add further administrators to the service account, right?

e.g. i create a service account and then allow a group of users to administer it (by adding their cloud credentials).

alesstimec (changes requested):
i think we are missing an authorization check when copying cloud credentials

internal/jimm/serviceaccount/serviceaccount.go
kian99: Correct, I'm actually not a fan of this approach, the alternative would be to require the user to provide the client ID+secret to gain access to manage a service account.

internal/jimm/cloudcredential.go
kian99: Perhaps it's a bit too inconsistent with some of the other business logic methods, but for service accounts the auth checks are all done at the API layer in `jujuapi` and then that decides whether to call the copyCredential method etc.

internal/jimm/cloudcredential_test.go
kian99: See my comment above.

internal/jimm/cloudcredential.go
alesstimec: perhaps we should move the authz check to make it consistent?

kian99: It's worth another discussion, but it's not related to the changes in this PR.

kian99: Made a card for it https://warthogs.atlassian.net/browse/JUJU-7389

internal/jimm/jimm.go
```diff
@@ -231,6 +232,13 @@ type AuditLogManager interface {
 	StartCleanup(ctx context.Context)
 }
 
+type ServiceAccountManager interface {
```
alesstimec, line 236: godoc, please

internal/jimm/serviceaccount/serviceaccount_test.go
```diff
@@ -0,0 +1,63 @@
+// Copyright 2025 Canonical.
+package serviceaccount_test
```
alesstimec, line 3: newline

internal/testutils/jimmtest/mocks/jimm_service_account_mock.go
```diff
@@ -0,0 +1,31 @@
+// Copyright 2025 Canonical.
+package mocks
```
alesstimec, line 3: newline, please

alesstimec (approved):
LGTM with three nitpicks

SimoneDutto (approved):
good

PR #1512, opened 2025-01-08
slim clouds

SimoneDutto:
## Description
In this pr we slim clouds, removing non-essential fields.

### DIFF
--- a/cmd/jimmctl/cmd/purge_logs_test.go
+++ b/cmd/jimmctl/cmd/purge_logs_test.go
@@ -1,4 +1,4 @@
-// Copyright 2024 Canonical.
+// Copyright 2025 Canonical.
 package cmd_test
 
 import (
--- a/internal/db/resource_test.go
+++ b/internal/db/resource_test.go
@@ -1,4 +1,4 @@
-// Copyright 2024 Canonical.
+// Copyright 2025 Canonical.
 package db_test
 
 import (
--- a/internal/dbmodel/cloud.go
+++ b/internal/dbmodel/cloud.go
@@ -1,4 +1,4 @@
-// Copyright 2024 Canonical.
+// Copyright 2025 Canonical.
 
 package dbmodel
 
@@ -26,29 +26,8 @@ type Cloud struct {
 	// cloud is hosted.
 	HostCloudRegion string
 
-	// AuthTypes is the authentication types supported by this cloud.
-	AuthTypes Strings
-
-	// Endpoint is the API endpoint URL for the cloud.
-	Endpoint string
-
-	// IdentityEndpoint is the API endpoint URL of the cloud identity
-	// service.
-	IdentityEndpoint string
-
-	// StorageEndpoint is the API endpoint URL of the cloud storage
-	// service.
-	StorageEndpoint string
-
 	// Regions contains the regions associated with this cloud.
 	Regions []CloudRegion `gorm:"foreignKey:CloudName;references:Name"`
-
-	// CACertificates contains the CA Certificates associated with this
-	// cloud.
-	CACertificates Strings
-
-	// Config contains the configuration associated with this cloud.
-	Config Map
 }
 
 // Tag returns a names.Tag for this cloud.
@@ -80,40 +59,11 @@ func (c Cloud) Region(name string) CloudRegion {
 	return CloudRegion{}
 }
 
-// ToJujuCloud converts the  Cloud object into a jujuparams.Cloud. The
-// cloud must have its regions association filled out.
-func (c Cloud) ToJujuCloud() jujuparams.Cloud {
-	var cl jujuparams.Cloud
-	cl.Type = c.Type
-	cl.HostCloudRegion = c.HostCloudRegion
-	cl.AuthTypes = []string(c.AuthTypes)
-	cl.Endpoint = c.Endpoint
-	cl.IdentityEndpoint = c.IdentityEndpoint
-	cl.StorageEndpoint = c.StorageEndpoint
-	cl.Regions = make([]jujuparams.CloudRegion, len(c.Regions))
-	cl.RegionConfig = make(map[string]map[string]interface{}, len(c.Regions))
-	for i, r := range c.Regions {
-		cl.Regions[i] = r.ToJujuCloudRegion()
-		if r.Config != nil {
-			cl.RegionConfig[r.Name] = map[string]interface{}(r.Config)
-		}
-	}
-	cl.CACertificates = []string(c.CACertificates)
-	cl.Config = map[string]interface{}(c.Config)
-	return cl
-}
-
 // FromJujuCloud updates a Cloud object with the details from the given
 // jujuparams.Cloud.
 func (c *Cloud) FromJujuCloud(cld jujuparams.Cloud) {
 	c.Type = cld.Type
 	c.HostCloudRegion = cld.HostCloudRegion
-	c.AuthTypes = Strings(cld.AuthTypes)
-	c.Endpoint = cld.Endpoint
-	c.IdentityEndpoint = cld.IdentityEndpoint
-	c.StorageEndpoint = cld.StorageEndpoint
-	c.CACertificates = Strings(cld.CACertificates)
-	c.Config = Map(cld.Config)
 	regions := make([]CloudRegion, 0, len(c.Regions))
 	for _, r := range cld.Regions {
 		reg := c.Region(r.Name)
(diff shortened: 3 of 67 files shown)


internal/jimm/cloud.go
```diff
@@ -20,35 +20,173 @@ import (
...
+		return api.Cloud(ctx, cl.ResourceTag(), &jCl)
+	})
+	if err != nil {
+		return jCl, errors.E(op, err)
+	}
+	jCl.IsControllerCloud = false // jimm doesn't know where is deployed, so none of the clouds should have this field set.
+	return jCl, nil
+}
+
+func (j *JIMM) GetClouds(ctx context.Context, user *openfga.User) (map[string]jujuparams.Cloud, error) {
```
alesstimec, line 76: godoc, please

internal/jimm/cloud.go
```diff
@@ -20,35 +20,173 @@ import (
...
-		return cl, nil
+	return jujuparams.CloudInfo{
+		CloudDetails: gCl.CloudDetails,
+		// TODO(Kian) CSS-6040 Determine whether to combine OpenFGA Tuples
+		// with Postgres data objects for a consolidated view.
+		Users: nil,
+	}, nil
+}
+
+func (j *JIMM) ListCloudsInfo(ctx context.Context, user *openfga.User, all bool) ([]jujuparams.ListCloudInfoResult, error) {
```
alesstimec, line 154: godoc, please

internal/jimmhttp/auth_handler_test.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2024 Canonical.
+// Copyright 2025 Canonical.
```
alesstimec, line 1: new line

internal/jimmhttp/rebac_admin/identities_integration_test.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2024 Canonical.
+// Copyright 2025 Canonical.
```
alesstimec, line 1: new line

cmd/jimmctl/cmd/purge_logs_test.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2024 Canonical.
+// Copyright 2025 Canonical.
```
alesstimec, line 1: new line, please

internal/jimm/cloud.go
```diff
@@ -20,35 +20,173 @@ import (
...
+	zapctx.Info(ctx, string(op))
 
 	var cl dbmodel.Cloud
 	cl.SetTag(tag)
 
 	if err := j.Database.GetCloud(ctx, &cl); err != nil {
-		return cl, errors.E(op, err)
+		return jujuparams.CloudInfo{}, errors.E(op, err)
+	}
+	// TODO (SimoneDutto): refactor this to use `user.IsCloudAdmin()`
```
alesstimec, line 117: where do we require admin access to the cloud?

alesstimec (approved):
The approach looks good to me.. a few comments..

internal/db/resource_test.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2024 Canonical.
+// Copyright 2025 Canonical.
 package db_test
```
kian99, line 2: newline

internal/dbmodel/cloud.go
```diff
@@ -26,29 +26,8 @@ type Cloud struct {
 	// cloud is hosted.
 	HostCloudRegion string
```
kian99, line 27: Why do we need to keep this?

internal/dbmodel/sql/postgres/019_simplify_cloud_table.up.sql
```diff
@@ -0,0 +1,3 @@
+-- remove non essential fields from cloud.
+ALTER TABLE clouds  DROP COLUMN auth_types, DROP COLUMN endpoint, DROP COLUMN identity_endpoint,
```
kian99, line 2: nit: add a space after the comment

internal/jimm/cloud.go
```diff
@@ -20,35 +20,173 @@ import (
 	ofganames "github.com/canonical/jimm/v3/internal/openfga/names"
 )
 
+// GetUserCloudAccess returns users access level for the specified cloud.
+func (j *JIMM) GetUserCloudAccess(ctx context.Context, user *openfga.User, cloud names.CloudTag) (string, error) {
```
kian99, line 24: I think there is already a method in the permission manager that does this. Can you use that instead?

internal/jimm/cloud.go
```diff
@@ -20,35 +20,173 @@ import (
...
+func (j *JIMM) GetCloud(ctx context.Context, user *openfga.User, tag names.CloudTag) (jujuparams.Cloud, error) {
+	const op = errors.Op("jimm.CloudInfo")
+	zapctx.Info(ctx, string(op))
+	var cl dbmodel.Cloud
+	cl.SetTag(tag)
+
+	if err := j.Database.GetCloud(ctx, &cl); err != nil {
+		return jujuparams.Cloud{}, errors.E(op, err)
+	}
+	// TODO (SimoneDutto): refactor this to use `user.IsCloudAdmin()`
```
kian99, line 44: Can we do that now?

internal/jimm/cloud.go
```diff
@@ -20,35 +20,173 @@ import (
...
+
+// GetCloud retrieves the cloud for the given cloud tag. If the cloud
+// cannot be found then an error with the code CodeNotFound is
+// returned. If the user does not have permission to view the cloud then an
+// error with a code of CodeUnauthorized is returned. If the user only has
+// add-model access to the cloud then the returned Users field will only
+// contain the authentcated user.
+func (j *JIMM) GetCloud(ctx context.Context, user *openfga.User, tag names.CloudTag) (jujuparams.Cloud, error) {
+	const op = errors.Op("jimm.CloudInfo")
+	zapctx.Info(ctx, string(op))
```
kian99, line 37: Was thinking about this sometime last year and whether we want it. If we do want it then it might be better suited to go inside the `jujuapi.FindMethod()` class of functions so that we don't clutter up these functions.

internal/jimm/cloud.go
```diff
@@ -20,35 +20,173 @@ import (
...
+			controllers = append(controllers, c.Controller)
+		}
+	}
+	err := j.firstSuccessfulController(ctx, controllers, func(api API) error {
+		return api.Cloud(ctx, cl.ResourceTag(), &jCl)
+	})
+	if err != nil {
+		return jCl, errors.E(op, err)
+	}
+	jCl.IsControllerCloud = false // jimm doesn't know where is deployed, so none of the clouds should have this field set.
```
kian99, line 72:
```suggestion
	jCl.IsControllerCloud = false // jimm doesn't know where it's deployed, so none of the clouds should have this field set.
```

internal/jimm/cloud.go
```diff
@@ -20,35 +20,173 @@ import (
...
+		return nil, err
+	}
+	results := make(map[string]jujuparams.Cloud, 0)
+	for _, cl := range clouds {
+		cloud, err := j.getCloudFromController(ctx, cl)
+		if err != nil {
+			return nil, errors.E(op, err)
+		}
+		results[cl.ResourceTag().String()] = cloud
+	}
```
kian99, line 96: I have a concern that this will be annoyingly slow. Not so slow that it's bad but since getting a cloud requires a sequential call to a controller... I've experienced running `juju <command>` to take several seconds sometimes so I hope we don't have that issue here.
Some caching in the future will of course help and then be similar in a way to what the database was doing.

internal/jimm/cloud.go
```diff
@@ -20,35 +20,173 @@ import (
...
+	const op = errors.Op("jimm.getCloudInfoFromController")
+	var gCl jujuparams.CloudInfo
+
+	controllers := make([]dbmodel.Controller, 0)
+	for _, cr := range cl.Regions {
+		for _, c := range cr.Controllers {
+			controllers = append(controllers, c.Controller)
+		}
+	}
+	err := j.firstSuccessfulController(ctx, controllers, func(api API) error {
```
kian99, line 139: Here we will always be calling the first controller in the list most frequently. We could randomise the list to do some round-robin if we want.

local/seed_db/main.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2024 Canonical.
+// Copyright 2025 Canonical.
```
kian99, line 1: newline

kian99 (commented):
I like how it's slimmed things but I'm a bit hesitant on performance, even though it's not a huge deal. I'm wondering if we want/need to slim down clouds. The gains seem marginal but curious to hear what you think after doing this.

kian99:
Points we discussed on this:
- Cloud fields are static and we don't get much gain here versus models where we were able to get rid of the watcher.
- This might be annoyingly slow if latency to controllers is high and we need to list all clouds.
- Returning the cloud info from the first controller that matches might not be accurate if different controllers are aware of different regions.
- Making this more performant would introduce a cache and mirror what we're currently doing with the database.

SimoneDutto:
> Points we discussed on this:
> 
> * Cloud fields are static and we don't get much gain here versus models where we were able to get rid of the watcher.
> * This might be annoyingly slow if latency to controllers is high and we need to list all clouds.
> * Returning the cloud info from the first controller that matches might not be accurate if different controllers are aware of different regions.
> * Making this more performant would introduce a cache and mirror what we're currently doing with the database.

After the discussion the decision is to close this pr. Moving forward we've decided not to remove from our db static fields.

PR #1513, opened 2025-01-08
fix(internal/jimm/controller.go): add default cloud region if none exist

alesstimec:
When adding a controller JIMM will query the controller for all known clouds and then add those clouds to its internal DB. If a cloud (e.g. MAAS or k8s) has no regions, JIMM would be unable to add a model to that cloud as it does not store association between clouds and controllers. To solve the issue we add a `default` region to clouds that do not have any regions and that cloud region will only have one controller associated with it.

## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests


internal/jimm/controller_test.go
```diff
@@ -176,6 +176,105 @@ func TestAddController(t *testing.T) {
...
+			// controller into JIMM.
+			ci.Users = []jujuparams.CloudUserInfo{{
+				UserName:    "alice@canonical.com",
+				DisplayName: "Alice",
+				Access:      "admin",
+			}, {
+				UserName:    "bob@canonical.com",
+				DisplayName: "Bob",
+				Access:      "add-model",
+			}}
```
kian99: Should we just leave this bit out?

kian99 (approved):

internal/jimm/controller_test.go
alesstimec: done

internal/jimm/controller.go
```diff
@@ -251,6 +251,18 @@ func (j *JIMM) AddController(ctx context.Context, user *openfga.User, ctl *dbmod
 	}
 
 	dbClouds := convertJujuCloudsToDbClouds(clouds)
+	for i, cloud := range dbClouds {
```
SimoneDutto, line 255: i would add some godoc explain why we are doing this

internal/jimm/controller_test.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2024 Canonical.
+// Copyright 2025 Canonical.
```
SimoneDutto, line 1: this is happening to me as well, do we want to change the cr header when we touch a file?

SimoneDutto (approved):
lgtm

internal/jimm/controller.go
alesstimec: done

PR #1514, opened 2025-01-09
chore: don't allow nil credential store

kian99:
## Description
Currently we allow a nil credential store for legacy functionality to store secrets in Postgres or Vault. Nowadays secrets go to the CredentialStore interface which can use an underlying Vault/Postgres/Other. In addition, secrets/sensitive data are no longer in dbmodel types.

Fixes [JUJU-7285](https://warthogs.atlassian.net/browse/JUJU-7285)

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

[JUJU-7285]: https://warthogs.atlassian.net/browse/JUJU-7285?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/db/secrets.go
```diff
@@ -126,6 +126,9 @@ func (d *Database) Get(ctx context.Context, tag names.CloudCredentialTag) (_ map
 	secret := dbmodel.NewSecret(tag.Kind(), tag.String(), nil)
 	err = d.GetSecret(ctx, &secret)
 	if err != nil {
+		if errors.ErrorCode(err) == errors.CodeNotFound {
+			return nil, nil
```
alesstimec, line 130: why are we returning no error if it is not found?

internal/jimm/cloudcredential.go
```diff
@@ -263,13 +247,10 @@ func (j *JIMM) updateControllerCloudCredential(
...
 	const op = errors.Op("jimm.updateControllerCloudCredential")
 
-	attr := cred.Attributes
-	if attr == nil {
-		var err error
-		attr, err = j.getCloudCredentialAttributes(ctx, cred)
-		if err != nil {
-			return nil, errors.E(op, err)
-		}
+	var err error
```
alesstimec: i don't think this line is needed

internal/jimm/types.go
```diff
@@ -0,0 +1,7 @@
+// Copyright 2025 Canonical.
+package jimm
```
alesstimec, line 3: missing new line

internal/jimm/types.go
```diff
@@ -0,0 +1,7 @@
...
+
+type ControllerCreds struct {
```
alesstimec, line 7: godoc, please

internal/testutils/jimmtest/gorm.go
```diff
@@ -322,7 +322,7 @@ func getOrCreateTemplateDatabase() (string, string, error) {
 
 	templateName, templateDSN, err := createTemplateDatabase()
 	if err != nil {
-		return "", "", errors.E(err, "error creating template database")
+		return "", "", fmt.Errorf("error creating template database: %w", err)
```
alesstimec, line 325: hmm.. yes.. i'm getting rather annoyed by our errors package.. not really fond of the way it handles errors..

alesstimec (approved):
LGTM with a few nitpicks. thanks

internal/db/secrets.go
kian99: This is the behaviour of our Vault client, that test where I set it to run against Postgres and Vault showed that our Postgres implementation was not consistent with our Vault implementation. So I've changed the Postgres implementation to match Vault.

SimoneDutto (approved):
lgtm, but i think we should go back to small prs. These latest prs (min included) are becoming difficult to review properly

kian99:
> lgtm, but i think we should go back to small prs

Yeah it will hopefully be better with the other managers because we can move whole files over and make changes rather than extracting bits and creating more lines changed than necessary.

PR #1515, opened 2025-01-10
feat: add ssh key table

kian99:
## Description
This PR adds a database migration and gorm model for users' public SSH keys. I reviewed https://github.com/canonical/jimm/pull/1287 when making this PR

Some justifications for the types chosen:
- `bytea` over `text` for the SSH public key. The Go ssh package has a `PublicKey` [interface](https://pkg.go.dev/golang.org/x/crypto/ssh#PublicKey) with a `Marshal` method returning a []byte. I expect this is what we will store in the database. The marshalled object also stores the key type.
- A separate key comment column allows us to persist the comment (not stored in the object mentioned above).
- No fingerprint column, the key fingerprint is a hash computed from the public key so we don't need to persist this separately.

Fixes [JUJU-7348](https://warthogs.atlassian.net/browse/JUJU-7348)

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

[JUJU-7348]: https://warthogs.atlassian.net/browse/JUJU-7348?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/dbmodel/sql/postgres/020_add_ssh_keys_table.up.sql
+++ b/internal/dbmodel/sql/postgres/020_add_ssh_keys_table.up.sql
@@ -1,0 +1,14 @@
+-- Add the ability to store users' SSH public keys
+
+CREATE TABLE ssh_keys (
+    id SERIAL PRIMARY KEY,
+    created_at TIMESTAMP WITH TIME ZONE DEFAULT NOW(),
+    updated_at TIMESTAMP WITH TIME ZONE,
+    public_key BYTEA NOT NULL,
+    key_comment VARCHAR(255),
+    identity_name TEXT NOT NULL,
+    FOREIGN KEY (identity_name) REFERENCES identities(name),
+    CONSTRAINT unique_identity_ssh_key UNIQUE(identity_name, public_key)
+);
+
+CREATE INDEX idx_ssh_keys_user_id ON ssh_keys (identity_name);
--- a/internal/dbmodel/sshkeys.go
+++ b/internal/dbmodel/sshkeys.go
@@ -1,0 +1,23 @@
+// Copyright 2025 Canonical.
+
+package dbmodel
+
+import "time"
+
+// SSHKey holds a user's public SSH key.
+type SSHKey struct {
+	// Note this doesn't use the standard gorm.Model to avoid soft-deletes.
+
+	ID        uint `gorm:"primarykey"`
+	CreatedAt time.Time
+	UpdatedAt time.Time
+
+	// IdentityName is the unique name (email or client-id) of this entity.
+	IdentityName string   `gorm:"uniqueIndex:unique_identity_ssh_key"`
+	Identity     Identity `gorm:"foreignKey:IdentityName;references:Name"`
+
+	// PublicKey holds the user's public SSH key.
+	PublicKey []byte `gorm:"uniqueIndex:unique_identity_ssh_key"`
+	// KeyComment holds a user provided comment.
+	KeyComment string
+}
--- a/internal/dbmodel/sshkeys_test.go
+++ b/internal/dbmodel/sshkeys_test.go
@@ -1,0 +1,35 @@
+// Copyright 2025 Canonical.
+
+package dbmodel_test
+
+import (
+	"testing"
+
+	qt "github.com/frankban/quicktest"
+
+	"github.com/canonical/jimm/v3/internal/dbmodel"
+)
+
+func TestSSHKeyUniqueConstraint(t *testing.T) {
+	c := qt.New(t)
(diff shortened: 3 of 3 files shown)


internal/dbmodel/sql/postgres/020_add_ssh_keys_table.up.sql
```diff
@@ -0,0 +1,14 @@
+-- Add the ability to store users' SSH public keys
+
+CREATE TABLE ssh_keys (
+    id SERIAL PRIMARY KEY,
+    created_at TIMESTAMP WITH TIME ZONE DEFAULT NOW(),
+    updated_at TIMESTAMP WITH TIME ZONE,
+    public_key BYTEA NOT NULL,
+    key_comment VARCHAR(255),
```
SimoneDutto, line 8: why 255? is there a hard limit?

internal/dbmodel/sql/postgres/020_add_ssh_keys_table.up.sql
```diff
@@ -0,0 +1,14 @@
...
+    identity_name TEXT NOT NULL,
+    FOREIGN KEY (identity_name) REFERENCES identities(name),
+    CONSTRAINT unique_identity_ssh_key UNIQUE(identity_name, public_key)
+);
+
+CREATE INDEX idx_ssh_keys_user_id ON ssh_keys (identity_name);
```
SimoneDutto, line 14: since we've discussed getting all user's keys and then checking only in this subset, i think this index is unnecessary. It is just a burden on the database

internal/dbmodel/sql/postgres/020_add_ssh_keys_table.up.sql
kian99: It's just a reasonable limit imo

internal/dbmodel/sql/postgres/020_add_ssh_keys_table.up.sql
kian99: This index would improve performance for queries of the form `SELECT * FROM ssh_keys WHERE user="kian@foo"`

SimoneDutto: yep, sorry. i've read "public_key"

SimoneDutto (approved):

internal/dbmodel/sql/postgres/020_add_ssh_keys_table.up.sql
alesstimec: 255 bytes should be enough for everybody ;)

internal/dbmodel/sql/postgres/020_add_ssh_keys_table.up.sql
```diff
@@ -0,0 +1,14 @@
+-- Add the ability to store users' SSH public keys
+
+CREATE TABLE ssh_keys (
+    id SERIAL PRIMARY KEY,
+    created_at TIMESTAMP WITH TIME ZONE DEFAULT NOW(),
+    updated_at TIMESTAMP WITH TIME ZONE,
+    public_key BYTEA NOT NULL,
```
alesstimec, line 7: i think juju displays hexadecimal md5 key signature when you run `juju ssh-keys` so i'm wondering if it would be worth storing that as well so that we can do the lookup

alesstimec (approved):
LGTM with a question about the key fingerprint being stored

internal/dbmodel/sql/postgres/020_add_ssh_keys_table.up.sql
kian99: Yes I was thinking about that too, but I don't expect users to have too many keys so we can just compute them from the public key

alesstimec: but if it would enable us to do the lookup based on key fingerprint without requiring the user to specify the username explicitly..

kian99: But I believe it would be incorrect to query the keys without the user specifying a username. It's a bit contrived but imagine I gave you my SSH key pair (or in the incredibly slim chance you happen to generate the same key). You could also upload the public key to the controller and then on login there are 2 keys that match.

It's the same principle as basic auth but without requiring a username. I don't know what the guarantees around uniqueness of SSH keys are (how long they are, their entropy, etc) but I'm fairly confident it would be wrong to query for keys without the "username" of the user attempting to login.

PR #1516, opened 2025-01-10
fix(internal/db/cloud.go): add `virtual` field to CloudRegion

alesstimec:
## Description

When we add a cloud reports no cloud regions we add a virtual "default" region to keep the
controller lookup logic the same when adding a model. Then only difference is that we do not send
the cloud region name in ModelCreateArgs to the controller.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

### DIFF
--- a/internal/dbmodel/cloud.go
+++ b/internal/dbmodel/cloud.go
@@ -1,4 +1,4 @@
-// Copyright 2024 Canonical.
+// Copyright 2025 Canonical.
 
 package dbmodel
 
@@ -181,6 +181,14 @@ type CloudRegion struct {
 	// Controllers contains any controllers that can provide service for
 	// this cloud-region.
 	Controllers []CloudRegionControllerPriority
+
+	// Virtual, if true, indicates that a cloud was reported to have
+	// no cloud regions so we created a virtual "default" regions for it.
+	// This is necessary because Juju seems inconsistent in how they manage
+	// clouds with no regions; e.g. maas or k8s clouds may not have regions,
+	// yet for some cloud juju creates a "default" region, while for others
+	// it does not.
+	Virtual bool
 }
 
 // ToJujuCloudRegion converts a CloudRegion into a jujuparams.CloudRegion.
--- a/internal/jimm/model_test.go
+++ b/internal/jimm/model_test.go
@@ -1054,6 +1054,106 @@ users:
 		OwnerTag: names.NewUserTag("alice@canonical.com").String(),
 	},
 	expectError: "no cloud specified for model; please specify one",
+}, {
+	name: "CreateModelOnACloudWithNoRegions",
+	// test-cloud has one virtual cloud region
+	env: `
+clouds:
+- name: test-cloud
+  type: test-provider
+  regions:
+  - name: default
+    virtual: true
+  users:
+  - user: alice@canonical.com
+    access: add-model
+cloud-credentials:
+- name: test-credential-1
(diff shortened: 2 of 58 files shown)


internal/jimm/model_test.go
```diff
@@ -1054,6 +1054,105 @@ users:
...
+    region: default
+    priority: 2
+`[1:],
+	updateCredential: func(_ context.Context, _ jujuparams.TaggedCredential) ([]jujuparams.UpdateCredentialModelResult, error) {
+		return nil, nil
+	},
+	grantJIMMModelAdmin: func(_ context.Context, _ names.ModelTag) error {
+		return nil
+	},
+	createModel: assertCreateModelArgs(&jujuparams.ModelCreateArgs{
```
kian99, line 1099: This test case and the corresponding `assertCreateModelArgs` are super confusing to read. Why is this pattern needed?

alesstimec: just continuing the pattern that was here before which is not all that confusing.. think of it as middleware: each function checks a certain aspect of the call and passes check to the next one

kian99: I took a look at the other test cases and I see what you mean that it's following the pattern. I think they're all extremely confusing, the whole test is really hard to read.

In any case, regarding the stuff inside `assertCreateModelArgs`. Why are we asserting that the fields match when we also have `expectModel` which is likely being used inside the test to verify the created model matches what is expected.

alesstimec: `expectModel` asserts that our database looks the way we expect it - it does not assert the arguments sent to the controller to create a new model. 
the `assertCreateModelArgs` asserts that what we send to the controller contains parameters we thing should be sent to the controller - in this test case the cloud region should not be sent since it's a virtual one

kian99: Ahh I understand, that is nice but tough to parse. Grabbing lunch then I'll take a closer look with that in mind. But seems fine to me

kian99 (approved):

internal/dbmodel/cloud.go
```diff
@@ -181,6 +181,14 @@ type CloudRegion struct {
...
 	// this cloud-region.
 	Controllers []CloudRegionControllerPriority
+
+	// Virtual, if true, indicates that a cloud was reported to have
+	// no cloud regions so we created a virtual "default" regions for it.
+	// This is necessary because Juju seems inconsistent in how they manage
+	// clouds with no regions; e.g. maas or k8s clouds may not have regions,
+	// yet for some cloud juju creates a "default" region, while for others
+	// it does not.
+	Virtual bool
```
SimoneDutto, line 191: premise: i don't know if you can add a region after a cloud has been created.

In case you can, we should mark the region as virtual and then remove the virtual region once a `real` region is added.

SimoneDutto: additional question: if we mark the region as `Virtual` we don't need to add a fake `default` region anymore as you did in the previous PR. Right?

alesstimec: i do not know if it is possible to add regions: theoretically we could add a controller that would bring with it additional regions for a cloud.. but then that becomes a more complicated issue - is this a public cloud, is it private, is it even the same cloud, etc..

on the second questions: yes, we still need to add a default "virtual" region

SimoneDutto (approved):

PR #1517, opened 2025-01-10
feat: add ssh jump server

SimoneDutto:
## Description
This pr introduces the ssh jump server.

It is the basic struct to start, we still are missing:
- handling host keys
- resolving actual controllers

But this is the initial blueprint. 

## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -349,6 +349,8 @@ require (
 require github.com/golang-migrate/migrate/v4 v4.17.1
 
 require (
+	github.com/anmitsu/go-shlex v0.0.0-20200514113438-38f4b401e2be // indirect
+	github.com/gliderlabs/ssh v0.3.8 // indirect
 	github.com/klauspost/compress v1.17.7 // indirect
 	github.com/moby/patternmatcher v0.6.0 // indirect
 	github.com/moby/sys/sequential v0.6.0 // indirect
--- a/go.mod
+++ b/go.mod
@@ -30,9 +30,11 @@ require (
 	github.com/coreos/go-oidc/v3 v3.11.0
 	github.com/dustinkirkland/golang-petname v0.0.0-20231002161417-6a283f1aaaf2
 	github.com/frankban/quicktest v1.14.6
+	github.com/gliderlabs/ssh v0.3.8
 	github.com/go-chi/chi/v5 v5.0.12
 	github.com/go-chi/render v1.0.2
 	github.com/go-macaroon-bakery/macaroon-bakery/v3 v3.0.1
+	github.com/golang-migrate/migrate/v4 v4.17.1
 	github.com/google/go-cmp v0.6.0
 	github.com/google/uuid v1.6.0
 	github.com/gorilla/sessions v1.2.1
@@ -314,7 +316,7 @@ require (
 	go.uber.org/atomic v1.11.0 // indirect
 	go.uber.org/mock v0.5.0 // indirect
 	go.uber.org/multierr v1.11.0 // indirect
-	golang.org/x/crypto v0.31.0 // indirect
+	golang.org/x/crypto v0.31.0
 	golang.org/x/exp v0.0.0-20240904232852-e7e105dedf7e // indirect
 	golang.org/x/net v0.30.0 // indirect
 	golang.org/x/sys v0.28.0 // indirect
@@ -346,11 +348,8 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.4.1 // indirect
 )
 
-require github.com/golang-migrate/migrate/v4 v4.17.1
-
 require (
 	github.com/anmitsu/go-shlex v0.0.0-20200514113438-38f4b401e2be // indirect
-	github.com/gliderlabs/ssh v0.3.8 // indirect
 	github.com/klauspost/compress v1.17.7 // indirect
 	github.com/moby/patternmatcher v0.6.0 // indirect
 	github.com/moby/sys/sequential v0.6.0 // indirect
--- a/internal/ssh/export_test.go
+++ b/internal/ssh/export_test.go
@@ -1,0 +1,4 @@
+// Copyright 2025 Canonical.
+package ssh
+
+type ForwardMessage = forwardMessage
--- a/internal/ssh/export_test.go
+++ b/internal/ssh/export_test.go
@@ -1,4 +1,5 @@
 // Copyright 2025 Canonical.
+
 package ssh
 
 type ForwardMessage = forwardMessage
--- a/internal/ssh/ssh.go
+++ b/internal/ssh/ssh.go
@@ -93,34 +93,36 @@ func directTCPIPHandler(resolver Resolver) func(srv *ssh.Server, conn *gossh.Ser
 			return
 		}
 
-		dChan, reqs, err := client.OpenChannel("direct-tcpip", gossh.Marshal(d))
+		dstChan, reqs, err := client.OpenChannel("direct-tcpip", gossh.Marshal(d))
 		if err != nil {
 			rejectConnectionAndLogError(ctx, newChan, "Failed to open destination channel", err)
 			return
 		}
-
+		// gossh.Request are requests sent outside of the normal stream of data (ex. pty-req for an interactive session).
+		// Since we only need the raw data to redirect, we can discard them.
 		go gossh.DiscardRequests(reqs)
 
-		ch, reqs, err := newChan.Accept()
+		srcDest, reqs, err := newChan.Accept()
 		if err != nil {
-			dChan.Close()
+			dstChan.Close()
 			return
 		}
-
+		// gossh.Request are requests sent outside of the normal stream of data (ex. pty-req for an interactive session).
+		// Since we only need the raw data to redirect, we can discard them.
 		go gossh.DiscardRequests(reqs)
 
 		go func() {
-			defer ch.Close()
-			defer dChan.Close()
-			_, err := io.Copy(ch, dChan)
+			defer srcDest.Close()
+			defer dstChan.Close()
+			_, err := io.Copy(srcDest, dstChan)
 			if err != nil {
 				rejectConnectionAndLogError(ctx, newChan, "Failed to copy data from src to dts", err)
 			}
 		}()
 		go func() {
-			defer ch.Close()
-			defer dChan.Close()
-			_, err := io.Copy(dChan, ch)
+			defer srcDest.Close()
+			defer dstChan.Close()
+			_, err := io.Copy(dstChan, srcDest)
 			if err != nil {
 				rejectConnectionAndLogError(ctx, newChan, "Failed to copy data from dst to src", err)
 			}
--- a/internal/ssh/ssh.go
+++ b/internal/ssh/ssh.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"net"
 
 	"github.com/gliderlabs/ssh"
 	"github.com/juju/zaputil/zapctx"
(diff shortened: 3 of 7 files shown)


go.mod
```diff
@@ -349,6 +349,8 @@ require (
 require github.com/golang-migrate/migrate/v4 v4.17.1
 
 require (
+	github.com/anmitsu/go-shlex v0.0.0-20200514113438-38f4b401e2be // indirect
+	github.com/gliderlabs/ssh v0.3.8 // indirect
```
kian99: Might need to run `go mod tidy`, I think this should be a direct dependency.

internal/ssh/ssh.go
```diff
@@ -0,0 +1,137 @@
...
+	"github.com/gliderlabs/ssh"
+	"github.com/juju/zaputil/zapctx"
+	"go.uber.org/zap"
+	gossh "golang.org/x/crypto/ssh"
+
+	"github.com/canonical/jimm/v3/internal/openfga"
+)
+
+// JUJU_SSH_DEFAULT_PORT is the default port we expect the juju controllers to respond on.
+const JUJU_SSH_DEFAULT_PORT = 2223
```
kian99: The default port is defined in the spec as 17022

internal/ssh/ssh.go
```diff
@@ -0,0 +1,137 @@
...
+
+// Resolver is the interface with the methods needed by the ssh jump server to route request.
+type Resolver interface {
+	// GetAddrFromModelUUID is the method to resolve the address of the controller to contact given the model UUID.
+	GetAddrFromModelUUID(ctx context.Context, user openfga.User, modelUUID string) (string, error)
```
kian99: I'd drop the Get prefix
```suggestion
	// AddrFromModelUUID is the method to resolve the address of the controller to contact given the model UUID.
	AddrFromModelUUID(ctx context.Context, user openfga.User, modelUUID string) (string, error)
```

internal/ssh/ssh_test.go
```diff
@@ -0,0 +1,175 @@
...
+	)
+
+	s.privateKey, err = gossh.ParsePrivateKey(keyPEM)
+	c.Assert(err, qt.IsNil)
+}
+
+// CleanUp doesn't exist in qtsuite, so it needs to be called manually
+func (s *sshSuite) CleanUp(c *qt.C) {
+	err := s.destinationJujuSSHServer.Close()
+	c.Assert(err, qt.IsNil)
```
kian99: Change this to c.Check() so that it doesn't stop the cleanup if one fails.

internal/ssh/ssh_test.go
```diff
@@ -0,0 +1,175 @@
...
+	err = s.jumpSSHServer.Close()
+	c.Assert(err, qt.IsNil)
+}
+
+func (s *sshSuite) TestSSHJump(c *qt.C) {
+	defer s.CleanUp(c)
```
kian99: suite cleanup doesn't run automatically?
I think there is a way to declare cleanup functions from inside the init() method

internal/ssh/ssh_test.go
```diff
@@ -0,0 +1,175 @@
...
+			"direct-tcpip": func(srv *gliderssh.Server, conn *gossh.ServerConn, newChan gossh.NewChannel, ctx gliderssh.Context) {
+				d := ssh.ForwardMessage{}
+				if err := gossh.Unmarshal(newChan.ExtraData(), &d); err != nil {
+					err := newChan.Reject(gossh.ConnectionFailed, "Failed to parse channel data")
+					c.Assert(err, qt.IsNil)
+					return
+				}
+				_, _, err := newChan.Accept()
+				c.Assert(err, qt.IsNil)
+				s.testF(d)
```
kian99: Let's give this a better name, like `testInDestinationServerF`

internal/ssh/ssh_test.go
```diff
@@ -0,0 +1,175 @@
...
+	// send forward message
+	msg := ssh.ForwardMessage{
+		DestAddr: "model1",
+		//nolint:gosec
+		DestPort: uint32(s.destinationServerPort),
+		SrcAddr:  "localhost",
+		SrcPort:  0,
+	}
+	s.testF = func(fm ssh.ForwardMessage) {
+		c.Assert(fm.DestAddr, qt.Equals, "model1")
```
kian99: Make it a Check so that it doesn't stop the test.

internal/ssh/ssh_test.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+	ch, _, err := client.OpenChannel("direct-tcpip", gossh.Marshal(&msg))
+	c.Assert(err, qt.IsNil)
+	defer ch.Close()
+	select {
+	case <-s.received:
+	case <-time.After(100 * time.Millisecond):
+		c.Fatalf("ssh jump test timeout")
```
kian99: Maybe c.Fail() so that it fails the test but allows the cleanup to continue?

internal/ssh/ssh_test.go
```diff
@@ -0,0 +1,175 @@
...
+		//nolint:gosec
+		DestPort: uint32(s.destinationServerPort),
+		SrcAddr:  "localhost",
+		SrcPort:  0,
+	}
+	s.testF = func(fm ssh.ForwardMessage) {
+		c.Assert(fm.DestAddr, qt.Equals, "model1")
+	}
+	ch, _, err := client.OpenChannel("direct-tcpip", gossh.Marshal(&msg))
+	c.Assert(err, qt.IsNil)
```
kian99: c.Check() here as well

internal/ssh/ssh_test.go
```diff
@@ -0,0 +1,175 @@
...
+			gossh.PublicKeys(s.privateKey),
+		},
+	})
+	c.Assert(err, qt.IsNil)
+
+	// send forward message
+	msg := ssh.ForwardMessage{
+		DestAddr: "model1",
+		//nolint:gosec
+		DestPort: 1,
```
kian99: Is it expected to fail because of the port right? Maybe add a comment on that

internal/utils/utils.go
```diff
@@ -21,3 +23,15 @@ func NewConversationID() string {
 	}
 	return hex.EncodeToString(buf)
 }
+
+// GetFreePort asks the kernel for a free open port that is ready to use.
+func GetFreePort() (int, error) {
```
kian99: nifty!

internal/ssh/export_test.go
```diff
@@ -0,0 +1,4 @@
+// Copyright 2025 Canonical.
+package ssh
```
alesstimec, line 3: new line

internal/ssh/ssh.go
alesstimec: also.. no need to export this variable

internal/ssh/ssh.go
```diff
@@ -0,0 +1,137 @@
...
+		}
+		if d.DestPort == 0 {
+			d.DestPort = JUJU_SSH_DEFAULT_PORT
+		}
+		addr, err := resolver.GetAddrFromModelUUID(ctx, openfga.User{}, d.DestAddr)
+		if err != nil {
+			rejectConnectionAndLogError(ctx, newChan, "Failed to resolve address from model uuid", err)
+			return
+		}
+		dest := fmt.Sprintf("%s:%d", addr, d.DestPort)
```
alesstimec:
```suggestion
		dest := net.JoinHostPort(addr, d.DestPort)
```

alesstimec: remember ipv6 addresses? :)

internal/ssh/ssh.go
```diff
@@ -0,0 +1,137 @@
...
+		// this is temporary. The way we dial to the controller will heavily change.
+		client, err := gossh.Dial("tcp", dest, &gossh.ClientConfig{
```
alesstimec, line 86: somewhere before this point we'll have to plug in some authorization: is this user authorized to ssh to this target?

internal/ssh/ssh.go
```diff
@@ -0,0 +1,137 @@
...
+			return
+		}
+
+		dChan, reqs, err := client.OpenChannel("direct-tcpip", gossh.Marshal(d))
+		if err != nil {
+			rejectConnectionAndLogError(ctx, newChan, "Failed to open destination channel", err)
+			return
+		}
+
+		go gossh.DiscardRequests(reqs)
```
alesstimec, line 107: a bit of explanation in the form of a godoc would be in order here and below.. why are we discarding certain requests..

internal/ssh/ssh_test.go
```diff
@@ -0,0 +1,175 @@
+// Copyright 2025 Canonical.
+package ssh_test
```
alesstimec, line 3: new line

internal/ssh/ssh_test.go
```diff
@@ -0,0 +1,175 @@
...
+	jumpSSHServer            ssh.Server
+	jumpServerPort           int
+	privateKey               gossh.Signer
+	testF                    func(fm ssh.ForwardMessage)
+	received                 chan bool
+}
+
+func (s *sshSuite) Init(c *qt.C) {
+	s.received = make(chan bool)
+	port, err := utils.GetFreePort()
```
alesstimec: i don't think there's a guarantee that the returned port won't be taken by another process before it's used in the gliderssh.Server ListenAndServe..

internal/ssh/ssh.go
SimoneDutto: yes, going to do that in https://warthogs.atlassian.net/browse/JUJU-7349

internal/ssh/ssh_test.go
SimoneDutto: no, is not guaranteed. If we see these tests being flaky I will try to implement a locking mechanism or smt else.
Is it fine for now to leave it like this with a comment?

SimoneDutto: the port is chosen randomly from the free ones, so you'd be very unlucky to have two times the same port

internal/ssh/ssh.go
SimoneDutto: added it.

internal/ssh/ssh_test.go
kian99: I think this wasn't changed but should've.

internal/ssh/ssh_test.go
```diff
@@ -0,0 +1,171 @@
...
+	// send forward message
+	msg := ssh.ForwardMessage{
+		DestAddr: "model1",
+		//nolint:gosec
+		DestPort: 1, // the test fails because there is no ssh server on this port.
+		SrcAddr:  "localhost",
+		SrcPort:  0,
+	}
+	s.testInDestinationServerF = func(fm ssh.ForwardMessage) {
+		c.Assert(fm.DestAddr, qt.Equals, "model1")
```
kian99: Another change to `Check`

SimoneDutto: done

internal/ssh/ssh_test.go
SimoneDutto: done

kian99 (approved):

internal/utils/utils.go
kian99: One other thing, you can consider moving this to the jimmtest package instead.

internal/ssh/ssh.go
```diff
@@ -0,0 +1,140 @@
...
+type Server struct {
+	*ssh.Server
+
+	resolver Resolver
+}
+
+// NewJumpSSHServer creates the jump server struct.
+func NewJumpSSHServer(ctx context.Context, port int, resolver Resolver) (Server, error) {
+	zapctx.Info(ctx, "NewSSHServer")
+	server := Server{
```
alesstimec, line 50: perhaps check that the resolver is not nil?

internal/ssh/ssh.go
```diff
@@ -0,0 +1,140 @@
...
+		go func() {
+			defer srcDest.Close()
+			defer dstChan.Close()
+			_, err := io.Copy(srcDest, dstChan)
+			if err != nil {
+				rejectConnectionAndLogError(ctx, newChan, "Failed to copy data from src to dts", err)
+			}
+		}()
+		go func() {
+			defer srcDest.Close()
```
alesstimec, line 127: won't we get a panic if we try to close the channel twice (in both goroutines)?

SimoneDutto:
```
// writeMu serializes calls to mux.conn.writePacket() and
// protects sentClose and packetPool. This mutex must be
// different from windowMu, as writePacket can block if there
// is a key exchange pending.
writeMu   sync.Mutex
sentClose bool
```
there is a mutex and a bool guarding the close msg

internal/utils/utils.go
SimoneDutto: done

internal/ssh/ssh.go
SimoneDutto: done

alesstimec (approved):

PR #1518, opened 2025-01-13
feat: ssh keys db crud

kian99:
## Description
This PR introduces add/list/remove methods to the `db` package for SSH keys. There is no read method because there is currently no use case where we need to lookup a single key.

Fixes [JUJU-7347](https://warthogs.atlassian.net/browse/JUJU-7347)

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

[JUJU-7347]: https://warthogs.atlassian.net/browse/JUJU-7347?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/db/sshkeys.go
```diff
@@ -0,0 +1,118 @@
+// Copyright 2025 Canonical.
+package db
```
alesstimec, line 3: new line, please

internal/db/sshkeys.go
```diff
@@ -0,0 +1,118 @@
...
+
+	var keys []dbmodel.SSHKey
+	if err := d.DB.Where("identity_name = ?", identityName).Find(&keys).Error; err != nil {
+		return errors.E(op, dbError(err))
+	}
+
+	// It is expected that we only have 1 key that matches this fingerprint
+	// because of the unique constraint between users and public keys.
+	for _, key := range keys {
+		fp, err := calculateSSHFingerprint(key.PublicKey)
```
alesstimec: ah, yes, i think we should be storing the fingerprint.. then this method would be much simpler and fingerprint is a static piece of data that never changes for a given key

internal/db/sshkeys_test.go
```diff
@@ -0,0 +1,153 @@
+// Copyright 2025 Canonical.
+package db_test
```
alesstimec, line 3: new line, please

internal/db/sshkeys_test.go
```diff
@@ -0,0 +1,153 @@
...
+	u, err := dbmodel.NewIdentity("bob@canonical.com")
+	c.Assert(err, qt.IsNil)
+	c.Assert(s.Database.DB.Create(u).Error, qt.IsNil)
+
+	key := dbmodel.SSHKey{
+		Identity:   *u,
+		PublicKey:  []byte("foo"),
+		KeyComment: "bar",
+	}
+	err = s.Database.AddSSHKey(context.Background(), &key)
```
alesstimec, line 30: we should test what that we get a sensible error if we try adding the key twice

alesstimec (commented):
should we store key fingerprints?

internal/db/sshkeys.go
kian99: We can, I just initially took the route of not storing it since I figured users would have a small number of keys. I'll let others chime in then see about changing it.

internal/db/sshkeys.go
```diff
@@ -0,0 +1,118 @@
...
+		}
+		if fp == fingerprint {
+			if err := d.DB.WithContext(ctx).Delete(key).Error; err != nil {
+				return errors.E(op, dbError(err))
+			}
+			return nil
+		}
+	}
+
+	return errors.E(op, errors.CodeNotFound, "key not found")
```
SimoneDutto:
```suggestion
	return errors.E(op, errors.CodeNotFound, return errors.E(op, errors.CodeNotFound, fmt.Sprintf("fingerprint for identity %s not found", identityName)))
```

internal/db/sshkeys.go
```diff
@@ -0,0 +1,118 @@
...
+	defer durationObserver()
+	defer servermon.ErrorCounter(servermon.DBQueryErrorCount, &err, string(op))
+
+	query := d.DB.Where("key_comment = ?", comment).Delete(&dbmodel.SSHKey{})
+	if err := query.Error; err != nil {
+		return errors.E(op, dbError(err))
+	}
+
+	if query.RowsAffected == 0 {
+		return errors.E(op, errors.CodeNotFound, "key not found")
```
SimoneDutto, line 75:
```suggestion
		return errors.E(op, errors.CodeNotFound, fmt.Sprintf("comment for identity %s not found", identityName))
```

internal/db/sshkeys.go
```diff
@@ -0,0 +1,118 @@
...
+		return errors.E(op, err)
+	}
+
+	durationObserver := servermon.DurationObserver(servermon.DBQueryDurationHistogram, string(op))
+	defer durationObserver()
+	defer servermon.ErrorCounter(servermon.DBQueryErrorCount, &err, string(op))
+
+	var keys []dbmodel.SSHKey
+	if err := d.DB.Where("identity_name = ?", identityName).Find(&keys).Error; err != nil {
+		return errors.E(op, dbError(err))
```
SimoneDutto, line 47: i get this dbError but it would be better to add an identifier to these logs. So it's easier to troubleshoot in case of issue.

internal/db/sshkeys.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+
+	if query.RowsAffected == 0 {
+		return errors.E(op, errors.CodeNotFound, "key not found")
+	}
+
+	return nil
+}
+
+// ListSSHKeys all a user's SSH keys.
```
SimoneDutto:
```suggestion
// ListSSHKeys lists all user's SSH keys.
```

internal/db/sshkeys.go
```diff
@@ -0,0 +1,118 @@
...
+func (d *Database) ListSSHKeys(ctx context.Context, identityName string) (keys []dbmodel.SSHKey, err error) {
```
SimoneDutto: ListSSHKeysForUser i think would be better

kian99: A bit torn on this because the method signature has an `identityName` and none of the other methods are "forUser" or specify user.

internal/db/sshkeys.go
kian99: That's what the `op` does I believe

SimoneDutto: i am referring to some identity identifier, like this `return errors.E(op, errors.CodeNotFound, fmt.Sprintf("comment for identity %s not found", identityName))`

kian99: I've added it to the remove key methods, but I don't think putting the user onto the error is necessarily needed for making debugging easier. Otherwise it just feels like we're adding all the input params onto the errors which can be excessive. Especially down in this method, maybe a method higher up can add more detail to the error in the future if needed.

SimoneDutto: fair enough!

internal/db/sshkeys.go
SimoneDutto: i think because the other ones are obvious, in the sense that you'd always a key for a user.
The listing can be done table wide, but here we are doing it per user

kian99: Fair enough, changing it.

internal/db/sshkeys.go
kian99: Changed it to store the fingerprint.

SimoneDutto (approved):
ty

internal/db/sshkeys.go
```diff
@@ -0,0 +1,98 @@
...
+	const op = errors.Op("db.AddSSHKey")
+	if err := d.ready(); err != nil {
+		return errors.E(op, err)
+	}
+
+	durationObserver := servermon.DurationObserver(servermon.DBQueryDurationHistogram, string(op))
+	defer durationObserver()
+	defer servermon.ErrorCounter(servermon.DBQueryErrorCount, &err, string(op))
+
+	if err := d.DB.WithContext(ctx).Create(sshKey).Error; err != nil {
```
alesstimec, line 24: should we verify that the ssh key fingerprint is not empty?

alesstimec: ah, i see we have a constraint..  i suppose it will be computed in the jimm package method (or the sshkey manager).

internal/db/sshkeys_test.go
```diff
@@ -0,0 +1,160 @@
...
+
+	// key2 with the same public-key but different owner should not be deleted.
+	key2 := dbmodel.SSHKey{Identity: *u2, PublicKey: publicKey.Marshal(), KeyComment: "bar", MD5Fingerprint: gossh.FingerprintLegacyMD5(publicKey)}
+	c.Assert(s.Database.DB.Create(&key2).Error, qt.IsNil)
+
+	var keyCount int64
+	c.Assert(s.Database.DB.Model(&dbmodel.SSHKey{}).Count(&keyCount).Error, qt.IsNil)
+	c.Assert(keyCount, qt.Equals, int64(2))
+
+	err = s.Database.RemoveSSHKeyByFingerprint(context.Background(), "bob@canonical.com", gossh.FingerprintLegacyMD5(publicKey))
```
alesstimec, line 108: we should also test for "key not found" error here and in the next test to be aligned with juju's key handling responses.

alesstimec (approved):
LGTM with a few comments

PR #1519, opened 2025-01-13
feat: add sshkeys manager

kian99:
## Description
In preparation for intercepting SSH commands in the model proxy, I've created the business logic package `sshkeys` and corresponding manager struct to handle logic related to user SSH keys.

This PR builds on #1518

Partially fixes [JUJU-7346](https://warthogs.atlassian.net/browse/JUJU-7346)

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests


SimoneDutto (approved):
nice

SimoneDutto (changes requested):
i think we're missing the mocks for the sshkey manager

kian99:
> i think we're missing the mocks for the sshkey manager

Yeah and I need to wire it up still

SimoneDutto (approved):
nice nice

internal/jimm/sshkeys/types.go
```diff
@@ -0,0 +1,22 @@
+// Copyright 2025 Canonical.
+
+package sshkeys
+
+import (
+	gossh "golang.org/x/crypto/ssh"
+)
+
+type PublicKey struct {
+	gossh.PublicKey
```
alesstimec, line 13: oh, please, let's not..

internal/jimm/sshkeys/types.go
```diff
@@ -0,0 +1,22 @@
+// Copyright 2025 Canonical.
+
+package sshkeys
+
+import (
+	gossh "golang.org/x/crypto/ssh"
+)
+
+type PublicKey struct {
```
alesstimec, line 12: godoc, please

alesstimec (commented):
Let's please not embed in imported type in one of our own.

internal/jimm/sshkeys/types.go
kian99: Why? I was following a similar approach to the gliderlabs/ssh work where they have their own interface type for a public key and embed the gossh public key type [here](https://github.com/gliderlabs/ssh/blob/d137aad99cd6f2d9495bfd98c755bec4e5dffb8c/wrap.go#L6). But then I needed a comment field so I turned it into a struct. Should I just make the `gossh.PublicKey` a full member variable rather than embed it?

alesstimec: because then our internal type depends on an imported type. should we update dependency and they change the PublicKey type it would auto-magically pull those changes in.. unlikely as it may be

kian99: Hmmm so maybe then we just define our own interface.
Should we copy the entire gossh.PublicKey interface (below):
```
type PublicKey interface {
	// Type returns the key format name, e.g. "ssh-rsa".
	Type() string

	// Marshal returns the serialized key data in SSH wire format, with the name
	// prefix. To unmarshal the returned data, use the ParsePublicKey function.
	Marshal() []byte

	// Verify that sig is a signature on the given data using this key. This
	// method will hash the data appropriately first. sig.Format is allowed to
	// be any signature algorithm compatible with the key type, the caller
	// should check if it has more stringent requirements.
	Verify(data []byte, sig *Signature) error
}
```
Or should we just copy the Type() and Marshal() methods?
Or something else?
I'm not sure I fully appreciate the risk of using an external type in our own type right now.

alesstimec: ah, nvm.. i see Verify uses the signature type as parameter.. no sense in making our own..

alesstimec (approved):

PR #1520, opened 2025-01-13
chore: move ListResources method to permission manager

kian99:
## Description
The `ListResources` method is used solely by the rebac-admin UI to list the resources that are known to JIMM. This method would need to go on a manager/business object eventually, and the permission manager made the most sense with the above in mind.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests


SimoneDutto:
the idea is to eventually move everything into a manager? or we can keep something in the jimm struct?

kian99:
> the idea is to eventually move everything into a manager? or we can keep something in the jimm struct?

Imo everything is inside a manager. The analogy of the kitchen drawer is that each utensil is contained within its own space, spoons, forks, knives, etc, all neatly arranged. We will for sure have things without a manager for some time yet but I think that will be an exception until we can clean them up.

internal/jimm/permissions/relations_test.go
```diff
@@ -266,3 +266,49 @@ func (s *permissionManagerSuite) TestListObjectRelations(c *qt.C) {
 		})
 	}
 }
+
+func (s *permissionManagerSuite) TestGetResources(c *qt.C) {
```
alesstimec: nitpick: s/TestGetResources/TestListResources/

alesstimec (approved):
LGTM with a nitpick

SimoneDutto (approved):
lgtm

PR #1522, opened 2025-01-14
add config ssh jump server

SimoneDutto:
## Description
In this PR we add configuration for SSH jump server.
Such as:
- port
- hostkey
- max concurrent connection: for this one we don't have an implementation yet, because it is not a default option in the gliderlab ssh server implementation

> Fly by move the types to a separate file to improve readability.

## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests


internal/ssh/ssh_test.go
```diff
@@ -71,13 +72,29 @@ func (s *sshSuite) Init(c *qt.C) {
 	port, err = jimmtest.GetFreePort()
 	c.Assert(err, qt.IsNil)
 	s.jumpServerPort = port
-	s.jumpSSHServer, err = ssh.NewJumpSSHServer(context.Background(), port, resolver{})
+	k, err := rsa.GenerateKey(rand.Reader, 2048)
+	c.Assert(err, qt.IsNil)
+	hostKey := pem.EncodeToMemory(
+		&pem.Block{
+			Type:  "RSA PRIVATE KEY",
```
kian99, line 79: There's no constant for this?

kian99 (approved):
I assume we'll have a follow up where we add the ability for the user to pass in these config options

internal/ssh/ssh.go
```diff
@@ -25,31 +25,16 @@ type Resolver interface {
...
-// Server is the custom struct to embed the gliderlabs.ssh server and a resolver.
-type Server struct {
-	*ssh.Server
-
-	resolver Resolver
-}
-
 // NewJumpSSHServer creates the jump server struct.
-func NewJumpSSHServer(ctx context.Context, port int, resolver Resolver) (Server, error) {
+func NewJumpSSHServer(ctx context.Context, config SSHServerConfig, resolver Resolver) (Server, error) {
```
alesstimec: suggestions:
- rename NewJumpSSHServer to NewJumpServer -> we know it's ssh because it is in the ssh package and it avoids repetition when somebody does `ssh.NewJumpSSHServer` (notice ssh appears twice in this call)
- likewise rename SSHServerConfig to ServerConfig for the same reason

internal/ssh/types.go
```diff
@@ -0,0 +1,25 @@
...
+}
+
+// Server is the custom struct to embed the gliderlabs.ssh server and a resolver.
+type Server struct {
+	*ssh.Server
+
+	resolver Resolver
+}
+
+type SSHServerConfig struct {
```
alesstimec:
```suggestion
type ServerConfig struct {
```
or just `Config`

internal/ssh/ssh_test.go
SimoneDutto: i've looked around and it's always a string, no types.

even in the godoc of this field:
`Type    string            // The type, taken from the preamble (i.e. "RSA PRIVATE KEY").`

cmd/jimmsrv/main.go
```diff
@@ -209,5 +209,18 @@ func start(ctx context.Context, s *service.Service) error {
 	})
 	s.Go(httpsrv.ListenAndServe)
 	zapctx.Info(ctx, "Successfully started JIMM server")
+
+	// // this is to show the integration, we will uncommented it once the ssh implementation is ready.
```
alesstimec: i would not add this in this PR as it might require a bit of discussion: do we start the jump server in main.go or in cmd/jimmsrv/service/service.go?

internal/ssh/types.go
```diff
@@ -0,0 +1,25 @@
+// Copyright 2025 Canonical.
+package ssh
```
alesstimec: new line, please

alesstimec (changes requested):

SimoneDutto:
> I assume we'll have a follow up where we add the ability for the user to pass in these config options

yep, i will open a PR for the k8s charm as well

cmd/jimmsrv/main.go
SimoneDutto: yes, it was just to show how that would have worked, we can do it in another time.

internal/ssh/types.go
```diff
@@ -0,0 +1,26 @@
...
+}
+
+// Server is the custom struct to embed the gliderlabs.ssh server and a resolver.
+type Server struct {
+	*ssh.Server
+
+	resolver Resolver
+}
+
+type Config struct {
```
alesstimec: godoc, please

internal/ssh/types.go
```diff
@@ -0,0 +1,26 @@
...
+// fowardMessage is the struct holding the information about the jump message received by the ssh client.
+type forwardMessage struct {
+	DestAddr string
+	DestPort uint32
+	SrcAddr  string
+	SrcPort  uint32
+}
+
+// Server is the custom struct to embed the gliderlabs.ssh server and a resolver.
+type Server struct {
```
alesstimec: i'm ok with types that are only used as parameters (wireformats, etc) living in types.go.. but i would really suggest keeping type definitions in the same file as their method implementations.. server type would possibly belong in internal/ssh/server.go making it easier to find it

alesstimec (commented):
generally lgtm, but i would suggest keeping type definitions in the same file as method implementations.

internal/ssh/types.go
SimoneDutto: done

internal/ssh/types.go
SimoneDutto: okok, moved them to the ssh.go

internal/ssh/ssh.go
```diff
@@ -40,16 +40,23 @@ type Server struct {
...
 }
 
-// NewJumpSSHServer creates the jump server struct.
-func NewJumpSSHServer(ctx context.Context, port int, resolver Resolver) (Server, error) {
-	zapctx.Info(ctx, "NewSSHServer")
+// Config is the struct holding the configuration for the jump server.
+type Config struct {
+	Port                     string
+	HostKey                  []byte
+	MaxConcurrentConnections string
```
alesstimec, line 47: why isn't the resolver field on the Config struct?

alesstimec (approved):
LGTM if you make the resolver a field on the Config type

internal/ssh/ssh.go
SimoneDutto: the resolver is not a config but a service needed to operate the ssh server. I would leave it as a separate argument.

Don't you agree?

PR #1523, opened 2025-01-14
feat: intercept ssh keyManager facade in model proxy

kian99:
## Description
Currently this PR builds on #1521.

This PR allows JIMM to intercept model calls for the SSH KeyManager facade and direct the calls to JIMM's business logic rather than forwarding the calls to the backing Juju controllers.

This is being done in order to support the SSH proxy efforts.
Ideally, in Juju 4 these methods would be done on the controller api and all this logic would move into the jujuapi package. I've left comments to that effect. If before that point we want to refactor the model proxy then the logic has been cleanly split so that the code you see inside `internal/rpcproxy/keymanager.go` is all the logic/code that would go inside of `jujuapi`.

I've manually tested the changes by spinning up JIMM and adding a Juju 3.5 controller and tried the following commands (not in this specific order):
```
juju list-ssh-keys
juju list-ssh-keys --full
juju add-ssh-key "$(cat somekey.pub)"
juju remove-ssh-key <comment>
juju remove-ssh-key <fingerprint>
```

Fixes [JUJU-7346](https://warthogs.atlassian.net/browse/JUJU-7346)

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

### DIFF
--- a/internal/rpcproxy/keymanager.go
+++ b/internal/rpcproxy/keymanager.go
@@ -28,6 +28,8 @@ type keyManagerFacade struct {
 	user *openfga.User
 }
 
+// ListKeys lists the authenticated user's SSH keys
+// in the format defined in args.
 func (s *keyManagerFacade) ListKeys(ctx context.Context, args jujuparams.ListSSHKeys) (jujuparams.StringsResults, error) {
 	keys, err := s.ListUserPublicKeys(ctx, s.user)
 	if err != nil {
@@ -52,6 +54,8 @@ func (s *keyManagerFacade) ListKeys(ctx context.Context, args jujuparams.ListSSH
 	return jujuparams.StringsResults{Results: []jujuparams.StringsResult{res}}, nil
 }
 
+// AddKeys saves the SSH keys defined in args and associates them
+// with the authenticated user.
 func (s *keyManagerFacade) AddKeys(ctx context.Context, args jujuparams.ModifyUserSSHKeys) (jujuparams.ErrorResults, error) {
 	var res []jujuparams.ErrorResult
 	errF := func(err error, msg string) jujuparams.ErrorResult {
@@ -77,6 +81,8 @@ func (s *keyManagerFacade) AddKeys(ctx context.Context, args jujuparams.ModifyUs
 	return jujuparams.ErrorResults{Results: res}, nil
 }
 
+// DeleteKeys removes saved keys associated with the authenticated user
+// and finds keys to remove either by comment or fingerprint.
 func (s *keyManagerFacade) DeleteKeys(ctx context.Context, args jujuparams.ModifyUserSSHKeys) (jujuparams.ErrorResults, error) {
 	var res []jujuparams.ErrorResult
 	errF := func(err error, msg string) jujuparams.ErrorResult {
--- a/internal/rpcproxy/keymanager.go
+++ b/internal/rpcproxy/keymanager.go
@@ -13,6 +13,7 @@ import (
 	"github.com/juju/utils/v3/ssh"
 	gossh "golang.org/x/crypto/ssh"
 
+	"github.com/canonical/jimm/v3/internal/errors"
 	"github.com/canonical/jimm/v3/internal/jimm/sshkeys"
 	"github.com/canonical/jimm/v3/internal/openfga"
 )
@@ -60,6 +61,7 @@ func (s *keyManagerFacade) AddKeys(ctx context.Context, args jujuparams.ModifyUs
 	var res []jujuparams.ErrorResult
 	errF := func(err error, msg string) jujuparams.ErrorResult {
 		return jujuparams.ErrorResult{Error: &jujuparams.Error{
+			Code:    string(errors.ErrorCode(err)),
 			Message: fmt.Sprintf("%s: %s", msg, err.Error()),
 		}}
 	}
@@ -87,6 +89,7 @@ func (s *keyManagerFacade) DeleteKeys(ctx context.Context, args jujuparams.Modif
 	var res []jujuparams.ErrorResult
 	errF := func(err error, msg string) jujuparams.ErrorResult {
 		return jujuparams.ErrorResult{Error: &jujuparams.Error{
+			Code:    string(errors.ErrorCode(err)),
 			Message: fmt.Sprintf("%s: %s", msg, err.Error()),
 		}}
 	}
--- a/internal/rpcproxy/keymanager.go
+++ b/internal/rpcproxy/keymanager.go
@@ -25,14 +25,14 @@ var isFingerprintRegexp = regexp.MustCompile("^[0-9a-f]{2}(:[0-9a-f]{2}){15}$")
 // the api layer logic for SSH key management methods that are currently
 // used by the rpcProxy.
 type keyManagerFacade struct {
-	SSHKeyManager
-	user *openfga.User
+	keyManager SSHKeyManager
+	user       *openfga.User
 }
 
 // ListKeys lists the authenticated user's SSH keys
 // in the format defined in args.
 func (s *keyManagerFacade) ListKeys(ctx context.Context, args jujuparams.ListSSHKeys) (jujuparams.StringsResults, error) {
-	keys, err := s.ListUserPublicKeys(ctx, s.user)
+	keys, err := s.keyManager.ListUserPublicKeys(ctx, s.user)
 	if err != nil {
 		return jujuparams.StringsResults{}, err
 	}
@@ -75,7 +75,7 @@ func (s *keyManagerFacade) AddKeys(ctx context.Context, args jujuparams.ModifyUs
 			PublicKey: out,
 			Comment:   comment,
 		}
-		if err := s.AddUserPublicKey(ctx, s.user, parsedKey); err != nil {
+		if err := s.keyManager.AddUserPublicKey(ctx, s.user, parsedKey); err != nil {
 			res = append(res, errF(err, fmt.Sprintf("Failed to add key (comment %s)", comment)))
 		}
 	}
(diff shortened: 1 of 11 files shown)


internal/rpcproxy/rpcproxy.go
```diff
@@ -710,3 +746,63 @@ func (p *clientProxy) handleAdminFacade(ctx context.Context, msg *message) (clie
 		return nil, nil, nil
 	}
 }
+
+// handleKeyManagerFacade processes the key manager facade call.
+func (p *clientProxy) handleKeyManagerFacade(ctx context.Context, msg *message) (clientResponse *message, err error) {
+	if p.user == nil {
+		return nil, errors.E("user not authenticated")
```
alesstimec, line 753: should we not return the error in `clientResponse`? i.e. fill in the `Error` and `ErrorCode` fields.

internal/rpcproxy/sshkeys.go
```diff
@@ -0,0 +1,123 @@
...
+// keyManagerFacade is intended to be a temporary struct used to emulate logic
+// that will eventually live in jujuapi. This struct contains all
+// the api layer logic for SSH key management methods that are currently
+// used by the rpcProxy.
+type keyManagerFacade struct {
+	SSHKeyManager
+	user *openfga.User
+}
+
+func (s *keyManagerFacade) ListKeys(ctx context.Context, args jujuparams.ListSSHKeys) (jujuparams.StringsResults, error) {
```
alesstimec: godoc, please

internal/rpcproxy/sshkeys.go
```diff
@@ -0,0 +1,123 @@
...
+	keys, err := s.ListUserPublicKeys(ctx, s.user)
+	if err != nil {
+		return jujuparams.StringsResults{}, err
+	}
+
+	var formatter func(key sshkeys.PublicKey) string
```
alesstimec: is this what juju does? it seems super suspicious.. i would expect formatting to happen in the cmd package

internal/rpcproxy/sshkeys.go
```diff
@@ -0,0 +1,123 @@
...
+
+	// The Juju CLI takes the first element from the slice of stringsResults.
+	res := jujuparams.StringsResult{}
+	for _, key := range keys {
+		res.Result = append(res.Result, formatter(key))
+	}
+	return jujuparams.StringsResults{Results: []jujuparams.StringsResult{res}}, nil
+}
+
+func (s *keyManagerFacade) AddKeys(ctx context.Context, args jujuparams.ModifyUserSSHKeys) (jujuparams.ErrorResults, error) {
```
alesstimec: godoc, please

internal/rpcproxy/sshkeys.go
```diff
@@ -0,0 +1,123 @@
+// Copyright 2025 Canonical.
```
alesstimec: could we rename this file to keymanagerfacade.go?

internal/rpcproxy/sshkeys.go
```diff
@@ -0,0 +1,123 @@
...
+		}
+		if err := s.AddUserPublicKey(ctx, s.user, jimmKey); err != nil {
+			res = append(res, errF(err, fmt.Sprintf("Failed to add key (comment %s)", comment)))
+		}
+	}
+
+	return jujuparams.ErrorResults{Results: res}, nil
+}
+
+func (s *keyManagerFacade) DeleteKeys(ctx context.Context, args jujuparams.ModifyUserSSHKeys) (jujuparams.ErrorResults, error) {
```
alesstimec: godoc, please

internal/rpcproxy/rpcproxy.go
kian99: That gets handled outside this function

internal/rpcproxy/sshkeys.go
kian99: What about just `keymanager.go`? Changing it to that for now.

internal/rpcproxy/sshkeys.go
kian99: Yes it's all done server side - https://github.com/juju/juju/blob/3.6/apiserver/facades/client/keymanager/keymanager.go#L111
It seems to be like this because of legacy reasons, i.e. the return type is a list of strings rather than some well defined type that the client can then format.

internal/rpcproxy/keymanager.go
```diff
@@ -0,0 +1,129 @@
...
+		res.Result = append(res.Result, formatter(key))
+	}
+	return jujuparams.StringsResults{Results: []jujuparams.StringsResult{res}}, nil
+}
+
+// AddKeys saves the SSH keys defined in args and associates them
+// with the authenticated user.
+func (s *keyManagerFacade) AddKeys(ctx context.Context, args jujuparams.ModifyUserSSHKeys) (jujuparams.ErrorResults, error) {
+	var res []jujuparams.ErrorResult
+	errF := func(err error, msg string) jujuparams.ErrorResult {
```
alesstimec, line 62: suggestion: just have `func(msg string) jujuparams.ErrorResult`.. you are already creating a message string every time you call `errF`, so appending the `err.Error()` each time, would be a minor change.

internal/rpcproxy/keymanager.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	return jujuparams.ErrorResults{Results: res}, nil
+}
+
+// DeleteKeys removes saved keys associated with the authenticated user
+// and finds keys to remove either by comment or fingerprint.
+func (s *keyManagerFacade) DeleteKeys(ctx context.Context, args jujuparams.ModifyUserSSHKeys) (jujuparams.ErrorResults, error) {
+	var res []jujuparams.ErrorResult
+	errF := func(err error, msg string) jujuparams.ErrorResult {
```
alesstimec, line 90: same suggestion

alesstimec (approved):

internal/rpcproxy/keymanager.go
kian99: Good point, as I was changing it I noticed `jujuparams.Error` also accepts an error code. If I want to fill that in I need to keep the error in the signature. I'm now populating the error code, so I've mostly left things the same, let me know what you think.

alesstimec: then perhaps 
```
func(msg, errorCode string) jujuparams.ErrorResult {
  		return jujuparams.ErrorResult{Error: &jujuparams.Error{
			Code:    string(errors.ErrorCode(err)),
			Message: msg,
		}}
}
```
but just a suggestion, i won't insist on it..

alesstimec (approved):

internal/rpcproxy/keymanager.go
```diff
@@ -0,0 +1,132 @@
...
+)
+
+var isFingerprintRegexp = regexp.MustCompile("^[0-9a-f]{2}(:[0-9a-f]{2}){15}$")
+
+// keyManagerFacade is intended to be a temporary struct used to emulate logic
+// that will eventually live in jujuapi. This struct contains all
+// the api layer logic for SSH key management methods that are currently
+// used by the rpcProxy.
+type keyManagerFacade struct {
+	SSHKeyManager
```
ale8k: Why is this embedded field? I often prefer named fields so it's really clear what struct we're calling methods on.

internal/rpcproxy/keymanager.go
```diff
@@ -0,0 +1,132 @@
...
+			Message: fmt.Sprintf("%s: %s", msg, err.Error()),
+		}}
+	}
+
+	for i, key := range args.Keys {
+		out, comment, _, _, err := gossh.ParseAuthorizedKey([]byte(key))
+		if err != nil {
+			res = append(res, errF(err, fmt.Sprintf("Failed to parse key (entry %d)", i)))
+		}
+		jimmKey := sshkeys.PublicKey{
```
ale8k: parsedKey* maybe a better name?

internal/rpcproxy/keymanager.go
```diff
@@ -0,0 +1,132 @@
...
+			}
+		}
+	}
+
+	return jujuparams.ErrorResults{Results: res}, nil
+}
+
+func marshalAuthorizedKeyWithComment(key sshkeys.PublicKey) string {
+	// Copied from gossh.MarshalAuthorizedKey with an addition for the comment.
+	// Errors from the buffer's Write..() methods are always nil.
```
ale8k, line 118: Move this up to the actual godoc please, also I've forogtten - what is the comment for?

internal/rpcproxy/keymanager.go
```diff
@@ -0,0 +1,132 @@
...
+	e := base64.NewEncoder(base64.StdEncoding, b)
+	_, _ = e.Write(key.Marshal())
+	e.Close()
+	b.WriteByte(' ')
+	b.WriteString(key.Comment)
+	b.WriteByte('\n')
+	return b.String()
+}
+
+func fingerprintWithComment(key sshkeys.PublicKey) string {
```
ale8k, line 135: Godoc again explaining why we need a comment would be good

internal/rpcproxy/rpcproxy_test.go
```diff
@@ -272,3 +280,139 @@ func TestProxySocketsAuditLogs(t *testing.T) {
 	c.Assert(auditLogs, qt.DeepEquals, expectedEvents)
 
 }
+
+func TestProxySocketsSSHKeys(t *testing.T) {
```
ale8k, line 288: A godoc for this test would be useful

internal/rpcproxy/keymanager.go
kian99: It's minor so I've left it as is.

ale8k (approved):
lgtm some minor

internal/rpcproxy/keymanager.go
kian99: An SSH key doesn't have a comment, but the OpenSSH authorized key format for public keys does and Juju stores the key comment and displays it back to the user when they list keys, helping to identify for a user where the key came from.

internal/rpcproxy/keymanager.go
kian99: Good catch, I'm not sure either, it's not necessary to be embedded, I've changed it.

PR #1524, opened 2025-01-15
implement auth in jump server

SimoneDutto:
## Description
In this PR we implement the authentication and authorization checks inside of the ssh jump server.
To do so:
- we require an ssh manager, able to interact with Jimm's state
- a new method in the ssh key manager to verify a public key fingerprint (the implementation will be done in a follow-up pr)

### DIFF
--- a/internal/jimm/ssh/ssh.go
+++ b/internal/jimm/ssh/ssh.go
@@ -1,0 +1,48 @@
+// Copyright 2025 Canonical.
+
+package ssh
+
+import (
+	"context"
+
+	"github.com/canonical/jimm/v3/internal/dbmodel"
+	"github.com/canonical/jimm/v3/internal/errors"
+	"github.com/canonical/jimm/v3/internal/openfga"
+)
+
+type identityManager interface {
+	FetchIdentity(ctx context.Context, id string) (*openfga.User, error)
+}
+
+type modelManager interface {
+	GetModel(ctx context.Context, uuid string) (dbmodel.Model, error)
+}
+
+type sshKeyManager interface {
+	VerifyPublicKey(ctx context.Context, user *openfga.User, fingerprint string) (bool, error)
+}
+
+// sshManager provides a means to manage ssh server within JIMM.
+type sshManager struct {
+	modelManager
+	identityManager
+	sshKeyManager
+}
+
+// NewSSHManager returns a new SSHManager that offers jimm functionality to the SSHJumpServer.
+func NewSSHManager(identityManager identityManager, modelManager modelManager, sshKeyManager sshKeyManager) (*sshManager, error) {
+	if identityManager == nil {
+		return nil, errors.E("identityManager cannot be nil")
+	}
+	if modelManager == nil {
+		return nil, errors.E("modelManager cannot be nil")
+	}
+	if sshKeyManager == nil {
+		return nil, errors.E("sshManager cannot be nil")
+	}
+	return &sshManager{
+		modelManager:    modelManager,
+		identityManager: identityManager,
+		sshKeyManager:   sshKeyManager,
+	}, nil
+}
--- a/internal/jimm/ssh/ssh.go
+++ b/internal/jimm/ssh/ssh.go
@@ -10,27 +10,30 @@ import (
 	"github.com/canonical/jimm/v3/internal/openfga"
 )
 
-type identityManager interface {
+// IdentityManager provides a means to fetch an identity from the identity service.
+type IdentityManager interface {
 	FetchIdentity(ctx context.Context, id string) (*openfga.User, error)
 }
 
-type modelManager interface {
+// ModelManager provides a means to fetch a model from the model service.
+type ModelManager interface {
 	GetModel(ctx context.Context, uuid string) (dbmodel.Model, error)
 }
 
-type sshKeyManager interface {
-	VerifyPublicKey(ctx context.Context, user *openfga.User, fingerprint string) (bool, error)
+// SSHKeyManager provides a means to manage ssh keys within JIMM.
+type SSHKeyManager interface {
+	VerifyPublicKey(ctx context.Context, claimUser string, publicKey []byte) (bool, error)
 }
 
 // sshManager provides a means to manage ssh server within JIMM.
 type sshManager struct {
-	modelManager
-	identityManager
-	sshKeyManager
+	ModelManager
+	IdentityManager
+	SSHKeyManager
 }
 
 // NewSSHManager returns a new SSHManager that offers jimm functionality to the SSHJumpServer.
-func NewSSHManager(identityManager identityManager, modelManager modelManager, sshKeyManager sshKeyManager) (*sshManager, error) {
+func NewSSHManager(identityManager IdentityManager, modelManager ModelManager, sshKeyManager SSHKeyManager) (*sshManager, error) {
 	if identityManager == nil {
 		return nil, errors.E("identityManager cannot be nil")
 	}
@@ -41,8 +44,8 @@ func NewSSHManager(identityManager identityManager, modelManager modelManager, s
 		return nil, errors.E("sshManager cannot be nil")
 	}
 	return &sshManager{
-		modelManager:    modelManager,
-		identityManager: identityManager,
-		sshKeyManager:   sshKeyManager,
+		ModelManager:    modelManager,
+		IdentityManager: identityManager,
+		SSHKeyManager:   sshKeyManager,
 	}, nil
 }
--- a/internal/jimm/ssh/ssh.go
+++ b/internal/jimm/ssh/ssh.go
@@ -53,6 +53,8 @@ type sshManager struct {
 	sshKeyManager   SSHKeyManager
 }
 
+// PublicKeyHandler is the method to verify the public key of the user. It first checks for the public key and then fetches the user.
+// It returns a user if successful.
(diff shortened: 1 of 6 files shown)


internal/jimm/ssh/ssh.go
```diff
@@ -0,0 +1,48 @@
...
+type identityManager interface {
+	FetchIdentity(ctx context.Context, id string) (*openfga.User, error)
+}
+
+type modelManager interface {
+	GetModel(ctx context.Context, uuid string) (dbmodel.Model, error)
+}
+
+type sshKeyManager interface {
+	VerifyPublicKey(ctx context.Context, user *openfga.User, fingerprint string) (bool, error)
```
kian99: First thought here is that the `user` should rather be something like `userClaim string`. This is the user they claim to be

internal/ssh/ssh.go
```diff
@@ -132,21 +160,38 @@ func directTCPIPHandler(resolver Resolver) func(srv *ssh.Server, conn *gossh.Ser
...
-				rejectConnectionAndLogError(ctx, newChan, "Failed to copy data from dst to src", err)
+				rejectConnectionAndLogError(ctx, newChan, "failed to copy data from dst to src", err)
 			}
 		}()
 		zapctx.Info(ctx, fmt.Sprintf("Proxying connection from %s:%d to %s:%d \n", d.SrcAddr, d.SrcPort, d.DestAddr, d.DestPort))
 	}
 }
 
+// fetchAndVerifySSHUser extracts the user from the context and checks the user has permission to ssh.
+func fetchAndVerifySSHUser(ctx ssh.Context, modelTag names.ModelTag) (*openfga.User, error) {
```
```suggestion
func fetchAndAuthorizeUser(ctx ssh.Context, modelTag names.ModelTag) (*openfga.User, error) {
```

kian99 (commented):
quick review, need to go over it again.

internal/jimm/ssh/ssh.go
SimoneDutto: it's fine that it's a `user` because we've already verify that the user is in our db by using the `FetchIdentity` with the `ctx.User()`

kian99: As discussed, let's opt to authenticate the user before we create an openfga.User object.

internal/jimm/ssh/ssh.go
```diff
@@ -0,0 +1,48 @@
...
+
+import (
+	"context"
+
+	"github.com/canonical/jimm/v3/internal/dbmodel"
+	"github.com/canonical/jimm/v3/internal/errors"
+	"github.com/canonical/jimm/v3/internal/openfga"
+)
+
+type identityManager interface {
```
kian99: We should export these types to indicate (at least to ourselves) that these interfaces form part of the API of our package and external packages rely on their definition.

internal/ssh/ssh.go
```diff
@@ -19,10 +20,18 @@ import (
...
+type SSHManager interface {
 	// AddrFromModelUUID is the method to resolve the address of the controller to contact given the model UUID.
-	AddrFromModelUUID(ctx context.Context, user openfga.User, modelUUID string) (string, error)
+	AddrFromModelUUID(ctx context.Context, user *openfga.User, modelTag names.ModelTag) (string, error)
+
+	// FetchIdentity
+	FetchIdentity(ctx context.Context, id string) (*openfga.User, error)
+
+	// VerifyPublicKey verifies the identityName is
+	VerifyPublicKey(ctx context.Context, user *openfga.User, fingerprint string) (bool, error)
```
kian99: I think we mentioned that we could discuss this in a separate PR, but verifying the public key, arguably should be done by using the raw key rather than the fingerprint.

internal/ssh/ssh.go
```diff
@@ -48,23 +57,33 @@ type Config struct {
...
 	}
 	server := Server{
 		Server: &ssh.Server{
 			Addr: fmt.Sprintf(":%s", config.Port),
 			ChannelHandlers: map[string]ssh.ChannelHandler{
-				"direct-tcpip": directTCPIPHandler(resolver),
+				"direct-tcpip": directTCPIPHandler(sshManager),
 			},
 			PublicKeyHandler: func(ctx ssh.Context, key ssh.PublicKey) bool {
+				user, err := sshManager.FetchIdentity(ctx, ctx.User())
```
kian99: Again here, I'd suggest first authenticating the user, then fetching their identity so that we don't hold an openfga.User object before the user has gone through authentication. Aiming to reduce the potential for security vulns.

internal/testutils/jimmtest/mocks/jimm_ssh_mock.go
```diff
@@ -0,0 +1,39 @@
...
+// SSHManager is an implementation of the sshManager interface.
+type SSHManager struct {
+	AddrFromModelUUID_ func(ctx context.Context, user *openfga.User, modelTag names.ModelTag) (string, error)
+	FetchIdentity_     func(ctx context.Context, id string) (*openfga.User, error)
+	VerifyPublicKey_   func(ctx context.Context, user *openfga.User, fingerprint string) (bool, error)
+}
+
+func (j SSHManager) FetchIdentity(ctx context.Context, id string) (*openfga.User, error) {
+	if j.FetchIdentity_ == nil {
+		return nil, nil
```
kian99: These errors should be "NotImplemented" errors.

internal/jimm/ssh/ssh.go
```diff
@@ -0,0 +1,51 @@
...
+
+// sshManager provides a means to manage ssh server within JIMM.
+type sshManager struct {
+	ModelManager
+	IdentityManager
+	SSHKeyManager
+}
+
+// NewSSHManager returns a new SSHManager that offers jimm functionality to the SSHJumpServer.
+func NewSSHManager(identityManager IdentityManager, modelManager ModelManager, sshKeyManager SSHKeyManager) (*sshManager, error) {
```
ale8k, line 32: This sshManager struct doesn't actually do anything other than wrap other managers, why do we need it?

internal/ssh/ssh.go
```diff
@@ -33,11 +42,11 @@ type forwardMessage struct {
 	SrcPort  uint32
 }
 
-// Server is the custom struct to embed the gliderlabs.ssh server and a resolver.
+// Server is the custom struct to embed the gliderlabs.ssh server and a sshManager.
 type Server struct {
 	*ssh.Server
 
-	resolver Resolver
+	sshManager SSHManager
```
ale8k: I think just put the other managers here scoped by interfaces?

ale8k (commented):
confused by the ssh manager, will wait for update

SimoneDutto:
> confused by the ssh manager, will wait for update

as discussed, the manager methods' implementation will be done in a followup. This is just the skeleton to satisfy the interface for the creation of the jump server

kian99 (approved):
lgtm, I'm okay with the sshManager wrapping the others with the assumption that it will grow to do more logic in the future. If it turns out not to, we can get rid of it.

SimoneDutto:
> lgtm, I'm okay with the sshManager wrapping the others with the assumption that it will grow to do more logic in the future. If it turns out not to, we can get rid of it.

yeah, same

internal/ssh/ssh.go
```diff
@@ -48,23 +57,34 @@ type Config struct {
...
 	server := Server{
 		Server: &ssh.Server{
 			Addr: fmt.Sprintf(":%s", config.Port),
 			ChannelHandlers: map[string]ssh.ChannelHandler{
-				"direct-tcpip": directTCPIPHandler(resolver),
+				"direct-tcpip": directTCPIPHandler(sshManager),
 			},
 			PublicKeyHandler: func(ctx ssh.Context, key ssh.PublicKey) bool {
+				claimUser := ctx.User()
+				if ok, err := sshManager.VerifyPublicKey(ctx, claimUser, key.Marshal()); !ok || err != nil {
```
ale8k: This logic could live purely on the ssh manager struct and be nicely encapsulated, easier to test too then.

also shouldn't it really be called sshauthoriser? it's not managing anything

internal/jimm/ssh/ssh.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+// sshManager provides a means to manage ssh server within JIMM.
+type sshManager struct {
+	modelManager    ModelManager
+	identityManager IdentityManager
+	sshKeyManager   SSHKeyManager
+}
+
+func (s *sshManager) PublicKeyHandler(ctx context.Context, claimUser string, key []byte) (*openfga.User, error) {
```
ale8k, line 58: godoc

internal/ssh/ssh.go
```diff
@@ -19,10 +20,19 @@ import (
...
+	// PublicKeyHandler is the method to verify the public key of the user. It returns a user if successful.
+	PublicKeyHandler(ctx context.Context, claimUser string, key []byte) (*openfga.User, error)
+}
+
+// TODO(simonedutto): this is going to change to reuse as much as our dial logic as we possibly can.
+// SSHResolver is the interface to resolve controller's addresses.
+type SSHResolver interface {
 	// AddrFromModelUUID is the method to resolve the address of the controller to contact given the model UUID.
-	AddrFromModelUUID(ctx context.Context, user openfga.User, modelUUID string) (string, error)
+	AddrFromModelUUID(ctx context.Context, user *openfga.User, modelTag names.ModelTag) (string, error)
```
ale8k, line 35: Nice

ale8k (approved):
lgtm just one godoc

PR #1525, opened 2025-01-17
chore: move model builder to separate file

kian99:
## Description
This PR is a mechanical change to move the model builder struct to a separate file so that future work to create a model manager which result in a smaller diff. The code has been copied verbatim. The model builder is an internal type so it has no "direct" tests (tested through model_test.go) and therefore there is no "modelbuilder_test.go".


internal/jimm/modelbuilder.go
```diff
@@ -0,0 +1,486 @@
+// Copyright 2025 Canonical.
+package jimm
```
alesstimec, line 3: new line, please

internal/jimm/modelbuilder.go
```diff
@@ -0,0 +1,486 @@
...
+	}
+	// the model should be deleted from the database regardless of the request
+	// context expiration
+	ctx := context.Background()
+	if derr := b.jimm.Database.DeleteModel(ctx, b.model); derr != nil {
+		zapctx.Error(ctx, "failed to delete model", zap.String("model", b.model.Name), zap.String("owner", b.model.Owner.Name), zaputil.Error(derr))
+	}
+}
+
+func (b *modelBuilder) UpdateDatabaseModel() *modelBuilder {
```
alesstimec, line 324: godoc, please

alesstimec (approved):
LGTM with a few nitpicks

ale8k (approved):
lgtm

PR #1526, opened 2025-01-17
feat: add model manager

kian99:
## Description
The model manager package in internal/jimm is the first for a Juju related domain that moves logic away from the large jimm struct into a dedicated package. This PR shows that the model manager relies on bits from what will eventually be the cloud and controller managers and also requires a way to dial Juju controllers.

This change is incomplete because I'd like the team to get a feel for whether they enjoy this approach of a smaller package with more dependencies OR whether they'd prefer a larger (i.e. single) Juju package that has less dependencies.

Fixes [JUJU-7383](https://warthogs.atlassian.net/browse/JUJU-7383)

## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests

[JUJU-7383]: https://warthogs.atlassian.net/browse/JUJU-7383?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/jimm/controller_test.go
+++ b/internal/jimm/controller_test.go
@@ -21,6 +21,7 @@ import (
 	"github.com/juju/juju/state"
 	"github.com/juju/names/v5"
 	semversion "github.com/juju/version"
+	"github.com/juju/version/v2"
 	"gopkg.in/macaroon.v2"
 
 	"github.com/canonical/jimm/v3/internal/dbmodel"
@@ -1469,6 +1470,12 @@ func TestInitiateMigration(t *testing.T) {
 			}
 		})
 	}
+}
+
+// newVersion wraps version.MustParse to return a *version.Number
+func newVersion(s string) *version.Number {
+	n := version.MustParse(s)
+	return &n
 }
 
 type result struct {
--- a/internal/jimm/juju/model/export_test.go
+++ b/internal/jimm/juju/model/export_test.go
@@ -1,0 +1,4 @@
+package model
+
+// ModelManager is a type alias to export modelManager for use in tests.
+type ModelManager = modelManager
--- a/internal/jimm/juju/model/model.go
+++ b/internal/jimm/juju/model/model.go
@@ -1,6 +1,6 @@
 // Copyright 2025 Canonical.
 
-package jimm
+package model
 
 import (
 	"context"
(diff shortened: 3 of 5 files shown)


internal/jimm/controller_test.go
```diff
@@ -1471,6 +1472,12 @@ func TestInitiateMigration(t *testing.T) {
 	}
 }
 
+// newVersion wraps version.MustParse to return a *version.Number
+func newVersion(s string) *version.Number {
```
kian99, line 1476: This test function was previously inside model_test.go so I've copied it here into controller_test.go

internal/jimm/juju/model/model.go
```diff
@@ -320,7 +433,7 @@ func (j *JIMM) ListModelSummaries(ctx context.Context, user *openfga.User, maski
 	}
 
 	// we query the model summaries for each controller
-	err = j.forEachController(ctx, uniqueControllers, func(c *dbmodel.Controller, a API) error {
+	err = mm.controllerManager.ForEachController(ctx, uniqueControllers, func(c *dbmodel.Controller, a API) error {
```
kian99, line 436: Alternatively to this, we could have the ForEachController method copy/pasted into the model manager. Or we could have a package like `utils` that provides shared functionality.

internal/jimm/juju/model/export_test.go
```diff
@@ -0,0 +1,4 @@
+package model
```
alesstimec, line 1: missing copyright

internal/jimm/juju/model/modelbuilder.go
```diff
@@ -1,35 +1,31 @@
-// Copyright 2025 Canonical.
-
-package jimm
+package model
```
alesstimec, line 1: missing copyright

internal/jimm/juju/model/modelbuilder.go
```diff
@@ -1,35 +1,31 @@
...
 
 import (
 	"context"
 	"fmt"
 
-	jujupermission "github.com/juju/juju/core/permission"
+	"github.com/canonical/jimm/v3/internal/dbmodel"
+	"github.com/canonical/jimm/v3/internal/errors"
+	"github.com/canonical/jimm/v3/internal/openfga"
```
alesstimec, line 9: imports need sorting

alesstimec (commented):
linter seems to be complaining

SimoneDutto:
i like this change, but i'm wondering if we will fall in circular dependency where we need to c/p code to move forward. I guess you are the one that worked the most on these managers. What do you think?

PR #1527, opened 2025-01-17
implement limit connections

SimoneDutto:
## Description
This is an implementation of the MaxConcurrentRequest for the ssh server.
The implementation is a copy/paste of https://cs.opensource.google/go/x/net/+/refs/tags/v0.34.0:netutil/listen.go with an additional timeout.
In fact the LimitListener hangs forever the `Accept()` until a seat is made available, making it hard to test and also counter intuitive, because we want to close connection when we reach the limit.

The error from the SSH client is:
image

> [!NOTE]  
> This is the first simple implementation until we get a better way of doing it by interacting with the gliderlab/ssh server. I am working on a issue + pr to make it possible.

## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests

### DIFF
--- a/internal/ssh/listener.go
+++ b/internal/ssh/listener.go
@@ -1,4 +1,35 @@
-// Copyright 2025 Canonical.
+// nolint:goheader
+// Copyright 2009 The Go Authors.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are
+// met:
+//
+//   - Redistributions of source code must retain the above copyright
+//
+// notice, this list of conditions and the following disclaimer.
+//   - Redistributions in binary form must reproduce the above
+//
+// copyright notice, this list of conditions and the following disclaimer
+// in the documentation and/or other materials provided with the
+// distribution.
+//   - Neither the name of Google LLC nor the names of its
+//
+// contributors may be used to endorse or promote products derived from
+// this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+// "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+// LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+// A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+// OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+// SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+// LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+// DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+// THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
 package ssh
 
 import (
--- a/internal/ssh/listener.go
+++ b/internal/ssh/listener.go
@@ -4,19 +4,15 @@
 // Redistribution and use in source and binary forms, with or without
 // modification, are permitted provided that the following conditions are
 // met:
-//
 //   - Redistributions of source code must retain the above copyright
-//
-// notice, this list of conditions and the following disclaimer.
+//     notice, this list of conditions and the following disclaimer.
 //   - Redistributions in binary form must reproduce the above
-//
-// copyright notice, this list of conditions and the following disclaimer
-// in the documentation and/or other materials provided with the
-// distribution.
+//     copyright notice, this list of conditions and the following disclaimer
+// 	   in the documentation and/or other materials provided with the
+//     distribution.
 //   - Neither the name of Google LLC nor the names of its
-//
-// contributors may be used to endorse or promote products derived from
-// this software without specific prior written permission.
+//     contributors may be used to endorse or promote products derived from
+//     this software without specific prior written permission.
 //
 // THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
 // "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
@@ -42,10 +38,10 @@ import (
 // This is a copypaste of netutil.LimiLister (link: https://cs.opensource.google/go/x/net/+/refs/tags/v0.34.0:netutil/listen.go),
 // but we add a timeout so when we are at the limit we actively close connections instead of waiting indefinetely. (Look at line 44)
 
-// LimitListenerWithTimeout returns a Listener that accepts at most n simultaneous
+// limitListenerWithTimeout returns a Listener that accepts at most n simultaneous
 // connections from the provided Listener, and it timeouts when the max
 // has been reached and no seats has been freed for the timeout period.
-func LimitListenerWithTimeout(l net.Listener, n int, timeout time.Duration) net.Listener {
+func limitListenerWithTimeout(l net.Listener, n int, timeout time.Duration) net.Listener {
 	return &limitListener{
 		Listener: l,
 		sem:      make(chan struct{}, n),
--- a/internal/ssh/listener.go
+++ b/internal/ssh/listener.go
@@ -78,6 +78,7 @@ func (l *limitListener) acquire() bool {
 }
 func (l *limitListener) release() { <-l.sem }
 
+// Accept waits for and returns the next connection to the listener, by checking the semaphore and the timeout.
 func (l *limitListener) Accept() (net.Conn, error) {
 	if !l.acquire() {
 		// If the semaphore isn't acquired because the listener was closed, expect
--- a/internal/ssh/listener.go
+++ b/internal/ssh/listener.go
@@ -1,0 +1,93 @@
+// Copyright 2025 Canonical.
+package ssh
+
+import (
+	"net"
+	"sync"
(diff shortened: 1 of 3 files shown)


internal/ssh/listener.go
```diff
@@ -0,0 +1,93 @@
...
+package ssh
+
+import (
+	"net"
+	"sync"
+	"time"
+)
+
+// N.B.:
+// This is a copypaste of netutil.LimiLister, but we add a timeout so when we are at the limit
```
kian99: I'd suggest adding a link to the source

internal/ssh/ssh.go
```diff
@@ -47,40 +49,68 @@ type forwardMessage struct {
 type Config struct {
 	Port                     string
 	HostKey                  []byte
-	MaxConcurrentConnections string
+	MaxConcurrentConnections int
+	AcceptConnectionTimeout  time.Duration
+}
+
+type Server struct {
+	*ssh.Server
```
kian99, line 59: There'll be opinions on embedding vs not embedding. If you don't _need_ to embed then it's usually clearer to avoid it imo.

internal/ssh/ssh.go
```diff
@@ -47,40 +49,68 @@ type forwardMessage struct {
...
 	}
 	server.AddHostKey(hostKey)
 
 	return server, nil
 }
 
+// ListenAndServe create a LimitListenerWithTimeout and Serve requests.
+func (srv Server) ListenAndServe() error {
+	ln, err := net.Listen("tcp", srv.Addr)
+	if srv.MaxConcurrentConnections == 0 {
```
kian99: I think these defaults should be set elsewhere. I'm thinking inside `NewJumpServer` inside a `SetConfigDefaults` functions. Or possibly even better, before we get to the ssh package.

internal/ssh/ssh_test.go
```diff
@@ -269,6 +270,36 @@ func (s *sshSuite) TestSSHFinalDestinationDialFail(c *qt.C) {
...
+			HostKeyCallback: gossh.FixedHostKey(s.hostKey.PublicKey()),
+			Auth: []gossh.AuthMethod{
+				gossh.PublicKeys(s.privateKey),
+			},
+			User: "alice",
+		})
+		c.Check(err, qt.IsNil)
+		clients = append(clients, client)
+	}
+	// this connection is sent when we are at maximum connection>
```
kian99: ?
```suggestion
	// this connection is dropped when we are at maximum connection>
```

internal/ssh/ssh.go
SimoneDutto: i like the setconfigdefault inside of this pkg. Tbf i'd expect the charm code to handle these defaults, so this is a fail-safe

internal/ssh/ssh.go
SimoneDutto: i think here is best to embed it because we are overriding the listenandserver with our custom listenandserver

internal/ssh/ssh.go
```diff
@@ -95,6 +96,20 @@ func NewJumpServer(ctx context.Context, config Config, sshAuthorizer SSHAuthoriz
 	return server, nil
 }
 
+// setConfigDefaults sets the default values for the configuration.
+func setConfigDefaults(config Config) Config {
+	if config.Port == "" {
+		config.Port = fmt.Sprint(jujuSSHDefaultPort)
+	}
+	if config.MaxConcurrentConnections == 0 {
```
```suggestion
	if config.MaxConcurrentConnections <= 0 {
```

internal/ssh/ssh.go
```diff
@@ -95,6 +96,20 @@ func NewJumpServer(ctx context.Context, config Config, sshAuthorizer SSHAuthoriz
...
+		config.MaxConcurrentConnections = 100
+	}
+	if config.AcceptConnectionTimeout == 0 {
```
```suggestion
	if config.AcceptConnectionTimeout <= 0 {
```

kian99 (approved):

internal/ssh/listener.go
```diff
@@ -0,0 +1,93 @@
+// Copyright 2025 Canonical.
+package ssh
```
alesstimec, line 29: you know

internal/ssh/listener.go
```diff
@@ -0,0 +1,93 @@
...
+// but we add a timeout so when we are at the limit we actively close connections instead of waiting indefinetely. (Look at line 44)
+
+// LimitListenerWithTimeout returns a Listener that accepts at most n simultaneous
+// connections from the provided Listener, and it timeouts when the max
+// has been reached and no seats has been freed for the timeout period.
+func LimitListenerWithTimeout(l net.Listener, n int, timeout time.Duration) net.Listener {
+	return &limitListener{
+		Listener: l,
+		sem:      make(chan struct{}, n),
+		done:     make(chan struct{}),
```
alesstimec, line 48: we could just use a context here instead of a channel.. modernize this code, so to speak

internal/ssh/listener.go
```diff
@@ -0,0 +1,93 @@
...
+	return err
+}
+
+type limitListenerConn struct {
+	net.Conn
+	releaseOnce sync.Once
+	release     func()
+}
+
+func (l *limitListenerConn) Close() error {
```
alesstimec, line 117: godoc

internal/ssh/listener.go
```diff
@@ -0,0 +1,93 @@
...
+
+	c, err := l.Listener.Accept()
+	if err != nil {
+		l.release()
+		return nil, err
+	}
+	return &limitListenerConn{Conn: c, release: l.release}, nil
+}
+
+func (l *limitListener) Close() error {
```
alesstimec, line 105: godoc

internal/ssh/listener.go
```diff
@@ -0,0 +1,93 @@
...
+	case l.sem <- struct{}{}:
+		return true
+	// we add a timeout here, so the connection is closed when the timeout has passed instead of waiting.
+	case <-time.After(l.timeout):
+		return false
+	}
+}
+func (l *limitListener) release() { <-l.sem }
+
+func (l *limitListener) Accept() (net.Conn, error) {
```
alesstimec, line 78: godoc

internal/ssh/listener.go
```diff
@@ -0,0 +1,125 @@
+// nolint:goheader
+// Copyright 2009 The Go Authors.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are
+// met:
+//
+//   - Redistributions of source code must retain the above copyright
+//
```
alesstimec: there seems to have been an issue with the copy&paste.. some new lines that should not be there..

internal/ssh/ssh.go
```diff
@@ -47,40 +49,83 @@ type forwardMessage struct {
 type Config struct {
 	Port                     string
 	HostKey                  []byte
-	MaxConcurrentConnections string
+	MaxConcurrentConnections int
+	AcceptConnectionTimeout  time.Duration
+}
+
+type Server struct {
```
alesstimec, line 58: godoc

internal/ssh/ssh.go
```diff
@@ -47,40 +49,83 @@ type forwardMessage struct {
...
+	*ssh.Server
+
+	MaxConcurrentConnections int
```
alesstimec: why are these fields exported?

internal/ssh/ssh.go
```diff
@@ -47,40 +49,83 @@ type forwardMessage struct {
...
+	AcceptConnectionTimeout  time.Duration
 }
 
 // NewJumpServer creates the jump server struct.
-func NewJumpServer(ctx context.Context, config Config, sshAuthorizer SSHAuthorizer, sshResolver SSHResolver) (*ssh.Server, error) {
+func NewJumpServer(ctx context.Context, config Config, sshAuthorizer SSHAuthorizer, sshResolver SSHResolver) (Server, error) {
```
alesstimec, line 66: why not return a pointer to Server?

internal/ssh/ssh.go
```diff
@@ -47,40 +49,83 @@ type forwardMessage struct {
...
 	return server, nil
 }
 
+// setConfigDefaults sets the default values for the configuration.
+func setConfigDefaults(config Config) Config {
+	if config.Port == "" {
+		config.Port = fmt.Sprint(jujuSSHDefaultPort)
+	}
+	if config.MaxConcurrentConnections <= 0 {
+		config.MaxConcurrentConnections = 100
```
alesstimec: add a const for this value, for consistency

internal/ssh/ssh.go
alesstimec: unexport these fields.. so the only way to sensibly create the server is via constructor, which calls setConfigDefaults.. then you can remove these checks.

internal/ssh/listener.go
```diff
@@ -0,0 +1,125 @@
...
+)
+
+// N.B.:
+// This is a copypaste of netutil.LimiLister (link: https://cs.opensource.google/go/x/net/+/refs/tags/v0.34.0:netutil/listen.go),
+// but we add a timeout so when we are at the limit we actively close connections instead of waiting indefinetely. (Look at line 44)
+
+// LimitListenerWithTimeout returns a Listener that accepts at most n simultaneous
+// connections from the provided Listener, and it timeouts when the max
+// has been reached and no seats has been freed for the timeout period.
+func LimitListenerWithTimeout(l net.Listener, n int, timeout time.Duration) net.Listener {
```
alesstimec: also, this does not need to be exported as it will only be used in this package

alesstimec (changes requested):

internal/ssh/ssh.go
SimoneDutto: here it makes no difference, the embedded ssh server is a pointer and there are two fields.

internal/ssh/ssh.go
```diff
@@ -47,40 +50,84 @@ type forwardMessage struct {
...
+	if config.AcceptConnectionTimeout <= 0 {
+		config.AcceptConnectionTimeout = defaultAcceptConnectionTimeout
+	}
+	return config
+}
+
+// ListenAndServe create a LimitListenerWithTimeout and Serve requests.
+func (srv Server) ListenAndServe() error {
+	ln, err := net.Listen("tcp", srv.Addr)
+	if srv.maxConcurrentConnections == 0 {
```
alesstimec: we don't need these checks anymore, do we? or they should live in the limitListenerWithTimeout function (which should probably just error out if max concurrent connections is 0 (OR less) or if accept connection timeout is 0 (OR less).

alesstimec (approved):
lgtm with one final suggestion

PR #1529, opened 2025-01-20
Juju 7464/reuse dial logic

SimoneDutto:
## Description
In this PR we reuse some logic from the rpc dial pkg to get addresses from Controller.

First commit is just extracting the GetAddressesAndTLSConfig from the Dial method and add some tests for it.

> I have decided to use a single SSHManager interface, because we don't even need two different interfaces for it. it's overengineering for 2 methods.

## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests

### DIFF
--- a/internal/jimm/ssh/ssh.go
+++ b/internal/jimm/ssh/ssh.go
@@ -59,12 +59,12 @@ type sshManager struct {
 func (s *sshManager) PublicKeyHandler(ctx context.Context, claimUser string, key []byte) (*openfga.User, error) {
 	zapctx.Info(ctx, "PublicKeyHandler")
 	if ok, err := s.sshKeyManager.VerifyPublicKey(ctx, claimUser, key); !ok || err != nil {
-		return nil, fmt.Errorf("cannot verify key for user %s: %s", claimUser, err.Error())
+		return nil, errors.E(err, "cannot verify key for user")
 	}
 	user, err := s.identityManager.FetchIdentity(ctx, claimUser)
 	if err != nil {
 		zapctx.Info(ctx, fmt.Sprintf("cannot find user %s", claimUser))
-		return nil, fmt.Errorf("cannot find user %s: %s", claimUser, err.Error())
+		return nil, errors.E(err, "cannot find user")
 	}
 	return user, nil
 }
@@ -75,12 +75,12 @@ func (s *sshManager) ResolveAddressesFromModelUUID(ctx context.Context, modelUUI
 
 	model, err := s.modelManager.GetModel(ctx, modelUUID)
 	if err != nil {
-		return nil, fmt.Errorf("cannot find model %s", modelUUID)
+		return nil, errors.E(err, "cannot find model")
 	}
 
 	addrs, _ := rpc.GetAddressesAndTLSConfig(ctx, &model.Controller)
 	if len(addrs) == 0 {
-		return nil, fmt.Errorf("cannot find addresses for model %s", modelUUID)
+		return nil, errors.E(err, "cannot find addresses for model's controller")
 	}
 
 	return addrs, nil
--- a/internal/jimm/ssh/ssh.go
+++ b/internal/jimm/ssh/ssh.go
@@ -11,6 +11,7 @@ import (
 	"github.com/canonical/jimm/v3/internal/dbmodel"
 	"github.com/canonical/jimm/v3/internal/errors"
 	"github.com/canonical/jimm/v3/internal/openfga"
+	"github.com/canonical/jimm/v3/internal/rpc"
 )
 
 // IdentityManager provides a means to fetch an identity from the identity service.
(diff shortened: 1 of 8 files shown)


internal/ssh/utils.go
kian99, line 1: I'd rename this file to just dial.go and move `rejectConnectionAndLogError` to the ssh.go file. While utils.go is fine when we don't know where to put something, if we can avoid having it at all in favour of more descriptive files, I'd vote for that.

internal/ssh/ssh.go
```diff
@@ -22,17 +21,13 @@ const juju_ssh_default_port = 17022
 
 type publicKeySSHUserKey struct{}
 
-// SSHAuthorizer is the interface to authorize users via public key.
-type SSHAuthorizer interface {
+// SSHManager is the interface to
```
kian99: docstring unfinished

kian99 (approved):
lgtm with 2 comments

internal/ssh/utils.go
SimoneDutto: sounds good

internal/jimm/ssh/ssh.go
```diff
@@ -67,3 +68,20 @@ func (s *sshManager) PublicKeyHandler(ctx context.Context, claimUser string, key
...
 	return user, nil
 }
+
+// ResolveAddressesFromModelUUID is the method to resolve the address of the controller to contact given the model UUID.
+func (s *sshManager) ResolveAddressesFromModelUUID(ctx context.Context, modelUUID string) ([]string, error) {
+	zapctx.Info(ctx, "ResolveAddressesFromModelUUID")
+
+	model, err := s.modelManager.GetModel(ctx, modelUUID)
+	if err != nil {
+		return nil, fmt.Errorf("cannot find model %s", modelUUID)
```
alesstimec: the returned error will tell us nothing about the underlying cause of error, which is not good - the model might not exist, database might be unreachable, etc.

internal/ssh/dial.go
```diff
@@ -0,0 +1,37 @@
+// Copyright 2025 Canonical.
+package ssh
```
alesstimec, line 3: new line, please

internal/ssh/dial.go
```diff
@@ -0,0 +1,37 @@
...
+// dialControllerSSHServer dials the controller ssh server, trying the addresses sequentially and returning a go ssh client.
+func dialControllerSSHServer(addrs []string, destPort uint32) (*gossh.Client, error) {
+	var client *gossh.Client
+	var err error
+	var errors []error
+	for _, addr := range addrs {
+		dest := net.JoinHostPort(addr, fmt.Sprint(destPort))
+		client, err = gossh.Dial("tcp", dest, &gossh.ClientConfig{
+			//nolint:gosec // this will be removed once we handle hostkeys
+			HostKeyCallback: gossh.InsecureIgnoreHostKey(),
```
alesstimec, line 25: i would not put this in our production code..

internal/ssh/dial.go
```diff
@@ -0,0 +1,37 @@
...
+				}),
+			},
+			Timeout: 5 * time.Second,
+		})
+		if err != nil {
+			errors = append(errors, err)
+		}
+	}
+	if client == nil {
+		return nil, fmt.Errorf("failed to dial controller ssh server: %v", errors)
```
alesstimec: this PR seems to change how we return errors.. is this an agreed direction?

internal/ssh/dial.go
SimoneDutto: this is just the ssh dial. So to be able to come here you need to be a user in jimm, have a public key registered in jimm, and then you can dial to controllers' ssh server from this code. 
Right now we don't have the ssh key of any controller, but i don't think it's a security issue

internal/ssh/dial.go
SimoneDutto: this is just the way to return errors in ssh dial, no change to the controller's dial

internal/jimm/ssh/ssh.go
SimoneDutto: sorry my oversight

internal/ssh/dial.go
kian99: Alternatively, we can just comment out the "server.ListenAndServer()" till we're all the way done with our implementation.

SimoneDutto: i don't see a way of exploiting this ssh server to do anything.
- first you need an ssh key added via juju commands
- then you need to have a model, with a registered jimm user with write permission on it
- then you need to have an ssh running on a controller, because we are resolving controller's addresses from model uuid
- then this dial is hardcoding a password "jwt"

internal/jimm/ssh/ssh.go
```diff
@@ -67,3 +68,20 @@ func (s *sshManager) PublicKeyHandler(ctx context.Context, claimUser string, key
...
 	return user, nil
 }
+
+// ResolveAddressesFromModelUUID is the method to resolve the address of the controller to contact given the model UUID.
+func (s *sshManager) ResolveAddressesFromModelUUID(ctx context.Context, modelUUID string) ([]string, error) {
+	zapctx.Info(ctx, "ResolveAddressesFromModelUUID")
+
+	model, err := s.modelManager.GetModel(ctx, modelUUID)
+	if err != nil {
+		return nil, fmt.Errorf("cannot find model %s: %w", modelUUID, err)
```
alesstimec:
```suggestion
		return nil, errors.E(err, "cannot find model"))
```

internal/jimm/ssh/ssh.go
```diff
@@ -67,3 +68,20 @@ func (s *sshManager) PublicKeyHandler(ctx context.Context, claimUser string, key
...
+	}
+
+	addrs, _ := rpc.GetAddressesAndTLSConfig(ctx, &model.Controller)
+	if len(addrs) == 0 {
+		return nil, fmt.Errorf("cannot find addresses for model %s", modelUUID)
```
alesstimec:
```suggestion
		return nil, errors.E(err, "cannot find addresses for model's controller")
```

internal/ssh/dial.go
alesstimec: well, no, it's about how we return and log errors.. i would suggest returning always returning a sensible error.. now, how the ssh server's handler treats, logs and returns these errors to its client is a different matter

alesstimec (approved):
LGTM.. BUT.. in the underlying layers i would suggest keeping error handling (and returning) as we do in other code.. the ssh server can then log and map these error as it likes..

internal/ssh/dial.go
SimoneDutto: ah ok, i've changed it to use the errorrs pkg with the msg decoupled from the error.

PR #1530, opened 2025-01-21
Juju 7405/add fingerprint endpoint

SimoneDutto:
## Description
In this PR we hook everything up in the main.go, and we add an endpoint to serve the public key fingerprint.

## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests

### DIFF
--- a/cmd/jimmsrv/main.go
+++ b/cmd/jimmsrv/main.go
@@ -218,7 +218,7 @@ func start(ctx context.Context, s *service.Service) error {
 	})
 	s.Go(httpsrv.ListenAndServe)
 	zapctx.Info(ctx, "Successfully started JIMM server")
-
+	maxConccurentConncetions, _ := strconv.Atoi(os.Getenv("JIMM_SSH_MAX_CONCURRENT_CONNECTIONS"))
 	sshManager, err := jimmssh.NewSSHManager(jimmsvc.JIMM().IdentityManager(), jimmsvc.JIMM(), jimmsvc.JIMM().SSHKeyManager())
 	if err != nil {
 		return err
@@ -226,7 +226,7 @@ func start(ctx context.Context, s *service.Service) error {
 	sshServer, err := ssh.NewJumpServer(ctx, ssh.Config{
 		Port:                     os.Getenv("JIMM_SSH_PORT"),
 		HostKey:                  []byte(hostKeyRaw),
-		MaxConcurrentConnections: os.Getenv("JIMM_SSH_MAX_CONCURRENT_CONNECTIONS"),
+		MaxConcurrentConnections: maxConccurentConncetions,
 	}, sshManager)
 	if err != nil {
 		return err
--- a/cmd/jimmsrv/main.go
+++ b/cmd/jimmsrv/main.go
@@ -18,7 +18,6 @@ import (
 
 	jimmsvc "github.com/canonical/jimm/v3/cmd/jimmsrv/service"
 	"github.com/canonical/jimm/v3/internal/errors"
-	jimmssh "github.com/canonical/jimm/v3/internal/jimm/ssh"
 	"github.com/canonical/jimm/v3/internal/logger"
 	"github.com/canonical/jimm/v3/internal/ssh"
 	"github.com/canonical/jimm/v3/version"
@@ -144,7 +143,7 @@ func start(ctx context.Context, s *service.Service) error {
 	}
 
 	hostKeyRaw := os.Getenv("JIMM_SSH_HOST_KEY")
-	fingerprint, err := ssh.GetFingerprintFromPrivateKey([]byte(hostKeyRaw))
+	fingerprints, err := ssh.GetFingerprintsFromPrivateKey([]byte(hostKeyRaw))
 	if err != nil {
 		return errors.E("Cannot parse hostkey")
 	}
@@ -154,15 +153,15 @@ func start(ctx context.Context, s *service.Service) error {
 	logSQL, _ := strconv.ParseBool(os.Getenv("JIMM_LOG_SQL"))
 
 	jimmsvc, err := jimmsvc.NewService(ctx, jimmsvc.Params{
-		ControllerUUID:        os.Getenv("JIMM_UUID"),
-		DSN:                   os.Getenv("JIMM_DSN"),
-		SSHHostKeyFingerprint: fingerprint,
-		ControllerAdmins:      strings.Fields(os.Getenv("JIMM_ADMINS")),
-		VaultRoleID:           os.Getenv("VAULT_ROLE_ID"),
-		VaultRoleSecretID:     os.Getenv("VAULT_ROLE_SECRET_ID"),
-		VaultAddress:          os.Getenv("VAULT_ADDR"),
-		VaultPath:             os.Getenv("VAULT_PATH"),
-		PublicDNSName:         os.Getenv("JIMM_DNS_NAME"),
+		ControllerUUID:      os.Getenv("JIMM_UUID"),
+		DSN:                 os.Getenv("JIMM_DSN"),
+		HostKeyFingerprints: fingerprints,
+		ControllerAdmins:    strings.Fields(os.Getenv("JIMM_ADMINS")),
+		VaultRoleID:         os.Getenv("VAULT_ROLE_ID"),
+		VaultRoleSecretID:   os.Getenv("VAULT_ROLE_SECRET_ID"),
+		VaultAddress:        os.Getenv("VAULT_ADDR"),
+		VaultPath:           os.Getenv("VAULT_PATH"),
+		PublicDNSName:       os.Getenv("JIMM_DNS_NAME"),
 		OpenFGAParams: jimmsvc.OpenFGAParams{
 			Scheme:    os.Getenv("OPENFGA_SCHEME"),
 			Host:      os.Getenv("OPENFGA_HOST"),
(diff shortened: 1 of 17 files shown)


docker-compose.common.yaml
```diff
@@ -43,6 +43,46 @@ services:
       JIMM_SECURE_SESSION_COOKIES: false
       JIMM_SESSION_COOKIE_MAX_AGE: 86400
       JIMM_SESSION_SECRET_KEY: Xz2RkR9g87M75xfoumhEs5OmGziIX8D88Rk5YW8FSvkBPSgeK9t5AS9IvPDJ3NnB
+      JIMM_SSH_PORT: 17022
+      JIMM_SSH_HOST_KEY: |-
```
SimoneDutto, line 47: i've generated this key in my CLI, it's just a test default private key

cmd/jimmsrv/main.go
```diff
@@ -141,19 +143,26 @@ func start(ctx context.Context, s *service.Service) error {
...
-		DSN:               os.Getenv("JIMM_DSN"),
-		ControllerAdmins:  strings.Fields(os.Getenv("JIMM_ADMINS")),
-		VaultRoleID:       os.Getenv("VAULT_ROLE_ID"),
-		VaultRoleSecretID: os.Getenv("VAULT_ROLE_SECRET_ID"),
-		VaultAddress:      os.Getenv("VAULT_ADDR"),
-		VaultPath:         os.Getenv("VAULT_PATH"),
-		PublicDNSName:     os.Getenv("JIMM_DNS_NAME"),
+		ControllerUUID:        os.Getenv("JIMM_UUID"),
+		DSN:                   os.Getenv("JIMM_DSN"),
+		SSHHostKeyFingerprint: fingerprint,
```
kian99: nit: I'd rename `SSHHostKeyFingerprint` to just `HostKeyFingerprint`

cmd/jimmsrv/main.go
```diff
@@ -209,6 +218,20 @@ func start(ctx context.Context, s *service.Service) error {
 	})
 	s.Go(httpsrv.ListenAndServe)
 	zapctx.Info(ctx, "Successfully started JIMM server")
-
+	maxConccurentConncetions, _ := strconv.Atoi(os.Getenv("JIMM_SSH_MAX_CONCURRENT_CONNECTIONS"))
+	sshManager, err := jimmssh.NewSSHManager(jimmsvc.JIMM().IdentityManager(), jimmsvc.JIMM(), jimmsvc.JIMM().SSHKeyManager())
```
kian99: Why are we creating the domain object here? I'd prefer if we created all the business logic objects in `internal/jimm`.

docker-compose.common.yaml
kian99: I wonder if we don't want to just generate one in the docker file or something?

internal/ssh/utils.go
```diff
@@ -0,0 +1,15 @@
+// Copyright 2025 Canonical.
+
+package ssh
+
+import (
+	gossh "golang.org/x/crypto/ssh"
+)
+
+func GetFingerprintFromPrivateKey(privateKey []byte) (string, error) {
```
kian99: Maybe we should generate multiple fingerprint types and return all of them for display?
MD5 and SHA-256 fingerprints, wdyt?

cmd/jimmsrv/main.go
SimoneDutto: yeah, it makes sense as well

internal/ssh/utils.go
SimoneDutto: yeah, why not!

docker-compose.common.yaml
SimoneDutto: having it hardcoded it's handier when you are qaing so you don't have to accept the new key everytime. I don't see an issue with having it here, just like with the session secret key

cmd/jimmsrv/main.go
```diff
@@ -141,19 +142,26 @@ func start(ctx context.Context, s *service.Service) error {
 		return errors.E("jimm session store secret must be at least 64 characters")
 	}
 
+	hostKeyRaw := os.Getenv("JIMM_SSH_HOST_KEY")
+	fingerprints, err := ssh.GetFingerprintsFromPrivateKey([]byte(hostKeyRaw))
```
alesstimec, line 149: i think we should differentiate between "JIMM_SSH_HOST_KEY not specified" and "Cannot parse hostkey"?

cmd/jimmsrv/main.go
```diff
@@ -141,19 +142,26 @@ func start(ctx context.Context, s *service.Service) error {
...
+	if err != nil {
+		return errors.E("Cannot parse hostkey")
```
alesstimec: could we please start with a lower case letter? i know, we're consistently inconsistent, but still..

cmd/jimmsrv/main.go
```diff
@@ -209,6 +217,17 @@ func start(ctx context.Context, s *service.Service) error {
 	})
 	s.Go(httpsrv.ListenAndServe)
 	zapctx.Info(ctx, "Successfully started JIMM server")
+	maxConccurentConncetions, _ := strconv.Atoi(os.Getenv("JIMM_SSH_MAX_CONCURRENT_CONNECTIONS"))
```
alesstimec, line 223: let's not ignore this error

cmd/jimmsrv/main.go
```diff
@@ -209,6 +217,17 @@ func start(ctx context.Context, s *service.Service) error {
...
+	sshServer, err := ssh.NewJumpServer(ctx, ssh.Config{
+		Port:                     os.Getenv("JIMM_SSH_PORT"),
+		HostKey:                  []byte(hostKeyRaw),
+		MaxConcurrentConnections: maxConccurentConncetions,
+	}, jimmsvc.JIMM().SSHManager())
+	if err != nil {
+		return err
+	}
+	s.Go(sshServer.ListenAndServe)
+	zapctx.Info(ctx, "Successfully started JIMM ssh server")
```
alesstimec, line 234: lower case, please

internal/jimmhttp/fingerprint.go
```diff
@@ -0,0 +1,27 @@
+// Copyright 2025 Canonical.
+package jimmhttp
```
alesstimec, line 3: you know..

internal/jimmhttp/fingerprint.go
```diff
@@ -0,0 +1,27 @@
...
+func WriteFingerprints(m map[string]string) func(http.ResponseWriter, *http.Request) {
+	return func(w http.ResponseWriter, r *http.Request) {
+		ctx := r.Context()
+		b, err := json.Marshal(m)
+		if err != nil {
+			zapctx.Error(ctx, "failed to marshal map", zap.Error(err))
+			return
+		}
+		w.Header().Set("Content-Type", "application/json")
+		_, err = w.Write(b)
```
alesstimec, line 24: also set the status `w.WriteHeader(http.StatusOK)`

internal/jimmhttp/fingerprint.go
```diff
@@ -0,0 +1,27 @@
...
+)
+
+// WriteFingerprints writes a map as JSON to the response.
+func WriteFingerprints(m map[string]string) func(http.ResponseWriter, *http.Request) {
+	return func(w http.ResponseWriter, r *http.Request) {
+		ctx := r.Context()
+		b, err := json.Marshal(m)
+		if err != nil {
+			zapctx.Error(ctx, "failed to marshal map", zap.Error(err))
+			return
```
alesstimec, line 21: set the status to internal server error before returning..
`.WriteHeader(http.StatusInternaServerError)`

internal/ssh/utils.go
```diff
@@ -0,0 +1,20 @@
...
+	gossh "golang.org/x/crypto/ssh"
+)
+
+// GetFingerprintsFromPrivateKey returns the fingerprints of the public key from the private key.
+func GetFingerprintsFromPrivateKey(privateKey []byte) (map[string]string, error) {
+	key, err := gossh.ParsePrivateKey(privateKey)
+	if err != nil {
+		return nil, err
+	}
+	m := make(map[string]string)
```
alesstimec:
```suggestion
	return map[string]string{
	  "SHA256": gossh.FingerprintSHA256(key.PublicKey()),
	  "MD5": gossh.FingerprintLegacyMD5(key.PublicKey()),
	}, nil
```

alesstimec (changes requested):
let's set responses statuses when returning fingerprints.. and there's a missing error check

cmd/jimmsrv/main.go
SimoneDutto: i though that if you missconfigured the max connections this field will be set to 0, becoming the default value as a consequence, which is alright, considering max concurrent connections is not a "life or death" configuration value. what do you think?

cmd/jimmsrv/main.go
SimoneDutto: also for logs we go for lowercase?

internal/jimmhttp/fingerprint.go
SimoneDutto: from `Write` godoc:
`If [ResponseWriter.WriteHeader] has not yet been called, Write calls
	// WriteHeader(http.StatusOK) before writing the data.`

internal/ssh/utils.go
```diff
@@ -6,10 +6,15 @@ import (
 	gossh "golang.org/x/crypto/ssh"
 )
 
-func GetFingerprintFromPrivateKey(privateKey []byte) (string, error) {
+// GetFingerprintsFromPrivateKey returns the fingerprints of the public key from the private key.
```
kian99:
```suggestion
// GetFingerprintsFromPrivateKey returns the fingerprints of the host key.
```

kian99 (approved):

cmd/jimmsrv/main.go
alesstimec: what would a 0 value mean for max concurrent connections? would it mean nobody would be able to use ssh or that we allow infinite concurrent connections?

alesstimec (approved):
LGTM with other reviewer's comments addressed

cmd/jimmsrv/main.go
SimoneDutto: basically it will trigger this code: https://github.com/canonical/jimm/blob/c0b0b11f1dc9f5db2fc9931d93f5ae83c8800f42/internal/ssh/ssh.go#L99-L110 and we'll set the default value

PR #1538, opened 2025-01-30
chore: fix permission manager tests

kian99:
## Description
Some of the permission manager tests were starting a whole JIMM instance just to test the domain object. Fix this to just start the permissionManagerSuite in tests that don't use a suite. No logic has been changed in the tests.

### DIFF
--- a/internal/jimm/permissions/access_test.go
+++ b/internal/jimm/permissions/access_test.go
@@ -17,7 +17,6 @@ import (
 	"github.com/canonical/jimm/v3/internal/db"
 	"github.com/canonical/jimm/v3/internal/dbmodel"
 	"github.com/canonical/jimm/v3/internal/errors"
-	"github.com/canonical/jimm/v3/internal/jimm"
 	"github.com/canonical/jimm/v3/internal/jimm/permissions"
 	"github.com/canonical/jimm/v3/internal/openfga"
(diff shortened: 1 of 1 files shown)


internal/jimm/permissions/access_test.go
```diff
@@ -394,15 +393,16 @@ func TestGrantModelAccess(t *testing.T) {
 		c.Run(tt.name, func(c *qt.C) {
 			ctx := context.Background()
 
-			j := jimmtest.NewJIMM(c, &jimm.Parameters{})
+			s := permissionManagerSuite{}
+			s.Init(c)
 
 			env := jimmtest.ParseEnvironment(c, tt.env)
-			env.PopulateDBAndPermissions(c, j.ResourceTag(), j.Database, j.OpenFGAClient)
+			env.PopulateDBAndPermissions(c, s.ctlTag, s.db, s.ofgaClient)
```
SimoneDutto, line 400: `s.cltTag` is an empty string right? just to make the env.PopulateDb method work

SimoneDutto (approved):

internal/jimm/permissions/access_test.go
kian99: No but it's just a random UUID.

alesstimec (approved):

PR #1543, opened 2025-02-07
feat: add weekly trivy scan for released container

kian99:
This PR adds to our weekly security workflow to scan the most recently released jimm container.
We also now release a container with the `latest` tag on each release.

### DIFF
--- a/.github/workflows/security-scan-weekly.yaml
+++ b/.github/workflows/security-scan-weekly.yaml
@@ -6,9 +6,9 @@ on:
   workflow_dispatch:
 
 jobs:
-  security-scan:
+  repo-security-scan:
     runs-on: [ubuntu-latest]
-    name: Security Scan
+    name: Repo Security Scan
     strategy:
       matrix:
         branch: [v3]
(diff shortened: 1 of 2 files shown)


.github/workflows/security-scan-weekly.yaml
```diff
@@ -23,3 +23,19 @@ jobs:
         with: 
           upload-github-security: "true"
           upload-sarif-artifact: "true"
+
+  container-security-scan:
```
ale8k: Can we put this in the comsys build tools scanner reusable workflow?

ale8k (changes requested):

.github/workflows/security-scan-weekly.yaml
kian99: Feels a bit unnecessary since it's just two steps no?

SimoneDutto (approved):

.github/workflows/security-scan-weekly.yaml
ale8k: Well we've got a weekly scan, which uploads SARIFs and now we've got another workflow doing exactly the same thing. Feels sensible to me to put it in there?

ale8k (approved):

.github/workflows/security-scan-weekly.yaml
kian99: Done.

PR #1547, opened 2025-02-10
refactor(dial): refactors to dial the controller as user instead of admin

alesstimec:
## Description

Refactors the way JIMM dials Juju controllers. Instead of dialing as the
admin user and acting as admin on behalf of the user that dialed JIMM,
JIMM will now dial controllers as that user.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests


internal/jimm/juju/model.go
```diff
@@ -614,6 +615,20 @@ func (j *JujuManager) ValidateModelUpgrade(ctx context.Context, user *openfga.Us
 	return nil
 }
 
+func convertModelRelationToAccessString(relation ofga.Relation) (string, error) {
```
kian99: I think you can use the function from the permissions package.

internal/jujuclient/dial.go
```diff
@@ -54,36 +56,101 @@ type Dialer struct {
...
+		switch user.GetControllerAccess(ctx, ctl.ResourceTag()) {
+		case ofganames.AdministratorRelation:
+			permissions[ctl.ResourceTag().String()] = "superuser"
+		default:
+			permissions[ctl.ResourceTag().String()] = "login"
+		}
+
+		if modelTag.Id() != "" {
+			// model permissions
+			switch user.GetModelAccess(ctx, modelTag) {
```
kian99: What do you think about simplifying to just checking for `write` and `read`? Admin access is only needed in Juju for granting other users with access so we shouldn't need the admin's access (although maybe deleting a model requires admin? Not sure).

alesstimec: I'd rather we correctly state user's permission level for the model in question. And, yes, to delete a model, i think one must be model admin.

go.mod
```diff
@@ -5,11 +5,11 @@ go 1.24.1
 // Juju based dependencies
 require (
 	github.com/juju/charm/v12 v12.1.1
-	github.com/juju/cmd/v3 v3.1.2
+	github.com/juju/cmd/v3 v3.2.0
 	github.com/juju/errors v1.0.0
 	github.com/juju/gnuflag v1.0.0
 	github.com/juju/http/v2 v2.0.1
-	github.com/juju/juju v0.0.0-20250306114340-d7a40b3e3801 // v3.6.4
+	github.com/juju/juju v0.0.0-20250327101341-0cd4a7bd2d86 // v3.6.4
```
kian99: Update the comment if this is an un-released version to indicate that.

internal/jimm/juju/applicationoffer.go
```diff
@@ -471,16 +482,40 @@ func (j *JujuManager) FindApplicationOffers(ctx context.Context, user *openfga.U
...
-	})
+	offerUUIDs, err := user.ListApplicationOffers(ctx, ofganames.ReaderRelation)
 	if err != nil {
 		return nil, errors.E(op, err)
 	}
 
-	offers, err := j.queryControllersForOffers(ctx, user, controllers, func(api API) ([]jujuparams.ApplicationOfferAdminDetailsV5, error) {
+	controllerPermissions := make(map[string]map[names.Tag]string)
+	for _, offerUUID := range offerUUIDs {
+		ot := names.NewApplicationOfferTag(offerUUID)
```
kian99: Much easier to read
```suggestion
		offerTag := names.NewApplicationOfferTag(offerUUID)
```

internal/jimm/juju/applicationoffer.go
```diff
@@ -471,16 +482,40 @@ func (j *JujuManager) FindApplicationOffers(ctx context.Context, user *openfga.U
...
+		offer.SetTag(ot)
+
+		if err := j.Database.GetApplicationOffer(ctx, &offer); err != nil {
+			if errors.ErrorCode(err) == errors.CodeNotFound {
+				continue
+			}
+			return nil, errors.E(op, err)
+		}
+
+		relation := user.GetApplicationOfferAccess(ctx, ot)
```
kian99: I'm not a fan of our methods that `GetAccess` because they do OpenFGA tree traversal logic I believe, right?
Since this is the `FindApplicationOffers` method, can't we just check for Read permission?

internal/jimm/juju/applicationoffer.go
```diff
@@ -513,10 +548,24 @@ func (j *JujuManager) ListApplicationOffers(ctx context.Context, user *openfga.U
 		if err := j.Database.GetModel(ctx, &m); err != nil {
 			return nil, errors.E(op, err)
 		}
-		controllers[m.Controller.ID] = &m.Controller
+		// check model permissions
+		modelRelation := user.GetModelAccess(ctx, m.ResourceTag())
```
kian99, line 546: Why for `ListApplicationOffers` do we populate the map with the model access, whereas in `FindApplicationOffers` we populate the permission map with the offers.

internal/jimm/juju/model.go
```diff
@@ -728,52 +741,81 @@ func (j *JujuManager) ListModels(ctx context.Context, user *openfga.User) ([]bas
...
-		// NOTE: Controller models are not included because we never relate
-		// controller models to users, and as such, they will not appear in the
-		// authorised uuid map.
-		for _, um := range ums {
-			mapModel, ok := modelsMap[um.UUID]
-			if !ok {
-				continue
+	eg := new(errgroup.Group)
+	for _, ctl := range controllers {
+		eg.Go(func() error {
```
kian99: Can you take a look at declaring this function a few lines above so that it's less nested or making it a true function.

internal/jujuapi/websocket_test.go
```diff
@@ -210,12 +217,20 @@ func (s *apiProxySuite) TestSessionTokenLoginProvider(c *gc.C) {
...
 	c.Assert(err, gc.IsNil)
 	var output bytes.Buffer
-	s.JIMMSuite.EnableDeviceFlow(aliceUser.Name)
-	conn, err := s.openCustomLoginProvider(c, &api.Info{
-		ModelTag:  s.Model.ResourceTag(),
-		SkipLogin: false,
-	}, "alice", api.NewSessionTokenLoginProvider("", &output, func(s string) {}))
-	c.Assert(err, gc.IsNil)
+	var conn api.Connection
+	for i := 0; i < 10; i++ {
```
kian99: Does this still need the loop?

internal/jujuapi/websocket_test.go
```diff
@@ -161,9 +162,15 @@ func (s *websocketSuite) openNoAssert(c *gc.C, d loginDetails) (api.Connection,
 
 func (s *websocketSuite) open(c *gc.C, info *api.Info, username string) api.Connection {
 	ld := loginDetails{info: info, username: username}
-	conn, err := s.openNoAssert(c, ld)
-	c.Assert(err, gc.Equals, nil)
-	return conn
+	for i := 0; i < 10; i++ {
```
kian99: Do we need this loop?

internal/jimm/juju/cache.go
```diff
@@ -57,9 +61,26 @@ func (d *cacheDialer) Dial(ctx context.Context, ctl *dbmodel.Controller, mt name
 	}
 }
 
-func (d *cacheDialer) dial(ctx context.Context, ctl *dbmodel.Controller, requiredPermissions map[string]string) (interface{}, error) {
+func mapKey(user *openfga.User, ctl *dbmodel.Controller, requiredPermissions map[string]string) string {
```
kian99, line 66: This needs a better name and some docstrings. It looks like we're creating a cached dialer per user per controller?

kian99 (commented):
There's a number of unrelated things going on in this PR by the look of things. Deleting some unused methods, upgrading the Juju version, the ability to dial as a user, changes to a number of methods to dial as the user making the call. I think if this was 2 or 3 PRs, it would be a lot easier to review thoroughly.

internal/jimm/juju/model.go
```diff
@@ -728,52 +741,81 @@ func (j *JujuManager) ListModels(ctx context.Context, user *openfga.User) ([]bas
...
-			um.Owner = mapModel.OwnerIdentityName
-			userModels = append(userModels, um)
-		}
-		return nil
-	})
-	if err != nil {
-		return nil, errors.E(op, fmt.Sprintf("failed to list models: %v", err))
+
+			zapctx.Debug(ctx, "dialing controller", zap.String("controller", ctl.UUID))
+			api, err := j.dial(ctx, user, &ctl, names.ModelTag{}, permissions...)
```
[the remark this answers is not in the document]
SimoneDutto: please rephrase your comments using english from the 1800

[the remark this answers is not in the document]
ale8k: Quiet copilot. This is fixed in 1.21+.

internal/jimm/juju/applicationoffer.go
alesstimec: juju requires that user must be model admin

go.mod
alesstimec: done

internal/jimm/juju/applicationoffer.go
alesstimec: done

internal/jimm/juju/applicationoffer.go
alesstimec: removed, you're right.. we list application offers above to which user has read access, so there's no need here

internal/jimm/juju/cache.go
alesstimec: that's exactly right. Since we now dial as user, we need to cache controller connection per controller, user and the set of permissions. that is why i disabled connection caching in this PR until we're absolutely sure we need it.

internal/jimm/juju/model.go
alesstimec: done

internal/jujuapi/websocket_test.go
alesstimec: removed

internal/jujuapi/websocket_test.go
alesstimec: nope.. removed

internal/jimm/juju/applicationoffer.go
```diff
@@ -513,10 +542,24 @@ func (j *JujuManager) ListApplicationOffers(ctx context.Context, user *openfga.U
...
```
kian99, line 546: Can we just do `IsModelAdmin()` and then continue based on the result?

internal/jimm/juju/cache.go
kian99: Can we still give this function a more descriptive name even though we're not currently using the cache.

internal/jimm/juju/cloud.go
```diff
@@ -364,20 +364,13 @@ func (j *JujuManager) addControllerCloud(ctx context.Context, ctl *dbmodel.Contr
...
-// returned from the dial operation. If the given function returns an error
-// that error will be returned with the code unmasked.
-func (j *JujuManager) doCloudAdmin(ctx context.Context, user *openfga.User, ct names.CloudTag, f func(*dbmodel.Cloud, API) error) error {
-	const op = errors.Op("jimm.doCloudAdmin")
+// RemoveCloud removes the given cloud from JAAS If the cloud is not found
+// then an error with the code CodeNotFound is returned. If the
+// authenticated user does not have admin access to the cloud then an error
+// with the code CodeUnauthorized is returned. If the RemoveClouds API call
+// returns an error the error code is not masked.
+func (j *JujuManager) RemoveCloud(ctx context.Context, user *openfga.User, ct names.CloudTag) error {
```
kian99, line 372: Does this work as expected if multiple controllers have access to the cloud?

internal/jimm/juju/model.go
```diff
@@ -251,11 +253,15 @@ func (j *JujuManager) ModelInfo(ctx context.Context, user *openfga.User, mt name
 		return nil, errors.E(op, err)
 	}
 
-	if ok, err := user.IsModelReader(ctx, mt); !ok || err != nil {
+	relation := user.GetModelAccess(ctx, mt)
```
kian99, line 256: I'd prefer to avoid these calls that get the access level, instead opting to check if the user has the minimum required access level.
In this case for ModelInfo, we could check if the user is a ModelReader.

internal/jimm/juju/model.go
```diff
@@ -728,52 +743,86 @@ func (j *JujuManager) ListModels(ctx context.Context, user *openfga.User) ([]bas
...
 		controllers = append(controllers, model.Controller)
 	}
 
-	// Call controllers for their models. We always call as admin, and we're
-	// filtering ourselves. We do this rather than send the user to be 100%
-	// certain that the models do belong to user according to OpenFGA. We could
-	// in theory rely on Juju correctly returning the models (by owner), but this
-	// is more reliable.
+	// The call to ListModels is currently broken in Juju because it always consults
+	// mongo for authorization. Because of that we have to resort to ModelInfo calls.
```
kian99, line 766: Can we put a link to the issue we've filed against Juju? Or a link to a PR that is aiming to fix it.
I'm a bit worried that turning list models into a loop that runs ModelInfo will be prohibitively slow in a real world scenario.

internal/jujuapi/modelmanager_test.go
```diff
@@ -96,24 +95,18 @@ func (s *modelManagerSuite) TestListModelSummaries(c *gc.C) {
...
 		ProviderType:    jimmtest.TestProviderType,
-		DefaultSeries:   "jammy",
 		Cloud:           jimmtest.TestCloudName,
 		CloudRegion:     jimmtest.TestCloudRegionName,
 		CloudCredential: jimmtest.TestCloudName + "/charlie@canonical.com/cred",
 		Owner:           "charlie@canonical.com",
 		Life:            life.Value(state.Alive.String()),
 		Status: base.Status{
-			Status: status.Available,
+			Status: "unavailable",
```
kian99, line 104: Why has this changed?

internal/jimm/juju/model.go
kian99: Couldn't we have this API call stick to using the `admin` user until we make the fix? That seems to be a lot easier.

internal/jujuclient/dial.go
```diff
@@ -54,36 +57,116 @@ type Dialer struct {
...
+
+		// we also state that the "admin" user is
+		// - controller superuser
+		// - model admin
+		userPermissions[ctl.ResourceTag().String()] = "superuser"
+		if modelTag.Id() != "" {
+			userPermissions[modelTag.String()] = "admin"
+		}
+	} else {
+		// if we are logging in as any other user, we need to check permissions
```
kian99, line 85: Why are we checking permissions so late, at the dial stage? In the business logic we are already checking if the user has access then adding it to the list of permissions required when dialing, it seems wrong to check again.
The logic also seems to be too forgiving, allowing the caller to provide a resource tag and an access level but optionally allows the access level to be empty. I would argue we should simplify this (if we need to do this) and the access level should never be empty.

kian99 (commented):
Almost lgtm, but several important comments.

internal/jimm/juju/model.go
alesstimec: the thing is that the user will see different level of detail depending on their access level

PR #1548, opened 2025-02-11
chore: refactor ssh listener server

kian99:
## Description
I found that the ssh server in jimm could be improved in some ways:
- Use an in-memory buffered connection instead of listening on a system port. Similar to our Juju tests.
- The internal ssh package was importing internal/jimm when it shouldn't.
- Refactored the sshManager interface so that dialing the Juju controller's SSH server can be mocked.
- Moved the limitListener to a separate package.
- Removed unused functions thanks to the changes above.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests


internal/ssh/limitlistener/listener.go
```diff
@@ -38,10 +38,10 @@ import (
 // This is a copypaste of netutil.LimiLister (link: https://cs.opensource.google/go/x/net/+/refs/tags/v0.34.0:netutil/listen.go),
 // but we add a timeout so when we are at the limit we actively close connections instead of waiting indefinetely. (Look at line 44)
 
-// limitListenerWithTimeout returns a Listener that accepts at most n simultaneous
+// ListenerWithTimeout returns a Listener that accepts at most n simultaneous
 // connections from the provided Listener, and it timeouts when the max
 // has been reached and no seats has been freed for the timeout period.
-func limitListenerWithTimeout(l net.Listener, n int, timeout time.Duration) net.Listener {
+func ListenerWithTimeout(l net.Listener, n int, timeout time.Duration) net.Listener {
```
SimoneDutto, line 44:
```suggestion
func ListenerWithLimitAndTimeout(l net.Listener, n int, timeout time.Duration) net.Listener {
```
I think it's better to express that there is both limit and timeout.

internal/ssh/ssh.go
```diff
@@ -144,52 +142,36 @@ func directTCPIPHandler(sshManager SSHManager) func(srv *ssh.Server, conn *gossh
...
-		srcDest, reqs, err := newChan.Accept()
+		clientConn, reqs, err := newChan.Accept()
 		if err != nil {
-			dstChan.Close()
+			rejectConnectionAndLogError(ctx, newChan, "failed to accept channel creation request", err)
 			return
 		}
-		// gossh.Request are requests sent outside of the normal stream of data (ex. pty-req for an interactive session).
-		// Since we only need the raw data to redirect, we can discard them.
 		go gossh.DiscardRequests(reqs)
```
SimoneDutto: why did you remove this comment? is it inaccurate or something?

internal/ssh/ssh.go
```diff
@@ -144,52 +142,36 @@ func directTCPIPHandler(sshManager SSHManager) func(srv *ssh.Server, conn *gossh
...
 		go func() {
-			defer srcDest.Close()
-			defer dstChan.Close()
-			_, err := io.Copy(srcDest, dstChan)
-			if err != nil {
-				rejectConnectionAndLogError(ctx, newChan, "failed to copy data from src to dts", err)
-			}
+			defer clientConn.Close()
+			defer controllerConn.Close()
+			_, _ = io.Copy(clientConn, controllerConn)
```
SimoneDutto: why ignoring the error here?

internal/ssh/ssh.go
```diff
@@ -144,52 +142,36 @@ func directTCPIPHandler(sshManager SSHManager) func(srv *ssh.Server, conn *gossh
...
 		go func() {
-			defer srcDest.Close()
-			defer dstChan.Close()
-			_, err := io.Copy(dstChan, srcDest)
-			if err != nil {
-				rejectConnectionAndLogError(ctx, newChan, "failed to copy data from dst to src", err)
-			}
+			defer clientConn.Close()
+			defer controllerConn.Close()
+			_, _ = io.Copy(controllerConn, clientConn)
```
SimoneDutto: why ignoring the error here?

internal/ssh/ssh_test.go
```diff
@@ -245,51 +258,27 @@ func (s *sshSuite) TestSSHJumpDialFail(c *qt.C) {
...
+	defer client.Close()
+
 	s.testInDestinationServerF = func(fm ssh.ForwardMessage) {
 		c.Check(fm.DestAddr, qt.Equals, "model1")
 	}
 	_, err = client.Dial("tcp", fmt.Sprintf("%s:%d", "model1", 1))
 	c.Assert(err, qt.ErrorMatches, ".*connect failed.*")
 }
 
-func (s *sshSuite) TestMaxConcurrentConnections(c *qt.C) {
```
SimoneDutto, line 263: i guess this test is to be removed with the in-memory connection right?

SimoneDutto: i don't feel like using an untested listener in our ssh server. Shouldn't we try to build a test for it?

internal/jimm/jimm.go
```diff
@@ -195,9 +196,8 @@ type SSHManager interface {
 	// PublicKeyHandler is the method to verify the public key of the user. It returns a user if successful.
 	PublicKeyHandler(ctx context.Context, claimUser string, key []byte) (*openfga.User, error)
 
-	// ControllerInfoFromModelUUID is the method to resolve the address of the controller to contact given the model UUID and
-	// a valid JWT To connect to the controller.
-	ControllerInfoFromModelUUID(ctx context.Context, modelUUID string, user *openfga.User) (ssh.ControllerInfo, error)
+	// DialControllerSSHServer dials the controller hosting the specific model UUID.
+	DialControllerSSHServer(ctx context.Context, modelUUID string, user *openfga.User) (*gossh.Client, error)
```
SimoneDutto: i preferred to dial the in `internal/ssh` instead of `internal/jimm/ssh`, but it's just a preference. I am curious to listen to other opinions

internal/ssh/ssh_test.go
```diff
@@ -217,17 +226,21 @@ func (s *sshSuite) TestSSHJumpPermissionFail(c *qt.C) {
 
 	for _, test := range tests {
 		c.Run(test.name, func(c *qt.C) {
-			client, err := gossh.Dial("tcp", fmt.Sprintf(":%d", s.jumpServerPort), &gossh.ClientConfig{
+			conn, err := s.jumpServerListener.Dial()
+			c.Assert(err, qt.IsNil)
+
+			sshConn, newChan, reqs, err := gossh.NewClientConn(conn, "", &gossh.ClientConfig{
```
SimoneDutto: can you create a private method in this file called `dialInMemory` or something, so you can explain why you are using these three functions to dial + you can reuse it in all tests, avoiding this boilerplate?

internal/testutils/jimmtest/utils.go
```diff
@@ -1,20 +0,0 @@
...
-
-package jimmtest
-
-import (
-	"errors"
-	"net"
-)
-
-// GetFreePort asks the kernel for a free open port that is ready to use.
-func GetFreePort() (int, error) {
```
SimoneDutto, line 11: bye bye nifty function

internal/ssh/limitlistener/listener.go
kian99: I think that's covered by the package name. Wdyt?

internal/jimm/jimm.go
kian99: This let's us mock the dial for testing so that's one reason.

internal/ssh/ssh.go
kian99: No that was a bit of an accident, I thought I would be removing the `go gossh.DiscardRequests(reqs)` line. I'll re-add it.

internal/ssh/ssh.go
kian99: Ah this is mb, I thought io.Copy returns an EOF error but I see the docstring says,
```
// A successful Copy returns err == nil, not err == EOF.
// Because Copy is defined to read from src until EOF, it does
// not treat an EOF from Read as an error to be reported.
```
So I'll revert this.

internal/ssh/ssh.go
kian99: Same as above.

internal/ssh/ssh_test.go
kian99: Yes good idea

internal/ssh/ssh_test.go
kian99: I realised that we probably don't want to test the listener in the SSH package. We also didn't copy the listeners tests when we copied the listener, so I'm working on a follow-up to copy the listener's existing tests + a new one for the timeout.

internal/testutils/jimmtest/utils.go
kian99: Sorry about that :cry:

internal/ssh/ssh.go
kian99: Although I think the way we handle the error is wrong, we can't reject the channel after accepting it.

internal/ssh/ssh_test.go
kian99: Added the tests in #1550

SimoneDutto (approved):
ty

internal/jimm/jimm.go
```diff
@@ -195,9 +196,8 @@ type SSHManager interface {
 	// PublicKeyHandler is the method to verify the public key of the user. It returns a user if successful.
 	PublicKeyHandler(ctx context.Context, claimUser string, key []byte) (*openfga.User, error)
 
-	// ControllerInfoFromModelUUID is the method to resolve the address of the controller to contact given the model UUID and
-	// a valid JWT To connect to the controller.
-	ControllerInfoFromModelUUID(ctx context.Context, modelUUID string, user *openfga.User) (ssh.ControllerInfo, error)
+	// DialControllerSSHServer dials the controller hosting the specific model UUID.
```
ale8k: This is worded weirdly... Uhm maybe something like "dials the hosting controller's SSH server for this model."

internal/jimm/jimm.go
ale8k: I also really hate how we pass users as extra arguments, we do it so often that we may as well a) put it on context with a specific key and ALWAYS expect a user to be there or b) have a new type which contains the user. Not for this PR but we should discuss this. Passing users as arguments is really ugly and annoying. It also couples our signatures and users to callers.

internal/jimm/ssh/ssh.go
```diff
@@ -15,6 +19,9 @@ import (
 	"github.com/canonical/jimm/v3/internal/rpc"
 )
 
+// jujuSSHDefaultPort is the default port we expect the juju controllers to respond on.
+const jujuSSHDefaultPort = 17022
```
ale8k, line 23: Another discussion for another time, we need a "default values file". Having random scattered vars for defaults doesn't feel good.

internal/jimm/ssh/ssh.go
```diff
@@ -112,3 +119,42 @@ func (s *sshManager) ControllerInfoFromModelUUID(ctx context.Context, modelUUID
 		JWT:       string(jwt),
 	}, nil
 }
+
+// DialControllerSSHServer determines which controller holds the desired
+// model and dials the controller returning an SSH connection to the controller.
+// We intentionally don't accept user input for the port to disallow the
+// user from probing for open ports on the Juju controller.
+func (s *sshManager) DialControllerSSHServer(ctx context.Context, modelUUID string, user *openfga.User) (*gossh.Client, error) {
```
ale8k: This is doing too much, we can discover the controller separately and pass the connection info. Would also make testing easier. Otherwise rename it to "DiscoverControllerByModelAndDialControllerSSHServer" if you know what I mean?  Cause really this is a side effect that the function name doesn't portray.

internal/jimm/ssh/ssh.go
```diff
@@ -112,3 +119,42 @@ func (s *sshManager) ControllerInfoFromModelUUID(ctx context.Context, modelUUID
...
+	connInfo, err := s.controllerInfoFromModelUUID(ctx, modelUUID, user)
+	if err != nil {
+		return nil, err
+	}
+	// TODO: Dial the controller and request it's SSH port
```
ale8k, line 126: Make a card please

internal/jimm/ssh/ssh.go
```diff
@@ -112,3 +119,42 @@ func (s *sshManager) ControllerInfoFromModelUUID(ctx context.Context, modelUUID
...
+	destPort := jujuSSHDefaultPort
+	var client *gossh.Client
+	var errs []error
+
+	for _, addr := range connInfo.Addresses {
+		dest := net.JoinHostPort(addr, fmt.Sprint(destPort))
+		client, err = gossh.Dial("tcp", dest, &gossh.ClientConfig{
+			User: "jimm",
+			//nolint:gosec // this will be removed once we handle hostkeys
+			HostKeyCallback: gossh.InsecureIgnoreHostKey(),
```
ale8k, line 138: TODO: don't ignore host key eventually? + Ticket? If we don't put todo in be harder to find and maybe put it on a line separate from the linting line, otherwise might get missed a bit

ale8k (changes requested):
Haven't gone through all but some comments on addressing the side effects should be done first, then will review again

internal/jimm/ssh/ssh.go
kian99: I don't agree with this one. It's looking up the controller and dialing. If we want the dialer to be something we define outside and pass in, we can do that in a separate PR. Also the docstring states that it finds the controller by the model.

internal/jimm/ssh/ssh.go
kian99: I'll just make a card.

internal/jimm/ssh/ssh.go
kian99: Brought the `controllerInfoFromModelUUID` out of the `DialControllerSSHServer` so that it's not doing two things in one.

internal/ssh/ssh_test.go
```diff
@@ -245,51 +249,31 @@ func (s *sshSuite) TestSSHJumpDialFail(c *qt.C) {
...
-	for _, client := range clients {
-		client.Close()
-	}
-}
-
 func TestIdentityManager(t *testing.T) {
 	qtsuite.Run(qt.New(t), &sshSuite{})
 }
+
+func inMemoryDial(c *qt.C, listener *bufconn.Listener, config *gossh.ClientConfig) *gossh.Client {
```
ale8k, line 281: lil godoc would go a long way here

internal/ssh/ssh_test.go
```diff
@@ -127,18 +120,31 @@ func (s *sshSuite) Init(c *qt.C) {
 				}
 				return userWithoutAccess, nil
 			},
-			ControllerInfoFromModelUUID_: func(ctx context.Context, modelUUID string, user *openfga.User) (jimmssh.ControllerInfo, error) {
-				if user == userWithAccess {
-					return jimmssh.ControllerInfo{Addresses: []string{""}, JWT: "valid-jwt"}, nil
+			DialControllerSSHServer_: func(ctx context.Context, modelUUID string, user *openfga.User) (*gossh.Client, error) {
```
ale8k: Snazzy

internal/testutils/jimmtest/mocks/jimm_ssh_mock.go
```diff
@@ -5,15 +5,16 @@ package mocks
...
 
+	gossh "golang.org/x/crypto/ssh"
+
 	"github.com/canonical/jimm/v3/internal/errors"
-	"github.com/canonical/jimm/v3/internal/jimm/ssh"
 	"github.com/canonical/jimm/v3/internal/openfga"
 )
 
 // SSHManager is an implementation of the SshManager interface.
 type SSHManager struct {
```
ale8k, line 16: Be nice to eventually just gomock interfaces for our services in packages that consume them

internal/ssh/ssh.go
```diff
@@ -144,52 +147,51 @@ func directTCPIPHandler(sshManager SSHManager) func(srv *ssh.Server, conn *gossh
...
-				rejectConnectionAndLogError(ctx, newChan, "failed to copy data from src to dts", err)
+				zapctx.Error(ctx, "ssh client to controller error", zap.Error(err))
 			}
+
 		}()
 		go func() {
-			defer srcDest.Close()
-			defer dstChan.Close()
-			_, err := io.Copy(dstChan, srcDest)
+			defer clientConn.Close()
```
ale8k, line 188: Nice qol

internal/ssh/ssh.go
```diff
@@ -131,9 +134,9 @@ func directTCPIPHandler(sshManager SSHManager) func(srv *ssh.Server, conn *gossh
 			rejectConnectionAndLogError(ctx, newChan, "failed to parse channel data", err)
 			return
 		}
-		if d.DestPort == 0 {
-			d.DestPort = jujuSSHDefaultPort
-		}
+
+		// TODO: Parse destAddr from a virtual hostname.
```
ale8k, line 138: Link ticket here would be good

internal/ssh/ssh.go
```diff
@@ -32,9 +32,12 @@ type SSHManager interface {
 	// PublicKeyHandler is the method to verify the public key of the user. It returns a user if successful.
 	PublicKeyHandler(ctx context.Context, claimUser string, key []byte) (*openfga.User, error)
 
-	// ControllerInfoFromModelUUID is the method to resolve the address of the controller to contact given the model UUID and
-	// a valid JWT To connect to the controller.
+	// ControllerInfoFromModelUUID uses the given model UUID to return the address of the controller to
+	// contact and a valid JWT To authenticate to the controller.
 	ControllerInfoFromModelUUID(ctx context.Context, modelUUID string, user *openfga.User) (jimmssh.ControllerInfo, error)
+
+	// DialControllerSSHServer dials the controller hosting the specific model UUID.
```
ale8k: Doc is wrong now we're using controller info

internal/ssh/ssh.go
```diff
@@ -32,9 +32,12 @@ type SSHManager interface {
 	// PublicKeyHandler is the method to verify the public key of the user. It returns a user if successful.
 	PublicKeyHandler(ctx context.Context, claimUser string, key []byte) (*openfga.User, error)
 
-	// ControllerInfoFromModelUUID is the method to resolve the address of the controller to contact given the model UUID and
-	// a valid JWT To connect to the controller.
+	// ControllerInfoFromModelUUID uses the given model UUID to return the address of the controller to
```
ale8k, line 35: ControllerInfoFromModelUUID returns controller connection details for the model. It probably shouldn't be returning a JWT but I guess a follow up.

internal/jimm/ssh/ssh.go
```diff
@@ -112,3 +119,39 @@ func (s *sshManager) ControllerInfoFromModelUUID(ctx context.Context, modelUUID
 		JWT:       string(jwt),
 	}, nil
 }
+
+// DialControllerSSHServer dials the controller and returns
+// an SSH connection.
+// We don't accept user input for the port to disallow the
```
ale8k: What does this mean? Surely this isn't the reason we don't allow the specifying of ports?

ale8k (approved):
lgtm some comments though, please do little godoc ones

PR #1549, opened 2025-02-11
fix tiobe scan

SimoneDutto:
In this PR we fix tiobe scan by:
- installing docker and docker compose
- running go test
- converting and copying coverage file to the right folder
- running the tiobe scan

I've run the pipeline to see it works: https://github.com/canonical/jimm/actions/runs/13260006484/job/37014210416
And you can check the results here https://canonical.tiobe.com/tiobeweb/TICS/Explorer.html#axes=ProjectGroup(PrivateRepos),Project(jimm)

> It takes 1h> to finish

## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests

### DIFF
--- a/.github/workflows/tiobe-scan.yaml
+++ b/.github/workflows/tiobe-scan.yaml
@@ -24,6 +24,23 @@ jobs:
         go install github.com/axw/gocov/gocov@v1.1.0
         go install github.com/AlekSi/gocov-xml@v1.1.0
 
+    - name: Install Docker
+      run: |
+        sudo apt-get update -y
+        sudo apt-get install -y \
+          ca-certificates \
+          curl \
+          gnupg \
+          lsb-release
+        sudo mkdir -p /etc/apt/keyrings
+        curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo gpg --dearmor -o /etc/apt/keyrings/docker.gpg
+        echo \
+          "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/docker.gpg] https://download.docker.com/linux/ubuntu \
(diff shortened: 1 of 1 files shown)


.github/workflows/tiobe-scan.yaml
```diff
@@ -24,6 +24,23 @@ jobs:
...
+          curl \
+          gnupg \
+          lsb-release
+        sudo mkdir -p /etc/apt/keyrings
+        curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo gpg --dearmor -o /etc/apt/keyrings/docker.gpg
+        echo \
+          "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/docker.gpg] https://download.docker.com/linux/ubuntu \
+          $(lsb_release -cs) stable" | sudo tee /etc/apt/sources.list.d/docker.list > /dev/null
+        sudo apt-get update -y
+        sudo snap install docker # installing via apt doesn't work, because you can't start systemctl services.
```
kian99, line 41: That's weird, why can't we start systemctl services? Let's also add a comment on the top of this step explaining why we're installing docker + compose

ale8k: Depending on the runner you use, you won't need to install docker. Are we using self-hosted here?

alesstimec (approved):

ale8k (approved):
lgtm but I don't think we need to install docker

.github/workflows/tiobe-scan.yaml
SimoneDutto: i have run too many tests and i can't find the exact run with the error.
But if i don't install docker via snap the error is something like "docker deamon is not running"
If i try to start the docker.service it says something like "service cannot be started, bla bla".
I have found this "snap install docker" in the ua-contract repo, and it worked.

PR #1550, opened 2025-02-11
fix: add limit listener tests

kian99:
## Description
In #1548 (which this PR builds on), the limit listener was moved to a separate package. In this PR I've done 2 things:
- Copied over the existing set of tests from the original implementation [here](https://cs.opensource.google/go/x/net/+/refs/tags/v0.34.0:netutil/listen_test.go).
- Added the test `TestLimitListenerTimeout`.
- Fixed a bug in the limit listener, when it reached saturation new connections would be dropped even if old connections were closed.

We're only using the limit listener so that clients who attempt to connect once we reach our maximum number of connections don't get blocked and instead get timed-out after a delay if no connection opens up. We can consider using the original implementation of the limit listener since clients shouldn't be blocked for too long with medium load as seats become free. With high load we may need to re-consider this.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

### DIFF
--- a/internal/ssh/limitlistener/listener.go
+++ b/internal/ssh/limitlistener/listener.go
@@ -29,6 +29,7 @@
 package limitlistener
 
 import (
+	"errors"
 	"net"
 	"sync"
 	"time"
@@ -36,7 +37,7 @@ import (
 
 // N.B.:
 // This is a copypaste of netutil.LimiLister (link: https://cs.opensource.google/go/x/net/+/refs/tags/v0.34.0:netutil/listen.go),
-// but we add a timeout so when we are at the limit we actively close connections instead of waiting indefinetely. (Look at line 44)
+// but we add a timeout so when we are at the limit we actively close connections instead of waiting indefinetely.
 
 // ListenerWithTimeout returns a Listener that accepts at most n simultaneous
 // connections from the provided Listener, and it timeouts when the max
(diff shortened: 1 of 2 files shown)


internal/ssh/limitlistener/listener.go
```diff
@@ -61,22 +62,23 @@ type limitListener struct {
 // acquire acquires the limiting semaphore. Returns true if successfully
 // acquired, false if the listener is closed and the semaphore is not
 // acquired.
```
SimoneDutto: can you update the godoc here?

SimoneDutto (approved):
nice work!

internal/ssh/limitlistener/listener_test.go
```diff
@@ -0,0 +1,409 @@
...
+		timeout         = 1 * time.Second
+	)
+
+	l, err := net.Listen("tcp", "127.0.0.1:0")
+	if err != nil {
+		t.Fatal(err)
+	}
+	l = ListenerWithTimeout(l, max, timeout)
+
+	acceptDone := make(chan struct{})
```
alesstimec, line 159: why do we need this channel?

internal/ssh/limitlistener/listener_test.go
```diff
@@ -0,0 +1,409 @@
...
+		timeout = 10 * time.Millisecond
+	)
+
+	l, err := net.Listen("tcp", "127.0.0.1:0")
+	if err != nil {
+		t.Fatal(err)
+	}
+	l = ListenerWithTimeout(l, max, timeout)
+
+	acceptDone := make(chan struct{})
```
alesstimec, line 265: same question

alesstimec (approved):
LGTM with a question

internal/ssh/limitlistener/listener_test.go
kian99: This test comes from the library code. The usefulness of this channel is as a way to ensure all the go routines are terminated at the end of the test, essentially like an err group. I could have used an errgroup but kept with the semantics of the library tests.
When the test ends, the defer closes the listener then waits on the channel. The go routine has a `defer close(acceptDone)` that will unblock the defer if the Go routine itself finishes.

PR #1552, opened 2025-02-12
chore: extract security scan workflow into jimm

kian99:
## Description
This PR removes the dependency on the comsys-build-tools repo and place the security scanning steps directly into JIMM's repo. This reduces the coupling to that repo and lowers the complexity with less branching logic based on input parameters.

### DIFF
--- a/.github/workflows/security-scan-weekly.yaml
+++ b/.github/workflows/security-scan-weekly.yaml
@@ -6,21 +6,79 @@ on:
   workflow_dispatch:
 
 jobs:
-  security-scan:
+  repo-security-scan:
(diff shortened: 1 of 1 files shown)


ale8k (approved):
lgtm - maybe just check if we can drop the mirror (idk if we can or if it does any harm really but yeah)

kian99:
> lgtm - maybe just check if we can drop the mirror (idk if we can or if it does any harm really but yeah)

Choosing to keep it for now as we introduced it to avoid rate-limiting issues on the upstream source.

SimoneDutto (approved):

PR #1553, opened 2025-02-25
chore: rename nested commands to be top-level

kian99:
## Description

This PR is a step towards moving `jimmctl` into the `jaas` plugin as described in [the spec](https://docs.google.com/document/d/14Xj3KkiTZaJLNQHTCMTecIfc3v3owgj2jGJuPU8I-QE/edit?tab=t.0).

In this PR we move all auth related commands to top-level commands. From `jimmctl auth group list` to `juju list-groups`. 

Additionally, all `relation` related commands were renamed to `permissions` i.e. `jimmctl auth relation add` -> `juju add-permission`.

Partially addresses [JUJU-7479](https://warthogs.atlassian.net/browse/JUJU-7479)

[JUJU-7479]: https://warthogs.atlassian.net/browse/JUJU-7479?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


alesstimec (approved):

cmd/jimmctl/cmd/group.go
```diff
@@ -21,10 +20,6 @@ import (
 )
 
 const (
-	groupDoc = `
-The group command enables group management for jimm
-`
-
 	addGroupDoc = `
```
ale8k: I think... Now this is going top level, and could become confusing to juju users let's get super explicit about these commands. And also format it somewhat similar to juju's existing commands like add-model:
```
❯ jj add-model -h
Usage: juju add-model [options] <model name> [cloud|region|(cloud/region)]

Summary:
Adds a workload model.
```
They don't say "the X command does Y", they just summarise it. Or should we change wording in a follow up?

cmd/jimmctl/cmd/relation.go
```diff
@@ -52,7 +52,7 @@ E.g. "user-Alice" or "controller-MyController"
         }
     ]
 
-Certain constraints apply when creating/removing a relation, namely:
+Certain constraints apply when creating/removing permissions, namely:
```
ale8k: Let's stick to one word, permission or access, we say here permissions but on line 87 we say access. I'd prefer we get our language totally ubiquitous as Juju is already a confusing beast.

cmd/jimmctl/cmd/relation.go
```diff
@@ -84,85 +84,69 @@ If target_object is an application offer, the relation can be one of:
...
-This will grant/revoke the relation to all users within TeamA:
+This will grant/revoke access to all users within TeamA:
 
 	group-TeamA#member administrator controller-MyController
 `
 
-	addRelationDoc = `
-The add command adds relation to jimm.
+	addPermissionDoc = `
+Adds access to a resource.
```
ale8k: Again we say addPermissionDoc, but then it says Adds access to a resource. It's also more "granting" rather than adding right? I'd go with sometihng more like Juju, they explain what you can grant to in their grant doc:
```
Grants access level to a Juju user for a model, controller, or application offer.
```

So when we settle on permission or access, let's say "<Grants/Adds> <permission/access> level to a JIMM user for a model, etc, etc,

SimoneDutto (commented):
you should update the scripts that use these commands.
For example: `./local/jimm/setup-service-account`, but maybe there are others.
This one about the setup-service account is important for terraform tests

ale8k (changes requested):
Instead of littering this PR with comments, I've left 3 with my main concerns, could you skim over the the role command too ensuring language is ubiquitous as it'll probably become the de facto language we use from then on (esp once JIMM see's actual users).

cmd/jimmctl/cmd/relation.go
ale8k: I also don't really mind which, but just some consistency and caution as I think this will end up in stone honestly.

cmd/jimmctl/cmd/relation.go
kian99: I would say we can't stick to one word here because the two words are complementary.
Just looking at the English definitions,

Permission
>the action of officially allowing someone to do a particular thing; consent or authorization.

Access
>the means or opportunity to approach or enter a place.

So granting _permission_ provides _access_, at least that seems right. So addressing your next comment, I'll change the language to `Grants access` since `Adds permission` sounds wrong (probably because I've got it in my head now that permission is sort of a verb).

Also a heads up that the help text for `add-permission` says `Grants access...` and the docstring for `list-permissions` says `Lists permissions` in case that's too terrible and we need to change it or else bad things will happen.

cmd/jimmctl/cmd/group.go
kian99: Good point, fixed for all the commands to just `Adds a group.` from `The add command adds group to jimm.`

ale8k (approved):
lgtm, I was thinking we can improve the purpose docs to be consistent and a little more detailed but I think that's for later.

cmd/jaas/cmd/relation.go
```diff
@@ -31,10 +31,13 @@ const (
 )
 
 const (
-	relationDoc = `
-relation command enables relation management for jimm
-`
 	genericConstraintsDoc = `
+This command work at a low-level and commands like 'juju grant'
```
alesstimec: s/work/works/

cmd/jaas/cmd/relation.go
```diff
@@ -84,85 +87,69 @@ If target_object is an application offer, the relation can be one of:
...
-	cmd.Register(newCheckRelationCommand())
-	cmd.Register(newListRelationsCommand())
-
-	return cmd
-}
-
-// newAddRelationCommand returns a command to add a relation.
-func newAddRelationCommand() cmd.Command {
-	cmd := &addRelationCommand{
+// NewAddPermissionCommand returns a command to add a relation.
```
alesstimec: "add a relation" vs "add a permission".. while "relation" is technically correct, we should also consider if we're introducing too much terminology to describe the same concept..

alesstimec (approved):
LGTM with a few comments

SimoneDutto (approved):

cmd/jaas/cmd/relation.go
kian99: Changed the use of `relation` in docstrings to `grant access`, `revoke access` or `lists permissions` where appropriate.

PR #1557, opened 2025-02-26
dump logs to github artifacts

SimoneDutto:
## Description
add step to github workflow to dump logs to a github artifact to make it easier to debug.

## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests

### DIFF
--- a/.github/actions/test-server/action.yaml
+++ b/.github/actions/test-server/action.yaml
@@ -37,6 +37,11 @@ outputs:
 runs:
   using: "composite"
   steps:
+    - name: Set container name
+      id: set-container-name
+      run: echo "container_name=$([[ '${{ inputs.jimm-version }}' == 'dev' ]] && echo 'jimm' || echo 'jimm-test')" >> $GITHUB_ENV
+      shell: bash
+
     - name: Login to GitHub Container Registry
       if: ${{ inputs.jimm-version != 'dev' }}
       uses: docker/login-action@v3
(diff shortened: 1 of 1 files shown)


alesstimec (approved):

.github/actions/test-server/action.yaml
```diff
@@ -136,3 +136,16 @@ runs:
       run: ./local/jimm/setup-service-account.sh
       working-directory: ${{ github.action_path }}/../../..
       shell: bash
+    
+    - name: Dump JIMM container logs
+      if: always()
```
kian99, line 146: The logs could be very long if this action is used in a workflow like the Terraform provider where it does integration testing.

kian99 (approved):

.github/actions/test-server/action.yaml
SimoneDutto: i'm working on making it a "post-setup" action.
because now the logs are just the one generated in this very workflow, without the things after that are using this workflow.

.github/actions/test-server/action.yaml
```diff
@@ -37,6 +37,11 @@ outputs:
 runs:
   using: "composite"
   steps:
+    - name: Set container name
+      id: set-container-name
+      run: echo "container_name=$([[ '${{ inputs.jimm-version }}' == 'dev' ]] && echo 'jimm' || echo 'jimm-test')" >> $GITHUB_ENV
```
kian99, line 42: Maybe we should update the docker compose so that both both profiles start jimm with the same name? Just `jimm`.

.github/actions/test-server/action.yaml
SimoneDutto: we can merge it as it now, so we have the debug log for our issue. and then work on the post setup

PR #1560, opened 2025-02-27
feat: surface all errors

kian99:
## Description
This is a test PR to see what kinds of things break if we were to surface all errors.

## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests

### DIFF
--- a/internal/errors/errors_test.go
+++ b/internal/errors/errors_test.go
@@ -1,8 +1,9 @@
-// Copyright 2024 Canonical.
+// Copyright 2025 Canonical.
 
 package errors_test
 
 import (
+	stderr "errors"
 	"testing"
 
 	qt "github.com/frankban/quicktest"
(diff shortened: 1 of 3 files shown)


internal/errors/errors_test.go
```diff
@@ -35,3 +36,18 @@ func TestE(t *testing.T) {
...
+	err := stderr.New("foo")
+
+	code := errors.Code("test code")
+	err = errors.E(errors.Op("test.op"), code, "reason", err)
+	c.Check(err.Error(), qt.Equals, `reason: foo`)
+	c.Check(errors.ErrorCode(err), qt.Equals, code)
+
+	err = errors.E(errors.Op("test.op2"), err, "more info")
+	c.Check(err.Error(), qt.Equals, `more info: reason: foo`)
+	c.Check(errors.ErrorCode(err), qt.Equals, code)
```
SimoneDutto, line 52: can you add two tests with errors concatenated?
- Use the output of errors.E for another errors.E to make sure we are seeing all output.
- Use the output of errors.E for another std.error.New() to see if it collects the error and the message or just the error

PR #1564, opened 2025-03-03
add model uuid to ssh key

SimoneDutto:
## Description

In this PR we add model uuid to ssh key table.
This is to be consistent with juju 3.6 behaviour.

The approach is to add a new struct to ssh keys method, called SSHKeyModel.
It contains two fields:
- ModelUUID: string
- All: boolean

And passing this struct to ssh keys methods to scope the operation on a single model or every models.

This is to facilitate behaviour where we want to query all ssh keys for user, independently of the model.
So, it should be easy to reuse this code for juju 4 as well.

I've QAed locally, and by landing this PR, the failing ssh keys tests in the terraform provider pass.
```
Running tool: /usr/local/go/bin/go test -timeout 120s -run ^TestAcc_ResourceSSHKey$ github.com/juju/terraform-provider-juju/internal/provider

ok  	github.com/juju/terraform-provider-juju/internal/provider	3.897s
```

### DIFF
--- a/internal/db/sshkeys.go
+++ b/internal/db/sshkeys.go
@@ -22,13 +22,18 @@ func (d *Database) AddSSHKey(ctx context.Context, sshKey *dbmodel.SSHKey) (err e
 	defer servermon.ErrorCounter(servermon.DBQueryErrorCount, &err, string(op))
 
 	if err := d.DB.WithContext(ctx).Create(sshKey).Error; err != nil {
-		return errors.E(op, dbError(err))
+		dbErr := dbError(err)
+		if errors.ErrorCode(dbErr) == errors.CodeAlreadyExists {
+			// we don't return an error if a user tries to add the same key twice.
+			return nil
+		}
+		return errors.E(op, dbErr)
 	}
 	return nil
 }
 
 // RemoveSSHKeyByFingerprint removes a user's ssh key identified by its fingerprint.
-func (d *Database) RemoveSSHKeyByFingerprint(ctx context.Context, identityName string, fingerprint string) (err error) {
+func (d *Database) RemoveSSHKeyByFingerprint(ctx context.Context, identityName string, model SSHKeyModelFilter, fingerprint string) (err error) {
 	const op = errors.Op("db.RemoveSSHKeyByFingerprint")
 
 	if err := d.ready(); err != nil {
@@ -40,6 +45,7 @@ func (d *Database) RemoveSSHKeyByFingerprint(ctx context.Context, identityName s
 	defer servermon.ErrorCounter(servermon.DBQueryErrorCount, &err, string(op))
 
 	query := d.DB.Where("identity_name = ?", identityName).
+		Where("model_uuid = ?", model.ModelUUID).
 		Where("md5_fingerprint = ?", fingerprint).
 		Delete(&dbmodel.SSHKey{})
 
@@ -55,7 +61,7 @@ func (d *Database) RemoveSSHKeyByFingerprint(ctx context.Context, identityName s
 }
 
 // RemoveSSHKeyByComment removes a user's ssh key identified by its comment.
-func (d *Database) RemoveSSHKeyByComment(ctx context.Context, identityName string, comment string) (err error) {
+func (d *Database) RemoveSSHKeyByComment(ctx context.Context, identityName string, model SSHKeyModelFilter, comment string) (err error) {
 	const op = errors.Op("db.RemoveSSHKeyByComment")
 
 	if err := d.ready(); err != nil {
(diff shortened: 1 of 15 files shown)


internal/dbmodel/sql/postgres/023_add_model_ssh_keys.up.sql
```diff
@@ -0,0 +1,7 @@
+-- add model association to user ssh keys
+ALTER TABLE ssh_keys DROP CONSTRAINT unique_identity_ssh_key;
+
+ALTER TABLE ssh_keys ADD COLUMN model_uuid TEXT,
+    ADD CONSTRAINT fk FOREIGN KEY(model_uuid) REFERENCES models(uuid) ON DELETE CASCADE;
```
kian99: Could we format this slightly differently 
```suggestion
ALTER TABLE ssh_keys 
    ADD COLUMN model_uuid TEXT,
    ADD CONSTRAINT fk FOREIGN KEY(model_uuid) REFERENCES models(uuid) ON DELETE CASCADE;
```

internal/db/sshkeys.go
```diff
@@ -22,13 +22,18 @@ func (d *Database) AddSSHKey(ctx context.Context, sshKey *dbmodel.SSHKey) (err e
...
+			return nil
+		}
+		return errors.E(op, dbErr)
 	}
 	return nil
 }
 
 // RemoveSSHKeyByFingerprint removes a user's ssh key identified by its fingerprint.
-func (d *Database) RemoveSSHKeyByFingerprint(ctx context.Context, identityName string, fingerprint string) (err error) {
+func (d *Database) RemoveSSHKeyByFingerprint(ctx context.Context, identityName string, model dbmodel.SSHKeyModel, fingerprint string) (err error) {
```
kian99: Let's rather pass in `modelUUID string` rather than a whole model object.

SimoneDutto: the name is poor but i didn't have any better, but the scope of this argument is what is described in the PR description

SimoneDutto: it is not the whole model, it is just two fields

kian99: Ah I see it now

internal/jimm/sshkeys/sshkeys_test.go
```diff
@@ -38,41 +40,40 @@ func (s *sshKeysManagerSuite) Init(c *qt.C) {
 	c.Assert(err, qt.IsNil)
 
 	s.db = db
-
+	user, _, _, model, _, _, _, _ := jimmtest.CreateTestControllerEnvironment(ctx, c, s.db)
```
kian99, line 43: I'm not the biggest fan of this test method if we can avoid it.

kian99 (approved):

internal/db/sshkeys.go
```diff
@@ -22,13 +22,18 @@ func (d *Database) AddSSHKey(ctx context.Context, sshKey *dbmodel.SSHKey) (err e
 	defer servermon.ErrorCounter(servermon.DBQueryErrorCount, &err, string(op))
 
 	if err := d.DB.WithContext(ctx).Create(sshKey).Error; err != nil {
-		return errors.E(op, dbError(err))
+		dbErr := dbError(err)
+		if errors.ErrorCode(dbErr) == errors.CodeAlreadyExists {
+			// we don't return an error if a user tries to add the same key twice.
+			return nil
```
alesstimec, line 28: :+1:

internal/dbmodel/sshkeys.go
```diff
@@ -16,10 +16,20 @@ type SSHKey struct {
...
 	// PublicKey holds the user's public SSH key.
 	PublicKey []byte `gorm:"uniqueIndex:unique_identity_ssh_key"`
 	// MD5Fingerprint is the MD5 fingerprint of the public key.
 	MD5Fingerprint string
 	// KeyComment holds a user provided comment.
 	KeyComment string
 }
+
+// SSHKeyModel holds the model UUID for the SSH Key or a flag to list all keys independent of the model.
+type SSHKeyModel struct {
```
alesstimec: could we move this type to the db package and keep dbmodel package only for those types that get stored in the database?

internal/db/sshkeys.go
alesstimec: how about we just have `dbmodel.SSHKey` as parameter?

internal/db/sshkeys.go
```diff
@@ -55,7 +61,7 @@ func (d *Database) RemoveSSHKeyByFingerprint(ctx context.Context, identityName s
 }
 
 // RemoveSSHKeyByComment removes a user's ssh key identified by its comment.
-func (d *Database) RemoveSSHKeyByComment(ctx context.Context, identityName string, comment string) (err error) {
+func (d *Database) RemoveSSHKeyByComment(ctx context.Context, identityName string, model dbmodel.SSHKeyModel, comment string) (err error) {
```
alesstimec: same here.. we could just have `dbmodel.SSHKey` as parameter

internal/dbmodel/sshkeys.go
alesstimec: or maybe we don't even need it.

alesstimec (approved):
LGTM with a few suggestions

internal/db/sshkeys.go
SimoneDutto: I can't express the `All` flag via the `dbmodel.SSHKey`

internal/dbmodel/sshkeys.go
SimoneDutto: i'm moved this struct to the db pkg

PR #1566, opened 2025-03-04
add option to bootstrap controller via charmed k8s action

SimoneDutto:
## Description
Add an option to bootstrap the controller via charmed action or not.
The idea is to set this option to false in juju action where we build juju manually and we can then bootstrap the controller with that build.

By making this changes, we are able to add smoke test to juju with a running jaas and a dev-built bootstrapped juju controller.

You can check it here: https://github.com/juju/juju/actions/runs/13655481588/job/38173523588?pr=19051

## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests

### DIFF
--- a/.github/actions/test-server/action.yaml
+++ b/.github/actions/test-server/action.yaml
@@ -23,6 +23,13 @@ inputs:
     description: 'Dump JIMM container logs at the end of the action.'
     required: false
     default: "false"
+  use-charmed-k8s-action:
+    description: |
+      If set to true the action will use the charmed-kubernetes/actions-operator to bootstrap a Juju controller via Juju snap.
+      If set to false the workflow calling this action needs to provide the Juju binary in the PATH, that the action will use 
+      to bootstrap a controller.
+    required: false
+    default: 'true'
 
 outputs:
   url:
@@ -87,7 +94,9 @@ runs:
         sudo lxd init --auto && \
         sudo chmod a+wr /var/snap/lxd/common/lxd/unix.socket && \
         lxc network set lxdbr0 ipv6.address none && \
-        sudo usermod -a -G lxd $USER
+        sudo usermod -a -G lxd $USER && \
+        sudo iptables -F FORWARD && \
+        sudo iptables -P FORWARD ACCEPT
       shell: bash
 
     - name: Setup cloud-init script for bootstraping Juju controllers
@@ -95,7 +104,8 @@ runs:
       shell: bash
       working-directory: ${{ github.action_path }}/../../..
       env:
-        SKIP_BOOTSTRAP: true
+        # skip bootstrap in case we use the charmed k8s action
+        SKIP_BOOTSTRAP: ${{ inputs.use-charmed-k8s-action == 'true'}} 
         CLOUDINIT_FILE: "cloudinit.temp.yaml"
 
     # See: https://github.com/charmed-kubernetes/actions-operator/issues/82
@@ -105,6 +115,7 @@ runs:
         python-version: ${{ inputs.python-version }}
 
     - name: Setup Juju Controller
+      if: ${{ inputs.use-charmed-k8s-action == 'true' }}
       uses: charmed-kubernetes/actions-operator@main
       with:
         provider: "lxd"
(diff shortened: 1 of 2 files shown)


.github/actions/test-server/action.yaml
```diff
@@ -87,15 +91,18 @@ runs:
         sudo lxd init --auto && \
         sudo chmod a+wr /var/snap/lxd/common/lxd/unix.socket && \
         lxc network set lxdbr0 ipv6.address none && \
-        sudo usermod -a -G lxd $USER
+        sudo usermod -a -G lxd $USER && \
+        sudo iptables -F FORWARD && \
```
alesstimec, line 98: how is this related to the boostrap config options?

SimoneDutto: I'll try my best to explain.
With this changes, we don't use the `charmed-kubernetes/actions-operator` action, but we rather bootstrap a controller using the lxd installed in this step.
- from ubuntu 22.04 we need this as well otherwise networking in lxd won't work.
- i've copied these commands from the action `charmed-kubernetes/actions-operator` https://github.com/charmed-kubernetes/actions-operator/blob/1c7c9a30d7d233e26e7a4fc1505cc44bbd937229/src/bootstrap/index.ts#L68-L73

A possible question you could have is:
Why was it working before then? Well before we were installing lxd in this step but then the `charmed-kubernetes/actions-operator` would have configure lxd again, so it was setting using this command as well.

Hope it's clear.

SimoneDutto: a thing i don't know is "what happens if they use this action in an ubuntu 20.04 machine?".
Is it worth to investigate? Or are we sure it is used basically by us on "newer versions"?

.github/actions/test-server/action.yaml
```diff
@@ -23,6 +23,10 @@ inputs:
     description: 'Dump JIMM container logs at the end of the action.'
     required: false
     default: "false"
+  bootstrap-controller-via-charmed-kubernetes-action:
+    description: 'Whether we bootstrap a controller to add to jimm via the charmed-kubernetes action.'
+    required: false
+    default: 'true'
```
kian99: I'd change this flag to `bootstrap-via-snap-juju`

SimoneDutto: it is not accurate though, because you could potentially install juju via snap and set this flag to false, and it will work as well.

kian99: Fair, but a shorter would be better imo. Maybe `use-released-juju-version`, and then update the description to mention when you might want to use this, e.g. "If this option is set to false, the workflow calling this action is expected to provide a Juju binary on the PATH that will be used to bootstrap a controller and add it to jimm".

SimoneDutto: let me brainstorm a name with my friend `gemini` (@alesstimec look away)

SimoneDutto:
```
use-charmed-k8s-action:
    description: |
      true: Use charmed-kubernetes/actions-operator to install juju and bootstrap a controller.
      false: Use existing Juju in the PATH and bootstrap a controller via QA scripts.
```
wdyt?

kian99: Name is fine, but the `true/false` description is confusing. I'd still suggest just giving it to the user in simple language, what happens if they set this value to false and why they might want to do that.

SimoneDutto: done.

alesstimec (approved):

.github/actions/test-server/action.yaml
```diff
@@ -23,6 +23,13 @@ inputs:
     description: 'Dump JIMM container logs at the end of the action.'
     required: false
     default: "false"
+  use-charmed-k8s-action:
+    description: |
+      If sets to true the action will use the charmed-kubernetes/actions-operator to bootstrap a Juju controller via Juju snap.
+      If sets to false the action you need to provide the juju binary in the PATH, and the action will 
+      bootstrap a controller with that juju binary.
```
kian99:
```suggestion
      If set to true the action will use the charmed-kubernetes/actions-operator to bootstrap a Juju controller via Juju snap.
      If set to false the workflow calling this action needs to provide the Juju binary in the PATH, that the action will use to bootstrap a controller.
```

kian99 (approved):

PR #1567, opened 2025-03-04
chore: copy jimmctl commands to jaas snap

kian99:
## Description
This PR is the start of the jimmctl migration to the jaas plugin. Here, I've copied all the commands verbatim to the jaas snap folder without wiring them up so that,
1. We maintain the existing functionality of the jimmctl snap
2. We can make follow-up PRs to shape the jaas plugin into it's new shape.

Unfortunately since this is a copy and not a move, I'm unsure how to preserve the history or whether its worth the effort.

### DIFF
--- a/cmd/jaas/cmd/addcloudtocontroller.go
+++ b/cmd/jaas/cmd/addcloudtocontroller.go
@@ -1,0 +1,220 @@
+// Copyright 2025 Canonical.
+
+package cmd
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/juju/cmd/v3"
+	"github.com/juju/gnuflag"
+	jujuapi "github.com/juju/juju/api"
+	"github.com/juju/juju/cloud"
+	jujucmd "github.com/juju/juju/cmd"
+	jujucmdcloud "github.com/juju/juju/cmd/juju/cloud"
+	jujucmdcommon "github.com/juju/juju/cmd/juju/common"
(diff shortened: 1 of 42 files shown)


alesstimec (commented):
I'd like us to preserve the history of the copied files.. so perhaps create a branch where you `git mv` files to the new location and then merge that back into your PR (see suggestions is https://stackoverflow.com/questions/16937359/git-copy-file-preserving-history)

kian99:
> I'd like us to preserve the history of the copied files.. so perhaps create a branch where you `git mv` files to the new location and then merge that back into your PR (see suggestions is https://stackoverflow.com/questions/16937359/git-copy-file-preserving-history)

@alesstimec I've tried this before and it didn't work, the comments on the post also suggest that it doesn't work in newer versions of git. This [other answer](https://stackoverflow.com/a/1043566) however sheds some light on how git can handle copies,
>Git does not do rename tracking nor copy tracking, which means it doesn't record renames or copies. What it does instead is rename and copy detection. You can request rename detection in git diff (and git show) by using the -M option, you can request additional copy detection in changed files by using the -C option instead, and you can request more expensive copy detection among all files with -C -C. See the [git-diff](http://git-scm.com/docs/git-diff) manpag

alesstimec (approved):
:shipit:

SimoneDutto (approved):
looking good, all the 6 thousands lines

PR #1568, opened 2025-03-04
make snap jaas executable

SimoneDutto:
## Description
`jaas` snap was not executable by itself as it is `jimmctl`, it was just a juju plugin. This pr makes jaas executable by itself.
The reason we need this is to make sure we can use jaas commands by doing `jaas add-service-account` instead `juju add-service-account`, and this is particularly useful when we build juju in devmode and we can't access snap jaas plugin commands via juju.

To accomplish that I've basically c/p jimmctl snapcraft.yaml

## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests

### DIFF
--- a/snaps/jaas/snapcraft.yaml
+++ b/snaps/jaas/snapcraft.yaml
@@ -3,8 +3,7 @@ summary: JAAS plugin
 description: Juju plugin for providing JAAS functionality to the Juju CLI.
 version: git
 grade: stable
-base: bare
-build-base: core20
+base: core20
 confinement: strict
 
 slots:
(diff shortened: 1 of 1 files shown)


ale8k (commented):
lgtm but @alesstimec @kian99 switch base to core20, so it'll be a bit bigger. Simone's issue was something within core20 is enabling exec hooks, hence it fixed it. The other plugs don't appear to be needed but we will need them when we move jimmctl into this. How do you guys feel on a core20 base?

SimoneDutto:
> lgtm but @alesstimec @kian99 switch base to core20, so it'll be a bit bigger. Simone's issue was something within core20 is enabling exec hooks, hence it fixed it. The other plugs don't appear to be needed but we will need them when we move jimmctl into this. How do you guys feel on a core20 base?

just as a note: it is the same as jimmctl is now

SimoneDutto:
we can close this one, because we've found another workaround

PR #1569, opened 2025-03-05
chore: ensure graceful server shutdown

kian99:
## Description
This PR fixes an issue where the JIMM server was not shutting down properly when receiving a SIGTERM signal. This was because all routines started by service.Go() need to end before service.Wait() will return. The SSH server was not being shutdown and blocking return. This is affecting our dev environment deployment where the service manager Pebble is waiting forever for JIMM to exit.

I've also made a change to avoid printing stacktraces in our production logger. These were present in our production logs and were a big clutter and making it very hard to read the logs. These weren't always turned on but I think this behaviour changed when we reworked our logger.

## Test instructions
I'll add a test for this in a follow-up but I've confirmed the issue by doing the following,
1. `docker compose logs -f jimm`

In a separate terminal 
1. `docker compose --profile dev up -d --wait`
3. `docker exec -it jimm bash`
4. `ps aux` (note that JIMM won't be PID 1 because of `air`)
5. `kill <jimm-pid>`
6. See that without these changes JIMM logs hang and with the change the server properly shuts down.

### DIFF
--- a/cmd/jimmsrv/main.go
+++ b/cmd/jimmsrv/main.go
@@ -30,7 +30,7 @@ func main() {
 	})
 	err := s.Wait()
 
-	zapctx.Error(context.Background(), "shutdown", zap.Error(err))
+	zapctx.Error(context.Background(), "jimm shutdown complete", zap.Error(err))
 	if _, ok := err.(*service.SignalError); !ok {
 		os.Exit(1)
 	}
@@ -207,19 +207,8 @@ func start(ctx context.Context, s *service.Service) error {
 		Handler:           jimmsvc,
 		ReadHeaderTimeout: time.Second * 5,
 	}
-	s.OnShutdown(func() {
-		ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
-		defer cancel()
 
-		zapctx.Warn(ctx, "server shutdown triggered")
-		err = httpsrv.Shutdown(ctx)
-		if err != nil {
-			zapctx.Error(ctx, "failed to shutdown server gracefully", zap.Error(err))
-		}
-		jimmsvc.Cleanup()
-	})
-	s.Go(httpsrv.ListenAndServe)
-	zapctx.Info(ctx, "Successfully started JIMM server")
+	// intentionally ignore the error because invalid values are handled by the jump server.
 	maxConccurentConncetions, _ := strconv.Atoi(os.Getenv("JIMM_SSH_MAX_CONCURRENT_CONNECTIONS"))
 
 	sshServer, err := ssh.NewJumpServer(ctx, ssh.Config{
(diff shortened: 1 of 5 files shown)


cmd/jimmsrv/main.go
```diff
@@ -207,30 +207,36 @@ func start(ctx context.Context, s *service.Service) error {
...
-		HostKey:                  []byte(hostKeyRaw),
-		MaxConcurrentConnections: maxConccurentConncetions,
-	}, jimmsvc.JIMM().SSHManager())
-	if err != nil {
-		return err
-	}
+	zapctx.Info(ctx, "Successfully started JIMM HTTP server")
 	s.Go(sshServer.ListenAndServe)
-	zapctx.Info(ctx, "Successfully started JIMM ssh server")
+	zapctx.Info(ctx, "Successfully started JIMM SSH server")
```
ale8k: Just "Started", you don't know if it's successful

cmd/jimmsrv/main.go
```diff
@@ -207,30 +207,36 @@ func start(ctx context.Context, s *service.Service) error {
...
 	s.OnShutdown(func() {
 		ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
 		defer cancel()
 
-		zapctx.Warn(ctx, "server shutdown triggered")
+		zapctx.Warn(ctx, "HTTP server shutdown triggered")
 		err = httpsrv.Shutdown(ctx)
 		if err != nil {
-			zapctx.Error(ctx, "failed to shutdown server gracefully", zap.Error(err))
+			zapctx.Error(ctx, "failed to shutdown HTTP server gracefully", zap.Error(err))
```
ale8k, line 227: shutdown implies graceful, can just say "failed to shutdown HTTP server" and same for SSH

cmd/jimmsrv/main.go
```diff
@@ -207,30 +207,36 @@ func start(ctx context.Context, s *service.Service) error {
 		Handler:           jimmsvc,
 		ReadHeaderTimeout: time.Second * 5,
 	}
+
+	maxConccurentConncetions, _ := strconv.Atoi(os.Getenv("JIMM_SSH_MAX_CONCURRENT_CONNECTIONS"))
```
ale8k, line 212: Spelling & why ignore err?

ale8k (changes requested):
Some comments to resolve

cmd/jimmsrv/main.go
kian99: No? Shutdown can be forced or graceful. In this case it wasn't graceful, something went wrong during shutdown.

cmd/jimmsrv/main.go
kian99: It was like this, I assume because the ssh server verifies this value and sets a default if it's out of bounds.

cmd/jimmsrv/main.go
kian99: I mean sure, but this wasn't the point of this PR.

cmd/jimmsrv/main.go
SimoneDutto: yep, kian is right

SimoneDutto (approved):

cmd/jimmsrv/main.go
ale8k: Usually the language "shutdown" just implies graceful? Terminate / Kill etc often are ungraceful? And it still attempted it gracefully?

ale8k (approved):

cmd/jimmsrv/main.go
kian99: No, inside the function, if the shutdown didn't complete by the time the context was cancelled then it returns an error to indicate shutdown didn't happen within the alloted time.

ale8k: Yup, if it was ungraceful, you wouldn't have any timer - it'd be "burn this thing with fire now!!" (See the Close() method)

PR #1570, opened 2025-03-05
chore: test graceful server shutdown

kian99:
## Description
This PR is a follow up to #1569 that tests that JIMM gracefully shuts down when it receives a SIGTERM signal.
In order to test shutdown we need to issue the signal to the JIMM process, this is most simply done when JIMM is PID 1 in a container and then we can do something like `docker stop jimm`.

This PR aims to do just that by modifying JIMM's Dockerfile so that we have one target which runs `air` and another target which just compiles and runs JIMM. I've labelled them `no-hot-reload` and `hot-reload` to indicate the purpose of air but I'm open to other naming suggestions.

In our Docker compose, we choose the target by suppling an environment variable, defaulting to the target with `air`. Our integration testing now uses the target without `air` where JIMM is PID 1 so that we can test shutdown.

## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests

### DIFF
--- a/.github/workflows/integration-test.yaml
+++ b/.github/workflows/integration-test.yaml
@@ -56,3 +56,19 @@ jobs:
         run: |
           # TODO(simonedutto): improve this test when juju ssh controller is implemented.
           nc -zv localhost 17022
+
+      - name: Stop the application container
+        run: docker stop jimm
+  
+      - name: Check logs for shutdown message
+        run: |
+          APP_CONTAINER=jimm
+          LOGS=$(docker logs "$APP_CONTAINER" -n 20)
+          echo "$LOGS"
+          
+          if echo "$LOGS" | grep -q "jimm shutdown complete"; then
+            echo "Graceful shutdown confirmed."
+          else
+            echo "Graceful shutdown message not found!"
+            exit 1
+          fi
--- a/local/jimm/Dockerfile
+++ b/local/jimm/Dockerfile
@@ -1,0 +1,13 @@
+# Installing delve in the image to avoid running go install in running containers
+FROM cosmtrek/air:latest AS dev
+
+RUN go install github.com/go-delve/delve/cmd/dlv@latest
+
+ENTRYPOINT [ "air" ]
+
+# Create a target without air that we use in smoke tests.
+FROM golang:1.24-bookworm as smoke
+
+COPY entrypoint.sh     /entry/entrypoint.sh
+
+CMD [ "/entry/entrypoint.sh" ]
--- a/.github/actions/test-server/action.yaml
+++ b/.github/actions/test-server/action.yaml
@@ -78,6 +78,8 @@ runs:
       working-directory: ${{ github.action_path }}/../../..
       run: make dev-env
       shell: bash
+      env: 
+        JIMM_TARGET: smoke
 
     - name: Retrieve server CA cert.
       id: fetch-cert
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -42,9 +42,12 @@ services:
     extends:
       file: docker-compose.common.yaml
       service: jimm-base
+    # In order to switch the target image you need to run compose with --build
+    # Changing target is mainly useful in CI, so local environments can just rely on the default.
     build:
-      context: .
-      dockerfile: ./local/Dockerfile
+      context: ./local/jimm/
+      dockerfile: Dockerfile
+      target: ${JIMM_TARGET:-dev}
     profiles: ["dev"]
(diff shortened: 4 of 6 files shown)


.github/workflows/test-graceful-shutdown.yaml
```diff
@@ -0,0 +1,44 @@
...
+
+# This workflow tests that JIMM shuts down gracefully, which is important
+# when running in certain orchestration environments, e.g. on K8s, with 
+# pebble, etc.
+# We don't test the development version of JIMM because it is complicated
+# to do so when running with 'air', instead opting to check the latest
+# image periodically.
+
+on:
+  workflow_dispatch:
```
SimoneDutto: we could run this when a tag is created, so each time we create a tag to release something we see the output of this test as well

kian99: Yeah but I chose not to overcomplicate it because it can't quite run when a tag is created, it needs to wait until after the workflow that creates the OCI image has successfully uploaded the image.
Possibly I can make this action run after that one, let me do that in fact.

SimoneDutto: but if we call this action from the release action we are sure latest is updated.
I wouldn't stress to much about it, but if it's easy enought it would be nice

SimoneDutto (approved):

.github/workflows/test-graceful-shutdown.yaml
kian99: Added it.

ale8k (commented):
We shouldn't do this, this should be a part of the smoke test and our smoke test should be running our rock.

kian99:
> We shouldn't do this, this should be a part of the smoke test and our smoke test should be running our rock.

Changed it as discussed. Not using the rock currently but the rock is tested with the k8s charm at least.

.github/workflows/integration-test.yaml
```diff
@@ -56,3 +56,19 @@ jobs:
         run: |
           # TODO(simonedutto): improve this test when juju ssh controller is implemented.
           nc -zv localhost 17022
+
+      - name: Stop the application container
+        run: docker stop jimm-test
```
SimoneDutto: here is now ` jimm-dev`

kian99: It's actually just `jimm` I believe.

SimoneDutto: yep, you are right

local/jimm/Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+RUN go install github.com/go-delve/delve/cmd/dlv@latest
+
+ENTRYPOINT [ "air" ]
+
+# Create a target without air that we use in smoke tests.
+FROM golang:latest as no-hot-reload
+
+COPY entrypoint.sh     /entry/entrypoint.sh
+
+CMD [ "/entry/entrypoint.sh" ]
```
ale8k, line 13: No need for a script, just put it in here please

local/jimm/Dockerfile
```diff
@@ -0,0 +1,13 @@
+# Installing delve in the image to avoid running go install in running containers
+FROM cosmtrek/air:latest AS hot-reload
```
ale8k: AS dev

next one AS smoke

ale8k (approved):
two minor comments then lgtm (approving but please do them)

local/jimm/Dockerfile
kian99: Unfortunately we can't, we can't build before the CMD/Entrypoint because we don't have the source code, it's mounted in. And I don't want a giant CMD.

local/jimm/Dockerfile
kian99: Sure

local/jimm/Dockerfile
```diff
@@ -0,0 +1,13 @@
+# Installing delve in the image to avoid running go install in running containers
+FROM cosmtrek/air:latest AS dev
+
+RUN go install github.com/go-delve/delve/cmd/dlv@latest
+
+ENTRYPOINT [ "air" ]
+
+# Create a target without air that we use in smoke tests.
+FROM golang:latest as smoke
```
alesstimec: should we instead go for a specific version? e.g. golang:1.24.1-alpine, so that we're aligned with out go.mod?

kian99: Yeah that's a good idea, I'll opt for 1.24-bookworm so that it's debian based rather than alpine with musl.

PR #1576, opened 2025-03-07
ignore logs for some paths

SimoneDutto:
## Description
In this PR we add a check for path to ignore in our HTTP logger.
Some of logs are really repetitive and annoying.

If you feel like they are somewhat useful feel free to drop this pr.

Example: 10 logs in 10 second
image

Example with the changes:
image

> fly-by change the log level for request with 200 status code, because it makes more sense.

## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests

### DIFF
--- a/internal/logger/http_logger.go
+++ b/internal/logger/http_logger.go
@@ -1,9 +1,10 @@
-// Copyright 2024 Canonical.
+// Copyright 2025 Canonical.
 
 package logger
 
 import (
 	"net/http"
+	"slices"
 	"time"
 
 	"github.com/go-chi/chi/v5/middleware"
@@ -20,6 +21,8 @@ type httpLogEntry struct {
 	request *http.Request
 }
 
+var pathsToIgnore = []string{"/", "/.well-known/jwks.json"}
+
 // Write is called when the request handler has finished.
 func (l *httpLogEntry) Write(status, bytes int, header http.Header, elapsed time.Duration, extra interface{}) {
 	fields := make([]zap.Field, 0)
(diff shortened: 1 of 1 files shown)


kian99 (commented):
I would change this slightly. The requests hitting the root and returning a 404 are from our docker compose, I wouldn't worry about that.
I also wouldn't do string matching in our log writer to avoid slowing it down. My suggestion would be to just skip logging HTTP 200 codes.

SimoneDutto:
> I would change this slightly. The requests hitting the root and returning a 404 are from our docker compose, I wouldn't worry about that. I also wouldn't do string matching in our log writer to avoid slowing it down. My suggestion would be to just skip logging HTTP 200 codes.

i woudn't do that, even when 200 is useful to see requests coming through to understand what's going on.
If you don't agree with removing the logs from some path, i'll just close the pr, no worries.

kian99:
> > I would change this slightly. The requests hitting the root and returning a 404 are from our docker compose, I wouldn't worry about that. I also wouldn't do string matching in our log writer to avoid slowing it down. My suggestion would be to just skip logging HTTP 200 codes.
> 
> i woudn't do that, even when 200 is useful to see requests coming through to understand what's going on. If you don't agree with removing the logs from some path, i'll just close the pr, no worries.

I would leave the 200 codes to metrics, we don't need a log line for each successful request. Even for non 200 codes tbh, I would probably just leave it to metrics.

SimoneDutto:
> > > I would change this slightly. The requests hitting the root and returning a 404 are from our docker compose, I wouldn't worry about that. I also wouldn't do string matching in our log writer to avoid slowing it down. My suggestion would be to just skip logging HTTP 200 codes.
> > 
> > 
> > i woudn't do that, even when 200 is useful to see requests coming through to understand what's going on. If you don't agree with removing the logs from some path, i'll just close the pr, no worries.
> 
> I would leave the 200 codes to metrics, we don't need a log line for each successful request. Even for non 200 codes tbh, I would probably just leave it to metrics.

I think it's pretty useful:
- in a dev env you have your log level set to debug, and see all the requests
- in a production env you have your log level higher and you don't see much other than failed requests, which is useful if you are analysing something

kian99:
I can see the benefits. So I think just close this PR and the extra few log lines we'll live with.

PR #1577, opened 2025-03-10
chore: rename jaas plugin commands

kian99:
## Description
This PR builds on the work in #1553 and renames more commands so that they match their expected usage as a plugin to the Juju CLI.

It also removes the `import-cloud-credential` command as agreed in the spec.

Partially addresses [JUJU-7479](https://warthogs.atlassian.net/browse/JUJU-7479)

[JUJU-7479]: https://warthogs.atlassian.net/browse/JUJU-7479?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cmd/jaas/cmd/addcloudtocontroller.go
```diff
@@ -27,16 +27,15 @@ import (
 
 const (
 	addCloudToControllerCommandDoc = `
-The add-cloud-to-controller command adds the specified cloud to a specific 
-controller on jimm.
+Adds the specified cloud to a specific controller on jimm.
```
alesstimec: suggestin: we should use uniform spelling of jimm.. either "JIMM" or "jimm", but it needs to be consistent in help text, at least.

alesstimec (approved):
LGTM with a note.

cmd/jaas/cmd/addcloudtocontroller.go
kian99: Yip, I'll change it to `JIMM`.

cmd/jaas/cmd/migratemodel.go
```diff
@@ -24,9 +24,9 @@ You may specify a model name (of the form owner/name) or model UUID.
 
 `
 	migrateModelCommandExample = `
-    jimmctl migrate mycontroller 2cb433a6-04eb-4ec4-9567-90426d20a004 fd469983-27c2-423b-bebf-84f616fb036b ...
```
SimoneDutto, line 27: few lines above you should change the description to match the current path that we have

cmd/jaas/cmd/relation.go
```diff
@@ -120,7 +120,7 @@ Verifies access to a resource.
 `
 
 	listPermissionsDoc = `
-List permissions known to jimm. Using the "target", "relation" and "object" flags, 
+List permissions known to JIMM. Using the "target", "relation" and "object" flags,
```
SimoneDutto, line 123:
```suggestion
Lists permissions known to JIMM. Using the "target", "relation" and "object" flags, 
```

internal/testutils/jimmtest/openfga.go
```diff
@@ -1,4 +1,5 @@
-// Copyright 2024 Canonical.
+// Copyright 2025 Canonical.
```
SimoneDutto, line 1: why is the copyright updated for some files and not for others?

SimoneDutto (approved):
this pr made me realize how many jimmctl commands we have I've never seen/used.

Some comments

internal/testutils/jimmtest/openfga.go
kian99: I don't know :D
Probably because of find-replace that I did. I won't bother trying to fix all of these rn.

cmd/jaas/cmd/migratemodel.go
kian99: I'm lost? What path?

cmd/jaas/cmd/relation.go
kian99: I'll fix this in a follow-up since it's small.

cmd/jaas/cmd/migratemodel.go
SimoneDutto: *pattern sorry

SimoneDutto: cmd/jaas/cmd/migratemodel.go:20 instead of 
"The migrate commands migrates a model, or many models between two controllers
registered within JIMM." -> "migrates a model, or many models"

PR #1581, opened 2025-03-18
chore: parse public DNS name input

kian99:
## Description
This PR validates the `PublicDNSName` variable supplied to JIMM on startup. This value is passed to Juju clients on login and used as part of TLS, see https://github.com/juju/juju/blob/3.6/juju/api.go#L194. The client will verify that the server's certificate contains the specified hostname. Normally when deploying JIMM via charms, the user will set this value manually.

But, when using an ingress charm like Traefik, it is the ingress that defines the URL for the service and JIMM accepts this value from Traefik. In the case of one of our deployments, this value was `https://some-domain.com/some-path` but the only part of that we actually want is the hostname, i.e. `some-domain.com`.

This PR now validates the config value by converting it to a URL, which we can then use to extract the host.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

### DIFF
--- a/cmd/jimmsrv/service/service.go
+++ b/cmd/jimmsrv/service/service.go
@@ -6,6 +6,7 @@ package service
 
 import (
 	"context"
+	"fmt"
 	"net/http"
 	"net/url"
 	"strconv"
@@ -484,13 +485,17 @@ func NewService(ctx context.Context, p Params) (*Service, error) {
 
 	macaroonDischarger, err := s.setupDischarger(p)
 	if err != nil {
-		return nil, errors.E(op, err, "failed to set up discharger")
+		return nil, errors.E(op, fmt.Errorf("failed to set up discharger: %v", err))
 	}
 	s.mux.Handle(localDischargePath+"/*", discharger.GetDischargerMux(macaroonDischarger, localDischargePath))
 
+	publicDNS, err := parseURLWithOptionalScheme(p.PublicDNSName)
+	if err != nil {
+		return nil, errors.E(op, fmt.Errorf("failed to parse public DNS name: %v", err))
+	}
 	params := jujuapi.Params{
 		ControllerUUID: p.ControllerUUID,
-		PublicDNSName:  p.PublicDNSName,
+		PublicDNSName:  publicDNS.Host,
 	}
 
 	// Websockets require extra care when cookies are used for authentication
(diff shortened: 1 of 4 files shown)


cmd/jimmsrv/service/service.go
```diff
@@ -511,6 +516,23 @@ func NewService(ctx context.Context, p Params) (*Service, error) {
 	return s, nil
 }
 
+// parseURLWithOptionalScheme parses an input string
+// that may exclude a scheme, i.e. missing "http://".
+// If no scheme is provided, "https" will be used.
+func parseURLWithOptionalScheme(addr string) (*url.URL, bool) {
```
SimoneDutto: isn't it better to return an `error`, just because other errors are still possible and we'd like to log them

cmd/jimmsrv/service/service.go
```diff
@@ -484,13 +485,17 @@ func NewService(ctx context.Context, p Params) (*Service, error) {
 
 	macaroonDischarger, err := s.setupDischarger(p)
 	if err != nil {
-		return nil, errors.E(op, err, "failed to set up discharger")
+		return nil, errors.E(op, fmt.Errorf("failed to set up discharger: %v", err))
 	}
 	s.mux.Handle(localDischargePath+"/*", discharger.GetDischargerMux(macaroonDischarger, localDischargePath))
 
+	publicDNS, ok := parseURLWithOptionalScheme(p.PublicDNSName)
```
SimoneDutto: didn't you changed recently juju to accept path as well. How does this change play out with it?

kian99: Nothing really to do with the path stuff here. You could consider this to be a Juju client bug as well though, the client is using the string that the server sends it without validating it. But we are also sending a junk value to the client so we should fix that.

cmd/jimmsrv/service/service.go
kian99: Yeah fair point, changed.

SimoneDutto (approved):
LGTM

SimoneDutto:
for what it's worth, we could add the parsing in the jimm's charm too

kian99:
> for what it's worth, we could add the parsing in the jimm's charm too

True, I'll leave it in the app for now. If we had other deployment mechanisms like snap, then the charm wouldn't be there to catch this.

alesstimec (approved):
LGTM, but agree with Simone - let's return an error and properly log it

PR #1582, opened 2025-03-18
feat(max-connections): implement max concurrent connections

SimoneDutto:
Implement max concurrent connections as suggested by gliderlabsssh maintainer here https://github.com/gliderlabs/ssh/issues/244

We can't send a custom message to the ssh client just yet because this is the handler before the ssh handshake, so we can just close the connection.

A nice addition is that in the ConnCallback we can handle other things as well, for example max connection durations.

> Fly by set max concurrent connection in docker compose to make it easier to QA the max locally.

## QA steps

spin up JIMM

`seq 16 | parallel -j 16 "ssh localhost -p 17022"`

```console
[1:10:50] ➜  jimm git:(implement-max-concurrent-connections) ✗ seq 16 | parallel -j 16 "ssh localhost -p 17022"
Connection closed by 127.0.0.1 port 17022
Connection closed by 127.0.0.1 port 17022
Connection closed by 127.0.0.1 port 17022
Connection closed by 127.0.0.1 port 17022
Connection closed by 127.0.0.1 port 17022
duttos@localhost: Permission denied (publickey).
duttos@localhost: Permission denied (publickey).
duttos@localhost: Permission denied (publickey).
duttos@localhost: Permission denied (publickey).
duttos@localhost: Permission denied (publickey).
duttos@localhost: Permission denied (publickey).
duttos@localhost: Permission denied (publickey).
duttos@localhost: Permission denied (publickey).
duttos@localhost: Permission denied (publickey).
duttos@localhost: Permission denied (publickey).
```


internal/ssh/ssh.go
```diff
@@ -219,3 +219,28 @@ func rejectConnectionAndLogError(ctx context.Context, newChan gossh.NewChannel,
 		zapctx.Error(ctx, msg, zap.Error(err))
 	}
 }
+
+func ConnCallback(maxConcurrentConnections int) ssh.ConnCallback {
+	mux := sync.Mutex{}
+	n := 0
```
kian99: Might be cleaner to use an atomic int here.

internal/ssh/ssh.go
```diff
@@ -219,3 +219,28 @@ func rejectConnectionAndLogError(ctx context.Context, newChan gossh.NewChannel,
...
+		mux.Lock()
+		defer mux.Unlock()
+		go func() {
+			<-ctx.Done()
+			mux.Lock()
+			defer mux.Unlock()
+			n--
+		}()
+		if n >= maxConcurrentConnections {
+			_, err := conn.Write([]byte("too many connections."))
```
kian99: Have you tested how this comes up when you are using e.g. openssh?

kian99 (commented):
almost lgtm

internal/ssh/ssh.go
SimoneDutto: yep, if you see in the description there is a console codeblock which is c/p from my terminal

internal/ssh/ssh.go
SimoneDutto: i could be wrong, but if i use `atomic int` and let's say my counter is `max concurrent -1`.
In the time between the first to `Load` and the first to `Add` we can let in more than one connection.

Making the `max concurrent` not completely right.
If the time between the Load and Add is very short, this is not a big deal, but it's not as correct as with the lock

internal/ssh/ssh.go
kian99: Some questions,
- How come we only get 10 lines of output but you requested 16 connections?
- Since we don't get the error in the output, is there a point to the conn.Write()?

internal/ssh/ssh.go
kian99: True, but you can fix that by calling Add(1), which will return the new value. If we are >= maxConcurrent, then Add(-1) to reverse it. In addition to Add(-1) inside the routine that is currently doing `n--`.
Try that and see if it's simpler.

internal/ssh/ssh.go
SimoneDutto: - there are 10 connections passing the max concurrent check and reaching the public key checks and 6 being refused.
```console
[1:10:52] ➜  jimm git:(implement-max-concurrent-connections) ✗ seq 16 | parallel -j 16 "ssh localhost -p 17022"
Connection closed by 127.0.0.1 port 17022
Connection closed by 127.0.0.1 port 17022
Connection closed by 127.0.0.1 port 17022
Connection closed by 127.0.0.1 port 17022
Connection closed by 127.0.0.1 port 17022
Connection closed by 127.0.0.1 port 17022
duttos@localhost: Permission denied (publickey).
duttos@localhost: Permission denied (publickey).
duttos@localhost: Permission denied (publickey).
duttos@localhost: Permission denied (publickey).
duttos@localhost: Permission denied (publickey).
duttos@localhost: Permission denied (publickey).
duttos@localhost: Permission denied (publickey).
duttos@localhost: Permission denied (publickey).
duttos@localhost: Permission denied (publickey).
duttos@localhost: Permission denied (publickey).
```
- well, it isn't shown by the ssh client, but if you are debugging the ssh server with telnet for example.
```console
[2:47:18] ➜  ~ telnet localhost 17022
Trying 127.0.0.1...
Connected to localhost.
Escape character is '^]'.
too many connections.Connection closed by foreign host.
```

kian99: Okay cool, then keep the error.
Fyi the output in the description has 5 "connection closed"

SimoneDutto: yy, i copied before changing the check for "<=", but know i've rerun the qa steps

internal/ssh/ssh.go
```diff
@@ -219,3 +219,29 @@ func rejectConnectionAndLogError(ctx context.Context, newChan gossh.NewChannel,
 		zapctx.Error(ctx, msg, zap.Error(err))
 	}
 }
+
+// ConnCallback returns a ConnCallback function that limits the number of concurrent connections.
+func ConnCallback(maxConcurrentConnections int) ssh.ConnCallback {
+	n := atomic.Int32{}
```
kian99, line 225: Nice, I assume we don't need to decrement in the case that `current > maxConcurrentConnections` because that will be handled by the routine that watches the done channel?

SimoneDutto: exactly

kian99 (approved):

ale8k (approved):

PR #1583, opened 2025-03-19
fix(max-concurrent-connections): fix max concurrent connections by closing the connection

SimoneDutto:
By returning `nil` we were closing the connection, but the context was not canceled. 
Making the context.Done stuck, now we close the connection before returning so the context is cancelled.

To test this, we repeat the dial test two times, so we are sure we have reset the counter in the ssh server back to 0.

### DIFF
--- a/internal/ssh/ssh.go
+++ b/internal/ssh/ssh.go
@@ -67,7 +67,6 @@ type Server struct {
 // NewJumpServer creates the jump server struct.
 func NewJumpServer(ctx context.Context, config Config, sshManager SSHManager) (Server, error) {
 	zapctx.Info(ctx, "NewJumpServer")
-
 	if sshManager == nil {
 		return Server{}, fmt.Errorf("Cannot create JumpSSHServer with a nil ssh manager.")
 	}
@@ -239,8 +238,10 @@ func ConnCallback(maxConcurrentConnections int) ssh.ConnCallback {
 			if err != nil {
 				zapctx.Error(ctx, "failed to write to connection", zap.Error(err))
 			}
-			// if ConnCallback returns a nil net.Conn, the connection is closed.
-			return nil
+			// The connection is close before returning, otherwise
+			// the context is not cancelled and the counter is not decremented.
+			conn.Close()
+			return conn
 		}
 		return conn
 	}
--- a/internal/ssh/ssh_test.go
+++ b/internal/ssh/ssh_test.go
@@ -39,8 +39,9 @@ type sshSuite struct {
 	privateKey         gossh.Signer
 	hostKey            gossh.Signer
 
-	received         chan bool
-	allowedModelUUID string
+	received                 chan bool
+	allowedModelUUID         string
+	maxConcurrentConnections int
 }
 
 func (s *sshSuite) Init(c *qt.C) {
@@ -109,11 +110,12 @@ func (s *sshSuite) Init(c *qt.C) {
 	c.Assert(err, qt.IsNil)
 
 	s.jumpServerListener = bufconn.Listen(1 * 1024)
+	s.maxConcurrentConnections = 10
 	jumpServer, err := ssh.NewJumpServer(context.Background(),
 		ssh.Config{
 			Port:                     "22",
 			HostKey:                  hostKey,
-			MaxConcurrentConnections: 10,
+			MaxConcurrentConnections: s.maxConcurrentConnections,
 		},
 		mocks.SSHManager{
 			PublicKeyHandler_: func(ctx context.Context, claimUser string, key []byte) (*openfga.User, error) {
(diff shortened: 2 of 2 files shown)


alesstimec (approved):

internal/ssh/ssh_test.go
```diff
@@ -274,30 +276,35 @@ func (s *sshSuite) TestSSHFinalDestinationDialFail(c *qt.C) {
...
-		HostKeyCallback: gossh.FixedHostKey(s.hostKey.PublicKey()),
-		Auth: []gossh.AuthMethod{
-			gossh.PublicKeys(s.privateKey),
-		},
-		User: "alice",
-	}
-	for range 10 {
+	// the reason we repeat this test 2 times, it's to make sure that closing the connections on
+	// the first iteration completely resets the counter on the ssh server side.
+	for range 2 {
```
ale8k, line 281: Can we not use t.Run here?

internal/ssh/ssh_test.go
```diff
@@ -274,30 +276,35 @@ func (s *sshSuite) TestSSHFinalDestinationDialFail(c *qt.C) {
...
-	clients := make([]*gossh.Client, 0, 10)
-	config := &gossh.ClientConfig{
-		HostKeyCallback: gossh.FixedHostKey(s.hostKey.PublicKey()),
-		Auth: []gossh.AuthMethod{
-			gossh.PublicKeys(s.privateKey),
-		},
-		User: "alice",
-	}
-	for range 10 {
+	// the reason we repeat this test 2 times, it's to make sure that closing the connections on
```
ale8k: is to make sure that*

internal/ssh/ssh_test.go
```diff
@@ -274,30 +276,35 @@ func (s *sshSuite) TestSSHFinalDestinationDialFail(c *qt.C) {
...
+			User: "alice",
+		}
+		for range s.maxConcurrentConnections {
+			client := inMemoryDial(c, s.jumpServerListener, config)
+			clients = append(clients, client)
+		}
+		jumpServerConn, err := s.jumpServerListener.Dial()
+		c.Assert(err, qt.IsNil)
+
+		_, _, _, err = gossh.NewClientConn(jumpServerConn, "", config)
```
ale8k, line 297: I expected to see "_, err = conn.Write([]byte("too many connections.\n"))" here. Why are we testing to see handshake failed?

ale8k (commented):
Might need this running over with me, but it's not testing what I expected the test to test.

internal/ssh/ssh_test.go
SimoneDutto: it's failing the handshake because the conn is closed in the connection callback.
"too many connections" is written in the conn, and you can see it if you are simply reading (for example with `telnet`). But here we are trying to handshake an ssh connection just like an ssh client would do

internal/ssh/ssh_test.go
SimoneDutto: we don't want to run the test two times, we want to run a test where we create the max conns and we close them in a single test run with a single server

internal/ssh/ssh.go
```diff
@@ -239,8 +238,10 @@ func ConnCallback(maxConcurrentConnections int) ssh.ConnCallback {
 			if err != nil {
 				zapctx.Error(ctx, "failed to write to connection", zap.Error(err))
 			}
-			// if ConnCallback returns a nil net.Conn, the connection is closed.
-			return nil
+			// The connection is close before returning.
+			// Otherwise, the context is not canceled and the counter is not decremented.
+			conn.Close()
+			return conn
```
kian99: What's the benefit of doing this versus `n.Add(-1); return nil`?

internal/ssh/ssh_test.go
kian99:
```suggestion
	// the reason we repeat this test 2 times is to make sure that closing the connections on
```

internal/ssh/ssh.go
```diff
@@ -239,8 +238,10 @@ func ConnCallback(maxConcurrentConnections int) ssh.ConnCallback {
 			if err != nil {
 				zapctx.Error(ctx, "failed to write to connection", zap.Error(err))
 			}
-			// if ConnCallback returns a nil net.Conn, the connection is closed.
-			return nil
+			// The connection is close before returning.
+			// Otherwise, the context is not canceled and the counter is not decremented.
```
```suggestion
			// The connection is close before returning, otherwise
			// the context is not cancelled and the counter is not decremented.
```

kian99 (approved):

internal/ssh/ssh.go
SimoneDutto: i'd prefer to have a single source of +1 and -1 to avoid strange bugs where we decrement twice. wdyt?

internal/ssh/ssh_test.go
ale8k: Maybe can discuss on call, not sure I'm fully following here.

ale8k (approved):

PR #1585, opened 2025-04-02
fix: missing check for cloud-credential owner

kian99:
## Description
This PR fixes a vulnerability identified in JIMM where a user can deploy a model using another user's cloud-credentials.
During model creation, the arguments include 3 key pieces of information:
1. The user making the request
2. The model owner
3. The cloud-credential to use
 
We check that the user making the request is either the same as the model owner or a JIMM superuser, who is allowed to create models on another user's behalf. But, importantly, we were not validating that the model owner had access to the cloud-credential.

Because JIMM doesn't model access to cloud-credentials, the only thing we can do is check that the owner of the cloud-credential matches the model owner.

This also means that if a JIMM admin is attempting to create a model on behalf of another user, say `Alice` is trying to create a model for `Bob`. Alice must specify the name of Bob's credential rather than her own. This behaviour was already possible but is now limited to JIMM superusers and while it doesn't seem particularly desirable, can be improved in a follow-up.

A note on the tests:
- I've added a test to cover the vulnerability.
- There was a test that verified a superuser could create models owned by other users, but this started failing after the fix. The test created a model for user `Bob` created by `Alice`, using `Alice's` credential. I've updated the test so that it uses a credential owned by `Bob`.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

### DIFF
--- a/internal/jimm/juju/modelbuilder.go
+++ b/internal/jimm/juju/modelbuilder.go
@@ -230,7 +230,7 @@ func (b *modelBuilder) WithCloudCredential(credentialTag names.CloudCredentialTa
 
 	// Verify ownership of cloud credential
 	if b.owner == nil || b.owner.Name != credentialTag.Owner().Id() {
-		b.err = errors.E("invalid owner for cloud credential", errors.CodeUnauthorized)
+		b.err = errors.E("model owner doesn't match cloud-credential owner", errors.CodeUnauthorized)
 		return b
 	}
 
--- a/internal/jimm/juju/modelbuilder.go
+++ b/internal/jimm/juju/modelbuilder.go
@@ -227,6 +227,13 @@ func (b *modelBuilder) WithCloudCredential(credentialTag names.CloudCredentialTa
 	if b.err != nil {
 		return b
 	}
+
+	// Verify ownership of cloud credential
+	if b.owner == nil || b.owner.Name != credentialTag.Owner().Id() {
+		b.err = errors.E("invalid owner for cloud credential", errors.CodeUnauthorized)
+		return b
+	}
+
 	credential := dbmodel.CloudCredential{
 		Name:              credentialTag.Name(),
 		CloudName:         credentialTag.Cloud().Id(),
--- a/internal/jimm/juju/model_test.go
+++ b/internal/jimm/juju/model_test.go
@@ -472,7 +472,7 @@ clouds:
     access: add-model
 cloud-credentials:
 - name: test-credential-1
-  owner: alice@canonical.com
+  owner: bob@canonical.com
   cloud: test-cloud
   auth-type: empty
 controllers:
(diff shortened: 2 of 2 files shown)


internal/jimm/juju/modelbuilder.go
```diff
@@ -227,6 +227,13 @@ func (b *modelBuilder) WithCloudCredential(credentialTag names.CloudCredentialTa
 	if b.err != nil {
 		return b
 	}
+
+	// Verify ownership of cloud credential
+	if b.owner == nil || b.owner.Name != credentialTag.Owner().Id() {
+		b.err = errors.E("invalid owner for cloud credential", errors.CodeUnauthorized)
```
alesstimec: i would return `cloud credential not found` as the user does not have access to it and it gives away no information about cloud credentials.

kian99: This also doesn't give information away I think. Because I haven't checked if the credential exists. I'm just validating the owner name of the credential (before we check if it exists).

alesstimec: ha, i see.. but it might give somebody the impression that those cloud creds exist..

alesstimec: also.. when would b.owner be nil? if that happens we're missing some information or have been running model builder steps in the wrong order?

kian99: I don't think b.owner should be nil at this point, but if it is I didn't want to panic.

kian99: >it might give somebody the impression that those cloud creds exist

Fair enough, will change.

kian99: Went to change this and felt like "cloud credential not found" is also misleading, because the credential may very well exist. So I've changed it to "model owner doesn't match cloud-credential owner" which is clear, and doesn't leak information about what exists.

alesstimec (approved):

internal/jimm/juju/modelbuilder.go
```diff
@@ -227,6 +227,13 @@ func (b *modelBuilder) WithCloudCredential(credentialTag names.CloudCredentialTa
...
 		return b
 	}
+
+	// Verify ownership of cloud credential
+	if b.owner == nil || b.owner.Name != credentialTag.Owner().Id() {
+		b.err = errors.E("model owner doesn't match cloud-credential owner", errors.CodeUnauthorized)
+		return b
+	}
+
 	credential := dbmodel.CloudCredential{
```
SimoneDutto, line 237: what if we do `CloudCredentialsForUser` which returns the cloud credentials where the user has a `can-add-model` access to the relative cloud?

SimoneDutto: otherwise one could have a `can-add-model` permision but without being the cloud credential owner he can't create a model, right?

kian99: We don't model permissions to cloud-credentials beyond simply the credential owner so I think you're conflating two separate things,
1. Permissions to cloud-credentials
2. Permissions to a cloud

The second point is modelled in OpenFGA but that is checked (or should be checked) elsewhere.
You can have a `can-add-model` permission, but without a cloud-credential that you own, you can't add a model and that is expected.

SimoneDutto: Ah ok, thank you!

SimoneDutto: so, you need to own a cloud credential regardeless, and nobody can give you permission to it? Not even a super-admin?

kian99: Correct

kian99: Well, like I mentioned in the description. A superuser, can deploy a model using someone else's cloud-credential.

SimoneDutto (approved):

PR #1586, opened 2025-04-03
fix(auditlog): redact parameters for login methods

alesstimec:
## Description

Adds login methods to the list of sensitive methods for audit logging so that parameters get redacted.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests


internal/jimm/auditlog/auditlog_test.go
```diff
@@ -72,29 +74,85 @@ func (s *auditLogManagerSuite) TestFindAuditEvents(c *qt.C) {
...
+		FacadeMethod: "UpdateCredentials",
+		Params:       sensitive_data,
 	}}
 	for i, event := range events {
 		e := event
 		s.manager.AddAuditLogEntry(&e)
 		events[i] = e
+
+		// for sensitive methods, we expect redacted parameters
+		if slices.Compare(events[i].Params, sensitive_data) == 0 {
```
ale8k: Love seeing this slices package, it makes me happy lol

internal/jimm/auditlog/auditlog_test.go
```diff
@@ -174,6 +242,7 @@ func (s *auditLogManagerSuite) TestFindAuditEvents(c *qt.C) {
 					c.Assert(err, qt.ErrorMatches, test.expectedError)
 				} else {
 					c.Assert(err, qt.Equals, nil)
+					c.Logf("received events: %+v", events)
```
ale8k: Can remove this

ale8k (approved):
LGTM, I give it a brief skim.

internal/jimm/auditlog/auditlog_test.go
```diff
@@ -72,29 +74,85 @@ func (s *auditLogManagerSuite) TestFindAuditEvents(c *qt.C) {
...
 	}}
 	for i, event := range events {
 		e := event
 		s.manager.AddAuditLogEntry(&e)
 		events[i] = e
+
+		// for sensitive methods, we expect redacted parameters
+		if slices.Compare(events[i].Params, sensitive_data) == 0 {
+			events[i].Params = []byte("{'params':'redacted'}")
+		}
```
SimoneDutto: i'm not getting why you would need to do this manually.
if the method `AddAuditLogEntry` is getting the pointer and then copying the result in the `events` slice.
Should it already be redacted? Right?

alesstimec: because at the end we compare the slice of received events to the predefined list.. and for sensitive methods we always expect to see "redacted"

SimoneDutto: yeah, but if the code which is redacting the entry is in `AddAuditLogEntry`. Why do we need to do it again in this loop?

alesstimec: because expected events in tests are based on this slice of events.. so we insert them into the DB with unredacted params and expect to receive them with redacted params

internal/jimm/auditlog/auditlog_test.go
```diff
@@ -72,29 +74,85 @@ func (s *auditLogManagerSuite) TestFindAuditEvents(c *qt.C) {
...
 	}, {
 		Time:         now.Add(3 * time.Hour),
 		IdentityTag:  s.priveligedUser.Identity.Tag().String(),
 		Model:        "TestModel",
 		FacadeMethod: "DestroyModel",
+		Params:       data,
+	}, {
+		Time:         now.Add(4 * time.Hour),
+		IdentityTag:  s.adminUser.Identity.Tag().String(),
+		FacadeMethod: "LoginWithSessionToken",
```
kian99: Don't you think these tests are a bit pointless? We're just testing all these value exists in a list. We can just test the redact method does the expected thing, in a separate test possibly.

alesstimec: done

internal/jimm/auditlog/auditlog_test.go
```diff
@@ -67,6 +67,67 @@ func (s *auditLogManagerSuite) Init(c *qt.C) {
...
+	}
+
+	tests := []struct {
+		about          string
+		entry          dbmodel.AuditLogEntry
+		expectedParams dbmodel.JSON
+	}{{
+		about:          "addcredentials",
+		entry:          newEntry("AddCredentials"),
+		expectedParams: auditlog.RedactJSON,
```
kian99: What do you think about patching the `sensitiveMethods` list and then you can just have 2 test cases, one where it redacts and one where it doesn't.

kian99 (approved):

internal/jimm/auditlog/auditlog_test.go
alesstimec: done

PR #1588, opened 2025-04-07
feat: compare security scan CVEs with KEV list

kian99:
KEVs or Known Exploited Vulnerabilities are vulnerabilities that are known exploitation attempts. In this change we create a short script to surface any KEVs in our CI. The KEVs are only printed in the output for now as the only alternative I can see is failing the workflow which I don't think is desirable.

Fixes [JUJU-7241](https://warthogs.atlassian.net/browse/JUJU-7241)

[JUJU-7241]: https://warthogs.atlassian.net/browse/JUJU-7241?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/security-scan-weekly.yaml
+++ b/.github/workflows/security-scan-weekly.yaml
@@ -58,6 +58,29 @@ jobs:
         uses: github/codeql-action/upload-sarif@v3
         with:
           sarif_file: 'govuln-results.sarif'
+      
+      #############
+      # KEV Check #
+      #############
+
+      - name: Fetch KEV list from CISA
+        shell: bash
+        run: |
+          curl -s -o ./scripts/kev.json https://www.cisa.gov/sites/default/files/feeds/known_exploited_vulnerabilities.json
+
+      - name: Compare Trivy results with KEV list
+        run: ./scripts/compare_kev_vulnerabilities.sh
+        shell: bash
+        env:
+          VULN_REPORT_FILE: trivy-results.sarif
+          KNOWN_CVES_FILE: kev.json
+
+      - name: Compare govuln results with KEV list
+        run: ./scripts/compare_kev_vulnerabilities.sh
+        shell: bash
+        env:
+          VULN_REPORT_FILE: govuln-results.sarif
+          KNOWN_CVES_FILE: kev.json
 
   container-security-scan:
     runs-on: [ubuntu-latest]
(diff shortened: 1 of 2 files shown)


alesstimec (approved):
LGTM

SimoneDutto (approved):

scripts/compare_kev_vulnerabilities.sh
```diff
@@ -0,0 +1,49 @@
...
+fi
+
+if [ ! -f "$KNOWN_CVES_FILE" ]; then
+  echo "Error: Known CVEs file '$KNOWN_CVES_FILE' not found."
+  exit 2
+fi
+
+# Extract CVE IDs from known vulnerabilities list
+known_cves="$(jq -r '.vulnerabilities[].cveID' "$KNOWN_CVES_FILE" | sort -u)"
+report_cves="$(jq -r '.runs[].results[].ruleId' "$VULN_REPORT_FILE" | sort -u)"
```
ale8k, line 38: It seems strange we're making a script to compare the lists. Has this not been done before / is this the "correct" way?

.github/workflows/security-scan-weekly.yaml
```diff
@@ -82,3 +117,21 @@ jobs:
...
+          curl -s -o ./kev.json https://www.cisa.gov/sites/default/files/feeds/known_exploited_vulnerabilities.json
+
+      - name: Compare Trivy container scan results with KEV list
+        run: ./scripts/compare_kev_vulnerabilities.sh 2>&1 | tee kev-trivy-container-scan.txt
+        shell: bash
+        env:
+          VULN_REPORT_FILE: jimm-container-trivy-results.sarif
+          KNOWN_CVES_FILE: kev.json
+
+      - name: Upload KEV Trivy Container scan results as artifact
```
ale8k, line 133: It may be ok to fail to workflow. KEVs should not appear often and when they do, is likely ok to block further merges until it's fixed. Otherwise further merges could potentially (even if very unlikely) further exacerbate the KEV.

scripts/compare_kev_vulnerabilities.sh
kian99: I looked into what others did,
- Anbox team made a simple script which I copied, see [here](https://github.com/canonical/ams/blob/e518782854c12977be04b8e6ffe4024fdb97566b/scripts/compare_kev_vulnerabilities.sh). I copied their script and improved it to use a sarif file for detected vulnerabilities so that we can re-use this for govuln checks and for Trivy scans.
- Canonical Contracts has [the following](https://github.com/canonical/contracts/tree/main/scripts/ssdlc/secscan) which is a more extensive set of Python scripts.
- Online there are some actions like [this one](https://github.com/advanced-security/dependabot-kev-action) that don't really fit our needs.

.github/workflows/security-scan-weekly.yaml
kian99: Yeah in a follow-up I changed it to exit 1 if a kev is detected.

PR #1590, opened 2025-04-09
chore: wire up new jaas commands

kian99:
## Description
This PR wires up the newly added commands in the jaas plugin so that they are available to the `jaas` cli.
In order for the commands to be available as top-level commands to the Juju CLI, i.e. so that you can do `juju <command>` rather than `juju jaas <command>` I've followed the same approach as before with symlinks created in the snapcraft.yaml. 

To ensure that future changes also update snapcraft.yaml, I've added a script that verifies all `jaas` commands are present in the snapcraft.yaml file and we run this check in CI.

Fixes [JUJU-7847](https://warthogs.atlassian.net/browse/JUJU-7847)

[JUJU-7847]: https://warthogs.atlassian.net/browse/JUJU-7847?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


alesstimec (approved):
LGTM. I suppose we'll want to generate "juju reference" documentation next?

SimoneDutto (approved):
LGTM

PR #1592, opened 2025-04-11
fix(offers): dial controller as user for GetApplicationOfferConsumeDetails

alesstimec:
## Description

For the GetApplicationOfferConsumeDetails we must dial the controller as the user - the JWT contains authorization claims for the user that is dialing the controller, so we can dial as "admin" on behalf of the user, because the controller is trying to determine whether the user has access to the offer.

## Engineering checklist

- [x] Covered by unit tests


internal/jujuclient/dial.go
```diff
@@ -76,14 +75,45 @@ func (d *Dialer) createLoginRequest(ctx context.Context, ctl *dbmodel.Controller
...
+
+	for k, v := range additionalPermissions {
+		p[k] = v
+	}
+	return d.createLoginRequest1(ctx, ctl.ResourceTag(), user.ResourceTag(), p)
+}
+
 // Dial implements jimm.Dialer.
-func (d *Dialer) Dial(ctx context.Context, ctl *dbmodel.Controller, modelTag names.ModelTag, requiredPermissions map[string]string) (juju.API, error) {
+func (d *Dialer) Dial(ctx context.Context, ctl *dbmodel.Controller, modelTag names.ModelTag, user *openfga.User, withPermissions map[string]string) (juju.API, error) {
```
kian99, line 116: Although I think the change to the Dial function's signature is fine for now. It would be good to revisit it, because it's getting too long imo.
Some ideas,
- We create `AdminDialer` and `UserDialer` structs whose `NewUserDialer` function accepts a dialer, user and permission map.
- We pass the user's permission in the context and extract them in the Dial method.

internal/jujuclient/dial.go
```diff
@@ -230,7 +273,17 @@ func (c *Connection) redial(ctx context.Context, requiredPermissions map[string]
...
-	api, err := c.dialer.Dial(ctx, c.ctl, c.mt, requiredPermissions)
+
+	grantedPermission := make(map[string]string)
+	for resource, permission := range requiredPermissions {
+		err := c.user.CheckPermission(ctx, resource, permission)
+		if err != nil {
+			return err
+		}
+		grantedPermission[resource] = permission
+	}
```
kian99, line 286: Can we add a comment above on why we check the permission. We discussed it in a previous PR and I've forgotten and it still doesn't strike me as obvious.

internal/openfga/user.go
```diff
@@ -280,6 +282,46 @@ func (u *User) ListApplicationOffers(ctx context.Context, relation ofga.Relation
...
+	case "superuser", string(params.ModelAdminAccess):
+		relation = ofganames.AdministratorRelation
+	case string(params.ModelWriteAccess):
+		relation = ofganames.WriterRelation
+	case string(params.ModelReadAccess):
+		relation = ofganames.ReaderRelation
+	case string(params.OfferConsumeAccess):
+		relation = ofganames.ConsumerRelation
+	case string("add-cloud"):
+		relation = ofganames.CanAddModelRelation
```
kian99, line 306: I think we already have a function that converts Juju permissions to OpenFGA relations

internal/jujuclient/dial.go
alesstimec: definitely agree.. this is a bit of a mess atm.

internal/openfga/user.go
alesstimec: in openfga/names we have the inverse function, but not this

internal/jujuclient/dial.go
alesstimec: done

internal/openfga/user.go
kian99: Had to look a bit but I think there is logic like this in `internal/jimm/permissions/access.go`, see `ToCloudRelation`, `ToModelRelation`, etc. Possibly those function should be here or something like that.

alesstimec: we'll have an import cycle if i use methods from internal/jimm/permissions

kian99: That's fine, can you make a follow-up to put this logic in a separate function and use it from `internal/jimm/permissions`? Or just look at whether we should at the least.

kian99 (approved):

SimoneDutto (approved):
LGTM.

how come we are changing the user we dial with but there isn't a test failing after this change? Don't we test for this kind of stuff?

alesstimec:
@SimoneDutto remember https://github.com/juju/juju/pull/19395/files#diff-1b8861e0b1a38163e5d335e547d1f87958e3cd55fe80d1fa428ab2ecaf574f26 ? currently there's no way to consume an application offer if you only have consume access to the offer.. i can add a test, but it will fail, because my fix to juju was reverted.

SimoneDutto:
> @SimoneDutto remember https://github.com/juju/juju/pull/19395/files#diff-1b8861e0b1a38163e5d335e547d1f87958e3cd55fe80d1fa428ab2ecaf574f26 ? currently there's no way to consume an application offer if you only have consume access to the offer.. i can add a test, but it will fail, because my fix to juju was reverted.

ok, let's remember to add a test once we land something in juju again

PR #1595, opened 2025-04-16
fix(applicationoffer): forwards call to grant/revoke offer access to controller.

alesstimec:
## Description

Forwards the call to grant or revoke offer access to the controller in charge of the offer. This should fix the problem with application offers to some extent, but will NOT fix the terraform provider - that still only adds relations in OpenFGA.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

### DIFF
--- a/internal/jimm/juju/applicationoffer.go
+++ b/internal/jimm/juju/applicationoffer.go
@@ -600,3 +600,17 @@ func (j *JujuManager) doApplicationOfferAdmin(ctx context.Context, user *openfga
 	}
 	return nil
 }
+
+// GrantOfferAccessOnController dials the controller in charge of the application offer and grants the user identified by `ut` the specified level of access to the offer.
+func (j *JujuManager) GrantOfferAccessOnController(ctx context.Context, user *openfga.User, ut names.UserTag, offerURL string, access jujuparams.OfferAccessPermission) error {
+	return j.doApplicationOfferAdmin(ctx, user, offerURL, func(offer *dbmodel.ApplicationOffer, api API) error {
+		return api.GrantApplicationOfferAccess(ctx, offerURL, ut, access)
+	})
+}
+
+// RevokeOfferAccessOnController dials the controller in charge of the application offer and revokes the user identified by `ut` the specified level of access to the offer.
+func (j *JujuManager) RevokeOfferAccessOnController(ctx context.Context, user *openfga.User, ut names.UserTag, offerURL string, access jujuparams.OfferAccessPermission) error {
+	return j.doApplicationOfferAdmin(ctx, user, offerURL, func(offer *dbmodel.ApplicationOffer, api API) error {
+		return api.RevokeApplicationOfferAccess(ctx, offerURL, ut, access)
+	})
+}
--- a/internal/jujuapi/applicationoffers.go
+++ b/internal/jujuapi/applicationoffers.go
@@ -5,6 +5,7 @@ package jujuapi
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/go-macaroon-bakery/macaroon-bakery/v3/bakery"
 	"github.com/juju/juju/core/crossmodel"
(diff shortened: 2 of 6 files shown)


internal/jimm/juju/applicationoffer.go
```diff
@@ -600,3 +600,17 @@ func (j *JujuManager) doApplicationOfferAdmin(ctx context.Context, user *openfga
...
 }
+
+// GrantOfferAccessOnController dials the controller in charge of the application offer and grants the user identified by `ut` specified level of access to the offer.
+func (j *JujuManager) GrantOfferAccessOnController(ctx context.Context, user *openfga.User, ut names.UserTag, offerURL string, access jujuparams.OfferAccessPermission) error {
+	return j.doApplicationOfferAdmin(ctx, user, offerURL, func(offer *dbmodel.ApplicationOffer, api API) error {
+		return api.GrantApplicationOfferAccess(ctx, offerURL, ut, access)
+	})
+}
+
+// RevokeOfferAccessOnController dials the controller in charge of the application offer and grants the user identified by `ut` specified level of access to the offer.
```
kian99:
```suggestion
// RevokeOfferAccessOnController dials the controller in charge of the application offer and revokes from the user identified by `ut` the specified level of access to the offer.
```

internal/jujuapi/applicationoffers.go
```diff
@@ -178,11 +178,17 @@ func (r *controllerRoot) modifyOfferAccess(ctx context.Context, change jujuparam
 		if err := r.jimm.PermissionManager().GrantOfferAccess(ctx, r.user, change.OfferURL, ut, change.Access); err != nil {
 			return errors.E(op, err)
 		}
+		if err := r.jimm.JujuManager().GrantOfferAccessOnController(ctx, r.user, ut, change.OfferURL, change.Access); err != nil {
```
kian99, line 182: Can you put a comment explaining that this is for pre Juju 3.6.6 controllers and why it works.

kian99 (approved):

internal/jimm/juju/applicationoffer.go
alesstimec: done

internal/jujuapi/applicationoffers.go
alesstimec: done

internal/jujuapi/applicationoffers.go
```diff
@@ -178,11 +178,20 @@ func (r *controllerRoot) modifyOfferAccess(ctx context.Context, change jujuparam
 		if err := r.jimm.PermissionManager().GrantOfferAccess(ctx, r.user, change.OfferURL, ut, change.Access); err != nil {
 			return errors.E(op, err)
 		}
+		// We grant access on the controller because pre 3.6.5 controllers are consulting their database to check if
```
kian99: It should be pre 3.6.6 I think because 3.6.5 is already out and has the same logic of checking the DB.

alesstimec: done

internal/jujuapi/applicationoffers.go
```diff
@@ -178,11 +178,20 @@ func (r *controllerRoot) modifyOfferAccess(ctx context.Context, change jujuparam
...
+			return errors.E(op, err)
+		}
 		return nil
 	case jujuparams.RevokeOfferAccess:
 		if err := r.jimm.PermissionManager().RevokeOfferAccess(ctx, r.user, change.OfferURL, ut, change.Access); err != nil {
 			return errors.E(op, err)
 		}
+		if err := r.jimm.JujuManager().RevokeOfferAccessOnController(ctx, r.user, ut, change.OfferURL, change.Access); err != nil {
+			return errors.E(op, err)
+		}
```
SimoneDutto: thinking out loud.

In case of removal is better to remove it first from jimm and then from juju, because if it fails we don't have an issue, we won't display this offer, etc, etc.

But in case of granting the permission, would it make sense to do the opposite? Before we add it to juju and then to jimm.
So, if juju fails the user doesn't see it when it lists offers and trigger the grant command again and we have a chance to try and re-add it to jimm/juju?

SimoneDutto (commented):
a question

internal/jujuapi/applicationoffers.go
alesstimec: done

SimoneDutto: If we have this new strategy we should ignore error like "duplicated grant", otherwise we add it to JIMM it errors saying it's duplicated and we don't fwd to juju.

alesstimec: i think the permission manager does not return an error if relation already exists.

SimoneDutto: Can you add a test for it?

alesstimec: done

SimoneDutto (approved):
ty

PR #1596, opened 2025-05-07
chore: parse virtual hostname

kian99:
## Description
This PR adds virtual hostname parsing to JIMM. The hostname sent by the client indicates which model the user is trying to access.

Additionally, I've changed the permission required for a user to SSH to `admin` from `write` to match Juju.

Fixes [JUJU-7577](https://warthogs.atlassian.net/browse/JUJU-7577)

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

[JUJU-7577]: https://warthogs.atlassian.net/browse/JUJU-7577?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/ssh/ssh.go
```diff
@@ -200,7 +208,7 @@ func fetchAndAuthorizeUser(ctx ssh.Context, modelTag names.ModelTag) (*openfga.U
 	if !ok {
 		return nil, fmt.Errorf("fo user in the context")
 	}
-	ok, err := user.IsModelWriter(ctx, modelTag)
+	ok, err := user.IsModelAdmin(ctx, modelTag)
```
ale8k, line 210: I see the PR description, but just curious why we want only admins to be allowed to ssh?

internal/ssh/ssh.go
```diff
@@ -200,7 +208,7 @@ func fetchAndAuthorizeUser(ctx ssh.Context, modelTag names.ModelTag) (*openfga.U
 	if !ok {
 		return nil, fmt.Errorf("fo user in the context")
```
ale8k: Not you but what is this error aha. Could you make it something more sensible please

internal/ssh/ssh.go
```diff
@@ -200,7 +208,7 @@ func fetchAndAuthorizeUser(ctx ssh.Context, modelTag names.ModelTag) (*openfga.U
...
 	}
-	ok, err := user.IsModelWriter(ctx, modelTag)
+	ok, err := user.IsModelAdmin(ctx, modelTag)
 	if err != nil {
 		return nil, fmt.Errorf("failed to resolve address from model uuid")
```
ale8k: Not you but this error line doesn't look right, shouldn't we just wrap the error and return it?

internal/ssh/ssh.go
```diff
@@ -159,6 +165,8 @@ func directTCPIPHandler(sshManager SSHManager) func(srv *ssh.Server, conn *gossh
 			return
 		}
 
+		// The port below is arbitrary as the controller will ignore
+		// it and proxy traffic to its terminating server.
```
ale8k: I think a little more detail is worth it, otherwise this comment assumes the reader fully understands our process ahead of reading this?
Maybe something like:
```
The port below is arbritary, as upon receiving the ChannelOpen request, the channel open message (which contains the port) is ignored and instead the controller sents traffic to its local terminating server.
```

internal/openfga/user.go
```diff
@@ -81,6 +81,15 @@ func (u *User) IsModelWriter(ctx context.Context, resource names.ModelTag) (bool
 	return isWriter, nil
 }
 
+// IsModelAdmin returns true if user has administrator relation to the model.
+func (u *User) IsModelAdmin(ctx context.Context, resource names.ModelTag) (bool, error) {
```
ale8k, line 85: Surprised we didn't have this.

ale8k (approved):
Some minor comments, but LGTM otherwise

ale8k (approved):
Some minor comments, but LGTM otherwise

internal/ssh/ssh.go
kian99: I don't see how that's any different with regards to "the reader fully understands our process ahead of reading this". You're still mentioning "local terminating server" so without understand how the Juju implementation works, both comments are the same.
I'll change it just `// The port below is arbitrary as the controller will ignore it`

internal/ssh/ssh.go
kian99: That matches Juju access rules.

internal/ssh/ssh.go
kian99: Yeah it looks wrong, will fix

internal/ssh/ssh.go
ale8k: I even written a clearer msg out, could just use it

SimoneDutto (approved):
nice

PR #1597, opened 2025-05-07
feat(ssh-manager): get ssh port to dial the controller via ControllerConfig facade

SimoneDutto:
Get the controller ssh port via ControllerConfig facade instead of hardcoding it.
Update mocks and add some tests.


internal/jimm/juju/controller_test.go
```diff
@@ -476,6 +476,56 @@ func TestEarliestControllerVersion(t *testing.T) {
...
+		},
+	})
+
+	env := jimmtest.ParseEnvironment(c, testControllerConfigEnv)
+	env.PopulateDBAndPermissions(c, j.ResourceTag(), j.Database, j.OpenFGAClient)
+
+	config, err := j.ControllerConfig(ctx, env.Controllers[0].Name)
+	c.Assert(err, qt.Equals, nil)
+	c.Assert(config, qt.Not(qt.IsNil))
+	c.Assert(config.SSHServerPort(), qt.Equals, 17022)
```
kian99, line 524: Would this work? Since the mock isn't returning anything besides controller-uuid

internal/jimm/juju/controller.go
```diff
@@ -723,3 +724,30 @@ func (j *JujuManager) InitiateMigration(ctx context.Context, user *openfga.User,
...
+	config := jujuparams.ControllerConfigResult{}
+	err = api.ControllerConfig(ctx, &config)
+	if err != nil {
+		return jujucontroller.Config(config.Config), errors.E(op, err)
+	}
+	// we need to check if the config inside is nil, because it is a map.
+	if config.Config == nil {
+		return jujucontroller.Config{}, errors.E(op, errors.CodeNotFound, "controller config not found")
+	}
+	return jujucontroller.Config(config.Config), nil
```
ale8k: Spot on, nice!

kian99 (approved):
Nice!

internal/jujuclient/controller.go
```diff
@@ -0,0 +1,21 @@
...
+	"context"
+
+	jujuerrors "github.com/juju/errors"
+	jujuparams "github.com/juju/juju/rpc/params"
+
+	"github.com/canonical/jimm/v3/internal/errors"
+)
+
+// ControllerConfig retrieves the controller configuration.
+func (c Connection) ControllerConfig(ctx context.Context, controllerConfig *jujuparams.ControllerConfigResult) error {
```
ale8k: Don't pass a config, just return one. Then in controller.go for juju manager, you won't need to worry if the map is nil.

internal/jimm/juju/controller.go
```diff
@@ -723,3 +724,30 @@ func (j *JujuManager) InitiateMigration(ctx context.Context, user *openfga.User,
...
+	}
+	defer api.Close()
+
+	config := jujuparams.ControllerConfigResult{}
+	err = api.ControllerConfig(ctx, &config)
+	if err != nil {
+		return jujucontroller.Config(config.Config), errors.E(op, err)
+	}
+	// we need to check if the config inside is nil, because it is a map.
+	if config.Config == nil {
```
ale8k: Left a comment in the client, we can prevent this check.

internal/jimm/ssh/ssh.go
```diff
@@ -51,11 +54,11 @@ type JWTGeneratorFactory interface {
 }
 
 // NewSSHManager returns a new SSHManager that offers jimm functionality to the SSHJumpServer.
-func NewSSHManager(identityManager IdentityManager, modelManager ModelManager, sshKeyManager SSHKeyManager, jwtFactory JWTGeneratorFactory) (*sshManager, error) {
+func NewSSHManager(identityManager IdentityManager, jujuManager JujuManager, sshKeyManager SSHKeyManager, jwtFactory JWTGeneratorFactory) (*sshManager, error) {
```
ale8k, line 57: Following what Juju does, we could have a pattern of params struct with a .Validate() method, I kinda like it in Juju.

ale8k (changes requested):

ale8k (changes requested):
LGTM, one minor comment on just creating the result in the client. Really don't like the &params pattern we do let's abandon it (and we have in some places already).

And to be super-stubborn, I've requested changes until then aha ;)

internal/jimm/juju/controller_test.go
SimoneDutto: yes, because the `Config` struct returns default values when they are not in the map

ale8k (approved):
Thank you for updating!

PR #1598, opened 2025-05-08
chore: create JWT token for SSH authentication

kian99:
## Description
This PR fixes how we generate JWTs when JIMM attempts to connect to a Juju controller via SSH. With this and some tweaks to Juju, I was able to successfully SSH into a machine by going through JIMM.

The full changes include:
1. Create separate JWT token generators for RPC login vs SSH login. Their needs are very different and the factory pattern we already have in place makes it easy to create one type of generator over the other.
2. Split the host and port when we fetch the controller's addresses.
3. Change our `setup-controller.sh` script to bootstrap with the `ssh-jump` feature flag.
4. Ensure the user's public key is added to the JWT we send Juju. This was requested by the Juju team, so that we can authenticate the user in Juju as well (even though the Juju database doesn't hold their public key).
5. Some drive-by fixes to naming.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

## Test instructions
Requires https://github.com/juju/juju/pull/19742 in Juju to land, then:
1. `make qa-lxd`
2. `juju add-model test`
3. `juju deploy ubuntu`
4. `juju show-model` get model UUID.
5. `ssh -J jimm-test@canonical.com@jimm.localhost:17022 ubuntu@0.$model_uuid.juju.local`

### DIFF
--- a/internal/jimm/jujuauth/factory.go
+++ b/internal/jimm/jujuauth/factory.go
@@ -2,9 +2,9 @@
 
 package jujuauth
 
-// Factory holds the necessary components for producing new stateful
-// Juju authenticator objects. Because these objects are
-// stateful, it is expected that a new one is used for each connection.
+// Factory holds the necessary components for producing
+// Juju authenticator objects. Currently a login token generator
+// and an SSH token generator are available.
 type Factory struct {
 	db            GeneratorDatabase
 	jwtService    JWTService
@@ -20,7 +20,16 @@ func NewFactory(db GeneratorDatabase, jwtService JWTService, accessChecker Gener
 	}
 }
 
-// New returns a new Juju token generator.
-func (f *Factory) New() TokenGenerator {
-	return newTokenGenerator(f.db, f.accessChecker, f.jwtService)
+// NewLoginGenerator returns a new token generator for Juju RPC login requests.
+// The LoginTokenGenerator is stateful and should be re-used for the lifetime
+// of a single connection, and recreated for each new connection.
+func (f *Factory) NewLoginGenerator() LoginTokenGenerator {
+	return newLoginTokenGenerator(f.db, f.accessChecker, f.jwtService)
+}
+
+// NewSSHGenerator returns a new token generator for Juju SSH connections.
+// The SSHToken generator is not stateful and can be re-used across
+// multiple connections.
+func (f *Factory) NewSSHGenerator() SSHTokenGenerator {
+	return newSSHTokenGenerator(f.jwtService)
 }
--- a/internal/jimm/jujuauth/sshtokens.go
+++ b/internal/jimm/jujuauth/sshtokens.go
@@ -1,0 +1,64 @@
+// Copyright 2025 Canonical.
+
+// Package jujuauth generates JWT tokens to
+// authenticate and authorize messages to Juju controllers.
+// This package is more specialised than a generic
+// JWT token generator as it crafts Juju specific
+// permissions that are added as claims to the JWT
+// and therefore exists in JIMM's business logic layer.
+package jujuauth
+
+import (
+	"context"
+	"encoding/base64"
+
+	"github.com/juju/juju/core/permission"
+	"github.com/juju/names/v5"
+
+	"github.com/canonical/jimm/v3/internal/jimmjwx"
+)
+
+// SSHTokenGenerator provides the ability to generate JWT tokens that
+// authenticate and authorize SSH connections to Juju controllers.
+type SSHTokenGenerator struct {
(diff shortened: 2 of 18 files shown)


internal/jimm/jujuauth/sshtokens.go
```diff
@@ -0,0 +1,64 @@
...
+// NewSSHToken generates a JWT token with the correct claims
+// for SSH authentication with Juju controllers.
+// It is expected that this function is called after
+// user authentication and authorization has taken place.
+func (s *SSHTokenGenerator) NewSSHToken(ctx context.Context, tokenArgs SSHTokenArgs) ([]byte, error) {
+	token, err := s.jwtService.NewJWT(ctx, jimmjwx.JWTParams{
+		Controller: tokenArgs.ControllerUUID,
+		User:       tokenArgs.User,
+		Access: map[string]string{
+			tokenArgs.ModelTag.String(): string(permission.AdminAccess),
```
alesstimec, line 56: shouldn't we verify that the user had admin access here?

internal/jimm/jujuauth/factory.go
```diff
@@ -20,7 +20,16 @@ func NewFactory(db GeneratorDatabase, jwtService JWTService, accessChecker Gener
...
+// of a single connection, and recreated for each new connection.
+func (f *Factory) NewLoginGenerator() LoginTokenGenerator {
+	return newLoginTokenGenerator(f.db, f.accessChecker, f.jwtService)
+}
+
+// NewSSHGenerator returns a new token generator for Juju SSH connections.
+// The SSHToken generator is not stateful and can be re-used across
+// multiple connections.
+func (f *Factory) NewSSHGenerator() SSHTokenGenerator {
+	return newSSHTokenGenerator(f.jwtService)
```
alesstimec, line 34: if you pass in the access checker here, you can check if user has admin access to the model when generating new ssh tokens

kian99: Yes we could, but we already do that before this point. I think it works better if the token generator just generates tokens and doesn't also check permissions (if possible).

kian99: What do you think?

internal/jimmjwx/jwt.go
```diff
@@ -142,14 +146,22 @@ func (j *JWTService) NewJWT(ctx context.Context, params JWTParams) ([]byte, erro
...
 		Subject(params.User).
 		Issuer(j.Host).
 		JwtID(jti).
 		Claim("access", params.Access).
-		Expiration(time.Now().Add(j.Expiry)).
-		Build()
+		Expiration(time.Now().Add(j.Expiry))
+
+	for k, v := range params.ExtraClaims {
+		if k == "access" {
```
alesstimec: declare a const "access"

internal/jimm/jujuauth/factory.go
alesstimec: just saying that we had the same "logic" in the model proxy and we've changed it.. i think it's better to check twice and be sure we're doing the right thing.. eventually somebody might decide to change the caller and NOT check if the user is admin..

internal/jimm/jujuauth/sshtokens.go
```diff
@@ -0,0 +1,64 @@
...
+		},
+		ExtraClaims: map[string]any{
+			"ssh_public_key": base64.StdEncoding.EncodeToString(tokenArgs.PublicKey),
```
alesstimec: declare a const "ssh_public_key"

internal/jimm/jujuauth/factory.go
kian99: The difference there is that the model proxy needs to check access dynamically. If the Juju controller responds with "access-required" we need to check for extra access claims. But we don't have those same constraints with the SSH tokens.

kian99: My lean is to check authorisation once and don't do the access check twice until we refactor/see a need.

alesstimec: hmm.. had a quick scan of the jump server implementation. not too fond of our authentication/authorization implementation - it's a bit disorganised imo.. so this is fine for now and we'll need to clean it up in a followup

alesstimec (approved):

internal/jimm/jujuauth/factory.go
SimoneDutto: i agree with always checking before generating a token. A follow-up is fine though

SimoneDutto (approved):
lgtm

PR #1599, opened 2025-05-08
fix: timeout jq-queries

ale8k:
This PR adds a simple timeout to the jq queries so that long running queries end after 5 seconds

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

### DIFF
--- a/cmd/jimmsrv/main.go
+++ b/cmd/jimmsrv/main.go
@@ -185,7 +185,7 @@ func start(ctx context.Context, s *service.Service) error {
 	if crossModelQueryTimeoutRaw != "" {
 		expiry, err := time.ParseDuration(crossModelQueryTimeoutRaw)
 		if err != nil {
-			zapctx.Error(ctx, "failed to parse cross model query timeout duration", zap.Error(err))
+			return errors.E("cannot parse cross model query timeout into duration")
 		} else {
 			crossModelQueryTimeout = expiry
 		}
--- a/cmd/jimmsrv/main.go
+++ b/cmd/jimmsrv/main.go
@@ -180,6 +180,17 @@ func start(ctx context.Context, s *service.Service) error {
 	}
 	jimmUUID := os.Getenv("JIMM_UUID")
 	publicDnsName := os.Getenv("JIMM_DNS_NAME")
+	crossModelQueryTimeout := 5 * time.Second
+	crossModelQueryTimeoutRaw := os.Getenv("JIMM_CROSS_MODEL_QUERY_TIMEOUT")
+	if crossModelQueryTimeoutRaw != "" {
+		expiry, err := time.ParseDuration(crossModelQueryTimeoutRaw)
+		if err != nil {
+			zapctx.Error(ctx, "failed to parse cross model query timeout duration", zap.Error(err))
+		} else {
+			crossModelQueryTimeout = expiry
+		}
+	}
+
 	jimmsvc, err := jimmsvc.NewService(ctx, jimmsvc.Params{
 		ControllerUUID:      jimmUUID,
 		DSN:                 os.Getenv("JIMM_DSN"),
@@ -228,6 +239,7 @@ func start(ctx context.Context, s *service.Service) error {
 			SSHPublicHostKey:            publicHostKey,
 			SSHMaxConcurrentConnections: maxConcurrentConnections,
 		},
+		CrossModelQueryTimeout: crossModelQueryTimeout,
 	})
 	if err != nil {
 		return err
@@ -240,8 +252,6 @@ func start(ctx context.Context, s *service.Service) error {
 		Handler:           jimmsvc,
 		ReadHeaderTimeout: time.Second * 5,
 	}
-
-	// intentionally ignore the error because invalid values are handled by the jump server.
 
 	sshServer, err := ssh.NewJumpServer(ctx, ssh.Config{
 		Port:                     sshPort,
--- a/internal/jimm/juju/model_status_parser.go
+++ b/internal/jimm/juju/model_status_parser.go
@@ -96,7 +96,7 @@ func (j *JujuManager) QueryModelsJq(ctx context.Context, modelUUIDs []string, jq
 			// query. As such, we simply append all to the errors field and continue to collect
 			// both erreoneous and valid query results.
 			if err, ok := v.(error); ok {
-				if stderrors.Is(v.(error), context.DeadlineExceeded) {
+				if stderrors.Is(err, context.DeadlineExceeded) {
 					return results, errors.E(op, fmt.Sprintf("jq query timed out after %.2f seconds", j.CrossModelQueryTimeout.Seconds()), err)
 				}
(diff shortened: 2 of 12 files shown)


internal/jimm/juju/model_status_parser.go
```diff
@@ -20,6 +23,10 @@ import (
 	"github.com/canonical/jimm/v3/pkg/api/params"
 )
 
+var (
+	jqQueryDeadline = time.Second * 5
```
kian99: What about rather making this a parameter on the struct then you don't need to make this a global.

internal/jimm/juju/model_status_parser_test.go
```diff
@@ -757,3 +758,69 @@ func TestQueryModelsJq(t *testing.T) {
...
+		Dialer: jimmtest.ModelDialerMap{
+			"10000000-0000-0000-0000-000000000000": &jimmtest.Dialer{
+				API: &jimmtest.API{
+					Status_: func(_ context.Context, _ []string) (*jujuparams.FullStatus, error) {
+						return &model1, nil
+					},
+					ListFilesystems_: func(ctx context.Context, machines []string) ([]jujuparams.FilesystemDetailsListResult, error) {
+						return []jujuparams.FilesystemDetailsListResult{
+							{
+								Result: []jujuparams.FilesystemDetails{
```
kian99: Do we need an actual result?

internal/jimm/juju/model_status_parser_test.go
```diff
@@ -757,3 +758,69 @@ func TestQueryModelsJq(t *testing.T) {
...
+
+	env := jimmtest.ParseEnvironment(c, crossModelQueryEnv)
+	env.PopulateDB(c, j.Database)
+
+	modelUUIDs := []string{
+		"10000000-0000-0000-0000-000000000000",
+	}
+
+	_, err := j.QueryModelsJq(ctx, modelUUIDs, "range(infinite)")
+	c.Assert(err, qt.ErrorMatches, "jq query timed out after 0.00 seconds")
```
kian99, line 795: Instead of a query like this, I think you can add a channel to synchronise the test.

kian99 (commented):
For the related Jira card, can you update it to mention that our jq queries can cause a memory leak and avoid closing it. I imagine that is still possible right?

internal/jimm/juju/model_status_parser.go
alesstimec: 5s might be a bit low for somebody like IS that has a lot of models..

internal/jimm/juju/model_status_parser_test.go
alesstimec: why does it say timed out after 0 seconds, if you patch the timout to 1 millisecond?

internal/jimm/juju/model_status_parser.go
alesstimec: please make this configurable (via charm config).. let jimm admin decide what is a sensible timout and we can default to something like 10, 20, 30 seconds..

alesstimec (changes requested):

internal/jimm/juju/model_status_parser.go
ale8k: Yeah I'll make it a param, I've no idea what would be best in terms of timing it out.

internal/jimm/juju/model_status_parser_test.go
ale8k: No no, was just c/p.

internal/jimm/juju/model_status_parser_test.go
ale8k: Not sure what you mean? Timing out instantly is simpler and shows us it is working?

ale8k:
> For the related Jira card, can you update it to mention that our jq queries can cause a memory leak and avoid closing it. I imagine that is still possible right?

Well, it is only possible up to the timeout. And you can only do so much. I do have a couple ideas to solve this but I'ma discuss with you guys first, see what you think.

internal/jimm/juju/model_status_parser.go
```diff
@@ -91,6 +96,9 @@ func (j *JujuManager) QueryModelsJq(ctx context.Context, modelUUIDs []string, jq
 			// query. As such, we simply append all to the errors field and continue to collect
 			// both erreoneous and valid query results.
 			if err, ok := v.(error); ok {
+				if stderrors.Is(v.(error), context.DeadlineExceeded) {
+					return results, errors.E(op, fmt.Sprintf("jq query timed out after %.2f seconds", j.CrossModelQueryTimeout.Seconds()), err)
```
kian99:
```suggestion
				if stderrors.Is(err, context.DeadlineExceeded) {
					return results, errors.E(op, fmt.Sprintf("jq query timed out after %.2f seconds", j.CrossModelQueryTimeout.Seconds()), err)
```

internal/jimm/juju/model_status_parser_test.go
kian99: Let's delete the filesystem details to make the test shorter to read.

internal/jimm/juju/model_status_parser_test.go
kian99: I understand things better now. The range(infinite) will make the query run forever and the short timeout will return.
What I was implying was the query might return before your timeout of 1 millisecond, then you need to synchronise the test somehow. That will only be an issue if we disallow these kinds of looping queries.

kian99 (approved):
lgtm with minor changes

internal/jimm/juju/model_status_parser.go
ale8k: Derp good spot

internal/jimm/juju/model_status_parser_test.go
ale8k: Ahhh yeah no no, I've kept it simple this PR as I'm working on some AST walking with depth complexity in a follow up. So the follow up will walk the tree before hand, figure out if it is "OK" and return immediately.

cmd/jimmsrv/main.go
```diff
@@ -155,6 +155,17 @@ func start(ctx context.Context, s *service.Service) error {
 
 	logSQL, _ := strconv.ParseBool(os.Getenv("JIMM_LOG_SQL"))
 
+	crossModelQueryTimeout := 5 * time.Second
+	crossModelQueryTimeoutRaw := os.Getenv("JIMM_CROSS_MODEL_QUERY_TIMEOUT")
+	if crossModelQueryTimeoutRaw != "" {
+		expiry, err := time.ParseDuration(crossModelQueryTimeoutRaw)
+		if err != nil {
+			zapctx.Error(ctx, "failed to parse cross model query timeout duration", zap.Error(err))
```
alesstimec: should we exit, too? i mean, this way JIMM will come up with unexpected configuration. If admin sets it, let them set it right - exit.

alesstimec (approved):
LGTM, but please exit if you can't parse the timeout

SimoneDutto (commented):
why aren't we using scriplets?

PR #1600, opened 2025-05-09
fix: ssh dial multiple addresses

kian99:
## Description
This PR fixes the `DialController` method in the SSH manager, which did not correctly break out of the loop if a dial succeeded. I've also added a test for the functionality.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

### DIFF
--- a/internal/jimm/ssh/ssh.go
+++ b/internal/jimm/ssh/ssh.go
@@ -64,6 +64,8 @@ func (d *BasicDialer) Dial(network string, addr string, config *gossh.ClientConf
 	return gossh.Dial(network, addr, config)
 }
 
+// SSHManagerParams contains the dependencies
+// needed to create the SSHManager service.
 type SSHManagerParams struct {
 	IdentityManager IdentityManager
 	JujuManager     JujuManager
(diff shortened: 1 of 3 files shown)


internal/jimm/ssh/ssh.go
```diff
@@ -51,25 +51,50 @@ type SSHKeyManager interface {
...
+
+// BasicDialer is a wrapper around the default Go x/crypto/ssh
+// dialer for cases where no changes are needed.
+type BasicDialer struct{}
+
+func (d *BasicDialer) Dial(network string, addr string, config *gossh.ClientConfig) (*gossh.Client, error) {
+	return gossh.Dial(network, addr, config)
+}
+
+type SSHManagerParams struct {
```
alesstimec, line 69: godoc, please

internal/jimm/ssh/ssh.go
```diff
@@ -51,25 +51,50 @@ type SSHKeyManager interface {
...
+	SSHKeyManager   SSHKeyManager
+	JWTFactory      *jujuauth.Factory
+	Dialer          SSHDialer
+}
+
 // NewSSHManager returns a new SSHManager that offers jimm functionality to the SSHJumpServer.
-func NewSSHManager(identityManager IdentityManager, jujuManager JujuManager, sshKeyManager SSHKeyManager, jwtFactory *jujuauth.Factory) (*sshManager, error) {
-	if identityManager == nil {
+func NewSSHManager(p SSHManagerParams) (*sshManager, error) {
+	if p.IdentityManager == nil {
```
alesstimec, line 78: could all of this perhaps live in SSHManagerParams validate method?

internal/jimm/ssh/ssh.go
```diff
@@ -51,25 +51,50 @@ type SSHKeyManager interface {
...
 	}
+	if p.Dialer == nil {
+		return nil, errors.E("dialer cannot be nil")
+	}
 	return &sshManager{
-		jujuManager:     jujuManager,
-		identityManager: identityManager,
-		sshKeyManager:   sshKeyManager,
-		jwtFactory:      jwtFactory,
+		jujuManager:     p.JujuManager,
```
alesstimec, line 102: seems like sshManager could just embed the SSHManagerParams?

alesstimec (approved):
LGTM with a few suggestions

internal/jimm/ssh/ssh.go
kian99: Yip, good idea.

internal/jimm/ssh/ssh.go
kian99: True, I'll do that

kian99: Ah, it can't right now, because the params are all exported, while the types on the actual manager are not exported. That could change in the future but I won't do it in this PR.

SimoneDutto (approved):
good

PR #1601, opened 2025-05-09
Implement controller config

SimoneDutto:
Add config manager to populate the config map to return when the ControllerConfig facade is called.

#QA

```
make qa-microk8s
juju controller-config
```

```console
[12:54:40] ➜  jimm git:(implement-controller-config) ✗ juju controller-config
Attribute                       Value
api-port                        80
controller-uuid                 3217dbc9-8ea9-4381-9e97-01eab0b3f6bb
public-dns-address              jimm.localhost
ssh-max-concurrent-connections  10
ssh-server-hostkey              |
  ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABgQCK/35j+/5hjCgaRjyMdd9AXOua+/rsD3dhN92VBJPqK/SkhPuprq+lvlKSyww3zFMLiImfxAByYQx9RJe+TkxerSWzXTAuzoBeGeCjqeEV00Xyyb+axN0WMoyZMRdSQ4/RXlVNlwGodcweCXJJ6Q/Yz00bivCsHM2kimSd7FHpHZP5Db5IG39nkuk0YqPjZPucW3w8vVEUMUtYwM3chZR3VocFal0ZFqWVJbvH7rQNJDXiB2kVyBR33eEQzsL/yAm8VUo5GVWG6v9fjEuE7kiIOosrGxBKwN3qVuOmUrLe/8g6V81pREGYpZ0z9z6HNWJg6SWe8uSsqaXkHO15ibYO3gp6nhvs+92Iex3gOcb6zj+mGf7PJZiD9eH7jBob+uL42If4fgvXLPj6l/V34MD4KDFDUVzJLrYPtwSPaqEdHQElWOl5lP4/HTYMw4humeqUJwiskcmxb8jiezRHrsgdKO3y7KO+77Rd5+mAw2pYl2kCDxR7YJyWSU7bHSWv4Hs=
ssh-server-port  17022
```


internal/jimm/config/config.go
```diff
@@ -0,0 +1,60 @@
...
+	}, nil
+}
+
+// GetConfig returns the configuration for the controller converted in the jujuparams format.
+func (c *configManager) GetConfig() (jujuparams.ControllerConfig, error) {
+	cfg := make(map[string]interface{})
+	cfg[jujucontroller.ControllerUUIDKey] = c.config.ControllerUUID
+	cfg[jujucontroller.APIPort] = c.config.APIPort
+	cfg[jujucontroller.SSHServerPort] = c.config.SSHPort
+	// TODO: update this to use the key coming from juju when we update the juju dependency.
```
kian99: Why would we use the key from Juju? We would only ever display jimm's key to the user.

local/jimm/add-microk8s-controller.sh
```diff
@@ -25,4 +25,4 @@ microk8s.kubectl patch -n controller-qa-microk8s svc/controller-service --type='
 CONTROLLER_NAME="${CONTROLLER_NAME:-qa-microk8s}"
 CLIENT_CREDENTIAL_NAME="${CLIENT_CREDENTIAL_NAME:-microk8s}"
 CUSTOM_CONTROLLER_HOSTNAME="${CUSTOM_CONTROLLER_HOSTNAME:-172.17.0.1:30040}"
-"$(dirname "${BASH_SOURCE[0]}")/add-controller.sh"
+CONTROLLER_NAME="$CONTROLLER_NAME" CLIENT_CREDENTIAL_NAME="$CLIENT_CREDENTIAL_NAME" CUSTOM_CONTROLLER_HOSTNAME="$CUSTOM_CONTROLLER_HOSTNAME" "$(dirname "${BASH_SOURCE[0]}")/add-controller.sh"
```
kian99, line 28: Good catch on this. Alternatively you could `export` the vars so that subprocesses receive them.
So,
```
export CONTROLLER_NAME="${CONTROLLER_NAME:-qa-microk8s}"
...
```

cmd/jimmsrv/main.go
```diff
@@ -195,6 +218,14 @@ func start(ctx context.Context, s *service.Service) error {
 		LogSQL:                    logSQL,
 		LogLevel:                  logLevel,
 		IsLeader:                  os.Getenv("JIMM_IS_LEADER") != "",
+		ControllerConfig: config.ControllerConfig{
+			ControllerUUID:              os.Getenv("JIMM_UUID"),
+			PublicDNSName:               os.Getenv("JIMM_DNS_NAME"),
```
kian99: In the future we should probably do validation on these, but not something for this PR.

cmd/jimmsrv/main.go
```diff
@@ -195,6 +218,14 @@ func start(ctx context.Context, s *service.Service) error {
 		LogSQL:                    logSQL,
 		LogLevel:                  logLevel,
 		IsLeader:                  os.Getenv("JIMM_IS_LEADER") != "",
+		ControllerConfig: config.ControllerConfig{
+			ControllerUUID:              os.Getenv("JIMM_UUID"),
```
kian99: What about parsing this variable once and re-using the result. Not a biggie but I prefer seeing the env vars parsed once, and I see we have this used elsewhere too.

internal/jimm/config/config.go
SimoneDutto: the keys must be the same to be able to access the hostkey of JIMM and Juju controller from the CLI seemlessly

local/jimm/add-microk8s-controller.sh
SimoneDutto: i didn't want to do it, because it messes with the bash you are executing the script from.
In the way i did it, nothing is set in your bash from this script

cmd/jimmsrv/main.go
SimoneDutto: yeah, why not!
i'll wait for the second review and i'll do it

internal/jimm/config/config.go
kian99: I'm not following, the CLI doesn't need the Juju controller's host key when connecting through jimm. Maybe we're talking about different things here.

SimoneDutto: when you do `juju ssh ` we will fetch the ssh port and ssh hostkey from the controller config to resolve the target and populate the temporary hostkey file.

If both juju and jimm list the config with the same key we can do what i've described without any changes between jimm flow and juju flow.

cmd/jimmsrv/main.go
```diff
@@ -155,6 +162,22 @@ func start(ctx context.Context, s *service.Service) error {
...
+	}
+	portInt, err := strconv.Atoi(port)
+	if err != nil {
+		return errors.E("failed to parse port", zap.Error(err))
+	}
+	sshPort := os.Getenv("JIMM_SSH_PORT")
+	if sshPort == "" {
+		return errors.E("empty ssh port from env variable")
+	}
+	sshPortInt, err := strconv.Atoi(sshPort)
```
ale8k, line 177: Could also check it's > 0

cmd/jimmsrv/main.go
```diff
@@ -195,6 +218,14 @@ func start(ctx context.Context, s *service.Service) error {
 		LogSQL:                    logSQL,
 		LogLevel:                  logLevel,
 		IsLeader:                  os.Getenv("JIMM_IS_LEADER") != "",
+		ControllerConfig: config.ControllerConfig{
```
ale8k, line 223: Ty for doing the struct

internal/jimm/config/config.go
```diff
@@ -0,0 +1,60 @@
...
+
+// NewConfigManager creates a new config manager with the given configuration.
+func NewConfigManager(config ControllerConfig) (*configManager, error) {
+	return &configManager{
+		config: config,
+	}, nil
+}
+
+// GetConfig returns the configuration for the controller converted in the jujuparams format.
+func (c *configManager) GetConfig() (jujuparams.ControllerConfig, error) {
```
ale8k: This manager shouldn't return a jujuparams.ControllerConfig (As it's a DTO), this is an API layer param, to keep the concerns separated just return configManager.config & then in the facade, map it over.

internal/jimm/config/config.go
ale8k: Not sure I'm following here either, I was under the same impression as Kian. We're defining our own keys? Is it because ControllerConfig uses setters/getters to access the map you want to use the same keys?

If that's the case, you can't access "ssh-server-hostkey" anyway?

local/jimm/add-microk8s-controller.sh
ale8k: This usually just worked, surprised you needed to add the vars?

ale8k (commented):
Looks good, but let's not return a jujuparam from the manager. It's mixing api with business/application logic. DTO's should be mapped in API layer. Granted there's a lot of contention on this topic, I often think "the mapping should happen where it is needed" if that helps persuade you.

internal/jimm/config/config.go
kian99: +1 to this ^^

internal/jimm/config/config.go
kian99: Ahh I understand now. You mean just use the same key-name/key-tag/thing. I.e. "ssh-host-key", not the actual key from Juju.

kian99: The fact that the key in key/value is the same word as in pki keys, made that way more confusing than it needed to be.

SimoneDutto: yeah!

ale8k (approved):
Approved as discussed to change returning DTO. Good work!

internal/jimm/config/config.go
SimoneDutto: done

internal/jujuapi/controller.go
```diff
@@ -220,9 +221,24 @@ func (r *controllerRoot) ModelStatus(ctx context.Context, args jujuparams.Entiti
...
+	config, err := r.jimm.ConfigManager().GetConfig()
+	if err != nil {
+		return jujuparams.ControllerConfigResult{}, errors.E(op, err)
+	}
+	cfg := make(map[string]interface{})
+	cfg[jujucontroller.ControllerUUIDKey] = config.ControllerUUID
+	cfg[jujucontroller.APIPort] = config.APIPort
+	cfg[jujucontroller.SSHServerPort] = config.SSHPort
+	// TODO: update this to use the key coming from juju when we update the juju dependency.
+	cfg["ssh-server-hostkey"] = config.SSHPublicHostKey
```
kian99, line 236: I've made it `ssh-host-key` in my WIP. We won't be able to reference a concrete type until the feature branch merges into 3.6 (unless I merge my change into 3.6).

kian99 (approved):

internal/jujuapi/controller.go
SimoneDutto: yep

PR #1603, opened 2025-05-22
fix(ssh-jump): fix key to get hostkey from controller config and get addrs to dial

SimoneDutto:
In this PR two small fixes:
- set the right key to extract the hostkey when calling ControllerConfig with JIMM and JUJU
- gather all available addresses. This is a fix for my setup (and potentially others), in which using the PublicAddress to ssh wasn't working. Isn't just better to always try all addresses and not to stop at the public one?

QA here: https://github.com/juju/juju/pull/19814

### DIFF
--- a/internal/rpc/dial.go
+++ b/internal/rpc/dial.go
@@ -68,14 +68,13 @@ func GetAddressesAndTLSConfig(ctx context.Context, ctl *dbmodel.Controller) ([]s
 			MinVersion: tls.VersionTLS12,
 		}
 	}
-
+	var addrs []string
 	if ctl.PublicAddress != "" {
 		// If there is a public-address configured it is almost
 		// certainly the one we want to use.
-		return []string{ctl.PublicAddress}, tlsConfig
+		addrs = append(addrs, ctl.PublicAddress)
 	}
 
-	var addrs []string
 	for _, hps := range ctl.Addresses {
 		for _, hp := range hps {
 			if maybeReachable(hp.Scope) {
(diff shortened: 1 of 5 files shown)


internal/rpc/dial.go
```diff
@@ -68,14 +68,13 @@ func GetAddressesAndTLSConfig(ctx context.Context, ctl *dbmodel.Controller) ([]s
 			MinVersion: tls.VersionTLS12,
 		}
 	}
-
+	var addrs []string
 	if ctl.PublicAddress != "" {
 		// If there is a public-address configured it is almost
 		// certainly the one we want to use.
-		return []string{ctl.PublicAddress}, tlsConfig
+		addrs = append(addrs, ctl.PublicAddress)
```
SimoneDutto, line 73: !!extra attention here!!

internal/rpc/dial.go
```diff
@@ -68,14 +68,13 @@ func GetAddressesAndTLSConfig(ctx context.Context, ctl *dbmodel.Controller) ([]s
 			MinVersion: tls.VersionTLS12,
 		}
 	}
-
+	var addrs []string
 	if ctl.PublicAddress != "" {
 		// If there is a public-address configured it is almost
 		// certainly the one we want to use.
```
kian99: You'll need to update this comment

internal/rpc/dial.go
kian99: Happy with this change, we'll try the public address first then the rest.

kian99 (approved):

alesstimec (approved):

PR #1606, opened 2025-05-30
feat(check-relations): add facade method to perform checks on multiple tuples

SimoneDutto:
## Description
In this PR we add a facade to perform multiple auth checks providing a list of tuples.
This is to facilitate the WebUI.

The implementation is very straight forward for now, and it is just a loop calling the single tuple facade.
The reason behind it is that:
- we can’t use BatchCheck because it was released in [release: v0.6.6 by ryanpq · Pull Request #188 · openfga/go-sdk](https://github.com/openfga/go-sdk/pull/188) and we use canonical openfga which is [ofga/go.mod at 2745ae6e630671d682c0596a0af5e0839eda0f10 · canonical/ofga](https://github.com/canonical/ofga/blob/2745ae6e630671d682c0596a0af5e0839eda0f10/go.mod#L9) 0.6.3
- we don't need to re-write the auth logic to perform auth check

The result is a list of ordered CheckResult containing allowed and the error if any.

Once this is landed we need to release the new jimm sdk as well.


internal/openfga/openfga.go
```diff
@@ -57,6 +57,13 @@ var (
 	ServiceAccountType Kind = jimmnames.ServiceAccountTagKind
 )
 
+type CheckResult struct {
```
alesstimec, line 63: godoc

alesstimec (commented):
LGTM. let's check with the web team, if this meets their requirements

alesstimec (approved):

internal/jimm/permissions/relations.go
```diff
@@ -75,6 +75,23 @@ func (j *permissionManager) CheckRelation(ctx context.Context, user *openfga.Use
 	return allowed, nil
 }
 
+// CheckRelations checks user permission and returns a slice of CheckResult for each tuple.
+// At the moment the implementation is a simple loop around CheckRelation.
+func (j *permissionManager) CheckRelations(ctx context.Context, user *openfga.User, tuples []apiparams.RelationshipTuple) ([]openfga.CheckResult, error) {
```
BarcoMasile, line 82: question: since this is an auth check, wouldn't it make sense to circuit break as soon as one of the result has `Allowed == false` and exit the loop, or even if an error comes back?

SimoneDutto: It depends on how it is used.
For example if you have a list of models in UI, and you want to list just the model the user has access to you call this endpoint and it returns the list of the models you have permission one.
If i return early or on the first error you wouldn't be able to do it.

BarcoMasile: Oh I see the use case now, thanks for clarifying!
My nitpick suggestion here would be to actually call it BatchCheck since CheckTuples makes me (at least) think that every check must be positive and it's correlated to the others

internal/jimm/permissions/relations.go
```diff
@@ -75,6 +77,23 @@ func (j *permissionManager) CheckRelation(ctx context.Context, user *openfga.Use
 	return allowed, nil
 }
 
+// CheckRelations checks user permission and returns a slice of CheckResult for each tuple.
```
kian99:
```suggestion
// CheckRelations checks user permissions and returns a slice of CheckResult for each tuple.
```

internal/jimm/permissions/relations.go
```diff
@@ -54,7 +54,9 @@ func (j *permissionManager) RemoveRelation(ctx context.Context, user *openfga.Us
 }
 
 // CheckRelation checks user permission and return true if the given tuple exists.
-// At the moment user is required be admin or checking its own relations
+// At the moment user is required be admin or checking its own relations.
+// TODO(simonedutto): this is a temporary implementation, once canonical/openfga supports BatchCheck
+// we can use that to improve performance.
```
kian99: Shouldn't this TODO be on the CheckRelations method instead?

internal/jujuapi/access_control_test.go
```diff
@@ -1427,6 +1427,18 @@ func (s *accessControlSuite) TestCheckRelationControllerAdministratorFlow(c *gc.
 		c.Assert(err, gc.IsNil)
 		c.Assert(res.Allowed, gc.Equals, tc.want)
 	}
+	// Check that the same tuples can be checked via the CheckRelations API
```
kian99, line 1430: I'm not the biggest fan of putting this test inside another test. It's a whole separate API method, I think it should get its own test unless you feel strongly that it's fine like this.

kian99 (commented):
Looking at OpenFGA's BatchCheck method [here](https://openfga.dev/api/service#/Relationship%20Queries/BatchCheck), each tuple being checked requires a "correlation ID" that is used in the response to correlate the result back to the appropriate input. If/when we switch to the native BatchCheck we will have to make a breaking change to our API. I suggest we either incorporate a correlation id to our request/response or simply make a PR to the canonical/ofga library to update it and then use that (and additionally check our OpenFGA charm is using a supported version). We would also need to update the OpenFGA docker image we use in our tests.

SimoneDutto:
> Looking at OpenFGA's BatchCheck method [here](https://openfga.dev/api/service#/Relationship%20Queries/BatchCheck), each tuple being checked requires a "correlation ID" that is used in the response to correlate the result back to the appropriate input. If/when we switch to the native BatchCheck we will have to make a breaking change to our API. I suggest we either incorporate a correlation id to our request/response or simply make a PR to the canonical/ofga library to update it and then use that (and additionally check our OpenFGA charm is using a supported version). We would also need to update the OpenFGA docker image we use in our tests.

Yeah yeah, i saw the `correlation_id` stuff. But it won't be a breaking change in our API.
We can always support our clients sending a list, we generate the correlation_id in jimm before sending the Batch request to Openfga and we recompose the list in our response.

To me making the client generate and keep somewhere the map with the `correlation_id` is a <little> worse, because we expose an implementation detail.
What do you think?

To update the upstream is a little different, because it isn't enough to just upgrade the openfga sdk, but you would also need to add the `BatchCheck` method + tests. So i would rather just open a generic issue.

Edit:
We can always change the implementation once we have the upgraded `BatchCheck` but for now this implementation would help the web team

internal/jimm/permissions/relations.go
SimoneDutto: Ah yeah, my bad

kian99:
> Yeah yeah, i saw the `correlation_id` stuff. But it won't be a breaking change in our API.
> We can always support our clients sending a list, we generate the correlation_id in jimm before sending the Batch request to Openfga and we recompose the list in our response.

Yeah true true. What do you envision? We get a list of tuples, we generate the correlation IDs and get the result from OpenFGA and then send back the results in the same order as was requested?

kian99 (approved):
Lgtm, I agree that the name should be changed to BatchCheck

SimoneDutto:
> > Yeah yeah, i saw the `correlation_id` stuff. But it won't be a breaking change in our API.
> > We can always support our clients sending a list, we generate the correlation_id in jimm before sending the Batch request to Openfga and we recompose the list in our response.
> 
> Yeah true true. What do you envision? We get a list of tuples, we generate the correlation IDs and get the result from OpenFGA and then send back the results in the same order as was requested?

yeah

SimoneDutto:
> Lgtm, I agree that the name should be changed to BatchCheck

I would agree on `BatchCheck` if we called the other `Check`. Since it's called `CheckRelation`, it makes sense to call this `CheckRelations` to be more consistent

PR #1607, opened 2025-06-03
chore: remove special case for service accounts

kian99:
## Description
In JIMM there exists a `serviceaccount` tag that is a special case used to manage a service account from a user's account. This was done so that a user could create cloud-credentials for a service account. Now a service account can be fully managed by Terraform so there is no need for all this special casing.

This PR removes all use of the `serviceaccount` tag in OpenFGA and in JIMM's types and all methods associated with it.

This PR cannot land until https://github.com/juju/terraform-provider-juju/pull/737 lands to ensure the Terraform Provider's integration tests with JIMM continue to work. I will test the changes with the Terraform provider on my fork to verify the tests will continue to work after this lands. See the first test run output [here](https://github.com/kian99/terraform-provider-juju/actions/runs/15417431456/job/43383484056).

Fixes [JUJU-7844](https://warthogs.atlassian.net/browse/JUJU-7844)

[JUJU-7844]: https://warthogs.atlassian.net/browse/JUJU-7844?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/jimmhttp/rebac_admin/resources.go
```diff
@@ -88,8 +88,6 @@ func validateAndConvertResourceFilter(typeFilter string) (string, error) {
 		return db.ControllersQueryKey, nil
 	case Model:
 		return db.ModelsQueryKey, nil
-	case ServiceAccount:
-		return db.ServiceAccountQueryKey, nil
```
kian99, line 92: @SimoneDutto I'm not sure about my removal of service account here, do you know if this should be kept or removed?

SimoneDutto: As far as i remember service account were just a resource like any other.

Digging a little into the code we were already avoiding creating service accounts via `identities.CreateIdentity`.
So i think we can get rid of service accounts.

SimoneDutto (approved):
LGTM

ale8k (approved):
Bye bye service account management :wave:

PR #1608, opened 2025-06-06
feat: allow anonymous user login

kian99:
## Description
Anoynmous login in Juju is used when controller's communicate for the purposes of cross-model relation. With the development of model-imports where a model that was not created through JIMM will now be managed by JIMM, we need to support proxying anonymous login requests through to controllers.

Fixes [JUJU-8020](https://warthogs.atlassian.net/browse/JUJU-8020)

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

[JUJU-8020]: https://warthogs.atlassian.net/browse/JUJU-8020?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/rpcproxy/rpcproxy.go
```diff
@@ -573,6 +575,9 @@ func checkPermissionsRequired(ctx context.Context, msg *message) (map[string]any
 func (p *controllerProxy) redoLogin(ctx context.Context, permissions map[string]any) error {
 	const op = errors.Op("rpc.redoLogin")
 
+	if p.anonymousLogin {
+		return errors.E(op, errors.CodeUnauthorized, "Anonymous login does not support re-authentication")
```
ale8k, line 583: Just so I'm clear... This is when the controller "anonymous logs-in", and we wouldn't redo it for permission checks because the checks in the model import flow are done at discharge? In anycase, this needs thoroughly godoc'ing because it looks really odd and is niche to offers.

internal/rpcproxy/rpcproxy.go
```diff
@@ -735,6 +740,20 @@ func (p *clientProxy) handleAdminFacade(ctx context.Context, msg *message) (clie
 
 		return controllerLoginMessageFnc(user)
 	case "Login":
+		var request params.LoginRequest
```
ale8k: I'd prefer this in a func called "handleAnonymousLogin" (whole lot can go in because we'll never support basic logic anyway), and then on that func godoc it so it's clear why we're supporting it and what it aims to achieve.

internal/rpcproxy/rpcproxylogin_test.go
```diff
@@ -173,6 +173,37 @@ func TestProxySocketsAdminFacade(t *testing.T) {
...
 		},
 		oauthAuthenticatorError: errors.E(errors.CodeUnauthorized),
+	}, {
+		about: "login with username/password fails",
+		messageToSend: rpcproxy.Message{
+			RequestID: 1,
+			Type:      "Admin",
+			Version:   3,
+			Request:   "Login",
+			Params:    []byte(`{"auth-tag":"user-bob"}`),
```
ale8k, line 183: Why is the auth-tag user-bob, why not user-admin?

ale8k (approved):
LGTM, just some comments on godoc'ing this more please. It'd be greatly appreciated!

internal/rpcproxy/rpcproxy.go
alesstimec: we can't do relogin because we can't gather additional facts (authorization checks) for an anonymous user..  well.. not entirely true - we could check if everyone@external has access to an application offer, for example

internal/rpcproxy/rpcproxy.go
```diff
@@ -735,6 +740,20 @@ func (p *clientProxy) handleAdminFacade(ctx context.Context, msg *message) (clie
...
+		err := json.Unmarshal(msg.Params, &request)
+		if err != nil {
+			return errorFnc(err)
+		}
+		user, err := names.ParseUserTag(request.AuthTag)
+		if err != nil {
+			return errorFnc(fmt.Errorf("invalid user tag: %v", err))
+		}
+		if user.Name() == api.AnonymousUsername {
```
alesstimec: and empty domain?

alesstimec (approved):

internal/rpcproxy/rpcproxy.go
kian99: Good catch, switched to using `user.Id()` which will contain the domain

internal/rpcproxy/rpcproxy.go
kian99: Changed.

internal/rpcproxy/rpcproxylogin_test.go
kian99: No reason, it could be anything.

internal/rpcproxy/rpcproxy.go
kian99: Exactly as Ales has explained. I've added a godoc for the function.

PR #1609, opened 2025-06-06
feat: add user mapping and model migrations tables

kian99:
## Description
This PR adds 2 new tables called `model_migrations` and `user_mappings` which will be used when migrating models to JIMM. In full this PR adds,
- 1 new SQL schema migration.
- The Go representations for the new structs in `internal/dbmodel`
- CRUD methods in `internal/db`, not all CRUD methods have been added, only those that make sense.

Fixes [JUJU-7950](https://warthogs.atlassian.net/browse/JUJU-7950)

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

## Test instructions

[JUJU-7950]: https://warthogs.atlassian.net/browse/JUJU-7950?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/db/modelmigration.go
+++ b/internal/db/modelmigration.go
@@ -1,0 +1,79 @@
+// Copyright 2025 Canonical.
+
+package db
+
+import (
+	"context"
+
+	"github.com/canonical/jimm/v3/internal/dbmodel"
+	"github.com/canonical/jimm/v3/internal/errors"
+	"github.com/canonical/jimm/v3/internal/servermon"
+)
+
+// AddModelMigration stores information about an incoming model migration.
+//   - returns an error with code errors.CodeAlreadyExists if
+//     a migration row with the same model UUID already exists.
+func (d *Database) AddModelMigration(ctx context.Context, modelMigration *dbmodel.ModelMigration) (err error) {
+	const op = errors.Op("db.AddModelMigration")
+	if err := d.ready(); err != nil {
+		return errors.E(op, err)
+	}
+
+	durationObserver := servermon.DurationObserver(servermon.DBQueryDurationHistogram, string(op))
+	defer durationObserver()
+	defer servermon.ErrorCounter(servermon.DBQueryErrorCount, &err, string(op))
+
+	db := d.DB.WithContext(ctx)
+
+	if err := db.Create(modelMigration).Error; err != nil {
+		return errors.E(op, dbError(err))
+	}
+	return nil
+}
+
+// GetModelMigration returns model migration information based on the model UUID.
+func (d *Database) GetModelMigration(ctx context.Context, modelMigration *dbmodel.ModelMigration) (err error) {
+	const op = errors.Op("db.GetModelMigration")
+	if err := d.ready(); err != nil {
+		return errors.E(op, err)
+	}
+
+	durationObserver := servermon.DurationObserver(servermon.DBQueryDurationHistogram, string(op))
+	defer durationObserver()
+	defer servermon.ErrorCounter(servermon.DBQueryErrorCount, &err, string(op))
+
+	db := d.DB.WithContext(ctx)
+	switch {
+	case modelMigration.ModelUUID.Valid:
+		db = db.Where("model_uuid = ?", modelMigration.ModelUUID.String)
+	default:
+		return errors.E(op, "missing uuid", errors.CodeBadRequest)
+	}
+
+	if err := db.First(&modelMigration).Error; err != nil {
+		err = dbError(err)
+		if errors.ErrorCode(err) == errors.CodeNotFound {
+			return errors.E(op, err, "model migration not found")
+		}
+		return errors.E(op, dbError(err))
+	}
+	return nil
+}
+
+// DeleteModelMigration removes a model migration entry from the database.
+func (d *Database) DeleteModelMigration(ctx context.Context, modelMigration *dbmodel.ModelMigration) (err error) {
+	const op = errors.Op("db.DeleteModelMigration")
+	if err := d.ready(); err != nil {
+		return errors.E(op, err)
+	}
+
+	durationObserver := servermon.DurationObserver(servermon.DBQueryDurationHistogram, string(op))
+	defer durationObserver()
+	defer servermon.ErrorCounter(servermon.DBQueryErrorCount, &err, string(op))
+
+	db := d.DB.WithContext(ctx)
+	if err := db.Delete(modelMigration).Error; err != nil {
+		return errors.E(op, dbError(err))
+	}
+	return nil
+}
--- a/internal/db/modelmigration.go
+++ b/internal/db/modelmigration.go
@@ -13,7 +13,7 @@ import (
 // AddModelMigration stores information about an incoming model migration.
 //   - returns an error with code errors.CodeAlreadyExists if
 //     a migration row with the same model UUID already exists.
-func (d *Database) AddModelMigration(ctx context.Context, modelMigration *dbmodel.ModelMigration) (err error) {
+func (d *Database) AddModelMigration(ctx context.Context, modelMigration *dbmodel.IncomingModelMigration) (err error) {
 	const op = errors.Op("db.AddModelMigration")
 	if err := d.ready(); err != nil {
 		return errors.E(op, err)
@@ -32,7 +32,7 @@ func (d *Database) AddModelMigration(ctx context.Context, modelMigration *dbmode
 }
 
 // GetModelMigration returns model migration information based on the model UUID.
-func (d *Database) GetModelMigration(ctx context.Context, modelMigration *dbmodel.ModelMigration) (err error) {
+func (d *Database) GetModelMigration(ctx context.Context, modelMigration *dbmodel.IncomingModelMigration) (err error) {
 	const op = errors.Op("db.GetModelMigration")
 	if err := d.ready(); err != nil {
 		return errors.E(op, err)
@@ -61,7 +61,7 @@ func (d *Database) GetModelMigration(ctx context.Context, modelMigration *dbmode
 }
 
 // DeleteModelMigration removes a model migration entry from the database.
-func (d *Database) DeleteModelMigration(ctx context.Context, modelMigration *dbmodel.ModelMigration) (err error) {
+func (d *Database) DeleteModelMigration(ctx context.Context, modelMigration *dbmodel.IncomingModelMigration) (err error) {
 	const op = errors.Op("db.DeleteModelMigration")
 	if err := d.ready(); err != nil {
 		return errors.E(op, err)
--- a/internal/db/modelmigration_test.go
+++ b/internal/db/modelmigration_test.go
@@ -5,22 +5,12 @@ package db_test
 import (
 	"context"
 	"database/sql"
-	"testing"
 
 	qt "github.com/frankban/quicktest"
 
-	"github.com/canonical/jimm/v3/internal/db"
 	"github.com/canonical/jimm/v3/internal/dbmodel"
 	"github.com/canonical/jimm/v3/internal/errors"
 )
-
-func TestAddModelMigrationUnconfiguredDatabase(t *testing.T) {
-	c := qt.New(t)
-	var d db.Database
-	err := d.AddModelMigration(context.Background(), &dbmodel.ModelMigration{})
-	c.Check(err, qt.ErrorMatches, `database not configured`)
-	c.Check(errors.ErrorCode(err), qt.Equals, errors.CodeServerConfiguration)
-}
 
 func (s *dbSuite) TestAddModelMigration(c *qt.C) {
 	err := s.Database.Migrate(context.Background())
--- a/internal/db/modelmigration_test.go
+++ b/internal/db/modelmigration_test.go
@@ -1,0 +1,155 @@
+// Copyright 2025 Canonical.
+
+package db_test
+
+import (
+	"context"
+	"database/sql"
+	"testing"
+
+	qt "github.com/frankban/quicktest"
+
+	"github.com/canonical/jimm/v3/internal/db"
+	"github.com/canonical/jimm/v3/internal/dbmodel"
+	"github.com/canonical/jimm/v3/internal/errors"
+)
+
+func TestAddModelMigrationUnconfiguredDatabase(t *testing.T) {
+	c := qt.New(t)
+	var d db.Database
+	err := d.AddModelMigration(context.Background(), &dbmodel.ModelMigration{})
+	c.Check(err, qt.ErrorMatches, `database not configured`)
+	c.Check(errors.ErrorCode(err), qt.Equals, errors.CodeServerConfiguration)
+}
+
+func (s *dbSuite) TestAddModelMigration(c *qt.C) {
+	err := s.Database.Migrate(context.Background())
+	c.Assert(err, qt.Equals, nil)
+
+	cloud := dbmodel.Cloud{
+		Name: "test-cloud",
+	}
+	err = s.Database.AddCloud(context.Background(), &cloud)
+	c.Assert(err, qt.IsNil)
+
+	controller := dbmodel.Controller{
+		Name:      "test-controller",
+		UUID:      "00000000-0000-0000-0000-000000000001",
+		CloudName: cloud.Name,
+	}
+	c.Assert(s.Database.DB.Create(&controller).Error, qt.IsNil)
+
+	migration := dbmodel.ModelMigration{
+		ModelUUID:          sql.NullString{String: "00000001-0000-0000-0000-000000000001", Valid: true},
+		TargetControllerID: controller.ID,
+		UserMapping:        dbmodel.JSON([]byte(`{"local": "external"}`)),
+	}
+	err = s.Database.AddModelMigration(context.Background(), &migration)
+	c.Assert(err, qt.Equals, nil)
+
+	var dbMigration dbmodel.ModelMigration
+	result := s.Database.DB.Where("model_uuid = ?", migration.ModelUUID).First(&dbMigration)
+	c.Assert(result.Error, qt.Equals, nil)
+	c.Assert(dbMigration.ModelUUID, qt.DeepEquals, migration.ModelUUID)
+	c.Assert(dbMigration.TargetControllerID, qt.Equals, controller.ID)
+	c.Assert(dbMigration.UserMapping, qt.DeepEquals, migration.UserMapping)
+}
+
+func (s *dbSuite) TestAddModelMigrationAlreadyExists(c *qt.C) {
+	err := s.Database.Migrate(context.Background())
+	c.Assert(err, qt.Equals, nil)
+
+	cloud := dbmodel.Cloud{
+		Name: "test-cloud",
+	}
+	err = s.Database.AddCloud(context.Background(), &cloud)
+	c.Assert(err, qt.IsNil)
+
+	controller := dbmodel.Controller{
+		Name:      "test-controller",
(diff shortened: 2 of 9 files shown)


internal/dbmodel/usermapping.go
```diff
@@ -0,0 +1,27 @@
...
+	CreatedAt time.Time
+	UpdatedAt time.Time
+
+	// ModelUUID is the UUID of the model the mapping applies to.
+	ModelUUID sql.NullString
+
+	// LocalUser is the local user that this mapping applies to.
+	LocalUser string
+
+	// ExtneralUserName is the external user that this local user maps to.
```
SimoneDutto:
```suggestion
	// ExternalUserName is the external user that this local user maps to.
```

internal/db/modelmigration.go
```diff
@@ -0,0 +1,79 @@
...
+import (
+	"context"
+
+	"github.com/canonical/jimm/v3/internal/dbmodel"
+	"github.com/canonical/jimm/v3/internal/errors"
+	"github.com/canonical/jimm/v3/internal/servermon"
+)
+
+// AddModelMigration stores information about an incoming model migration.
+//   - returns an error with code errors.CodeAlreadyExists if
```
SimoneDutto, line 14: i like this type of comment.

internal/dbmodel/modelmigration.go
```diff
@@ -0,0 +1,25 @@
...
+
+	// ModelUUID is the UUID of the incoming model.
+	ModelUUID sql.NullString
+
+	// TargetControllerID is the target controller for the model undegroing migration.
+	TargetControllerID uint
+	TargetController   Controller
+
+	// UserMapping holds a mapping of local users to external users.
+	UserMapping JSON
```
SimoneDutto, line 27: Why do we hold this JSON reference?
At the end of the day we use the JSON to create the user mappings in our db, so we can get rid of the original JSON mapping right?

internal/dbmodel/usermapping.go
```diff
@@ -0,0 +1,27 @@
...
+	"time"
+)
+
+// UserMapping holds information per model on
+// mapping local users to external users.
+type UserMapping struct {
+	// Note this doesn't use the standard gorm.Model to avoid soft-deletes.
+	ID        uint `gorm:"primarykey"`
+	CreatedAt time.Time
+	UpdatedAt time.Time
```
SimoneDutto, line 16: There isn't any method to update the `UserMapping`.
If we'll create that since we are not using `gorm.Model` remember to manually handle the UpdateAt.

internal/dbmodel/sql/postgres/024_add_model_migration_tables.up.sql
```diff
@@ -0,0 +1,27 @@
...
+
+CREATE TABLE model_migrations (
+    id SERIAL PRIMARY KEY,
+    created_at TIMESTAMP WITH TIME ZONE DEFAULT NOW(),
+    model_uuid TEXT NOT NULL UNIQUE,
+    target_controller_id INTEGER REFERENCES controllers (id),
+    user_mapping JSONB NOT NULL
+);
+
+CREATE INDEX idx_model_migrations_model_uuid ON model_migrations (model_uuid);
```
SimoneDutto: I don't think you need this one, because you already have `model_uuid TEXT NOT NULL UNIQUE`, unless you are doing this to improve queries performance, which I don't think would ever be a problem on this table.

SimoneDutto (commented):
looking good, some comments

internal/dbmodel/sql/postgres/024_add_model_migration_tables.up.sql
```diff
@@ -0,0 +1,27 @@
...
+    id SERIAL PRIMARY KEY,
+    created_at TIMESTAMP WITH TIME ZONE DEFAULT NOW(),
+    updated_at TIMESTAMP WITH TIME ZONE,
+    model_uuid TEXT NOT NULL,
+    local_user TEXT NOT NULL,
+    external_user TEXT NOT NULL REFERENCES identities(name),
+    CONSTRAINT unique_user_mappings_key UNIQUE(model_uuid, local_user)
+);
+
+CREATE INDEX idx_model_user_mappings_local_user ON user_mappings (model_uuid, local_user);
```
SimoneDutto: I don't think you need this either. It would basically index each row.

Maybe you are looking for CONSTRAINT

"The preferred way to add a unique constraint to a table is ALTER TABLE ... ADD CONSTRAINT" 
https://www.postgresql.org/docs/9.4/indexes-unique.html

internal/dbmodel/modelmigration.go
kian99: Correct. This row for model-migration will be deleted when the migration completes/aborts. If it succeeds we will take the user-mapping JSON and apply it to the user-mapping table.

internal/dbmodel/sql/postgres/024_add_model_migration_tables.up.sql
kian99: True, and also 
>[PostgreSQL automatically creates a unique index when a unique constraint or primary key is defined for a table.](https://www.postgresql.org/docs/current/indexes-unique.html#:~:text=PostgreSQL%20automatically%20creates%20a%20unique,mechanism%20that%20enforces%20the%20constraint.)

internal/dbmodel/sql/postgres/024_add_model_migration_tables.up.sql
kian99: When we query for a user mapping, we will always query by the model-uuid and the local-user (these are the 2 pieces of information in the macaroon, alongside the offer-uuid). That's why I created the multicolumn index.

internal/dbmodel/modelmigration.go
```diff
@@ -0,0 +1,25 @@
...
+// ModelMigration holds the information for a pending model migration.
+type ModelMigration struct {
+	// Note this doesn't use the standard gorm.Model to avoid soft-deletes.
+	ID        uint `gorm:"primarykey"`
+	CreatedAt time.Time
+
+	// ModelUUID is the UUID of the incoming model.
+	ModelUUID sql.NullString
+
+	// TargetControllerID is the target controller for the model undegroing migration.
```
ale8k: s/undegroing/undergoing

internal/dbmodel/modelmigration.go
```diff
@@ -0,0 +1,25 @@
...
+
+package dbmodel
+
+import (
+	"database/sql"
+	"time"
+)
+
+// ModelMigration holds the information for a pending model migration.
+type ModelMigration struct {
```
ale8k: This needs a different name, this isn't model migrations, this is "ModelMigrationTargetAndInitialUserMappingsToStoreOnCompletion" (Just describing it aha, but a better name is my point) and I also think you should godoc its explicit purpose (specifically UserMapping) and how it'll be used. We'll lose context in time with this, so I'd prefer us be kind to ourselves

internal/dbmodel/sql/postgres/024_add_model_migration_tables.up.sql
ale8k: The table would always be small right? So wouldn't the db sequentially scanning be quicker than an index? I think remove this index

ale8k: See: https://funivan.com/post/index-scan-vs-sequential-scan/ for more

internal/db/modelmigration.go
```diff
@@ -0,0 +1,79 @@
...
+		return errors.E(op, err)
+	}
+
+	durationObserver := servermon.DurationObserver(servermon.DBQueryDurationHistogram, string(op))
+	defer durationObserver()
+	defer servermon.ErrorCounter(servermon.DBQueryErrorCount, &err, string(op))
+
+	db := d.DB.WithContext(ctx)
+
+	if err := db.Create(modelMigration).Error; err != nil {
```
alesstimec, line 28: should we verify the UserMapping field to make sure it contains something we expect and not a random json blob?

internal/db/modelmigration.go
```diff
@@ -0,0 +1,79 @@
...
+		return errors.E(op, dbError(err))
+	}
+	return nil
+}
+
+// GetModelMigration returns model migration information based on the model UUID.
+func (d *Database) GetModelMigration(ctx context.Context, modelMigration *dbmodel.ModelMigration) (err error) {
```
ale8k: I know we usually pass the dbmodel obj, but would love to see us just return a dbmodel.ModelMigration and the func take a uuid. Really hate this style and want us to move away from it honestly.

internal/dbmodel/usermapping_test.go
```diff
@@ -0,0 +1,44 @@
...
+
+	qt "github.com/frankban/quicktest"
+
+	"github.com/canonical/jimm/v3/internal/dbmodel"
+)
+
+// TestUserMapping_UniqueConstraint tests that the unique constraint on
+// UserMapping is enforced correctly ensuring two rows with the same
+// ModelUUID and LocalUser cannot be created.
+func TestUserMapping_UniqueConstraint(t *testing.T) {
```
ale8k: Interesting we using _ here, looks like AI added this because we don't usually?

internal/db/modelmigration_test.go
```diff
@@ -0,0 +1,155 @@
...
+
+	var dbMigration dbmodel.ModelMigration
+	result := s.Database.DB.Where("model_uuid = ?", migration.ModelUUID).First(&dbMigration)
+	c.Assert(result.Error, qt.Equals, nil)
+	c.Assert(dbMigration.ModelUUID, qt.DeepEquals, migration.ModelUUID)
+	c.Assert(dbMigration.TargetControllerID, qt.Equals, controller.ID)
+	c.Assert(dbMigration.UserMapping, qt.DeepEquals, migration.UserMapping)
+}
+
+func (s *dbSuite) TestAddModelMigrationAlreadyExists(c *qt.C) {
```
alesstimec: this case could well be covered in the test above, which would avoid duplicating the same setup..

internal/dbmodel/sql/postgres/024_add_model_migration_tables.up.sql
```diff
@@ -0,0 +1,25 @@
+-- Add tables to handle model migrations
+-- The model_migration table contains information
+-- about ongoing migrations and their target controller.
+-- The UserMapping table is used to map local users
+-- from migrated models to their external counterparts.
```
ale8k, line 5: Where is the version bump? Or have you purposefully left it until a later PR?

internal/db/modelmigration_test.go
```diff
@@ -0,0 +1,155 @@
...
+	c.Assert(s.Database.DB.Create(&controller).Error, qt.IsNil)
+
+	migration := dbmodel.ModelMigration{
+		ModelUUID:          sql.NullString{String: "00000001-0000-0000-0000-000000000001", Valid: true},
+		TargetControllerID: controller.ID,
+		UserMapping:        dbmodel.JSON([]byte(`{"local":"external"}`)),
+	}
+	c.Assert(s.Database.DB.Create(&migration).Error, qt.IsNil)
+
+	err = s.Database.DeleteModelMigration(context.Background(), &migration)
```
alesstimec, line 113: perhaps also try to delete model migration that does not exist?

internal/dbmodel/modelmigration_test.go
```diff
@@ -0,0 +1,40 @@
...
+
+	m2 := dbmodel.ModelMigration{
+		ModelUUID: sql.NullString{
+			String: "00000001-0000-0000-0000-000000000001",
+			Valid:  true,
+		},
+		TargetControllerID: ctl.ID,
+		UserMapping:        dbmodel.JSON([]byte(`{"new-local": "new-external"}`)),
+	}
+	c.Assert(db.Create(&m2).Error, qt.ErrorMatches, ".*duplicate key value violates unique constraint \"model_migrations_model_uuid_key\".*")
```
ale8k: @alesstimec often prefers we store result of errors and then c.Assert(err...), shouldn't we keep to doing that?

internal/db/usermapping_test.go
```diff
@@ -0,0 +1,105 @@
...
+	c.Assert(err, qt.IsNil)
+	c.Assert(s.Database.DB.Create(&u).Error, qt.IsNil)
+
+	userMapping := &dbmodel.UserMapping{
+		ModelUUID:        modelUUID,
+		LocalUser:        localUser,
+		ExternalUserName: identityName,
+	}
+
+	err = s.Database.AddUserMapping(ctx, userMapping)
```
alesstimec, line 33: try adding it twice and assert that you get already exists error

internal/db/usermapping_test.go
```diff
@@ -0,0 +1,105 @@
...
+
+	userMapping := &dbmodel.UserMapping{
+		ModelUUID:        modelUUID,
+		LocalUser:        localUser,
+		ExternalUserName: identityName,
+	}
+	c.Assert(s.Database.AddUserMapping(ctx, userMapping), qt.IsNil)
+
+	// Delete
+	err = s.Database.DeleteUserMapping(ctx, userMapping)
```
alesstimec, line 103: try deleting user mapping that does not exist

internal/db/modelmigration.go
kian99: Yes but not at the DB layer

internal/db/modelmigration_test.go
kian99: Fair, I'll remove the duplication.

internal/db/usermapping.go
```diff
@@ -0,0 +1,77 @@
...
+	durationObserver := servermon.DurationObserver(servermon.DBQueryDurationHistogram, string(op))
+	defer durationObserver()
+	defer servermon.ErrorCounter(servermon.DBQueryErrorCount, &err, string(op))
+
+	db := d.DB.WithContext(ctx)
+	switch {
+	case userMapping.ModelUUID.Valid && userMapping.LocalUser != "":
+		db = db.Where("model_uuid = ? AND local_user = ?", userMapping.ModelUUID.String, userMapping.LocalUser)
+	default:
+		return errors.E(op, "missing model uuid and/or local user", errors.CodeBadRequest)
```
ale8k: Be nicer to say exactly what is missing with something like:
```go
switch {
case !userMapping.ModelUUID.Valid && userMapping.LocalUser == "":
	return errors.E(op, "missing both model UUID and local user", errors.CodeBadRequest)
case !userMapping.ModelUUID.Valid:
	return errors.E(op, "missing model UUID", errors.CodeBadRequest)
case userMapping.LocalUser == "":
	return errors.E(op, "missing local user", errors.CodeBadRequest)
default:
	if err := db.First(&userMapping).Error; err != nil {
		err = dbError(err)
		if errors.ErrorCode(err) == errors.CodeNotFound {
			return errors.E(op, err, "user mapping not found")
		}
		return errors.E(op, dbError(err))
	}
}
```

internal/db/modelmigration_test.go
```diff
@@ -0,0 +1,155 @@
...
+	"testing"
+
+	qt "github.com/frankban/quicktest"
+
+	"github.com/canonical/jimm/v3/internal/db"
+	"github.com/canonical/jimm/v3/internal/dbmodel"
+	"github.com/canonical/jimm/v3/internal/errors"
+)
+
+func TestAddModelMigrationUnconfiguredDatabase(t *testing.T) {
```
ale8k: We add these all over the place, I don't think we should be duplicating it constantly... Maybe just delete this?

internal/dbmodel/usermapping_test.go
kian99: Haha I was wondering if anyone would notice this. I actually added it because I saw AI adding somewhere else in Juju and I liked it. Splits the thing I'm testing with the specifics.

internal/db/modelmigration_test.go
kian99: Added it, but I see it doesn't return an error when deleting an item that doesn't exist. So, do you still want it?

internal/dbmodel/usermapping_test.go
ale8k: I really like it (Imo) but we discussed it once witn Juju and they hated it, and I think I tried to push for it for us and rightfully so, @alesstimec wasn't keen because it's just not how we do things and stood out like a sore thumb. having said that, I'm 1000% for continuing like this if the guys are on board?

internal/dbmodel/modelmigration.go
kian99: Here are some name options `ActiveModelMigrations` or `ModelMIgrationJobs` or `OngoingModelMigrations`. But I'm not sure about including details about the fields in the table in the name.

kian99: I've updated the godoc though.

internal/dbmodel/sql/postgres/024_add_model_migration_tables.up.sql
kian99: It doesn't matter much either way as you say the table will likely be small. But the expected access pattern is such that we are always querying by model-uuid and local user. In any case, I'll just make it an index on the model UUID.

internal/db/modelmigration.go
kian99: I'd prefer to just follow the pattern, so leaving this as is.

internal/db/modelmigration_test.go
kian99: Sure

internal/db/usermapping.go
kian99: Will do

internal/dbmodel/sql/postgres/024_add_model_migration_tables.up.sql
kian99: Which version bump? Our migrations run up to the latest, we don't need to specify the number in a file somewhere anymore.

internal/dbmodel/usermapping_test.go
kian99: I don't mind, but since it's different I've just opted to make it consistent now.

internal/dbmodel/modelmigration_test.go
kian99: Left this as is, Ales didn't comment on it.

SimoneDutto (approved):
Lgtm

internal/dbmodel/sql/postgres/024_add_model_migration_tables.up.sql
ale8k: What I'm saying is no index is probably best for this

internal/dbmodel/modelmigration.go
ale8k: Whatever you think if best, I just didn't think calling it ModelMigration was right. Happy with any of them honestly.

internal/dbmodel/usermapping_test.go
ale8k: Aww, we can discuss it another time though I am really in favour of it.

internal/dbmodel/sql/postgres/024_add_model_migration_tables.up.sql
kian99: Would definitely require some actual measurements to know for sure, but usually an index helps so I'll leave this as is.

ale8k (approved):
lgtm

PR #1610, opened 2025-06-10
feat(discharge): change discharge endpoint to try to resolve local users to external users

SimoneDutto:
The new behaviour in the CheckThirdPartyCaveat method is to call the
OfferAuthorizer to check permissions. The OfferAuthorizer will then be
in charge of trying to resolve the local user to an external user by
consulting the user mapping in incoming migrations.

### DIFF
--- a/internal/discharger/discharger.go
+++ b/internal/discharger/discharger.go
@@ -1,4 +1,4 @@
-// Copyright 2024 Canonical.
+// Copyright 2025 Canonical.
 
 package discharger
 
@@ -18,10 +18,8 @@ import (
 	"go.uber.org/zap"
 
 	"github.com/canonical/jimm/v3/internal/db"
-	"github.com/canonical/jimm/v3/internal/dbmodel"
 	"github.com/canonical/jimm/v3/internal/errors"
-	"github.com/canonical/jimm/v3/internal/openfga"
-	ofganames "github.com/canonical/jimm/v3/internal/openfga/names"
+	"github.com/canonical/jimm/v3/internal/jimm"
 )
 
 var defaultDischargeExpiry = 15 * time.Minute
@@ -33,17 +31,22 @@ type MacaroonDischargerConfig struct {
 	ControllerUUID         string
 }
 
-func NewMacaroonDischarger(cfg MacaroonDischargerConfig, db *db.Database, ofgaClient *openfga.OFGAClient) (*MacaroonDischarger, error) {
+// NewMacaroonDischarger creates a new MacaroonDischarger instance with the provided configuration, database, and offer authorizer.
+func NewMacaroonDischarger(cfg MacaroonDischargerConfig, db *db.Database, offerAuthorizer jimm.OfferAuthorizer) (*MacaroonDischarger, error) {
+	op := errors.Op("discharger.NewMacaroonDischarger")
 	var kp bakery.KeyPair
 	if cfg.PublicKey == "" || cfg.PrivateKey == "" {
 		return nil, errors.E("missing bakery private/public key")
 	} else {
 		if err := kp.Private.UnmarshalText([]byte(cfg.PrivateKey)); err != nil {
-			return nil, errors.E(err, "cannot unmarshal private key")
+			return nil, errors.E(op, err, "cannot unmarshal private key")
 		}
 		if err := kp.Public.UnmarshalText([]byte(cfg.PublicKey)); err != nil {
-			return nil, errors.E(err, "cannot unmarshal public key")
+			return nil, errors.E(op, err, "cannot unmarshal public key")
 		}
+	}
+	if offerAuthorizer == nil {
+		return nil, errors.E(op, "userMappingManager cannot be nil")
 	}
 
 	checker := checkers.New(jjmacaroon.MacaroonNamespace)
@@ -62,16 +65,16 @@ func NewMacaroonDischarger(cfg MacaroonDischargerConfig, db *db.Database, ofgaCl
 	)
 
 	return &MacaroonDischarger{
-		ofgaClient: ofgaClient,
-		bakery:     b,
-		kp:         kp,
+		offerAuthorizer: offerAuthorizer,
+		bakery:          b,
+		kp:              kp,
 	}, nil
 }
 
 type MacaroonDischarger struct {
-	ofgaClient *openfga.OFGAClient
-	bakery     *bakery.Bakery
-	kp         bakery.KeyPair
+	bakery          *bakery.Bakery
+	kp              bakery.KeyPair
+	offerAuthorizer jimm.OfferAuthorizer
 }
 
 // GetDischargerMux returns a mux that can handle macaroon bakery requests for the provided discharger.
@@ -98,7 +101,7 @@ func GetDischargerMux(macaroonDischarger *MacaroonDischarger, rootPath string) *
 // a declared caveat declaring offer uuid:
 //
 //	declared offer-uuid <offer uuid>
-func (md *MacaroonDischarger) CheckThirdPartyCaveat(ctx context.Context, req *http.Request, cavInfo *bakery.ThirdPartyCaveatInfo, _ *httpbakery.DischargeToken) ([]checkers.Caveat, error) {
+func (md *MacaroonDischarger) CheckThirdPartyCaveat(ctx context.Context, _ *http.Request, cavInfo *bakery.ThirdPartyCaveatInfo, _ *httpbakery.DischargeToken) ([]checkers.Caveat, error) {
 	caveatTokens := strings.Split(string(cavInfo.Condition), " ")
 	if len(caveatTokens) != 3 {
 		zapctx.Error(ctx, "caveat token length incorrect", zap.Int("length", len(caveatTokens)))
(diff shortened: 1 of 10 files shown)


internal/discharger/discharger.go
```diff
@@ -33,7 +32,15 @@ type MacaroonDischargerConfig struct {
 	ControllerUUID         string
 }
 
-func NewMacaroonDischarger(cfg MacaroonDischargerConfig, db *db.Database, ofgaClient *openfga.OFGAClient) (*MacaroonDischarger, error) {
+// UserMappingManager is an interface that defines methods for mapping local users to external users.
+type UserMappingManager interface {
+	// MapLocalUserToExternalUser maps a local user to an external user
+	// based on the user tag and the offer tag. It retrieves the model_uuid from the offer,
+	// and uses it to find the external user in the user mapping table.
+	MapLocalUserToExternalUser(ctx context.Context, userTag names.UserTag, offerTag names.ApplicationOfferTag) (*dbmodel.Identity, error)
```
kian99: The map is from model-uuid + local-user -> external user. Nothing to do with application offer. If you want the manager to handle finding the model from the offer, that could work but if possible I would avoid to keep things a bit clearer.

internal/discharger/discharger.go
```diff
@@ -112,25 +124,32 @@ func (md *MacaroonDischarger) CheckThirdPartyCaveat(ctx context.Context, req *ht
...
-
+	var identity *dbmodel.Identity
 	offerTag := names.NewApplicationOfferTag(offerUUID)
-
-	i, err := dbmodel.NewIdentity(userTag.Id())
-	if err != nil {
-		return nil, err
+	// If the user is a local user, we can try to map it to an external user.
+	if userTag.IsLocal() {
+		identity, err = md.userMappingManager.MapLocalUserToExternalUser(ctx, userTag, offerTag)
```
kian99: Before this we need to determine the model UUID from the offer UUID unless you expect the domain service will do that?

SimoneDutto: yeah, basically the user mapping manager will figure out the model UUID from the offer.

In my head the discharger shouldn't have any business logic, it is the duty of the usermappingmanager to resolve thing with the info the discharger has (usertag+offertag).
In this way the mapping is very trasparent for the discharger, plus we don't need to use the `db` in the discharger or pass another manager.

Let me know what you think though.

kian99: Yeah that's okay with me

internal/testutils/jimmtest/suite.go
```diff
@@ -246,7 +247,7 @@ func setupMacaroonDischarger(c *gc.C, uuid string, db *db.Database, ofgaClient *
 		PrivateKey:             "ly/dzsI9Nt/4JxUILQeAX79qZ4mygDiuYGqc2ZEiDEc=",
 		PublicKey:              "izcYsQy3TePp6bLjqOo3IRPFvkQd2IKtyODGqC6SdFk=",
 	}
-	macaroonDischarger, err := discharger.NewMacaroonDischarger(cfg, db, ofgaClient)
+	macaroonDischarger, err := discharger.NewMacaroonDischarger(cfg, db, ofgaClient, usermapper.UserMapppingManagerStub{})
```
kian99: I have a thought looking at this. Why not change the `userMapper` interface to something like `CrossModelAuthChecker` or `OfferAuthorizer` and just put all the logic that requires the DB and OpenFGA into a manager, then the macaroon discharger can handle the logic for parsing the macaroon caveats and then just pass the offer-uuid and user to JIMM. And then when you're reading the discharger logic, you never see anything about mapping local to external users because that's business logic.

SimoneDutto: i though about that, but i didn't want to refactor the discharger in this PR. But if you are happy with it, i'll do it

internal/jimm/jimm.go
```diff
@@ -202,6 +203,12 @@ type ConfigManager interface {
 	GetConfig() (config.ControllerConfig, error)
 }
 
+// OfferAuthorizer provides methods to check if a user is a consumer of an application offer.
+type OfferAuthorizer interface {
```
kian99, line 209: These interfaces only exist because we have multiple "api" entrypoints that use them, i.e. the Juju API and the Rebac Admin UI. Since the offer authorizer interface is already defined in the discharger, you could exclude this interface and just keep the OfferManager struct. This is how I imagine the rest of the manager's should go but it's just a bit of work that hasn't been done yet.
If you do choose to keep the struct around, on the `JIMM` struct, comment basically the above explaining why this is a struct and the rest are interfaces.
Also okay with leaving this as is.

internal/jimm/offer/authorizer.go
```diff
@@ -0,0 +1,83 @@
...
+	if err != nil {
+		return "", err
+	}
+	incomingMigration := dbmodel.IncomingModelMigration{
+		ModelUUID: sql.NullString{
+			String: offer.Model.UUID.String,
+			Valid:  true,
+		},
+	}
+	err = offerAuth.store.GetIncomingModelMigration(ctx, &incomingMigration)
```
kian99: Don't use this table, as it indicates a model migration has not yet completed. Use the user mapping table.

internal/jimm/offer/authorizer_test.go
```diff
@@ -0,0 +1,167 @@
...
+	"github.com/canonical/jimm/v3/internal/testutils/jimmtest"
+)
+
+type offerAuthorizerSuite struct {
+	offerAuthorizer *offer.OfferAuthorizer
+
+	offerUUID string
+}
+
+const offerAuthEnv = `clouds:
```
kian99, line 29: Can you simplify this env? I.e. it doesn't look like `fred@canonical.com` is used.

internal/testutils/jimmtest/mocks/jimm_offerauthorizer_mock.go
```diff
@@ -0,0 +1,23 @@
...
+package mocks
+
+import (
+	"context"
+
+	"github.com/juju/names/v5"
+)
+
+// OfferAuthorizer is a mock implementation of the OfferAuthorizer interface.
+type OfferAuthorizer struct {
```
kian99, line 12: Although we're not using it, we should probably also add a getter for the offer authorizer in `internal/testutils/jimmtest/jimm_mock.go`.

kian99 (approved):

internal/discharger/discharger.go
```diff
@@ -33,7 +31,8 @@ type MacaroonDischargerConfig struct {
 	ControllerUUID         string
 }
 
-func NewMacaroonDischarger(cfg MacaroonDischargerConfig, db *db.Database, ofgaClient *openfga.OFGAClient) (*MacaroonDischarger, error) {
+// NewMacaroonDischarger creates a new MacaroonDischarger instance with the provided configuration, database, and offer authorizer.
+func NewMacaroonDischarger(cfg MacaroonDischargerConfig, db *db.Database, offerAuthorizer jimm.OfferAuthorizer) (*MacaroonDischarger, error) {
```
ale8k, line 35: Whilst you're here, can remove db as a param and pass *dbrootkeystore.RootKeys.

ale8k: Also can make a params struct with a .validate(), then  remove line 47 to 49.

And if we're gonna use errors.E here, let's add an op code please.

internal/jimm/offer/authorizer.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/juju/names/v5"
+
+	"github.com/canonical/jimm/v3/internal/db"
+	"github.com/canonical/jimm/v3/internal/dbmodel"
+	"github.com/canonical/jimm/v3/internal/errors"
+	"github.com/canonical/jimm/v3/internal/openfga"
+)
+
+type offerAuthorizer struct {
+	store   *db.Database
```
ale8k, line 18: Doesn't need entire DB, interface would be nice here.

internal/jimm/offer/authorizer_test.go
```diff
@@ -0,0 +1,163 @@
...
+		c.Assert(allowed, qt.Equals, test.allowed)
+		if test.expectedError != "" {
+			c.Assert(err, qt.ErrorMatches, test.expectedError)
+		} else {
+			c.Assert(err, qt.IsNil)
+		}
+	}
+}
+
+func TestDischarger(t *testing.T) {
```
ale8k: This isn't the discharger ;). s/TestDischarger/TestOfferAuthorizerSuite

internal/jimm/jimm.go
```diff
@@ -569,3 +585,8 @@ func (j *JIMM) JujuManager() JujuManager {
 func (j *JIMM) ConfigManager() ConfigManager {
 	return j.configManager
 }
+
+// OfferAuthorizer returns a manager that enables checking if a user is a consumer of an application offer.
```
ale8k: Returns an authoriser*, let's keep language as the service describes. It's just called an OfferAuthoriser

ale8k (approved):

ale8k (approved):
Some very minor comments, stamping LGTM, but please do them. Nice work!

internal/jimm/offer/authorizer.go
SimoneDutto: yeah, true but basically all of our managers are importing db and openfga client.

I would mock the database layer once we have gomock or something handier.

internal/discharger/discharger.go
SimoneDutto: added op.
Didn't want to change from db to dbRootKeys to avoid scope-creep in this pr

PR #1611, opened 2025-06-10
chore: partially move to the Juju api client

kian99:
## Description
This PR is fairly involved, below is an explanation of the goal followed by a list of challenges and solution followed by other minor changes.

This PR partially modifies the `jujuclient` package and changes the cloud related methods to illustrate what a move to the juju api client might look like. Currently our `jujuclient` invokes facade calls directly, requiring logic for unpacking the response across different facade versions. Juju has logic for this already in various api client packages like `github.com/juju/juju/api/client/cloud`.

The greatest benefit of this change is a slimming down of the logic required in JIMM and the ability to remove tests from the `jujuclient` package since we are just a wrapper around Juju's API client.

Challenges and solutions:
1. The Juju client methods refer to types from `"github.com/juju/juju/cloud"` whereas our current implementation passed around rpc types from `"github.com/juju/juju/rpc/params"`. In order to facilitate the move, I opted to pass around the more "concrete" type internally and convert from the RPC type -> concrete type in the API layer, i.e. in our `jujuapi` package.
2. Related to 2.,  there was a challenge in how to convert from the rpc type that we receive from the client to the concrete type. The converter function exists in Juju in 2 places ([the client](https://github.com/juju/juju/blob/3.6/api/client/cloud/cloud.go#L406) and [apiserver](https://github.com/juju/juju/blob/3.6/apiserver/facades/client/cloud/cloud.go#L1033)) but neither are exported and are identical to each other so I've also copied it into JIMM.
3. The Juju client's methods don't accept a context but our lower-level facade caller does. I didn't want to remove this capability. Where initially I used Juju's [api.Open](https://github.com/juju/juju/blob/3.6/api/apiclient.go#L108) method to obtain an object that can make facade calls (one could also use Juju's SimpleConnector), the lack of support for context was problematic. ~~But we can fortunately use our existing `Connection` struct which satisfies the [APICallCloser](https://github.com/juju/juju/blob/3.6/api/base/clientfacade.go#L8) interface and already supports the use of a context.~~ After a discussion, it doesn't make sense to use a context for an individual message in a websocket connection because a cancellation on the client side doesn't signal anything to the server. Instead the context is tied to the life of the websocket connection. This does still mean that we can't use Juju's `SimpleConnector` or `api.Open` function without at least some further consideration/tweaks because those methods don't accept a context and use `context.Background` internally.

Some other changes that I've made:
- I've removed the "redial" logic in the jujuclient as I'm confident it's not necessary. When making API calls to Juju we should know exactly what permissions are required beforehand. The redialling logic is only necessary for the model proxy where we don't know what command the client is trying to execute and the initial token added to their request may be missing permissions.
- Removed the `cachedialer` after a discussion with the team considering that each connection may require unique permissions. A cache dialer is useful but will require a new implementation for full correctness.
- Removed API client methods for `GrantCloudAccess` and `RevokeCloudAccess` since they are no longer necessary.

## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests


internal/jujuapi/types.go
```diff
@@ -0,0 +1,46 @@
+// Copyright 2025 Canonical.
+
+package jujuapi
+
+import (
+	"github.com/juju/juju/cloud"
+	jujuparams "github.com/juju/juju/rpc/params"
+)
+
+func cloudFromParams(cloudName string, p jujuparams.Cloud) cloud.Cloud {
```
SimoneDutto, line 10: Did you manage to find a way to test when this function changes in Juju?

SimoneDutto, line 10: I know it's private but i think a godoc explaining where is coming from and why we should be careful to update it would help.

SimoneDutto (approved):
I think cancelling the context on the connection instead of the single facade can be tricky, but if you guys are ok with it, so am i.

internal/jujuapi/types.go
kian99: Without exporting the method in Juju, I looked at how we could test it in JIMM.
```go
func (s *typesSuite) TestCloudFromParams(c *gc.C) {
	// Check the number of fields on the RPC params struct to ensure we update our logic if it changes.
	cloudParams := jujuparams.Cloud{}
	t := reflect.TypeOf(cloudParams)
	c.Assert(t.NumField(), gc.Equals, 12, gc.Commentf("The number of fields on %s/%s has changed, "+
		"be sure to update the CloudFromParams method then update this test.", t.PkgPath(), t.Name()))

	// Check the number of fields on the concrete Cloud struct to ensure we update our logic if it changes.
	cloud := jujucloud.Cloud{}
	t = reflect.TypeOf(cloud)
	c.Assert(t.NumField(), gc.Equals, 14, gc.Commentf("The number of fields on %s/%s has changed, "+
		"be sure to update the CloudFromParams method then update this test.", t.PkgPath(), t.Name()))
}
```
I came up with this to ensure we are alerted to changes in either struct. This is a basic measure of ensuring the fields are mapped correctly but doesn't account for changes in the logic of the mapping function. And even if I test our mapper, we have no way of knowing if Juju's mapping logic has changed.
Another idea is that we would pick up issues during e2e tests if our mapping logic doesn't work as expected.

Thinking about it from another perspective. Clients that have already been released will use the mapping logic and fields that existed at the time of the release. If there is a change in Juju (say the addition of a new field), they would still need to continue to support old clients so our "outdated" mapping logic should not "break" but we might not support new features from new clients. Ultimately we should see about getting this field exported, I have yet to ask about that.

ale8k (approved):
lgtm. Might be good to tag this commit when it merges as "cache-dialer-removed" or something? So we can refer back to it when we need to. I guess we can just go back to the year too? I mention this in case we realise later that we do require cache dialing.

PR #1612, opened 2025-06-11
Juju 7944/prepare model migration

ale8k:
## Description
feat: prepareModelMigration method
    
This method prepares JIMM for an incoming model migration. It stores the model uuid, target
controller name and usermap, to be referred to later when the migrate methods require these details.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

### DIFF
--- a/internal/db/modelmigration.go
+++ b/internal/db/modelmigration.go
@@ -14,7 +14,7 @@ import (
 //   - returns an error with code errors.CodeAlreadyExists if
 //     a migration row with the same model UUID already exists.
 func (d *Database) AddIncomingModelMigration(ctx context.Context, modelMigration *dbmodel.IncomingModelMigration) (err error) {
-	const op = errors.Op("db.AddModelMigration")
+	const op = errors.Op("db.AddIncomingModelMigration")
 	if err := d.ready(); err != nil {
 		return errors.E(op, err)
 	}
@@ -33,7 +33,7 @@ func (d *Database) AddIncomingModelMigration(ctx context.Context, modelMigration
 
 // GetIncomingModelMigration returns model migration information based on the model UUID.
 func (d *Database) GetIncomingModelMigration(ctx context.Context, modelMigration *dbmodel.IncomingModelMigration) (err error) {
-	const op = errors.Op("db.GetModelMigration")
+	const op = errors.Op("db.GetIncomingModelMigration")
 	if err := d.ready(); err != nil {
 		return errors.E(op, err)
 	}
@@ -62,7 +62,7 @@ func (d *Database) GetIncomingModelMigration(ctx context.Context, modelMigration
 
 // DeleteIncomingModelMigration removes a model migration entry from the database.
 func (d *Database) DeleteIncomingModelMigration(ctx context.Context, modelMigration *dbmodel.IncomingModelMigration) (err error) {
-	const op = errors.Op("db.DeleteModelMigration")
+	const op = errors.Op("db.DeleteIncomingModelMigration")
 	if err := d.ready(); err != nil {
 		return errors.E(op, err)
 	}
--- a/internal/db/modelmigration.go
+++ b/internal/db/modelmigration.go
@@ -50,7 +50,7 @@ func (d *Database) GetModelMigration(ctx context.Context, modelMigration *dbmode
 		return errors.E(op, "missing uuid", errors.CodeBadRequest)
 	}
 
-	if err := db.First(&modelMigration).Error; err != nil {
+	if err := db.Preload("TargetController").First(&modelMigration).Error; err != nil {
 		err = dbError(err)
 		if errors.ErrorCode(err) == errors.CodeNotFound {
 			return errors.E(op, err, "model migration not found")
--- a/internal/db/modelmigration.go
+++ b/internal/db/modelmigration.go
@@ -10,10 +10,10 @@ import (
 	"github.com/canonical/jimm/v3/internal/servermon"
 )
 
-// AddModelMigration stores information about an incoming model migration.
+// AddIncomingModelMigration stores information about an incoming model migration.
 //   - returns an error with code errors.CodeAlreadyExists if
 //     a migration row with the same model UUID already exists.
-func (d *Database) AddModelMigration(ctx context.Context, modelMigration *dbmodel.IncomingModelMigration) (err error) {
+func (d *Database) AddIncomingModelMigration(ctx context.Context, modelMigration *dbmodel.IncomingModelMigration) (err error) {
 	const op = errors.Op("db.AddModelMigration")
 	if err := d.ready(); err != nil {
 		return errors.E(op, err)
@@ -31,8 +31,8 @@ func (d *Database) AddModelMigration(ctx context.Context, modelMigration *dbmode
 	return nil
 }
 
-// GetModelMigration returns model migration information based on the model UUID.
-func (d *Database) GetModelMigration(ctx context.Context, modelMigration *dbmodel.IncomingModelMigration) (err error) {
+// GetIncomingModelMigration returns model migration information based on the model UUID.
+func (d *Database) GetIncomingModelMigration(ctx context.Context, modelMigration *dbmodel.IncomingModelMigration) (err error) {
 	const op = errors.Op("db.GetModelMigration")
 	if err := d.ready(); err != nil {
 		return errors.E(op, err)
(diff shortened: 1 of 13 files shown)


internal/db/modelmigration.go
```diff
@@ -60,8 +60,8 @@ func (d *Database) GetModelMigration(ctx context.Context, modelMigration *dbmode
 	return nil
 }
 
-// DeleteModelMigration removes a model migration entry from the database.
-func (d *Database) DeleteModelMigration(ctx context.Context, modelMigration *dbmodel.IncomingModelMigration) (err error) {
+// DeleteIncomingModelMigration removes a model migration entry from the database.
+func (d *Database) DeleteIncomingModelMigration(ctx context.Context, modelMigration *dbmodel.IncomingModelMigration) (err error) {
 	const op = errors.Op("db.DeleteModelMigration")
```
kian99: You'll need to delete this op strings as well

internal/jujuapi/jimm.go
```diff
@@ -517,3 +521,42 @@ func (r *controllerRoot) Version(ctx context.Context) (apiparams.VersionResponse
...
+		return errors.E(op, "invalid model tag", err)
+	}
+
+	if !names.IsValidControllerName(args.TargetControllerName) {
+		return errors.E(op, "invalid controller name")
+	}
+	userMap := map[string]string{}
+	if err := json.Unmarshal([]byte(args.UserMapping), &userMap); err != nil {
+		return errors.E(op, "invalid user map, unable to unmarshal")
+	}
```
kian99: Would be good not to unmarshal fields from a struct that was already unmarshalled. Indicates our type could be improved.

internal/db/modelmigration.go
ale8k: F good spot

internal/jujuapi/jimm.go
ale8k: Good point. I think we might actually be able to just use a map[string]string, I'll test now

ale8k: Yeah it worked fine, will update

internal/jujuapi/jimm_test.go
kian99, line 1: Can we turn these into table tests? Each test in `jujuapi` is an integration test that uses the connsuite and spins up a separate mongo db and with lots of small tests it adds up and makes things slower. An alternative for tests that validate the unhappy path would be to take them out the suite and mock out the application layer since we can do that now.

internal/jimm/juju/jimm_test.go
```diff
@@ -711,3 +710,107 @@ func TestInitiateInternalMigration(t *testing.T) {
...
+
+	env := jimmtest.ParseEnvironment(c, PrepareModelMigrationTestEnv)
+	env.PopulateDB(c, j.Database)
+	dbUser := env.User("alice@canonical.com").DBObject(c, j.Database)
+	user := openfga.NewUser(&dbUser, nil)
+	user.JimmAdmin = true
+
+	modelUUID := env.Models[0].DBObject(c, j.Database).UUID.String
+	userMapping := map[string]string{"alice": "alice@canonical.com"}
+	targetController := env.Controllers[1].DBObject(c, j.Database)
```
kian99: use `Controllers[0[` if you simply the env above.

internal/jimm/juju/jimm_test.go
```diff
@@ -711,3 +710,107 @@ func TestInitiateInternalMigration(t *testing.T) {
 		})
 	}
 }
+
+const PrepareModelMigrationTestEnv = `clouds:
```
kian99: Doesn't need to be exported.
You can simplify this yaml by removing a bunch of bits including the second controller, the model and the cloud-credentials.

internal/jujuapi/jimm_test.go
ale8k: I'm really not a fan of table tests and I remember us having this conversation before but not our outcomes.

kian99: That's fine, you can just move these into a separate suite like I've done in [this PR](https://github.com/canonical/jimm/pull/1616/commits/21b4368826004cf1965e5b86a74c0dad5bb76161#diff-2ff06f5078ea01152ea1838df4dbcec744a7f55d667bb3972742dfd29ca25b4fR22), that doesn't rely on the connSuite. Instead of a table test you can just run the different scenarios sequentially like I've done in the linked test.

ale8k: Do we want the mix integration and mock tests in the same package? I understand your concern about a new mongodb, but we'll have to be aware that when we approach tests we should check the type within the suite now...

kian99: I think it's fine, I'd hate to continue to write integration tests just to test some small piece of validation. We can put the mock tests into a separate file with `_mock_test.go` possibly (since most of the tests are integration). But open to anything sensible on that front.

ale8k: I think before changing something we've previously had strongly consistent it's better to discuss as a team and not on a PR comment thread. Let's chat tomorrow.

pkg/api/params/params.go
```diff
@@ -545,3 +545,11 @@ type VersionResponse struct {
 	Version string `json:"version" yaml:"version"`
 	Commit  string `json:"commit" yaml:"commit"`
 }
+
+// PrepareModelMigrationRequest holds the details to prepare JIMM
+// for a model migration.
+type PrepareModelMigrationRequest struct {
+	ModelTag             string            `json:"model-tag" yaml:"model-tag"`
+	TargetControllerName string            `json:"target-controller-name" yaml:"target-controller-name"`
```
SimoneDutto, line 553: What about `TargetControllerTag`, just to be consistent

internal/jimm/juju/jimm.go
```diff
@@ -267,3 +269,38 @@ func (j *JujuManager) InitiateInternalMigration(ctx context.Context, user *openf
...
+func (j *JujuManager) PrepareModelMigration(
+	ctx context.Context,
+	user *openfga.User,
+	modelUUID string,
+	targetControllerName string,
+	userMapping map[string]string,
+) error {
+	const op = errors.Op("jujumanager.PrepareModelMigration")
+
+	err := j.Database.Transaction(func(d *db.Database) error {
```
SimoneDutto, line 284: Do we need a transaction for this? I think that if the controller doesn't exist the `AddIncomingModelMigration` will just return an error.

SimoneDutto (approved):
LGTM, 2 comments

pkg/api/params/params.go
ale8k: Controller tags are UUID and we want name, if we could know the controller UUID to send then definitely?

internal/jimm/juju/jimm.go
ale8k: What if the controller is removed between the get and add?

internal/jujuapi/jimm_test.go
ale8k: Updated, ptal.

internal/jujuapi/jimm_test.go
```diff
@@ -907,3 +907,19 @@ func (s *jimmSuite) TestVersion(c *gc.C) {
...
+
+	ctlName := "prepare-model-migration-controller"
+	s.AddController(c, ctlName, s.APIInfo(c))
+
+	err := client.PrepareModelMigration(&apiparams.PrepareModelMigrationRequest{
+		ModelTag:             names.NewModelTag("5650ac3f-8332-437f-874f-089e0e447e7f").String(),
+		TargetControllerName: ctlName,
+		UserMapping:          map[string]string{"alice": "alice@canonical.com"}, // `{"alice": "alice@canonical.com"}`,
+	})
+	c.Assert(err, gc.IsNil)
```
kian99, line 924: We can check that an entry was created in the DB.

internal/jujuapi/jimm_unit_test.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+func (s *jimmSuite) TestPrepareModelMigration_InvalidUserMapping(c *gc.C) {
+	ctx := context.Background()
+
+	jujuManager := mocks.JujuManager{
+		PrepareModelMigration_: func(ctx context.Context, user *openfga.User, modelUUID, targetControllerName string, userMapping map[string]string) error {
+			return nil
+		},
+	}
```
kian99: You could delete this and the others since it should never reach the the point where it calls the PrepareModelMigration method.

kian99 (approved):
Otherwise lgtm

internal/jujuapi/jimm_test.go
ale8k: I thought about this but I'd just be testing the db logic again. It's tested in application layer, so will leave this.

PR #1613, opened 2025-06-11
chore: use better type for storing usermapping

kian99:
## Description
Use a stronger type than plain JSON for storing the user mapping in the `IncomingModelMigration` table. Instead of `JSON` I've switched it to our `StringMap` type which is a map of `string` to `string`.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests


ale8k (approved):
lgtm

alesstimec (approved):

PR #1615, opened 2025-06-11
fix(soft-delete): remove soft-delete from cloud_credentials

SimoneDutto:
Before: we were soft-deleting credentials. This is an issue because when you delete and re-add a credential it isn't re-added but can't be used because it is soft-deleted.
So this PR remove the soft-delete.

By removing the soft-delete we can't `RevokeCredentials` with `force` anymore, because models hold a reference to credentials and if we delete a cloud credential with a model still referencing it, postgresql returns an error and it isn't good to break our state to allow `force`.

### DIFF
--- a/internal/jimm/juju/cloudcredential.go
+++ b/internal/jimm/juju/cloudcredential.go
@@ -47,7 +47,7 @@ func (j *JujuManager) GetCloudCredential(ctx context.Context, user *openfga.User
 
 // RevokeCloudCredential checks that the credential with the given path
 // can be revoked  and revokes the credential.
-func (j *JujuManager) RevokeCloudCredential(ctx context.Context, user *dbmodel.Identity, tag names.CloudCredentialTag, force bool) error {
+func (j *JujuManager) RevokeCloudCredential(ctx context.Context, user *dbmodel.Identity, tag names.CloudCredentialTag) error {
 	const op = errors.Op("jimm.RevokeCloudCredential")
 
 	if user.Name != tag.Owner().Id() {
@@ -75,8 +75,10 @@ func (j *JujuManager) RevokeCloudCredential(ctx context.Context, user *dbmodel.I
 	if err != nil {
 		return errors.E(op, err)
 	}
-	// if the cloud credential is still used by any model we return an error
-	if len(models) > 0 && !force {
+	// Before we accepted the force flag to remove the credential regardless of the references count.
+	// Now we want to ensure that the credential is not used by any models before removing it to maintain
+	// referential integrity.
+	if len(models) > 0 {
 		return errors.E(op, errors.CodeBadRequest, fmt.Sprintf("cloud credential still used by %d model(s)", len(models)))
 	}
 
--- a/internal/jujuapi/cloud.go
+++ b/internal/jujuapi/cloud.go
@@ -59,19 +59,6 @@ func init() {
 	}
 }
 
-// RevokeCredentials implements the RevokeCredentials method used in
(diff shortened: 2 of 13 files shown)


internal/jujuapi/cloud.go
```diff
@@ -59,19 +59,6 @@ func init() {
 	}
 }
 
-// RevokeCredentials implements the RevokeCredentials method used in
-// version 1 & 2 of the Cloud facade.
-func (r *controllerRoot) RevokeCredentials(ctx context.Context, args jujuparams.Entities) (jujuparams.ErrorResults, error) {
```
SimoneDutto, line 64: This wasn't used, so i've deleted it.

internal/jimm/juju/cloudcredential.go
```diff
@@ -75,8 +75,11 @@ func (j *JujuManager) RevokeCloudCredential(ctx context.Context, user *dbmodel.I
 	if err != nil {
 		return errors.E(op, err)
 	}
-	// if the cloud credential is still used by any model we return an error
-	if len(models) > 0 && !force {
+	// If the cloud credential is still used by any model we return an error.
+	// Before we were accepting the force flag to remove the credentials regardeless of the references count.
+	// Now we want to ensure that the credential is not used by any model before removing it to avoid messing
+	// up with our state.
```
kian99:
```suggestion
	// Before we accepted the force flag to remove the credential regardless of the references count.
	// Now we want to ensure that the credential is not used by any models before removing it to maintain
	// referential integrity.
```

internal/jujuapi/cloud.go
kian99: What do you mean it wasn't used? It's an exported facade method?

SimoneDutto: I can be wrong but usually to register a facade method we do
```go
updateCloudMethod := rpc.Method(r.UpdateCloud)
r.AddMethod("Cloud", 7, "UpdateCloud", updateCloudMethod)
```

This wasn't registered to any facade, it was just an exported method.

kian99: Ah fair enough! Sounds good then

kian99 (approved):

internal/jujuapi/cloud.go
alesstimec: shouldn't we register the method instead??

internal/jujuapi/export_test.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2024 Canonical.
+// Copyright 2025 Canonical.
```
alesstimec, line 1: why? nothing changed in this file..

SimoneDutto: the golang-cilint was automatically doing it everytime i saved in my vscode, so i've left it

internal/jujuapi/cloud.go
SimoneDutto: https://github.com/search?q=repo%3Ajuju%2Fjuju%20RevokeCredentials&type=code

in juju they don't have it, they just have RevokeCredentialsCheckModels, which we correctly register.
So i think it was a left-over

kian99: It looks like Juju doesn't implement it (at least in 3.6), like the docstring says, it was used in "version 1 & 2 of the Cloud facade." and if 3.6 doesn't support that then neither should we (since we only support 3.6 controllers)

alesstimec (approved):

PR #1616, opened 2025-06-11
chore: implement Prechecks on the MigrationTarget facade

kian99:
## Description
This PR starts to implement the `MigrationTarget` facade on JIMM. It introduces the `Prechecks` method used to start a model migration. There is a TODO around modifying the model description that we receive, that will be handled in a separate PR.

Fixes [JUJU-8031](https://warthogs.atlassian.net/browse/JUJU-8031)

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

[JUJU-8031]: https://warthogs.atlassian.net/browse/JUJU-8031?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/jimm/jimm.go
+++ b/internal/jimm/jimm.go
@@ -13,6 +13,7 @@ import (
 	"github.com/juju/juju/api/base"
 	jujucloud "github.com/juju/juju/cloud"
 	jujucontroller "github.com/juju/juju/controller"
+	"github.com/juju/juju/core/migration"
 	jujuparams "github.com/juju/juju/rpc/params"
 	"github.com/juju/names/v5"
 	"github.com/juju/version"
@@ -236,6 +237,9 @@ type JujuManager interface {
 	UnsetModelDefaults(ctx context.Context, user *dbmodel.Identity, cloudTag names.CloudTag, region string, keys []string) error
 	UpdateMigratedModel(ctx context.Context, user *openfga.User, modelTag names.ModelTag, targetControllerName string) error
 	ValidateModelUpgrade(ctx context.Context, u *openfga.User, mt names.ModelTag, force bool) error
+
+	// Migration related methods
+	Prechecks(ctx context.Context, user *openfga.User, model migration.ModelInfo) error
 
 	// Other methods
 
--- a/internal/jimm/jimm.go
+++ b/internal/jimm/jimm.go
@@ -239,6 +239,9 @@ type JujuManager interface {
 	ValidateModelUpgrade(ctx context.Context, u *openfga.User, mt names.ModelTag, force bool) error
 
 	// Migration related methods
+	// The migration methods below are sorted roughly
+	// in the order they are expected to be called.
+	PrepareModelMigration(ctx context.Context, user *openfga.User, modelUUID string, targetControllerName string, userMapping map[string]string) error
 	Prechecks(ctx context.Context, user *openfga.User, model migration.ModelInfo) error
 
 	// Other methods
@@ -268,7 +271,6 @@ type JujuManager interface {
 	RevokeOfferAccessOnController(ctx context.Context, user *openfga.User, ut names.UserTag, offerURL string, access jujuparams.OfferAccessPermission) error
 	UpdateCloud(ctx context.Context, u *openfga.User, ct names.CloudTag, cloud jujucloud.Cloud) error
 	UpdateCloudCredential(ctx context.Context, u *openfga.User, args juju.UpdateCloudCredentialArgs) ([]jujuparams.UpdateCredentialModelResult, error)
-	PrepareModelMigration(ctx context.Context, user *openfga.User, modelUUID string, targetControllerName string, userMapping map[string]string) error
 
 	// These are methods on the Juju manager that don't need to be mocked and can be removed from this interface later.
 	UpdateMetrics(ctx context.Context)
--- a/internal/jimm/juju/migrationtarget.go
+++ b/internal/jimm/juju/migrationtarget.go
@@ -16,6 +16,9 @@ import (
 )
 
 // Prechecks checks that the model can be migrated to the target controller.
+// It does this by calling the method of the same name on the target Juju controller.
+// As part of all model migrations passing through JIMM, it modifies the model description
+// to replace any local user references with their external mapping.
 func (j *JujuManager) Prechecks(ctx context.Context, user *openfga.User, model migration.ModelInfo) error {
 	const op = errors.Op("jimm.Prechecks")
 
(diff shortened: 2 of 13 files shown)


internal/jimm/juju/migrationtarget.go
```diff
@@ -0,0 +1,78 @@
...
+	if err != nil {
+		return errors.E(op, fmt.Errorf("failed to run pre-checks for migration: %w", err))
+	}
+	return nil
+}
+
+// modifyMigrationInfo modifies the description of the model migration
+// to replace any local user references with their external mapping.
+func (j *JujuManager) modifyMigrationInfo(model *migration.ModelInfo, userMapping dbmodel.StringMap) error {
+	if model.Owner.Domain() != "" && model.Owner.Domain() != names.LocalUserDomain {
```
SimoneDutto: if `Owner` is a UserTag you should have a method called `IsLocal` which is basically doing the checks you are doing here.

internal/jimm/juju/migrationtarget_test.go
```diff
@@ -0,0 +1,180 @@
...
+	modelDescription := description.NewModel(descriptionArgs)
+	userArgs := description.UserArgs{
+		Name:        names.NewUserTag("bob"),
+		DisplayName: "bob",
+		Access:      "admin",
+	}
+	modelDescription.AddUser(userArgs)
+	modelInfo := migration.ModelInfo{
+		UUID:                   "00000001-0000-0000-0000-000000000001",
+		Owner:                  names.NewUserTag("bob"),
```
SimoneDutto: here you are not creating the `Owner` with the `owner` passed as an argument.

internal/jimm/juju/migrationtarget.go
```diff
@@ -0,0 +1,78 @@
...
+
+	"github.com/juju/juju/core/migration"
+	"github.com/juju/names/v5"
+
+	"github.com/canonical/jimm/v3/internal/dbmodel"
+	"github.com/canonical/jimm/v3/internal/errors"
+	"github.com/canonical/jimm/v3/internal/openfga"
+)
+
+// Prechecks checks that the model can be migrated to the target controller.
```
SimoneDutto, line 18: I think the name should express clearer that we are also mapping the owner here.
`Prechecks` to me sounds like something that should just return a boolean.

Also add to the godoc that we are doing the mapping

internal/jujuapi/migrationtarget.go
```diff
@@ -0,0 +1,60 @@
...
+
+		return []int{4}
+	}
+}
+
+// Prechecks implements the Prechecks method of the MigrationTarget facade.
+func (r *controllerRoot) Prechecks(ctx context.Context, args jujuparams.MigrationModelInfo) error {
+	const op = errors.Op("jujuapi.Prechecks")
+
+	if !r.user.JimmAdmin {
```
SimoneDutto, line 44: proposal: since only jimm admins can call all of these facade methods, can we gate the access at `MigrationTarget` level? I don't know if we have a way of doing it right now, but it would be cool.

SimoneDutto (commented):
looks good, just a few comments

internal/jujuapi/migrationtarget.go
kian99: We would have to build something for that, as we can't currently.

internal/jimm/juju/migrationtarget.go
kian99: I've updated the godoc but I chose the name to match Juju since we are going to be mapping the owner every time we receive a model description during the migration.

SimoneDutto (approved):

internal/jimm/jimm.go
```diff
@@ -236,6 +237,9 @@ type JujuManager interface {
 	UpdateMigratedModel(ctx context.Context, user *openfga.User, modelTag names.ModelTag, targetControllerName string) error
 	ValidateModelUpgrade(ctx context.Context, u *openfga.User, mt names.ModelTag, force bool) error
 
+	// Migration related methods
+	Prechecks(ctx context.Context, user *openfga.User, model migration.ModelInfo) error
```
ale8k, line 245: Could you move preparemodelmigration above this? And can we keep the order in which they run here with a little godoc explaining it? I think it'd help us for future aha.

internal/jimm/juju/migrationtarget.go
```diff
@@ -0,0 +1,81 @@
...
+func (j *JujuManager) Prechecks(ctx context.Context, user *openfga.User, model migration.ModelInfo) error {
+	const op = errors.Op("jimm.Prechecks")
+
+	modelMigration := dbmodel.IncomingModelMigration{
+		ModelUUID: sql.NullString{
+			String: model.UUID,
+			Valid:  true,
+		},
+	}
+	err := j.Database.GetIncomingModelMigration(ctx, &modelMigration)
```
ale8k: Let's not call it modelMigration, I think this reads a bit clearer:
```go
func (j *jujuManager) getUserMapping(modelUUID string) (StringMap, error) {
	incomingModelMigration := dbmodel.IncomingModelMigration{
		ModelUUID: sql.NullString{
			String: model.UUID,
			Valid:  true,
		},
	}
	if err := j.Database.GetIncomingModelMigration(ctx, &incomingModelMigration); err != nil {
		return errors.E(op, fmt.Errorf("failed to get model migration %q: %w", model.UUID, err))
       }
       
       return ncomingModelMigration.UserMapping
}

````

Because logic wise we don't care about anything other than the mapping, and it'll read nicely this way.
(Inline function probably better?)

File: internal/jimm/juju/migrationtarget.go [hunk 7]
```diff
@@ -0,0 +1,81 @@
...
+// to replace any local user references with their external mapping.
+func (j *JujuManager) modifyMigrationInfo(model *migration.ModelInfo, userMapping dbmodel.StringMap) error {
+	if !model.Owner.IsLocal() {
+		// If the owner is not a local user, we do not modify it.
+		// This is useful when migrating a model from one JIMM
+		// controller to another, where the owner is already an external user.
+		return nil
+	}
+	newOwnerFound := false
+	for originalUser, newUser := range userMapping {
```
ale8k: localUser, externalUser is clearer

ale8k: Do we need to loop? Can't we just do userMapping[localUser] ?

internal/jujuapi/migrationtarget.go
```diff
@@ -0,0 +1,60 @@
...
+		return errors.E(op, err)
+	}
+	model := migration.ModelInfo{
+		UUID:                   args.UUID,
+		Name:                   args.Name,
+		Owner:                  ownerTag,
+		AgentVersion:           args.AgentVersion,
+		ControllerAgentVersion: args.ControllerAgentVersion,
+	}
+	return r.jimm.JujuManager().Prechecks(ctx, r.user, model)
```
ale8k: This error would miss the op from where it is called. Error needs checking and wrapping in errors.E if present?

internal/jujuapi/migrationtarget_test.go
```diff
@@ -0,0 +1,46 @@
...
+		Name:                   "test-model",
+		ControllerAgentVersion: version.MustParse("3.5.0"),
+		AgentVersion:           version.MustParse("3.5.0"),
+		ModelDescription:       modelDescription,
+	}
+	client := migrationtarget.NewClient(conn)
+	err := client.Prechecks(model)
+	c.Assert(err, gc.ErrorMatches, `.*model migration not found`)
+
+	// TODO: Run a prepare model migration call and then call preChecks again.
```
ale8k: This can be done now?

ale8k (approved):
LGTM now, thanks!

PR #1619, opened 2025-06-13
chore: implement Abort method for migrations

kian99:
## Description
This PR implements the `Abort` migration on the `MigrationTarget` facade for migrating models through JIMM. The method is essentially just a pass-through to the Juju controller. 

One bit that is noteworthy is that after the API call to Juju completes successfully, we delete the `IncomingModelMigration` entry from our database. If the deletion operation fails we do not return an error because the abort was already successful on the target controller. This does imply we could have a dangling resource in JIMM. That will need to be handled in a separate PR (I've added a note in [JUJU-8030](https://warthogs.atlassian.net/browse/JUJU-8030) about this).

Fixes [JUJU-8084](https://warthogs.atlassian.net/browse/JUJU-8084)
## Engineering checklist
- [ ] Documentation updated
- [x] Covered by unit tests

[JUJU-8084]: https://warthogs.atlassian.net/browse/JUJU-8084?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[JUJU-8030]: https://warthogs.atlassian.net/browse/JUJU-8030?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/jimm/juju/migrationtarget.go
+++ b/internal/jimm/juju/migrationtarget.go
@@ -10,11 +10,51 @@ import (
 	"github.com/juju/juju/core/migration"
 	"github.com/juju/names/v5"
 	"github.com/juju/version/v2"
+	"github.com/juju/zaputil/zapctx"
+	"go.uber.org/zap"
 
 	"github.com/canonical/jimm/v3/internal/dbmodel"
 	"github.com/canonical/jimm/v3/internal/errors"
 	"github.com/canonical/jimm/v3/internal/openfga"
 )
+
+// AbortMigration aborts a model migration with the given model UUID.
+// It does this by calling the Abort methodon the target Juju controller.
+// It also deletes the migration record from the database, but does not return an error
+// if the deletion fails, as the migration has already been aborted on the target controller.
+func (j *JujuManager) AbortMigration(ctx context.Context, user *openfga.User, modelUUID string) error {
+	const op = errors.Op("jimm.Abort")
+
+	incomingModel := dbmodel.IncomingModelMigration{
+		ModelUUID: sql.NullString{
+			String: modelUUID,
+			Valid:  true,
+		},
+	}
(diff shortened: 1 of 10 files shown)


SimoneDutto (approved):
LGTM

internal/jimm/juju/migrationtarget.go
```diff
@@ -9,12 +9,53 @@ import (
...
+	defer api.Close()
+
+	err = api.Abort(modelUUID)
+	if err != nil {
+		return errors.E(op, fmt.Errorf("failed to abort migration: %w", err))
+	}
+
+	err = j.Database.DeleteIncomingModelMigration(ctx, &incomingModel)
+	if err != nil {
+		// Don't return an error if we fail to delete the migration record,
```
alesstimec, line 52: is this error something we'd want to monitor?

alesstimec (approved):

internal/jimm/juju/migrationtarget.go
kian99: That's a good question. I can't say definitively whether to add a metric. I could add a metric called `failed_cleanup_incoming_migration`? Or more specifically `failed_cleanup_incoming_migration_after_abort`. It just seems like it doesn't add much. 
Alternatively, I imagine we'll have a background job that cleans up these rows after a period of time, the issue there is that "period of time" might be too long when you want to restart your migration.
Possibly just something to address in [JUJU-8030](https://warthogs.atlassian.net/browse/JUJU-8030). But let me know what you think.

PR #1620, opened 2025-06-13
chore: implement CheckMachines method

kian99:
## Description
This PR implements the `CheckMachines` method on the `MigrationTarget` facade for migrating models through JIMM. The implementation is mostly a pass-through to calling the target Juju controller.

The one area of interest is in the `jujuapi` implementation, where we take the `[]errors` returned by the Juju controller and turn these back into error structs according to Juju's RPC params. One might wonder whether we need to ensure the error code is maintained, beyond just the error message as is done in this PR. Taking a look at the Juju client source code [here](https://github.com/juju/juju/blob/3.6/api/controller/migrationtarget/client.go#L290) shows that only the error message is extracted and the error code and error info are not inspected so we are safe in what we are doing. In fact, because the Juju client package does not extract the code and info fields, we do not have this information at all.

Fixes [JUJU-8088](https://warthogs.atlassian.net/browse/JUJU-8088)

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

[JUJU-8088]: https://warthogs.atlassian.net/browse/JUJU-8088?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/jimm/jimm.go
+++ b/internal/jimm/jimm.go
@@ -243,6 +243,7 @@ type JujuManager interface {
 	// in the order they are expected to be called.
 	PrepareModelMigration(ctx context.Context, user *openfga.User, modelUUID string, targetControllerName string, userMapping map[string]string) error
 	Prechecks(ctx context.Context, user *openfga.User, model migration.ModelInfo) error
+	CheckMachines(ctx context.Context, user *openfga.User, modelUUID string) ([]error, error)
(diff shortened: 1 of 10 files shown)


SimoneDutto (approved):

ale8k (approved):
lgtm

PR #1621, opened 2025-06-17
feat: proxy http endpoints for model migration

kian99:
## Description
This PR introduces HTTP handlers for Juju's migration endpoints located at `migration/*` which are used to send charms/resources/tools from one controller to another when a model is migrating. The handlers are implemented in a similar fashion to the existing `HTTPProxyHandler` in `internal/jimmhttp/httpproxy_handler.go` with the caveat that I have refactored those existing handlers to move the application logic they were performing into the appropriate place. The new handlers feature slightly different authentication and logic because the model UUID is contained in a header value rather than as a path segment.

I've kept the handler tests as integration tests. While I have moved the application logic associated with fetch a model migration and then fetching the controller details into the domain layer, I have not opted to mock out the proxyHTTP call so the tests remain the same for the time being.

I have also refactored the proxy logic to use the standard library's `net/http/httputil.ReverseProxy` struct which provides an extensive implementation of HTTP reverse proxy logic. Currently the approach when reverse proxying is to either:
- Try the controller's public address if it is not empty.
- Randomise and select 1 of the controller's addresses that are potentially reachable depending on the network scope.

In a follow-up PR I will add additional logic to the proxy to try the controller's addresses in sequential order.

Fixes [JUJU-8120](https://warthogs.atlassian.net/browse/JUJU-8120)

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

[JUJU-8120]: https://warthogs.atlassian.net/browse/JUJU-8120?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cmd/jimmsrv/service/service.go
+++ b/cmd/jimmsrv/service/service.go
@@ -499,12 +499,16 @@ func NewService(ctx context.Context, p Params) (*Service, error) {
 	// Websockets require extra care when cookies are used for authentication
 	// to avoid CSRF attacks. https://portswigger.net/web-security/websockets/cross-site-websocket-hijacking
 	websocketCors := middleware.NewWebsocketCors(p.CorsAllowedOrigins)
+	// Juju API handlers
 	s.mux.Handle("/api", websocketCors.Handler(jujuapi.APIHandler(ctx, s.jimm, params)))
 	s.mux.Handle("/model/*", websocketCors.Handler(http.StripPrefix("/model", jujuapi.ModelHandler(ctx, s.jimm, params))))
-	mountHandler(
-		"/model/{uuid}/{type:charms|applications}",
-		jimmhttp.NewHTTPProxyHandler(s.jimm),
-	)
+	// Uploading local charms
+	mountHandler("/model/{uuid}/{type:charms|applications}", jimmhttp.NewHTTPProxyHandler(s.jimm))
+	// HTTP Migration endpoints
+	mountHandler("/migrate", jimmhttp.NewMigrationHTTPProxyHandler(s.jimm))
+	// Log transfer endpoint
+	// TODO: Implement log transfer, this uses a websocket rather than HTTP.
+	// s.mux.Handle("/migrate/logtransfer", jimmhttp.NewHTTPProxyHandler(s.jimm).Routes())
 
 	// serve the ssh public key fingerprint
 	s.mux.Get("/ssh/public-key-fingerprints", jimmhttp.WriteFingerprints(p.HostKeyFingerprints))
--- a/internal/jimm/jimm.go
+++ b/internal/jimm/jimm.go
@@ -247,6 +247,10 @@ type JujuManager interface {
 	ValidateModelUpgrade(ctx context.Context, u *openfga.User, mt names.ModelTag, force bool) error
 
 	// Migration related methods
+
+	// ControllerDetailsForIncomingModel retrieves details about the
+	// target controller for a model that is being migrated.
+	ControllerDetailsForIncomingModel(ctx context.Context, modelUUID string) (dbmodel.Controller, string, string, error)
 	// The migration methods below are sorted roughly
 	// in the order they are expected to be called.
 	PrepareModelMigration(ctx context.Context, user *openfga.User, modelUUID string, targetControllerName string, userMapping map[string]string) error
@@ -270,6 +274,7 @@ type JujuManager interface {
 	GetCloud(ctx context.Context, u *openfga.User, tag names.CloudTag) (dbmodel.Cloud, error)
 	GetCloudCredential(ctx context.Context, user *openfga.User, tag names.CloudCredentialTag) (*dbmodel.CloudCredential, error)
 	GetCloudCredentialAttributes(ctx context.Context, u *openfga.User, cred *dbmodel.CloudCredential, hidden bool) (attrs map[string]string, redacted []string, err error)
+	ControllerDetailsForModel(ctx context.Context, modelUUID string) (dbmodel.Controller, string, string, error)
 	GrantOfferAccessOnController(ctx context.Context, user *openfga.User, ut names.UserTag, offerURL string, access jujuparams.OfferAccessPermission) error
 	InitiateInternalMigration(ctx context.Context, user *openfga.User, modelNameOrUUID string, targetController string) (jujuparams.InitiateMigrationResult, error)
 	InitiateMigration(ctx context.Context, user *openfga.User, spec jujuparams.MigrationSpec) (jujuparams.InitiateMigrationResult, error)
--- a/internal/jimm/jimm.go
+++ b/internal/jimm/jimm.go
@@ -250,7 +250,7 @@ type JujuManager interface {
 
 	// ControllerDetailsForIncomingModel retrieves details about the
 	// target controller for a model that is being migrated.
-	ControllerDetailsForIncomingModel(ctx context.Context, modelUUID string) (juju.ControllerDetails, error)
+	ControllerDetailsForIncomingModel(ctx context.Context, modelUUID string) (juju.ControllerConnectionDetails, error)
 	// The migration methods below are sorted roughly
 	// in the order they are expected to be called.
 	PrepareModelMigration(ctx context.Context, user *openfga.User, modelUUID string, targetControllerName string, userMapping map[string]string) error
@@ -274,7 +274,7 @@ type JujuManager interface {
 	GetCloud(ctx context.Context, u *openfga.User, tag names.CloudTag) (dbmodel.Cloud, error)
 	GetCloudCredential(ctx context.Context, user *openfga.User, tag names.CloudCredentialTag) (*dbmodel.CloudCredential, error)
 	GetCloudCredentialAttributes(ctx context.Context, u *openfga.User, cred *dbmodel.CloudCredential, hidden bool) (attrs map[string]string, redacted []string, err error)
-	ControllerDetailsForModel(ctx context.Context, modelUUID string) (juju.ControllerDetails, error)
+	ControllerDetailsForModel(ctx context.Context, modelUUID string) (juju.ControllerConnectionDetails, error)
 	GrantOfferAccessOnController(ctx context.Context, user *openfga.User, ut names.UserTag, offerURL string, access jujuparams.OfferAccessPermission) error
 	InitiateInternalMigration(ctx context.Context, user *openfga.User, modelNameOrUUID string, targetController string) (jujuparams.InitiateMigrationResult, error)
 	InitiateMigration(ctx context.Context, user *openfga.User, spec jujuparams.MigrationSpec) (jujuparams.InitiateMigrationResult, error)
--- a/internal/jimm/jimm.go
+++ b/internal/jimm/jimm.go
@@ -250,7 +250,7 @@ type JujuManager interface {
 
 	// ControllerDetailsForIncomingModel retrieves details about the
 	// target controller for a model that is being migrated.
-	ControllerDetailsForIncomingModel(ctx context.Context, modelUUID string) (dbmodel.Controller, string, string, error)
+	ControllerDetailsForIncomingModel(ctx context.Context, modelUUID string) (juju.ControllerDetails, error)
 	// The migration methods below are sorted roughly
 	// in the order they are expected to be called.
 	PrepareModelMigration(ctx context.Context, user *openfga.User, modelUUID string, targetControllerName string, userMapping map[string]string) error
@@ -274,7 +274,7 @@ type JujuManager interface {
 	GetCloud(ctx context.Context, u *openfga.User, tag names.CloudTag) (dbmodel.Cloud, error)
 	GetCloudCredential(ctx context.Context, user *openfga.User, tag names.CloudCredentialTag) (*dbmodel.CloudCredential, error)
 	GetCloudCredentialAttributes(ctx context.Context, u *openfga.User, cred *dbmodel.CloudCredential, hidden bool) (attrs map[string]string, redacted []string, err error)
-	ControllerDetailsForModel(ctx context.Context, modelUUID string) (dbmodel.Controller, string, string, error)
+	ControllerDetailsForModel(ctx context.Context, modelUUID string) (juju.ControllerDetails, error)
 	GrantOfferAccessOnController(ctx context.Context, user *openfga.User, ut names.UserTag, offerURL string, access jujuparams.OfferAccessPermission) error
 	InitiateInternalMigration(ctx context.Context, user *openfga.User, modelNameOrUUID string, targetController string) (jujuparams.InitiateMigrationResult, error)
 	InitiateMigration(ctx context.Context, user *openfga.User, spec jujuparams.MigrationSpec) (jujuparams.InitiateMigrationResult, error)
--- a/internal/jimm/juju/controller.go
+++ b/internal/jimm/juju/controller.go
@@ -772,6 +772,10 @@ func (j *JujuManager) ControllerDetailsForModel(ctx context.Context, modelUUID s
 		return ControllerDetails{}, errors.E(op, err)
 	}
 
+	if username == "" || password == "" {
+		return ControllerDetails{}, errors.E(op, errors.CodeNotFound, fmt.Errorf("missing credentials for controller %q", model.Controller.Name))
+	}
+
 	return ControllerDetails{
 		Controller: model.Controller,
 		Credentials: ControllerCreds{
--- a/internal/jimm/juju/controller.go
+++ b/internal/jimm/juju/controller.go
@@ -750,7 +750,7 @@ func (j *JujuManager) ControllerConfig(ctx context.Context, controllerName strin
 
 // ControllerDetailsForModel returns the controller details for the specified model
 // including the controller, and the admin credentials (username and password) for the controller.
-func (j *JujuManager) ControllerDetailsForModel(ctx context.Context, modelUUID string) (dbmodel.Controller, string, string, error) {
+func (j *JujuManager) ControllerDetailsForModel(ctx context.Context, modelUUID string) (ControllerDetails, error) {
 	const op = errors.Op("jimm.ControllerDetailsForModel")
 
 	model := dbmodel.Model{
@@ -761,13 +761,22 @@ func (j *JujuManager) ControllerDetailsForModel(ctx context.Context, modelUUID s
 	}
 	err := j.Database.GetModel(ctx, &model)
 	if err != nil {
-		return dbmodel.Controller{}, "", "", errors.E(op, err)
+		if errors.ErrorCode(err) == errors.CodeNotFound {
+			return ControllerDetails{}, errors.E(op, errors.CodeNotFound, fmt.Sprintf("migrating model %q not found", modelUUID))
+		}
+		return ControllerDetails{}, errors.E(op, err)
 	}
 
 	username, password, err := j.CredentialStore.GetControllerCredentials(ctx, model.Controller.Name)
 	if err != nil {
-		return dbmodel.Controller{}, "", "", errors.E(op, err)
+		return ControllerDetails{}, errors.E(op, err)
 	}
 
-	return model.Controller, username, password, nil
+	return ControllerDetails{
+		Controller: model.Controller,
+		Credentials: ControllerCreds{
+			AdminIdentityName: username,
+			AdminPassword:     password,
+		},
+	}, nil
 }
--- a/internal/jimm/juju/controller.go
+++ b/internal/jimm/juju/controller.go
@@ -747,3 +747,27 @@ func (j *JujuManager) ControllerConfig(ctx context.Context, controllerName strin
 	}
 	return jujucontroller.Config(cfg.Config), nil
 }
+
+// ControllerDetailsForModel returns the controller details for the specified model
+// including the controller, and the admin credentials (username and password) for the controller.
+func (j *JujuManager) ControllerDetailsForModel(ctx context.Context, modelUUID string) (dbmodel.Controller, string, string, error) {
+	const op = errors.Op("jimm.ControllerDetailsForModel")
+
+	model := dbmodel.Model{
+		UUID: sql.NullString{
+			String: modelUUID,
+			Valid:  true,
+		},
+	}
+	err := j.Database.GetModel(ctx, &model)
+	if err != nil {
+		return dbmodel.Controller{}, "", "", errors.E(op, err)
+	}
+
+	username, password, err := j.CredentialStore.GetControllerCredentials(ctx, model.Controller.Name)
+	if err != nil {
+		return dbmodel.Controller{}, "", "", errors.E(op, err)
+	}
+
+	return model.Controller, username, password, nil
+}
--- a/internal/jimm/juju/controller.go
+++ b/internal/jimm/juju/controller.go
@@ -750,7 +750,7 @@ func (j *JujuManager) ControllerConfig(ctx context.Context, controllerName strin
 
 // ControllerDetailsForModel returns the controller details for the specified model
 // including the controller, and the admin credentials (username and password) for the controller.
-func (j *JujuManager) ControllerDetailsForModel(ctx context.Context, modelUUID string) (ControllerDetails, error) {
+func (j *JujuManager) ControllerDetailsForModel(ctx context.Context, modelUUID string) (ControllerConnectionDetails, error) {
 	const op = errors.Op("jimm.ControllerDetailsForModel")
 
 	model := dbmodel.Model{
@@ -762,25 +762,19 @@ func (j *JujuManager) ControllerDetailsForModel(ctx context.Context, modelUUID s
 	err := j.Database.GetModel(ctx, &model)
 	if err != nil {
 		if errors.ErrorCode(err) == errors.CodeNotFound {
-			return ControllerDetails{}, errors.E(op, errors.CodeNotFound, fmt.Sprintf("migrating model %q not found", modelUUID))
+			return ControllerConnectionDetails{}, errors.E(op, errors.CodeNotFound, fmt.Sprintf("migrating model %q not found", modelUUID))
 		}
-		return ControllerDetails{}, errors.E(op, err)
+		return ControllerConnectionDetails{}, errors.E(op, err)
 	}
 
 	username, password, err := j.CredentialStore.GetControllerCredentials(ctx, model.Controller.Name)
 	if err != nil {
-		return ControllerDetails{}, errors.E(op, err)
+		return ControllerConnectionDetails{}, errors.E(op, err)
 	}
 
 	if username == "" || password == "" {
-		return ControllerDetails{}, errors.E(op, errors.CodeNotFound, fmt.Errorf("missing credentials for controller %q", model.Controller.Name))
+		return ControllerConnectionDetails{}, errors.E(op, errors.CodeNotFound, fmt.Errorf("missing credentials for controller %q", model.Controller.Name))
 	}
 
-	return ControllerDetails{
-		Controller: model.Controller,
-		Credentials: ControllerCreds{
-			AdminIdentityName: username,
-			AdminPassword:     password,
-		},
-	}, nil
+	return toControllerConnectionDetails(model.Controller, username, password), nil
 }
--- a/internal/jimm/juju/controller_test.go
+++ b/internal/jimm/juju/controller_test.go
@@ -1492,3 +1492,57 @@ func (c *testControllerClient) InitiateMigration(spec controller.MigrationSpec) 
 func (c *testControllerClient) Close() error {
 	return nil
 }
+
+const testControllerDetailsForModelEnv = `clouds:
+- name: test-cloud
+  type: test
+  regions:
+  - name: test-region-1
+cloud-credentials:
+- name: test-cred
+  cloud: test-cloud
+  owner: alice@canonical.com
+  type: empty
+controllers:
+- name: controller-1
+  uuid: 00000001-0000-0000-0000-000000000001
+  cloud: test-cloud
+  region: test-region-1
+  agent-version: 3.3
+  public-address: test-address.com
+models:
+- name: model-1
+  uuid: 00000002-0000-0000-0000-000000000003
+  controller: controller-1
+  cloud: test-cloud
+  region: test-region-1
+  cloud-credential: test-cred
+  owner: alice@canonical.com
+`
+
+func TestControllerDetailsForModel(t *testing.T) {
+	c := qt.New(t)
+	ctx := context.Background()
+
+	invalidUUID := "invalid-uuid"
+	validUUID := "00000002-0000-0000-0000-000000000003"
+
+	j := newTestJujuManager(c, nil)
+
+	env := jimmtest.ParseEnvironment(c, testControllerDetailsForModelEnv)
+	env.PopulateDB(c, j.Database)
+
+	err := j.CredentialStore.PutControllerCredentials(ctx, "controller-1", "test-user", "test-password")
+	c.Assert(err, qt.IsNil)
+
+	_, _, _, err = j.ControllerDetailsForModel(ctx, invalidUUID)
+	c.Assert(err, qt.ErrorMatches, `model not found`)
+
+	ctl, username, password, err := j.ControllerDetailsForModel(ctx, validUUID)
+	c.Assert(err, qt.IsNil)
+	c.Assert(ctl.ID, qt.Not(qt.Equals), 0)
+	c.Assert(ctl.Name, qt.Equals, "controller-1")
+	c.Assert(ctl.PublicAddress, qt.Equals, "test-address.com")
+	c.Assert(username, qt.Equals, "test-user")
+	c.Assert(password, qt.Equals, "test-password")
+}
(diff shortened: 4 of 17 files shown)


internal/jimm/jimm.go
```diff
@@ -262,6 +266,7 @@ type JujuManager interface {
 	GetCloud(ctx context.Context, u *openfga.User, tag names.CloudTag) (dbmodel.Cloud, error)
 	GetCloudCredential(ctx context.Context, user *openfga.User, tag names.CloudCredentialTag) (*dbmodel.CloudCredential, error)
 	GetCloudCredentialAttributes(ctx context.Context, u *openfga.User, cred *dbmodel.CloudCredential, hidden bool) (attrs map[string]string, redacted []string, err error)
+	ControllerDetailsForModel(ctx context.Context, modelUUID string) (dbmodel.Controller, string, string, error)
```
alesstimec: where are we using this? i can't see, but i may be blind :)

kian99: Ah good catch, I didn't wire it up. Now fixed.

internal/jimm/juju/migrationtarget_test.go
```diff
@@ -142,6 +146,37 @@ func TestCheckMachines_MissingIncomingModel(t *testing.T) {
...
+
+	j := newTestJujuManager(c, nil)
+
+	env := jimmtest.ParseEnvironment(c, testMigrationEnv)
+	env.PopulateDBAndPermissions(c, j.ResourceTag(), j.Database, j.OpenFGAClient)
+
+	err := j.CredentialStore.PutControllerCredentials(ctx, "test1", "test-user", "test-password")
+	c.Assert(err, qt.IsNil)
+
+	_, _, _, err = j.ControllerDetailsForIncomingModel(ctx, migratingModelUUID)
```
alesstimec: can we add a test case where we don't have credentials for the controller?

internal/jimmhttp/httpproxy_handler.go
```diff
@@ -55,23 +54,24 @@ func (hph *HTTPProxyHandler) ProxyHTTP(w http.ResponseWriter, req *http.Request)
...
-		writeError(ctx, w, http.StatusNotFound, err, "cannot get model")
-		return
-	}
-	u, p, err := hph.jimm.CredentialStore.GetControllerCredentials(ctx, model.Controller.Name)
+
+	controller, username, password, err := hph.jimm.JujuManager().ControllerDetailsForModel(ctx, modelUUID)
 	if err != nil {
-		writeError(ctx, w, http.StatusNotFound, err, "cannot retrieve credentials")
+		if errors.ErrorCode(err) == errors.CodeNotFound {
+			writeError(ctx, w, http.StatusNotFound, err, "controller details not found")
```
alesstimec: i'd just return "model not found".. it could be that we don't have the model in our db.. i wouldn't expect a case where we know the model, but can't find a controller for it..

cmd/jimmsrv/service/service.go
```diff
@@ -499,12 +499,16 @@ func NewService(ctx context.Context, p Params) (*Service, error) {
 	// Websockets require extra care when cookies are used for authentication
 	// to avoid CSRF attacks. https://portswigger.net/web-security/websockets/cross-site-websocket-hijacking
 	websocketCors := middleware.NewWebsocketCors(p.CorsAllowedOrigins)
+	// Juju API handlers
```
ale8k, line 502: Much love ❤️

internal/jimm/juju/migrationtarget.go
```diff
@@ -86,6 +86,34 @@ func (j *JujuManager) CheckMachines(ctx context.Context, user *openfga.User, mod
...
+		ModelUUID: sql.NullString{
+			String: modelUUID,
+			Valid:  true,
+		},
+	}
+
+	err := j.Database.GetIncomingModelMigration(ctx, &incomingModel)
+	if err != nil {
+		if errors.ErrorCode(err) == errors.CodeNotFound {
+			return dbmodel.Controller{}, "", "", errors.E(op, errors.CodeNotFound, fmt.Sprintf("migrating model %q not found", modelUUID))
```
ale8k: To stop us creating empty dbmodel.Controller twice here, maybe just a pointer would be better with nil?

internal/jimmhttp/httpproxy_handler.go
```diff
@@ -55,23 +54,24 @@ func (hph *HTTPProxyHandler) ProxyHTTP(w http.ResponseWriter, req *http.Request)
 
 	modelUUID := chi.URLParam(req, "uuid")
 	if modelUUID == "" {
-		writeError(ctx, w, http.StatusUnprocessableEntity, errgo.New("cannot parse path"), "cannot parse path")
+		writeError(ctx, w, http.StatusUnprocessableEntity, errors.E("cannot parse path"), "cannot parse path")
```
ale8k: Can make little string var above and pass to both err and msg.

internal/jimmhttp/migrationproxy_handler.go
```diff
@@ -0,0 +1,85 @@
...
+func (hph *MigrationHTTPProxyHandler) ProxyHTTP(w http.ResponseWriter, req *http.Request) {
+	ctx := req.Context()
+
+	modelUUID := req.Header.Get(jujuparams.MigrationModelHTTPHeader)
+	if modelUUID == "" {
+		errMsg := fmt.Sprintf("missing %s header value", jujuparams.MigrationModelHTTPHeader)
+		writeError(ctx, w, http.StatusBadRequest, errors.E(errMsg), errMsg)
+		return
+	}
+
```
ale8k, line 68: Should validate model uuid here, no reason to continue and call controller details if the uuid is invalid.

internal/jimm/juju/migrationtarget.go
```diff
@@ -86,6 +86,34 @@ func (j *JujuManager) CheckMachines(ctx context.Context, user *openfga.User, mod
 	return machineErrors, nil
 }
 
+// ControllerDetailsForIncomingModel retrieves the target controller details for a model that is being migrated.
+// It returns the controller information, username, and password for the target controller.
+func (j *JujuManager) ControllerDetailsForIncomingModel(ctx context.Context, modelUUID string) (dbmodel.Controller, string, string, error) {
+	const op = errors.Op("jimm.GetControllerForIncomingModel")
```
ale8k: Op code needs updating.

internal/jimm/juju/controller.go
```diff
@@ -747,3 +747,27 @@ func (j *JujuManager) ControllerConfig(ctx context.Context, controllerName strin
...
+func (j *JujuManager) ControllerDetailsForModel(ctx context.Context, modelUUID string) (dbmodel.Controller, string, string, error) {
+	const op = errors.Op("jimm.ControllerDetailsForModel")
+
+	model := dbmodel.Model{
+		UUID: sql.NullString{
+			String: modelUUID,
+			Valid:  true,
+		},
+	}
+	err := j.Database.GetModel(ctx, &model)
```
ale8k, line 762: If there was an error, model.Controller is zeroed already. No need to return `dbmodel.Controller{}`

internal/jimm/juju/controller_test.go
```diff
@@ -1492,3 +1492,57 @@ func (c *testControllerClient) InitiateMigration(spec controller.MigrationSpec)
...
+func TestControllerDetailsForModel(t *testing.T) {
+	c := qt.New(t)
+	ctx := context.Background()
+
+	invalidUUID := "invalid-uuid"
+	validUUID := "00000002-0000-0000-0000-000000000003"
+
+	j := newTestJujuManager(c, nil)
+
+	env := jimmtest.ParseEnvironment(c, testControllerDetailsForModelEnv)
```
ale8k, line 1532: This doesn't need a db? You just care `GetModel` and `GetControllerDetails` returns "something"?

Can leave it, but I really think we shouldn't be going to DB at this point (like our old discussions about sandwich integration testing).

internal/jimm/juju/controller_test.go
```diff
@@ -1492,3 +1492,57 @@ func (c *testControllerClient) InitiateMigration(spec controller.MigrationSpec)
...
+	env.PopulateDB(c, j.Database)
+
+	err := j.CredentialStore.PutControllerCredentials(ctx, "controller-1", "test-user", "test-password")
+	c.Assert(err, qt.IsNil)
+
+	_, _, _, err = j.ControllerDetailsForModel(ctx, invalidUUID)
+	c.Assert(err, qt.ErrorMatches, `model not found`)
```
ale8k: This is what I meant with my comment above, this error is nothing to do with "ControllerDetailsForModel", and let's say you change the DB error for `GetModel`, suddenly this test is broken despite it actually working perfectly fine.

internal/jimm/juju/migrationtarget.go
ale8k: Opcode needs updating

internal/jimm/juju/controller.go
```diff
@@ -747,3 +747,27 @@ func (j *JujuManager) ControllerConfig(ctx context.Context, controllerName strin
 	}
 	return jujucontroller.Config(cfg.Config), nil
 }
+
+// ControllerDetailsForModel returns the controller details for the specified model
+// including the controller, and the admin credentials (username and password) for the controller.
+func (j *JujuManager) ControllerDetailsForModel(ctx context.Context, modelUUID string) (dbmodel.Controller, string, string, error) {
```
ale8k: Prefer this and `ControllerDetailsForIncomingModel` to have a struct response with details inside. 4 return values is a bit much and less clear. Just a simple

```go
type ControllerDetails struct {
    Ctrl dbmodel.Controller
    Username string
    Password string
}
```

Will be fine.

internal/jimm/juju/migrationtarget_test.go
```diff
@@ -142,6 +146,37 @@ func TestCheckMachines_MissingIncomingModel(t *testing.T) {
 	c.Assert(err, qt.ErrorMatches, `.*model migration not found`)
 }
 
+func TestControllerDetailsForIncomingModel(t *testing.T) {
```
ale8k, line 148: In controller_test.go we check for model not found, but we don't check for `errors.E(op, errors.CodeNotFound, fmt.Sprintf("migrating model %q not found", modelUUID))` here.

internal/jimm/juju/migrationtarget.go
ale8k: Our code for the controller.go variant which just return ctrl details has:

```go
	err := j.Database.GetModel(ctx, &model)
	if err != nil {
		return dbmodel.Controller{}, "", "", errors.E(op, err)
	}
```

Which propogates the db error, creating a change detector. But here we're creating an explicit error not linked to the db error. Perhaps changing controller.go where it fails to get the model to have an explicit error would be good to make it consistent with this?

internal/jimm/juju/migrationtarget.go
```diff
@@ -86,6 +86,34 @@ func (j *JujuManager) CheckMachines(ctx context.Context, user *openfga.User, mod
 	return machineErrors, nil
 }
 
+// ControllerDetailsForIncomingModel retrieves the target controller details for a model that is being migrated.
+// It returns the controller information, username, and password for the target controller.
+func (j *JujuManager) ControllerDetailsForIncomingModel(ctx context.Context, modelUUID string) (dbmodel.Controller, string, string, error) {
```
ale8k: Can share same struct as suggested earlier here.

ale8k: Also shouldn't it be "ControllerDetailsFromIncomingModel"?

internal/jimmhttp/httpproxy_handler.go
```diff
@@ -55,23 +54,24 @@ func (hph *HTTPProxyHandler) ProxyHTTP(w http.ResponseWriter, req *http.Request)
...
 		return
 	}
-	model, err := hph.jimm.JujuManager().GetModel(ctx, modelUUID)
-	if err != nil {
-		writeError(ctx, w, http.StatusNotFound, err, "cannot get model")
-		return
-	}
-	u, p, err := hph.jimm.CredentialStore.GetControllerCredentials(ctx, model.Controller.Name)
+
```
ale8k, line 62: Should validate model uuid here, no reason to continue and call controller details if the uuid is invalid.

internal/rpc/httpproxy.go
```diff
@@ -7,98 +7,117 @@ import (
...
 )
 
-type httpOptions struct {
-	TLSConfig *tls.Config
-	URL       url.URL
+const (
+	defaultScheme = "https"
+)
+
+type ControllerProxy struct {
```
ale8k: `s/ControllerProxy/ControllerProxtDetails`

internal/jimmhttp/migrationproxy_handler.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/canonical/jimm/v3/internal/errors"
+	"github.com/canonical/jimm/v3/internal/jimm"
+	"github.com/canonical/jimm/v3/internal/middleware"
+	"github.com/canonical/jimm/v3/internal/rpc"
+)
+
+// MigrationProxyHandler is an handler that provides proxying,
+// specifically for HTTP-based model migration endpoints. These
+// are used primarily to send data like charms/resources/tools
```
ale8k, line 22: It'd be good to be super explicit about what migration endpoints we're handling, even if it'd handle more. I say this because:
```go
const (
	// all endpoints managed by this handler
	migrationProxyEndpoints = "/*"
)
```

Isn't very helpful aha 😝

internal/jimmhttp/migrationproxy_handler.go
```diff
@@ -0,0 +1,85 @@
...
+	hph.Router.Use(func(h http.Handler) http.Handler {
+		return middleware.AuthenticateWithSessionTokenViaBasicAuth(h, hph.jimm.LoginManager())
+	})
+	hph.Router.Use(middleware.AuthorizeUserAsJIMMAdmin)
+}
+
+// ProxyHTTP extracts the model uuid from an HTTP header to proxy the request to the right controller.
+func (hph *MigrationHTTPProxyHandler) ProxyHTTP(w http.ResponseWriter, req *http.Request) {
+	ctx := req.Context()
+
```
ale8k, line 61: Lil debug line saying this is starting the path (`req.URL.Path`) would be good. If we ever had issues in a prod deploy, we've got no visibility as to what endpoint has been hit / not been hit.

internal/rpc/httpproxy.go
```diff
@@ -7,98 +7,117 @@ import (
...
-		for _, v := range vv {
-			w.Header().Add(k, v)
+func getControllerAddresses(ctl dbmodel.Controller) ([]*url.URL, error) {
+	urls := make([]*url.URL, 0, 1)
+	if ctl.PublicAddress != "" {
+		address := ctl.PublicAddress
+		if !strings.Contains(address, "://") {
+			address = defaultScheme + "://" + address // ensure the address has a scheme
 		}
+		newURL, err := url.Parse(address)
```
ale8k, line 98: Shouldn't we be parsing the public address, and then doing:
```go
	u.Scheme = defaultScheme
```
What you're doing here is what url.Parse should do for us...

internal/rpc/httpproxy.go
```diff
@@ -7,98 +7,117 @@ import (
...
+type ControllerProxy struct {
+	Controller dbmodel.Controller
+	Username   string
+	Password   string
 }
 
-// ProxyHTTP proxies the request to the controller using the info contained in dbmodel.Controller.
-// It tries for a controller, if it errors, it logs the error and go to the next, if no controller responds it returns a 504.
-func ProxyHTTP(ctx context.Context, ctl *dbmodel.Controller, w http.ResponseWriter, req *http.Request) error {
+func ProxyHTTP(ctx context.Context, ctl ControllerProxy, w http.ResponseWriter, req *http.Request) {
```
ale8k: Godoc

internal/testutils/jimmtest/mocks/jimm_juju_controller_mock.go
```diff
@@ -16,13 +16,15 @@ import (
...
 // ControllerService is an implementation of the jujuapi.ControllerService interface.
 type ControllerService struct {
-	AddController_             func(ctx context.Context, u *openfga.User, ctl *dbmodel.Controller, creds juju.ControllerCreds) error
-	ControllerInfo_            func(ctx context.Context, name string) (*dbmodel.Controller, error)
-	EarliestControllerVersion_ func(ctx context.Context) (version.Number, error)
-	ListControllers_           func(ctx context.Context, user *openfga.User) ([]dbmodel.Controller, error)
-	RemoveController_          func(ctx context.Context, user *openfga.User, controllerName string, force bool) error
-	SetControllerDeprecated_   func(ctx context.Context, user *openfga.User, controllerName string, deprecated bool) error
-	ControllerConfig_          func(ctx context.Context, controllerName string) (jujucontroller.Config, error)
+	AddController_                     func(ctx context.Context, u *openfga.User, ctl *dbmodel.Controller, creds juju.ControllerCreds) error
```
ale8k, line 19: We need to change these service names to just JujuManager, etc. In future. It's confusing to see a mock struct called controller service despite an actual controller service not existing.

internal/rpc/httpproxy.go
```diff
@@ -7,98 +7,117 @@ import (
...
-				URL:       createURLWithNewHost(*req.URL, fmt.Sprintf("%s:%d", hp.Value, hp.Port)),
-			})
-			if err == nil {
-				return nil
-			} else {
-				zapctx.Error(ctx, "failed to proxy request: continue to next addr", zaputil.Error(err))
-			}
-		}
+
+	if len(urls) == 0 {
```
ale8k: This should be done on line 43.

internal/rpc/httpproxy.go
```diff
@@ -7,98 +7,117 @@ import (
...
+	// transport is the default HTTP transport config with TLS configuration.
+	transport := &http.Transport{
+		TLSClientConfig:       tlsConfig,
+		Proxy:                 http.ProxyFromEnvironment,
+		ForceAttemptHTTP2:     true,
+		MaxIdleConns:          100,
+		IdleConnTimeout:       90 * time.Second,
+		TLSHandshakeTimeout:   10 * time.Second,
+		ExpectContinueTimeout: 1 * time.Second,
 	}
```
ale8k: Can change this to:
```go
	transport := http.DefaultTransport.(*http.Transport).Clone()
	transport.TLSClientConfig = tlsConfig
```

internal/rpc/httpproxy.go
```diff
@@ -7,98 +7,117 @@ import (
...
+	if len(urls) == 0 {
+		zapctx.Error(ctx, "no controller addresses found", zap.String("controller", ctl.Controller.Name))
+		http.Error(w, "no controller addresses found", http.StatusInternalServerError)
+		return
 	}
 
-	return errgo.New("couldn't reach a valid address for controller")
-}
+	if len(urls) > 1 {
+		rand.Shuffle(len(urls), func(i, j int) {
```
ale8k, line 64: Shouldn't this be done when getting the addresses? Why does ProxyHTTP care?

internal/rpc/httpproxy.go
```diff
@@ -7,98 +7,117 @@ import (
...
+	}
 
-func doRequest(ctx context.Context, w http.ResponseWriter, req *http.Request, opt httpOptions) error {
-	client := &http.Client{
-		Transport: &http.Transport{
-			TLSClientConfig: opt.TLSConfig,
+	// TODO: Consider implementing a better load balancing mechanism that handles
+	// multiples URLs and handles failing backends gracefully e.g. try send to first
+	// URL and on failure, try second, etc.
+	proxy := &httputil.ReverseProxy{
```
ale8k, line 73: It needs to use our zap logger, see:
```go
    // ErrorLog specifies an optional logger for errors
    // that occur when attempting to proxy the request.
    // If nil, logging is done via the log package's standard logger.
    ErrorLog *log.Logger
```

ale8k: Also, do any of these endpoints need instant streaming? Because if we ever proxy one where user experience is important. We'd need flush interval to be set to a negative value I think? Probably not a cause for concern right now but something to be aware of?

ale8k (commented):
I think this needs a bit more work but only minor things really. Once that's done, I think it's a LGTM! Nice work!

internal/jimm/juju/controller.go
kian99: Changed.

internal/jimm/juju/controller_test.go
kian99: We don't have a mock database, so we need a db for the j.Database.GetModel() call to work.
Unless I'm missing what you're saying.

internal/jimm/juju/controller_test.go
kian99: Depends on whether we want to enforce the error strings/types returned by this layer.
For example, if the model isn't in the DB, we get some error, if the `ControllerDetailsForModel` is expected to return a "not found" error in this scenario for the layer above (the API layer) then we should test and enforce that. The API layer could expect this error code in order to return an HTTP 404 or an HTTP 500.

It would be very easy to accidentally replace the error code from the DB layer in the domain layer here. In saying that, I'm not checking the error code, but I should so I'll check for that instead of just the error string.

internal/jimm/juju/migrationtarget.go
kian99: I think `for` incoming model makes more sense, but that's just me. The controller is one that is managed by JIMM, the model is incoming and on its way to that controller.

internal/jimm/juju/migrationtarget.go
kian99: Resolving now that we're using a differerent struct.

kian99: Yeah it was an oversight on my part, controller.go should have the same logic, good spot.

internal/jimm/juju/migrationtarget_test.go
kian99: Yip, good spot, I think the relevant check is for the error code, the error message isn't as important (that would just be a change detector test then). Will update the two tests to be consistent.

internal/jimmhttp/migrationproxy_handler.go
kian99: Done

internal/rpc/httpproxy.go
kian99: Renamed to `ControllerDetails` since the package name is already httpproxy. Then it will ready httpproxy.ControllerDetails which seems clear enough.

internal/rpc/httpproxy.go
kian99: I'd prefer to keep this one as is.

internal/rpc/httpproxy.go
kian99: This is changing in my follow-up so feel free to leave it for now but it is done to load-balance requests between the various controller nodes.

internal/rpc/httpproxy.go
kian99: Most of the time (probably always?) the address doesn't have a scheme, and without a scheme, `Parse` will fail.

internal/jimmhttp/migrationproxy_handler.go
kian99: This feels like it would be a better fit for a middleware (on the response), wdyt? Can do in a separate PR

internal/jimmhttp/httpproxy_handler.go
```diff
@@ -55,23 +55,31 @@ func (hph *HTTPProxyHandler) ProxyHTTP(w http.ResponseWriter, req *http.Request)
...
+	controllerDetails, err := hph.jimm.JujuManager().ControllerDetailsForModel(ctx, modelUUID)
 	if err != nil {
-		writeError(ctx, w, http.StatusNotFound, err, "cannot retrieve credentials")
+		if errors.ErrorCode(err) == errors.CodeNotFound {
+			writeError(ctx, w, http.StatusNotFound, err, "controller details not found")
+			return
+		}
+		writeError(ctx, w, http.StatusInternalServerError, err, "failed to get controller details")
```
alesstimec, line 75: are we giving away implementation details? or should return a vague error about failing to proxy the request and log the error?

internal/jimmhttp/migrationproxy_handler.go
```diff
@@ -0,0 +1,82 @@
...
+	if modelUUID == "" {
+		errMsg := fmt.Sprintf("missing %s header value", jujuparams.MigrationModelHTTPHeader)
+		writeError(ctx, w, http.StatusBadRequest, errors.E(errMsg), errMsg)
+		return
+	}
+
+	controllerDetails, err := hph.jimm.JujuManager().ControllerDetailsForIncomingModel(ctx, modelUUID)
+	if err != nil {
+		if errors.ErrorCode(err) == errors.CodeNotFound {
+			writeError(ctx, w, http.StatusNotFound, err, "controller details not found")
```
alesstimec: this, too, feels like a "model not found" error, rather than "controller details not found"

internal/jimmhttp/migrationproxy_handler.go
```diff
@@ -0,0 +1,82 @@
...
+			return
+		}
+		writeError(ctx, w, http.StatusInternalServerError, err, "cannot retrieve controller")
```
alesstimec, line 75: "failed to proxy the request" and log the error?

internal/jimm/juju/migrationtarget_test.go
kian99: Added, was good to add this too because our credential store implementation returns no error if a credential isn't found so we needed to check if the username/password aren't empty. I don't know why it's like that though.

internal/jimmhttp/httpproxy_handler.go
kian99: I don't think it's bad to return an "accurate" error. At least without some kind of tracing/correlation IDs, we'd have a tough time understanding errors if we make them more obscure.

internal/jimmhttp/migrationproxy_handler.go
kian99: As above, I think if we do that we might make life harder when debugging, especially if we don't return a correlation ID/use tracing to match the user's error with our logs/data.

internal/jimm/juju/controller_test.go
ale8k: An interface would work instead is what I meant, then that can be mocked.

kian99: We've never mocked the database at this layer before, I would prefer not to start now.

ale8k: That's fair, we can discuss go mocking this layer another time. Resolving.

internal/jimm/juju/migrationtarget.go
ale8k: I just meant you're getting them "from" the model, and not looking them up. But it's just semantics, resolving .

internal/rpc/httpproxy.go
ale8k: Ah yeah nice, that's better.

internal/jimmhttp/migrationproxy_handler.go
ale8k: A lil debug line specific to this for now is fine. But agreed an actual logging middleware would be good. Chi has one IIRC for request logging.

internal/rpc/httpproxy.go
ale8k: So then we can just do u.Scheme = defaultScheme always?

internal/rpc/httpproxy.go
ale8k: But we're using the default anyway and there's built in code to clone the default transport? I disagree, let's use the cloning built into the transport and just update the config rather than setting it all manually.

internal/jimmhttp/migrationproxy_handler.go
kian99: Sure, adding.

internal/rpc/httpproxy.go
kian99: I'm not following, without running Parse() we don't have a URL object, just a string. And without a scheme, `Parse()` will return an error.

internal/rpc/httpproxy.go
kian99: Sure, changing.

internal/jimmhttp/httpproxy_handler_test.go
```diff
@@ -112,18 +112,32 @@ func (s *httpProxySuite) TestHTTPProxyHandler(c *gc.C) {
...
 			url:            fmt.Sprintf("/model/%s/charms", "54d9f921-c45a-4825-8253-74e7edc28066"),
 			modelUUID:      "54d9f921-c45a-4825-8253-74e7edc28066",
 			statusExpected: http.StatusNotFound,
-			bodyExpected:   "Not Found - model not found",
+			bodyExpected:   "Not Found - migrating model .* not found",
 		},
 	}
 
 	for _, test := range tests {
+		c.Log(test.description)
```
ale8k, line 140: Do we need this?

internal/jimmhttp/httpproxy_handler_test.go
```diff
@@ -112,18 +112,32 @@ func (s *httpProxySuite) TestHTTPProxyHandler(c *gc.C) {
 			statusExpected: http.StatusOK,
 			bodyExpected:   "OK",
 		},
+		{
+			description: "invalid model UUID",
```
ale8k, line 116: Thanks for this

internal/jimmhttp/migrationproxy_handler.go
```diff
@@ -43,7 +40,9 @@ func NewMigrationHTTPProxyHandler(jimm *jimm.JIMM) *MigrationHTTPProxyHandler {
 // Routes returns the grouped routers routes with group specific middlewares.
 func (hph *MigrationHTTPProxyHandler) Routes() chi.Router {
 	hph.SetupMiddleware()
-	hph.Router.HandleFunc(migrationProxyEndpoints, hph.ProxyHTTP)
+	hph.Router.HandleFunc("/charms/*", hph.ProxyHTTP)
```
ale8k, line 43: Ty for this too

ale8k (approved):
LGTM, to for all the updates & sorry to be a hassle 😝

internal/jimmhttp/httpproxy_handler_test.go
kian99: Yes, when we use gocheck and you run subtests there is no c.Run() like there is with the standard library testing.T. So logging the subtest description is the best way to see which subtest failed. You'll see this often in Juju too.

internal/jimm/juju/controller.go
```diff
@@ -747,3 +747,40 @@ func (j *JujuManager) ControllerConfig(ctx context.Context, controllerName strin
 	}
 	return jujucontroller.Config(cfg.Config), nil
 }
+
+// ControllerDetailsForModel returns the controller details for the specified model
+// including the controller, and the admin credentials (username and password) for the controller.
+func (j *JujuManager) ControllerDetailsForModel(ctx context.Context, modelUUID string) (ControllerDetails, error) {
```
alesstimec: nitpick: you could return pointer to `ControllerDetails`, which would save you from having to do `return ControllerDetails{}, err`

internal/jimm/juju/migrationtarget.go
```diff
@@ -88,6 +88,44 @@ func (j *JujuManager) CheckMachines(ctx context.Context, user *openfga.User, mod
 	return machineErrors, nil
 }
 
+// ControllerDetailsForIncomingModel retrieves the target controller details for a model that is being migrated.
+// It returns the controller information, username, and password for the target controller.
+func (j *JujuManager) ControllerDetailsForIncomingModel(ctx context.Context, modelUUID string) (ControllerDetails, error) {
```
alesstimec: nitpick: same with comment about returning a pointer.

internal/jimm/juju/types.go
```diff
@@ -2,9 +2,18 @@
...
 // ControllerCreds represent the admin username and password
 // used to authenticate with a Juju controller via basic auth.
 type ControllerCreds struct {
 	AdminIdentityName string
 	AdminPassword     string
 }
+
+// ControllerDetails contains the details of a Juju controller,
+// including the controller itself and the credentials used to access it.
+type ControllerDetails struct {
```
alesstimec: perhaps a `ControllerConnectionDetails` type that wouldn't include `dbmodel.Controller` with all its details, but only the data we need to connect to the controller?

alesstimec (approved):
LGTM with a few nitpicks/suggestions

internal/jimm/juju/controller.go
kian99: Personally, I try to avoid returning a pointer unless it's necessary e.g. I'm returning a mutex in the object, etc. That will indicate to the caller they must pass the object around as a pointer. As is, we don't need to return a pointer so I'll leave it.

internal/jimm/juju/types.go
kian99: Added

PR #1622, opened 2025-06-18
feat(model-migration): implement Activate from the MigrationTarget facade

SimoneDutto:
# Description
This PR implements the Activate method on the MigrationTarget facade for migrating models through JIMM. The implementation is:
- check IncomingMigration
- pass the request through
- (in another PR) set the model to `alive`
- delete IncomingMigration

### DIFF
--- a/internal/jimm/juju/migrationtarget.go
+++ b/internal/jimm/juju/migrationtarget.go
@@ -8,11 +8,13 @@ import (
 	"fmt"
 
 	"github.com/juju/juju/core/migration"
+	coremigration "github.com/juju/juju/core/migration"
 	"github.com/juju/names/v5"
 	"github.com/juju/version/v2"
 	"github.com/juju/zaputil/zapctx"
 	"go.uber.org/zap"
 
+	"github.com/canonical/jimm/v3/internal/db"
 	"github.com/canonical/jimm/v3/internal/dbmodel"
 	"github.com/canonical/jimm/v3/internal/errors"
 	"github.com/canonical/jimm/v3/internal/openfga"
@@ -171,5 +173,58 @@ func (j *JujuManager) modifyMigrationInfo(model *migration.ModelInfo, userMappin
 	model.Owner = names.NewUserTag(newOwner)
 	// TODO: Replace fields on the model description including the model owner and users.
 
+	return nil
+}
+
+// Activate gets the model migration, proxies the Activate call to the target controller,
+// and then deletes the model migration from the database.
+func (j *JujuManager) Activate(ctx context.Context, modelTag names.ModelTag, migrationInfo coremigration.SourceControllerInfo, relatedModels []string) error {
+	const op = errors.Op("jimm.Activate")
+
+	modelMigration := dbmodel.IncomingModelMigration{
+		ModelUUID: sql.NullString{
+			String: modelTag.Id(),
+			Valid:  true,
+		},
+	}
+	err := j.Database.GetIncomingModelMigration(ctx, &modelMigration)
+	if err != nil {
+		return errors.E(op, fmt.Errorf("failed to get model migration for model %q: %w", modelTag.Id(), err))
+	}
+	api, err := j.dialController(ctx, &modelMigration.TargetController)
+	if err != nil {
+		return errors.E(op, fmt.Errorf("failed to dial controller: %w", err))
+	}
+	defer api.Close()
+
+	err = api.Activate(modelTag.Id(), migrationInfo, relatedModels)
+	if err != nil {
+		return errors.E(op, fmt.Errorf("failed to activate model %q: %w", modelTag.Id(), err))
+	}
+
+	// This is done in a transaction to ensure that the model migration is only deleted
+	// if user mappings have been created.
+	err = j.Database.Transaction(func(db *db.Database) error {
+		for localUser, externalUser := range modelMigration.UserMapping {
+			userMapping := &dbmodel.UserMapping{
+				ModelUUID:        modelMigration.ModelUUID,
+				LocalUser:        localUser,
+				ExternalUserName: externalUser,
+			}
+			err = db.AddUserMapping(ctx, userMapping)
+			if err != nil {
+				return errors.E(op, fmt.Errorf("failed to add user mapping for model %q: %w", modelTag.Id(), err))
+			}
+		}
+		// TODO(SimoneDutto): set the model we've created in Import to active.
+		err = db.DeleteIncomingModelMigration(ctx, &modelMigration)
+		if err != nil {
+			return errors.E(op, fmt.Errorf("failed to delete model migration for model %q: %w", modelTag.Id(), err))
+		}
+		return nil
+	})
+	if err != nil {
+		return errors.E(op, fmt.Errorf("failed to activate model %q: %w", modelTag.Id(), err))
+	}
 	return nil
 }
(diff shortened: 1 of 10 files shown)


internal/jimm/juju/migrationtarget.go
```diff
@@ -173,3 +175,42 @@ func (j *JujuManager) modifyMigrationInfo(model *migration.ModelInfo, userMappin
...
+
+		err = api.Activate(modelTag.Id(), migrationInfo, relatedModels)
+		if err != nil {
+			return errors.E(op, fmt.Errorf("failed to activate model %q: %w", modelTag.Id(), err))
+		}
+		// TODO(SimoneDutto): set the model we've created in Import to active.
+		err = db.DeleteIncomingModelMigration(ctx, &modelMigration)
+		if err != nil {
+			return errors.E(op, fmt.Errorf("failed to delete model migration for model %q: %w", modelTag.Id(), err))
+		}
```
kian99, line 223: If we are deleting this now there are a few things to do,
- Are there any steps that occur after this? E.g. transferring migrate-logs? If so those operations will need to look at the models table rather than the incomingModelMigration table. This one isn't for you, but more for everyone to keep in mind regarding whether an operation takes place before or after Activate().
- I don't think we have created a model during Import, we'd need to create one now.
- We'll need to convert the IncomingModelMigration userMapping into rows in the UserMapping table.

internal/jimm/juju/migrationtarget.go
```diff
@@ -173,3 +175,42 @@ func (j *JujuManager) modifyMigrationInfo(model *migration.ModelInfo, userMappin
...
+	err := j.Database.Transaction(func(db *db.Database) error {
+		err := db.GetIncomingModelMigration(ctx, &modelMigration)
+		if err != nil {
+			return errors.E(op, fmt.Errorf("failed to get model migration for model %q: %w", modelTag.Id(), err))
+		}
+		api, err := j.dialController(ctx, &modelMigration.TargetController)
+		if err != nil {
+			return errors.E(op, fmt.Errorf("failed to dial controller: %w", err))
+		}
+		defer api.Close()
```
kian99: Generally you don't hold a transaction open during an API call. That could lead to the transaction being open for a long time e.g. if there are networking issues and blocking other operations and connections to the db. Rather get the incoming model migration outside a transaction, make the API call and if it was successful, in a transaction create the user mapping entries and delete the incoming model migration row.

SimoneDutto: true!

internal/jimm/juju/migrationtarget.go
SimoneDutto: - yeah, from this method on we should rely on `Model`
- we don't have `Import` implemented yet, in a follow-up
- done

internal/jimm/juju/migrationtarget.go
```diff
@@ -173,3 +175,56 @@ func (j *JujuManager) modifyMigrationInfo(model *migration.ModelInfo, userMappin
...
+	defer api.Close()
+
+	err = api.Activate(modelTag.Id(), migrationInfo, relatedModels)
+	if err != nil {
+		return errors.E(op, fmt.Errorf("failed to activate model %q: %w", modelTag.Id(), err))
+	}
+
+	// This is done in a transaction to ensure that the model migration is only deleted
+	// if user mappings have been created.
+	err = j.Database.Transaction(func(db *db.Database) error {
```
kian99, line 207: Now this is tricky, if the call to "Activate" succeeds but our call to create a the new DB table fails, we are sending an error back to the source Juju controller saying "activate failed" but in reality it was just our DB that failed. This is unlikely but you can see the problem.

On the one hand I'm inclined to say it's so rare that we can leave it like this or simply not return an error as long as the "Activate" call succeeds.

The most correct approach, I think, would be one that is eventually consistent, in that we have a background routine that polls for the progress of active migrations and completes them (i.e. converts the user mapping, etc).

internal/jimm/juju/migrationtarget.go
kian99: On point 2,
Should we create the model during `Import` or at the point in time where we delete the `IncomingModelMigration` row? Before we delete the row we can create the model by querying Juju for the model details.

internal/jimm/juju/migrationtarget.go
ale8k: We really need distributed transactions still. Or, more over, a "call juju with try on condition" solution at the least...

internal/jimm/juju/migrationtarget.go
ale8k: I think for us creating it at the deletion of `IncomingModelMigration` makes the most sense, as we can guarantee the migration details have been deleted and model created in the same transaction. It's also worth checking when Juju adds it to their DB, but for us, I think keeping thing in transactions where we can is best such that we do not bite ourselves in the foot with potential future problems.

internal/jimm/juju/migrationtarget_test.go
```diff
@@ -304,6 +304,105 @@ func TestAdoptResources_NoIncomingModelMigration(t *testing.T) {
...
+		ModelUUID: modelMigration.ModelUUID,
+		LocalUser: "bob",
+	}
+
+	err = j.Database.GetUserMapping(ctx, &userMapping)
+	c.Assert(err, qt.IsNil)
+	c.Assert(userMapping.ExternalUserName, qt.Equals, "alice@canonical.com")
+}
+
+func TestActivate_APIFailure(t *testing.T) {
```
ale8k, line 363: What about db failure?

internal/jujuapi/migrationtarget_unit_test.go
```diff
@@ -217,3 +217,56 @@ func (s *migrationTargetUnitSuite) TestAdoptResources(c *gc.C) {
...
+	args := jujuparams.ActivateModelArgs{
+		ModelTag:        names.NewModelTag("00000001-0000-0000-0000-000000000001").String(),
+		ControllerTag:   names.NewControllerTag("00000001-0000-0000-0000-000000000002").String(),
+		ControllerAlias: "controller-1",
+		CrossModelUUIDs: []string{"related-model-1", "related-model-2"},
+	}
+	// Validate access denied without JIMM admin permissions.
+	err := cr.Activate(ctx, args)
+	c.Assert(err, gc.ErrorMatches, `unauthorized`)
+	// Validate the precheck method is called when the user is a JIMM admin.
```
ale8k: Split these into their own tests, so it's clear from test names what we're checking. This test is called `TestActivate`, but is checking 4 failure scenarios.

ale8k (approved):
LGTM, I think the distributed transaction problem isn't yours to solve in this PR. @kian99 I think let's discuss this at another date and keep this PR in mind as another of many places where JIMM is susceptible to be left in a poor place should one of the calls fail? What do you think?

internal/jimm/juju/migrationtarget.go
SimoneDutto: After a little bit of digging in the juju codebase:
- they have a field in the state for models called `migration-mode`.
- they set the migration mode to `MigrationNone` during activate
- they filter listing models based of `migration-mode` to avoid displaying migrating models.

I think we should create the model in the `Import` facade with migration-mode set to migration none.
To make sure we are mimicking the same behaviour we have in Juju.

kian99 (approved):

internal/jimm/juju/migrationtarget_test.go
SimoneDutto: i've added a test for missing migration, i'm not really sure how to test for db failure since we are not mocking the database.
Do you have any idea?

PR #1623, opened 2025-06-18
feat: proxy log transfer websocket handler

kian99:
## Description
This PR builds on #1621.

This PR implements a websocket handler at `/migrate/logtransfer` that is used during model migration to transfer the model logs to a new controller. In JIMM, we authenticate that the incoming user is a controller superuser and that a model exists for the specified model UUID. Then we connect to the target controller and proxy the websocket messages between the source and target controller.

Fixes [JUJU-8122](https://warthogs.atlassian.net/browse/JUJU-8122)

## Engineering checklist
- [ ] Documentation updated
- [x] Covered by unit tests

[JUJU-8122]: https://warthogs.atlassian.net/browse/JUJU-8122?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/jimmhttp/websocket.go
+++ b/internal/jimmhttp/websocket.go
@@ -50,8 +50,13 @@ func (h *WSHandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 		return
 	}
 
+	// Set the request context with additional values that will be used by the websocket server.
+
+	// Set the path of the request to the context for handling different endpoints.
 	ctx = context.WithValue(ctx, contextPathKey{}, req.URL.EscapedPath())
+	// Set the migrating model UUID if it exists in the request header.
 	ctx = context.WithValue(ctx, migratingModelUUIDKey{}, req.Header.Get(jujuparams.MigrationModelHTTPHeader))
+	// Set the client version from the request header, expected by Juju controllers.
 	ctx = context.WithValue(ctx, clientVersionKey{}, req.Header.Get(jujuparams.JujuClientVersion))
 
 	conn, err := h.Upgrader.Upgrade(w, req, nil)
--- a/internal/jimmhttp/websocket.go
+++ b/internal/jimmhttp/websocket.go
@@ -1,4 +1,4 @@
-// Copyright 2024 Canonical.
+// Copyright 2025 Canonical.
 
 package jimmhttp
 
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	"github.com/gorilla/websocket"
+	jujuparams "github.com/juju/juju/rpc/params"
 	"github.com/juju/zaputil/zapctx"
 	"go.uber.org/zap"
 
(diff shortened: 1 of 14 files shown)


SimoneDutto (approved):
LGTM

internal/jujuapi/streamproxy.go
```diff
@@ -45,7 +45,7 @@ func (s streamProxier) Authenticate(ctx context.Context, w http.ResponseWriter,
 }
 
 // ServeWS implements jimmhttp.WSServer.
-func (s streamProxier) ServeWS(ctx context.Context, clientConn *websocket.Conn) {
```
ale8k, line 48: Whilst you're here, let's update this godoc. We're clear on controller proxying, but not on model. Just explain it does path -> path.

internal/jimmhttp/websocket.go
```diff
@@ -49,7 +50,10 @@ func (h *WSHandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 		return
 	}
 
-	ctx = context.WithValue(ctx, contextPathKey("path"), req.URL.EscapedPath())
+	ctx = context.WithValue(ctx, contextPathKey{}, req.URL.EscapedPath())
+	ctx = context.WithValue(ctx, migratingModelUUIDKey{}, req.Header.Get(jujuparams.MigrationModelHTTPHeader))
```
ale8k, line 58: Godoc when these are set and how. They look random otherwise? I'd ever say let us have another func called "populateCtx" or something which does this, and then that func is godoc'd with what context values and how/when.

internal/jujuapi/api.go
```diff
@@ -38,15 +38,24 @@ func ModelHandler(ctx context.Context, jimm *jimm.JIMM, p Params) http.Handler {
...
 		Upgrader: websocketUpgrader,
-		Server: &streamProxier{apiServer: apiServer{
+		Server: &streamModelProxier{apiServer: apiServer{
 			jimm: jimm,
 		}},
 	})
 	return mux
 }
+
+func LogTransferHandler(ctx context.Context, jimm *jimm.JIMM, p Params) http.Handler {
```
ale8k, line 55: Godoc

internal/jujuapi/streamcontrollerproxy.go
```diff
@@ -0,0 +1,120 @@
...
+	logTransferPath = "/migrate/logtransfer"
+)
+
+// streamControllerProxier serves any HTTP endpoints
+// served against a controller's root address rather than
+// a model specific path segment i.e. /model/{uuid}/*.
+// Messages are handled by proxying them between the
+// controller and client.
+type streamControllerProxier struct {
+	apiServer
```
ale8k: We're seemingly only using jimm on the API server here, and we've implemented ServeWS, do we need an apiServer or just a JIMM?

internal/jujuclient/dial.go
```diff
@@ -362,6 +362,25 @@ func (c *Connection) ConnectStream(path string, attrs url.Values) (base.Stream,
 // values are used as URL query values when making the initial
 // HTTP request. Headers passed in will be added to the HTTP
 // request.
-func (c *Connection) ConnectControllerStream(path string, attrs url.Values, headers http.Header) (base.Stream, error) {
-	return nil, errors.E(errors.CodeNotImplemented)
+func (c *Connection) ConnectControllerStream(path string, attrs url.Values, extraHeaders http.Header) (base.Stream, error) {
+	const op = errors.Op("jujuclient.ConnectControllerStream")
+
+	user, pass, err := c.dialer.ControllerCredentialsStore.GetControllerCredentials(c.ctx, c.ctl.Name)
```
ale8k, line 368: Because we're getting controller credentials here, and it's sensitive, this might be a candidate for an auth check? (If it's hassle don't bother)

ale8k (approved):
Approving but please do not use an apiServer on the controller stream unless we need it? And furthermore, would be nice to just pass the services it needs rather than all of JIMM. So possibly do that too please.

PR #1624, opened 2025-06-20
feat: add multi-backend transport implementation

kian99:
## Description
This PR builds on #1621.
This PR introduces a `multiBackendTransport` struct that implements the `http.RoundTripper` interface. It attempts to forward a request to multiple backends in succession with logic to fail early if a request fails partway through. Additionally, the transport will also do round-robin selection of the next backend if the struct is re-used.

I tried to come up with many failure scenarios to cover all the important cases e.g. 2 backends and the first is unreachable, 2 backends and the first fails partway through, etc. Let me know if you can think of anything that is missing.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

### DIFF
--- a/internal/rpc/httpproxy.go
+++ b/internal/rpc/httpproxy.go
@@ -8,7 +8,7 @@ import (
 	"crypto/x509"
 	"fmt"
 	"log"
-	"math/rand"
+	"math/rand/v2"
 	"net/http"
 	"net/http/httputil"
 	"net/url"
@@ -43,6 +43,13 @@ func ProxyHTTP(ctx context.Context, ctl juju.ControllerConnectionDetails, w http
 		return
 	}
 
+	// Shuffle for initial load balancing
+	if len(urls) > 1 {
+		rand.Shuffle(len(urls), func(i, j int) {
+			urls[i], urls[j] = urls[j], urls[i]
+		})
+	}
+
 	var tlsConfig *tls.Config
 	if ctl.CACertificate != "" {
 		cp := x509.NewCertPool()
@@ -57,25 +64,25 @@ func ProxyHTTP(ctx context.Context, ctl juju.ControllerConnectionDetails, w http
 		}
 	}
 
-	transport := http.DefaultTransport.(*http.Transport).Clone()
-	transport.TLSClientConfig = tlsConfig
+	baseTransport := http.DefaultTransport.(*http.Transport).Clone()
+	baseTransport.TLSClientConfig = tlsConfig
 
-	if len(urls) > 1 {
-		rand.Shuffle(len(urls), func(i, j int) {
-			urls[i], urls[j] = urls[j], urls[i]
-		})
+	// Create custom transport that handles multiple backends
+	multiBackendTransport, err := newMultiBackendTransport(baseTransport, urls)
+	if err != nil {
+		zapctx.Error(ctx, "failed to create multiBackendTransport", zap.Error(err))
+		http.Error(w, fmt.Sprintf("failed to create multiBackendTransport: %v", err), http.StatusInternalServerError)
+		return
 	}
 
-	// TODO: Consider implementing a better load balancing mechanism that handles
-	// multiples URLs and handles failing backends gracefully e.g. try send to first
-	// URL and on failure, try second, etc.
 	adminUsername := names.NewUserTag(ctl.Credentials.AdminIdentityName).String()
 	proxy := &httputil.ReverseProxy{
 		Rewrite: func(pr *httputil.ProxyRequest) {
-			pr.SetURL(urls[0])
+			// The multiBackendTransport will handle URL selection and failover
+			// We just need to set up basic auth here
 			pr.Out.SetBasicAuth(adminUsername, ctl.Credentials.AdminPassword)
 		},
-		Transport: transport,
+		Transport: multiBackendTransport,
 		ErrorLog:  log.New(&proxyErrorLogger{}, "", 0), // flag=0 to avoid printing extra info that zap already gives us
 	}
 	proxy.ServeHTTP(w, req)
--- a/internal/rpc/multitransport.go
+++ b/internal/rpc/multitransport.go
@@ -1,0 +1,98 @@
+// Copyright 2025 Canonical.
+
+package rpc
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+
+	"github.com/juju/zaputil/zapctx"
+	"go.uber.org/zap"
+)
+
+// multiBackendTransport is a custom HTTP transport that attempts to make requests
+// to multiple backend URLs in succession until one succeeds. It will not
+// continue to retry if a request fails partway through (e.g., if the body
+// was partially read), as we don't store the original request body.
+//
+// Reusing a multiBackendTransport object will ensure that requests
+// are made in a round-robin fashion across the provided URLs.
+//
+// If a new multiBackendTransport is created per request, it will always start
+// from the first URL, so the caller should randomise the list of URLs first.
+type multiBackendTransport struct {
+	// baseTransport is the underlying transport used for actual HTTP requests
+	baseTransport http.RoundTripper
+	// urls contains the list of backend urls to try in order
+	urls []*url.URL
+	// currentURLIndex tracks which URL should be tried next (for round-robin)
+	currentURLIndex int
+}
+
+// RoundTrip implements the http.RoundTripper interface and attempts to make
+// requests to multiple backends in succession.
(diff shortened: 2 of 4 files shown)


internal/rpc/multitransport.go
```diff
@@ -0,0 +1,99 @@
...
+	// urls contains the list of backend urls to try in order
+	urls []*url.URL
+	// currentURLIndex tracks which URL should be tried next (for round-robin)
+	currentURLIndex int
+}
+
+// RoundTrip implements the http.RoundTripper interface and attempts to make
+// requests to multiple backends in succession.
+func (m *multiBackendTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	if len(m.urls) == 0 {
```
alesstimec: perhaps a constructor where we can check that `urls` are not empty?

internal/rpc/httpproxy.go
```diff
@@ -7,98 +7,134 @@ import (
...
+		return
+	}
+
+	if len(urls) == 0 {
+		zapctx.Error(ctx, "no controller addresses found", zap.String("controller", ctl.Controller.Name))
+		http.Error(w, "no controller addresses found", http.StatusInternalServerError)
+		return
+	}
+
+	// Shuffle for initial load balancing
```
alesstimec, line 46: do we need to shuffle if we're doing round-robin anyways?

internal/rpc/multitransport.go
kian99: Moved the check to the constructor.

internal/rpc/httpproxy.go
kian99: It's only round-robin if you keep the multiTransport object around. Currently we recreate it for each HTTP proxy request coming in because technically the controller's addresses can be changing dynamically (due to HA). To make this smarter and do actual round-robin we would need a way to update the object's list of addresses whenever we notice a controller's addresses have changed.

internal/rpc/multitransport.go
```diff
@@ -0,0 +1,98 @@
...
+		lastErr = err
+	}
+
+	// All backends failed
+	return nil, fmt.Errorf("all backends failed, last error: %w", lastErr)
+}
+
+// newMultiBackendTransport creates a new MultiBackendTransport with the given
+// base transport and URLs.
+func newMultiBackendTransport(baseTransport http.RoundTripper, urls []*url.URL) (*multiBackendTransport, error) {
```
ale8k: Please keep constructors at the top, but below the structure definition.

internal/rpc/httpproxy.go
```diff
@@ -57,25 +64,25 @@ func ProxyHTTP(ctx context.Context, ctl juju.ControllerConnectionDetails, w http
...
-	transport.TLSClientConfig = tlsConfig
+	baseTransport := http.DefaultTransport.(*http.Transport).Clone()
+	baseTransport.TLSClientConfig = tlsConfig
 
-	if len(urls) > 1 {
-		rand.Shuffle(len(urls), func(i, j int) {
-			urls[i], urls[j] = urls[j], urls[i]
-		})
+	// Create custom transport that handles multiple backends
+	multiBackendTransport, err := newMultiBackendTransport(baseTransport, urls)
```
ale8k, line 71: Do we need a new instance for each request? We could have a very specific transport which manages the retrieval of controller addresses from inside the RoundTrip, that is, based on URL X, we look up controller Y, find addresses Z, and let the logic run. Then, we don't need to construct a new transport and tls config each time. I see this as equivalent to using a shared HTTP client, you wouldn't want to configure it over and over. Also, doing it this way could move the shuffle logic into the RoundTrip.

It also doesn't need a full base transport. It just needs a TLS config to construct it and the constructor can clone the default transport and update the config right? We're doing it in http proxy which doesn't care about that logic.

internal/rpc/multitransport.go
```diff
@@ -0,0 +1,98 @@
...
+	currentURLIndex int
+}
+
+// RoundTrip implements the http.RoundTripper interface and attempts to make
+// requests to multiple backends in succession.
+func (m *multiBackendTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	var lastErr error
+
+	// Try each URL in succession
+	for i := 0; i < len(m.urls); i++ {
```
ale8k, line 54: Can't you do

```go
for _, url := range m.urls
```

Then you won't need line 45?

internal/rpc/multitransport.go
```diff
@@ -0,0 +1,98 @@
...
+		// Calculate the URL index to try (round-robin starting from currentURLIndex)
```
ale8k, line 55: I'd put this logic in a separate method on the transport with a nice godoc and name to clearly separate the sequential roundtripper from the cloning and target updating of a request.

internal/rpc/multitransport_test.go
```diff
@@ -0,0 +1,203 @@
...
+			urls: []*url.URL{
+				partialReadServerURL,
+				successServerURL,
+			},
+			shouldError: true,
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.description, func(t *testing.T) {
```
ale8k: c.Run? And then you won't need line 121?...

internal/rpc/httpproxy.go
kian99: Going to the DB in the transport is one way to handle it. But I wouldn't want to set out and design that now. Currently the safest thing to do is create a new instance for each request, easy enough too since we need to fetch the model+controller from the DB anyway.

internal/rpc/multitransport.go
kian99: I think keeping it as is makes it easier to understand everything that it's doing, which is fairly contained and straightforward. So I'd prefer to keep it like this.

internal/rpc/multitransport.go
kian99: You need the index for the round-robin logic. Which is why we don't range over the urls.

internal/rpc/multitransport_test.go
kian99: Thanks for that, done.

internal/rpc/multitransport.go
```diff
@@ -0,0 +1,98 @@
...
+			zap.String("backend", targetURL.String()),
+			zap.Error(err),
+			zap.Int("attempt", i+1),
+			zap.Int("total_backends", len(m.urls)))
+
+		lastErr = err
+	}
+
+	// All backends failed
+	return nil, fmt.Errorf("all backends failed, last error: %w", lastErr)
```
SimoneDutto, line 97: why dont' we `errors.Join`?

internal/rpc/multitransport.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/juju/zaputil/zapctx"
+	"go.uber.org/zap"
+)
+
+// multiBackendTransport is a custom HTTP transport that attempts to make requests
+// to multiple backend URLs in succession until one succeeds. It will not
+// continue to retry if a request fails partway through (e.g., if the body
+// was partially read), as we don't store the original request body.
+//
+// Reusing a multiBackendTransport object will ensure that requests
```
SimoneDutto, line 22: to make the code simpler you could just drop this feature since we don't have a use case for it just yet.
But feel free to leave it since it's tested.

internal/rpc/httpproxy.go
```diff
@@ -43,6 +43,13 @@ func ProxyHTTP(ctx context.Context, ctl juju.ControllerConnectionDetails, w http
 		return
 	}
 
+	// Shuffle for initial load balancing
+	if len(urls) > 1 {
+		rand.Shuffle(len(urls), func(i, j int) {
+			urls[i], urls[j] = urls[j], urls[i]
+		})
+	}
```
SimoneDutto, line 51: what about shuffling them in the `newMultiBackendTransport`?

SimoneDutto (approved):
lgtm

internal/rpc/httpproxy.go
kian99: I was initially doing that but it makes it harder to test the round robin so I decided to move it out.

internal/rpc/multitransport.go
kian99: Could do that, I'm inclined to leave it like this just purely because I think it's not too much of a different.

internal/rpc/httpproxy.go
ale8k: Yeah fair enough, a good follow up would be to single instance this.

ale8k (approved):
LGTM, I do think we should follow up and not have an instance per req of the transport though.

PR #1625, opened 2025-06-23
feat: implement advisory (none table linked) locks for jimm

ale8k:
Implements advisory locks in JIMM. Has a single const id for bootstrap lock.

See [locks](https://www.postgresql.org/docs/current/explicit-locking.html#There%20are%20two%20ways%20to%20acquire%20an%20advisory%20lock) for usage.

## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests

### DIFF
--- a/internal/db/advisory_locks.go
+++ b/internal/db/advisory_locks.go
@@ -8,7 +8,7 @@ import (
 	"fmt"
 )
 
-// LockDI represents a unique identifier for an advisory lock in the database.
+// LockID represents a unique identifier for an advisory lock in the database.
 type LockID int64
 
 const (
@@ -16,7 +16,7 @@ const (
 	ControllerBootstrapLock LockID = iota + 1
 )
 
-// LockAdvisory attempts to acquire an advisory lock with the given ID.
+// lockAdvisory attempts to acquire an advisory lock with the given ID.
 // It returns an error if the lock is already held by another session.
 func (d *Database) lockAdvisory(ctx context.Context, id LockID) error {
 	var success bool
@@ -31,7 +31,7 @@ func (d *Database) lockAdvisory(ctx context.Context, id LockID) error {
 	return nil
 }
 
-// UnlockAdvisory releases an advisory lock with the given ID.
+// unlockAdvisory releases an advisory lock with the given ID.
 // It returns an error if the lock was not held or could not be released.
 func (d *Database) unlockAdvisory(ctx context.Context, id LockID) error {
 	var released bool
--- a/internal/db/advisory_locks.go
+++ b/internal/db/advisory_locks.go
@@ -18,7 +18,7 @@ const (
 
 // LockAdvisory attempts to acquire an advisory lock with the given ID.
 // It returns an error if the lock is already held by another session.
-func (d *Database) LockAdvisory(ctx context.Context, id LockID) error {
+func (d *Database) lockAdvisory(ctx context.Context, id LockID) error {
 	var success bool
 	err := d.DB.WithContext(ctx).Raw("SELECT pg_try_advisory_lock(?)", id).Scan(&success).Error
 	if err != nil {
@@ -33,7 +33,7 @@ func (d *Database) LockAdvisory(ctx context.Context, id LockID) error {
 
 // UnlockAdvisory releases an advisory lock with the given ID.
 // It returns an error if the lock was not held or could not be released.
-func (d *Database) UnlockAdvisory(ctx context.Context, id LockID) error {
+func (d *Database) unlockAdvisory(ctx context.Context, id LockID) error {
 	var released bool
 
 	err := d.DB.WithContext(ctx).Raw("SELECT pg_advisory_unlock(?)", id).Scan(&released).Error
@@ -46,4 +46,16 @@ func (d *Database) UnlockAdvisory(ctx context.Context, id LockID) error {
 	}
 
 	return nil
+}
+
+// LockBootstrap acquires the advisory lock for controller bootstrap operations.
+// It returns an error if the lock cannot be acquired.
+func (d *Database) LockBootstrap(ctx context.Context) error {
+	return d.lockAdvisory(ctx, ControllerBootstrapLock)
+}
+
+// UnlockBootstrap releases the advisory lock for controller bootstrap operations.
+// It returns an error if the lock was not held or could not be released.
+func (d *Database) UnlockBootstrap(ctx context.Context) error {
+	return d.unlockAdvisory(ctx, ControllerBootstrapLock)
 }
--- a/internal/db/advisory_locks.go
+++ b/internal/db/advisory_locks.go
@@ -13,7 +13,7 @@ type LockID int64
 
 const (
 	// ControllerBootstrapLock is the advisory lock ID used for controller bootstrap operations.
-	ControllerBootstrapLock LockID = 1001
+	ControllerBootstrapLock LockID = iota + 1
 )
 
 // LockAdvisory attempts to acquire an advisory lock with the given ID.
--- a/internal/db/advisory_locks.go
+++ b/internal/db/advisory_locks.go
@@ -1,0 +1,45 @@
+// Copyright 2025 Canonical.
+
+package db
+
+import (
+	"context"
+	"errors"
+	"fmt"
+)
+
+// LockDI represents a unique identifier for an advisory lock in the database.
+type LockID int64
+
+const (
+	// ControllerBootstrapLock is the advisory lock ID used for controller bootstrap operations.
+	ControllerBootstrapLock LockID = 1001
+)
+
+func (d *Database) LockAdvisory(ctx context.Context, id LockID) error {
+	var success bool
+	err := d.DB.WithContext(ctx).Raw("SELECT pg_try_advisory_lock(?)", id).Scan(&success).Error
+	if err != nil {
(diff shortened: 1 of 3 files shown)


internal/db/advisory_locks.go
```diff
@@ -0,0 +1,45 @@
...
+
+// LockDI represents a unique identifier for an advisory lock in the database.
+type LockID int64
+
+const (
+	// ControllerBootstrapLock is the advisory lock ID used for controller bootstrap operations.
+	ControllerBootstrapLock LockID = 1001
+)
+
+func (d *Database) LockAdvisory(ctx context.Context, id LockID) error {
```
alesstimec: godoc, please

internal/db/advisory_locks.go
```diff
@@ -0,0 +1,45 @@
...
+		return fmt.Errorf("error acquiring advisory lock: %w", err)
+	}
+	if !success {
+		return errors.New("lock is already held")
+	}
+
+	return nil
+}
+
+func (d *Database) UnlockAdvisory(ctx context.Context, id LockID) error {
```
alesstimec: godoc please

internal/db/advisory_locks_test.go
```diff
@@ -0,0 +1,63 @@
...
+	"testing"
+	"time"
+
+	qt "github.com/frankban/quicktest"
+	"github.com/frankban/quicktest/qtsuite"
+	"gorm.io/driver/postgres"
+	"gorm.io/gorm"
+
+	"github.com/canonical/jimm/v3/internal/db"
+	jimmdb "github.com/canonical/jimm/v3/internal/db"
```
alesstimec: why is this package imported as "db" (L14) and "jimmdb" (L15)?

internal/db/advisory_locks_test.go
```diff
@@ -0,0 +1,63 @@
...
+
+	db1 := &db.Database{
+		DB: gdb1,
+	}
+
+	db2 := &db.Database{
+		DB: gdb2,
+	}
+
+	// Acquite lock in db session 1.
```
alesstimec: Acquire?

internal/db/advisory_locks.go
```diff
@@ -0,0 +1,45 @@
...
+	"errors"
+	"fmt"
+)
+
+// LockDI represents a unique identifier for an advisory lock in the database.
+type LockID int64
+
+const (
+	// ControllerBootstrapLock is the advisory lock ID used for controller bootstrap operations.
+	ControllerBootstrapLock LockID = 1001
```
kian99: You could use an iota here, something like the 
```suggestion
	ControllerBootstrapLock LockID = iota + 1
```

internal/db/advisory_locks.go
kian99: What do you think about not exporting the lock IDs and adding methods like `LockBootstrap` and `UnlockBootstrap` instead?

internal/db/advisory_locks_test.go
ale8k: Ah I originally named the variable db, then called them 1 and 2. Will fix.

internal/db/advisory_locks_test.go
ale8k: Fixed.

internal/db/advisory_locks_test.go
```diff
@@ -0,0 +1,62 @@
...
+	"gorm.io/driver/postgres"
+	"gorm.io/gorm"
+
+	"github.com/canonical/jimm/v3/internal/db"
+	"github.com/canonical/jimm/v3/internal/testutils/jimmtest"
+)
+
+type advisoryLocksSuite struct{}
+
+func (s *advisoryLocksSuite) TestAdvisory_LockAndUnlock(c *qt.C) {
```
kian99, line 29: This test might be overcomplicating things. I think you can look at the migration code which does,
```go
	sqlDB, err := db.DB()
	if err != nil {
		return fmt.Errorf("failed to obtain raw DB: %w", err)
	}
	conn, err := sqlDB.Conn(ctx)
	if err != nil {
		return fmt.Errorf("failed to obtain DB conn: %w", err)
	}
```
to get a new DB connection, then you won't need to modify jimmtest to get 2 separate DB connections

internal/db/advisory_locks.go
ale8k: Sure, added.

internal/db/advisory_locks.go
ale8k: Possibly less error prone? But it's a case of wrong const vs wrong method used, aren't both equally risky?

internal/db/advisory_locks_test.go
ale8k: How do I get from conn -> gorm.DB?

internal/db/advisory_locks.go
kian99: Not too concerned about making it less error-prone, but it feels easier to read
```go
db.LockBootstrap(ctx)
```
vs
```go
db.LockAdvisory(ctx, db.ControllerBootstrapLock)
```

ale8k: Yeah I don't mind, I'll update it now

internal/db/advisory_locks_test.go
```diff
@@ -0,0 +1,65 @@
...
+		DB: gdb2,
+	}
+
+	// Acquire lock in db session 1.
+	err = db1.LockBootstrap(ctx)
+	c.Assert(err, qt.IsNil, qt.Commentf("Failed to acquire lock"))
+
+	// Attempt to acquire lock in session 2, should fail.
+	err = db2.LockBootstrap(ctx)
+	c.Assert(
```
kian99, line 51: Feel free to ignore this one, but without a restriction on line length in jimm, I wouldn't mind if you kept this as a single line to match the other assertions.

internal/db/advisory_locks_test.go
```diff
@@ -0,0 +1,65 @@
...
+	qt "github.com/frankban/quicktest"
+	"github.com/frankban/quicktest/qtsuite"
+	"gorm.io/driver/postgres"
+	"gorm.io/gorm"
+
+	"github.com/canonical/jimm/v3/internal/db"
+	"github.com/canonical/jimm/v3/internal/testutils/jimmtest"
+)
+
+type advisoryLocksSuite struct{}
```
kian99: Now that you are not doing anything custom, you could switch this test to use the `dbSuite` like many of the other tests in the `db` package.

kian99 (approved):

internal/db/advisory_locks_test.go
ale8k: Ah it's more a habit and preference I tend to do. If you don't mind I'd prefer to leave it?

internal/db/advisory_locks_test.go
ale8k: Agreed! Updated

internal/db/advisory_locks.go
```diff
@@ -0,0 +1,61 @@
...
+	"context"
+	"errors"
+	"fmt"
+)
+
+// LockID represents a unique identifier for an advisory lock in the database.
+type LockID int64
+
+const (
+	// ControllerBootstrapLock is the advisory lock ID used for controller bootstrap operations.
```
alesstimec, line 15: perhaps a godoc that we should NOT EVER change this ID..

internal/db/advisory_locks.go
```diff
@@ -0,0 +1,61 @@
...
+package db
+
+import (
+	"context"
+	"errors"
+	"fmt"
+)
+
+// LockID represents a unique identifier for an advisory lock in the database.
+type LockID int64
```
alesstimec, line 12: this does not need to be exported as it is not used by any of the exported methods.

internal/db/advisory_locks.go
```diff
@@ -0,0 +1,61 @@
...
+
+const (
+	// ControllerBootstrapLock is the advisory lock ID used for controller bootstrap operations.
+	ControllerBootstrapLock LockID = iota + 1
```
alesstimec, line 16: this also does not need to be exported

internal/db/advisory_locks_test.go
```diff
@@ -0,0 +1,68 @@
...
+		DB: jimmtest.PostgresDB(c, time.Now),
+	}
+	s.Database = db
+}
+
+func (s *advisoryLocksSuite) TestAdvisory_LockAndUnlock(c *qt.C) {
+	ctx := c.Context()
+
+	sqldb, err := s.Database.DB.DB()
+	c.Assert(err, qt.IsNil, qt.Commentf("Failed to get SQL DB connection"))
```
alesstimec, line 33: if you are going to put `qt.Commentf` everywhere AND capitalize the first letter,  could we at least make it a full sentence and end with a full stop (".")? here and everywhere else?

alesstimec (approved):
LGTM modulo a few comments

internal/db/advisory_locks.go
ale8k: Fs I just merged and didn't see these comments. I'll add this in this one https://github.com/canonical/jimm/pull/1627/ if that's ok?

internal/db/advisory_locks.go
ale8k: As comment above, I'll add to other PR. Sorry.

PR #1627, opened 2025-06-24
feat: add job tracker db model & methods

ale8k:
## Description
Adds the db model and methods for the job tracker.

I made slight adjustment away from the spec to the table name to leverage GORMs automatic pluralisation of entry/entries. 

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

## Test instructions
N/A

### DIFF
--- a/internal/db/job_tracker.go
+++ b/internal/db/job_tracker.go
@@ -1,0 +1,137 @@
+// Copyright 2025 Canonical.
+
+package db
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/google/uuid"
+	"github.com/juju/zaputil/zapctx"
+	"go.uber.org/zap"
+
+	"github.com/canonical/jimm/v3/internal/dbmodel"
+	"github.com/canonical/jimm/v3/internal/errors"
+	"github.com/canonical/jimm/v3/internal/servermon"
+)
+
+// AddJob adds a new job to the job tracker.
+// It returns the job ID and an error if the job already exists or if the creation fails
+func (d *Database) AddJob(ctx context.Context, jobType string) (jobId uuid.UUID, err error) {
+	const op = errors.Op("db.AddJob")
+	if err := d.ready(); err != nil {
+		return jobId, errors.E(op, err)
+	}
+
+	durationObserver := servermon.DurationObserver(servermon.DBQueryDurationHistogram, string(op))
+	defer durationObserver()
+	defer servermon.ErrorCounter(servermon.DBQueryErrorCount, &err, string(op))
+
+	db := d.DB.WithContext(ctx)
+
+	entry := dbmodel.NewJobTrackerEntry(jobType)
+	if err := db.Create(entry).Error; err != nil {
+		err := dbError(err)
+		if errors.ErrorCode(err) == errors.CodeAlreadyExists {
+			zapctx.Debug(ctx, "job already exists", zap.String("jobID", entry.JobID.String()))
+			return jobId, errors.E(op, fmt.Sprintf("job %s already exists", entry.JobID), err)
+		}
+		return jobId, errors.E(op, err)
+	}
+
+	jobId = entry.JobID
+	return jobId, nil
+}
+
+// StopJob marks a job as stopped.
+// It returns an error if the job does not exist or if the update fails.
+func (d *Database) StopJob(ctx context.Context, jobId uuid.UUID) (err error) {
+	const op = errors.Op("db.StopJob")
+	if err := d.ready(); err != nil {
+		return errors.E(op, err)
+	}
+
+	durationObserver := servermon.DurationObserver(servermon.DBQueryDurationHistogram, string(op))
+	defer durationObserver()
+	defer servermon.ErrorCounter(servermon.DBQueryErrorCount, &err, string(op))
+
+	db := d.DB.WithContext(ctx)
+	if err := db.Model(&dbmodel.JobTrackerEntry{}).Where("job_id = ?", jobId).Update("stop", true).Error; err != nil {
+		return dbError(err)
+	}
+
+	return nil
+}
+
+// SetJobSuccessful sets the job status to successful.
+// It returns an error if the job does not exist or if the update fails.
+func (d *Database) SetJobRunning(ctx context.Context, jobId uuid.UUID) (err error) {
+	entry := dbmodel.JobTrackerEntry{
+		JobID: jobId,
+	}
+	entry.SetRunning()
+	return d.updateJobStatus(ctx, entry)
+}
+
+// SetJobSuccessful sets the job status to successful.
+// It returns an error if the job does not exist or if the update fails.
+func (d *Database) SetJobSuccessful(ctx context.Context, jobId uuid.UUID) (err error) {
+	entry := dbmodel.JobTrackerEntry{
+		JobID: jobId,
+	}
+	entry.SetSuccessful()
+	return d.updateJobStatus(ctx, entry)
+}
+
+// SetJobFailed sets the job status to failed and records the error message.
+// It returns an error if the job does not exist or if the update fails.
+func (d *Database) SetJobFailed(ctx context.Context, jobId uuid.UUID, jobErr error) (err error) {
+	const op = errors.Op("db.SetJobFailed")
+	entry := dbmodel.JobTrackerEntry{
+		JobID: jobId,
+	}
+	if err := entry.SetFailed(jobErr); err != nil {
+		return errors.E(op, err)
+	}
+
+	return d.updateJobStatus(ctx, entry)
+}
+
+func (d *Database) updateJobStatus(ctx context.Context, entry dbmodel.JobTrackerEntry) (err error) {
+	const op = errors.Op("db.updateJobStatus")
+	if err := d.ready(); err != nil {
+		return errors.E(op, err)
+	}
+
+	durationObserver := servermon.DurationObserver(servermon.DBQueryDurationHistogram, string(op))
+	defer durationObserver()
+	defer servermon.ErrorCounter(servermon.DBQueryErrorCount, &err, string(op))
+
+	db := d.DB.WithContext(ctx)
+	if err := db.Model(&entry).Select("status", "error").Updates(entry).Error; err != nil {
+		return dbError(err)
+	}
+
+	return nil
+}
+
+// GetJobStatus retrieves the status of a job by its ID.
+// It returns the job status and an error if the job does not exist or if the query fails.
+func (d *Database) GetJobStatus(ctx context.Context, jobId uuid.UUID) (status dbmodel.JobStatus, err error) {
+	const op = errors.Op("db.GetJobStatus")
+	if err := d.ready(); err != nil {
+		return status, errors.E(op, err)
+	}
+
+	durationObserver := servermon.DurationObserver(servermon.DBQueryDurationHistogram, string(op))
+	defer durationObserver()
+	defer servermon.ErrorCounter(servermon.DBQueryErrorCount, &err, string(op))
+
+	db := d.DB.WithContext(ctx)
+	entry := &dbmodel.JobTrackerEntry{}
+	if err := db.First(entry, "job_id = ?", jobId).Error; err != nil {
+		return status, dbError(err)
+	}
+
+	return entry.GetStatus(), nil
+}
--- a/internal/db/job_tracker.go
+++ b/internal/db/job_tracker.go
@@ -29,7 +29,10 @@ func (d *Database) AddJob(ctx context.Context, jobType string) (jobId uuid.UUID,
 
 	db := d.DB.WithContext(ctx)
 
-	entry := dbmodel.NewJobTrackerEntry(jobType)
+	entry, err := dbmodel.NewJobTrackerEntry(jobType)
+	if err != nil {
+		return jobId, errors.E(op, fmt.Sprintf("failed to create new job tracker entry: %v", err))
+	}
 	if err := db.Create(entry).Error; err != nil {
 		err := dbError(err)
 		if errors.ErrorCode(err) == errors.CodeAlreadyExists {
@@ -56,8 +59,14 @@ func (d *Database) StopJob(ctx context.Context, jobId uuid.UUID) (err error) {
 	defer servermon.ErrorCounter(servermon.DBQueryErrorCount, &err, string(op))
 
 	db := d.DB.WithContext(ctx)
-	if err := db.Model(&dbmodel.JobTrackerEntry{}).Where("job_id = ?", jobId).Update("stop", true).Error; err != nil {
+
+	result := db.Model(&dbmodel.JobTrackerEntry{}).Where("job_id = ?", jobId).Update("stop_signal", true)
+	if err := result.Error; err != nil {
 		return dbError(err)
+	}
+
+	if result.RowsAffected == 0 {
+		return errors.E(op, errors.CodeNotFound, fmt.Sprintf("job %s not found", jobId))
 	}
 
 	return nil
@@ -70,7 +79,7 @@ func (d *Database) SetJobRunning(ctx context.Context, jobId uuid.UUID) (err erro
 		JobID: jobId,
 	}
 	entry.SetRunning()
-	return d.updateJobStatus(ctx, entry)
+	return d.updateJob(ctx, entry)
 }
 
 // SetJobSuccessful sets the job status to successful.
@@ -80,7 +89,7 @@ func (d *Database) SetJobSuccessful(ctx context.Context, jobId uuid.UUID) (err e
 		JobID: jobId,
 	}
 	entry.SetSuccessful()
-	return d.updateJobStatus(ctx, entry)
+	return d.updateJob(ctx, entry)
 }
 
 // SetJobFailed sets the job status to failed and records the error message.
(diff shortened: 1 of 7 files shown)


internal/dbmodel/job_tracker.go
```diff
@@ -0,0 +1,87 @@
...
+	StatusFailed     JobStatus = "failed"
+)
+
+// JobTrackerEntry represents a job that is being tracked within an instance of JIMM.
+type JobTrackerEntry struct {
+	// JobID is the unique identifier for the job. This is not to be set manually, please
+	// use the constructor [NewJobTrackerEntry] instead.
+	JobID uuid.UUID `gorm:"type:uuid;primaryKey"`
+	// JobType is the type of the job, e.g., bootstrap, destroy, etc.
+	JobType string `gorm:"type:varchar(128);not null"`
```
kian99, line 32: I wouldn't add the DB types here because we don't use gorm to generate our migrations.

internal/db/job_tracker.go
```diff
@@ -0,0 +1,137 @@
...
+	if err := d.ready(); err != nil {
+		return errors.E(op, err)
+	}
+
+	durationObserver := servermon.DurationObserver(servermon.DBQueryDurationHistogram, string(op))
+	defer durationObserver()
+	defer servermon.ErrorCounter(servermon.DBQueryErrorCount, &err, string(op))
+
+	db := d.DB.WithContext(ctx)
+	if err := db.Model(&entry).Select("status", "error").Updates(entry).Error; err != nil {
```
kian99: Why do we need to do a select and then update?
Would just an update suffice?

internal/db/job_tracker.go
```diff
@@ -0,0 +1,137 @@
...
+	return d.updateJobStatus(ctx, entry)
+}
+
+// SetJobSuccessful sets the job status to successful.
+// It returns an error if the job does not exist or if the update fails.
+func (d *Database) SetJobSuccessful(ctx context.Context, jobId uuid.UUID) (err error) {
+	entry := dbmodel.JobTrackerEntry{
+		JobID: jobId,
+	}
+	entry.SetSuccessful()
```
kian99, line 91: These methods for setting successful/running/failed are exported in the `dbmodel` package so that they can be used here. But presumably they wouldn't be used elsewhere?
What about changing the signature for `updateJobStatus` to `(ctx context.Context, jobID string, status JobStatus)` then you don't need the exported methods on the JobTrackerEntry.

internal/dbmodel/sql/postgres/026_job_tracker.up.sql
```diff
@@ -0,0 +1,20 @@
...
+CREATE TABLE IF NOT EXISTS job_tracker_entries (
+    job_id UUID PRIMARY KEY,
+    job_type VARCHAR(128) NOT NULL,
+    stop BOOLEAN NOT NULL DEFAULT FALSE,
+    status job_tracker_status NOT NULL DEFAULT 'pending',
+    error TEXT,
+
+    created_at TIMESTAMP,
+    updated_at TIMESTAMP,
+    deleted_at TIMESTAMP
```
kian99: Should we get rid of this to avoid soft-deletes?

internal/dbmodel/job_tracker.go
ale8k: They're ignored sure, but it's nice to see within the model.  It also means should we ever go and use GORM for SQL generation, we're safe. Which I think is a good thing.

internal/db/job_tracker.go
ale8k: It's the nature of updates, if a value is zeroed on the model, and is nullable and has no default in the schema, then GORM ignores it during updates. So, specifically selecting them prior to the update fixes this. You can also do a where/update, but this is clearer to me.

internal/db/job_tracker.go
ale8k: I like having explicit methods, prevents developer error of writing just any old string and then encountering a DB error despite no error.

internal/dbmodel/sql/postgres/026_job_tracker.up.sql
ale8k: I thought it might've been nice to query old deleted jobs in the future where they have an error. If we hard delete it's lost for good... I guess I'm not against hard deleting them other than that?

alesstimec: that, or use `gorm.Model`

internal/db/job_tracker.go
```diff
@@ -0,0 +1,137 @@
...
+	durationObserver := servermon.DurationObserver(servermon.DBQueryDurationHistogram, string(op))
+	defer durationObserver()
+	defer servermon.ErrorCounter(servermon.DBQueryErrorCount, &err, string(op))
+
+	db := d.DB.WithContext(ctx)
+
+	entry := dbmodel.NewJobTrackerEntry(jobType)
+	if err := db.Create(entry).Error; err != nil {
+		err := dbError(err)
+		if errors.ErrorCode(err) == errors.CodeAlreadyExists {
```
alesstimec, line 38: afaics this would never return this type of error as we create the entry on L32 - it will be assigned a unique UUID (unlikely to clash with an existing one).

internal/db/job_tracker.go
```diff
@@ -0,0 +1,137 @@
...
+	if err := d.ready(); err != nil {
+		return errors.E(op, err)
+	}
+
+	durationObserver := servermon.DurationObserver(servermon.DBQueryDurationHistogram, string(op))
+	defer durationObserver()
+	defer servermon.ErrorCounter(servermon.DBQueryErrorCount, &err, string(op))
+
+	db := d.DB.WithContext(ctx)
+	if err := db.Model(&dbmodel.JobTrackerEntry{}).Where("job_id = ?", jobId).Update("stop", true).Error; err != nil {
```
alesstimec: why not use the same pattern as for methods below?
```
entry.SetStopping()
return d.updateJobStatus(ctx, entry)
```
also, what would call this method? if the job manager is trying to stop a job, i'd expect it to set a `stopping` status, not `stopped`.

internal/dbmodel/job_tracker.go
```diff
@@ -0,0 +1,87 @@
...
+// [statusPending], [StatusRunning], [StatusSuccessful], or [statusFailed].
+// It is stored as a custom enum in the database.
+// StatusFailed must be set using the [JobTrackerEntry.SetFailed] method.
+type JobStatus string
+
+const (
+	StatusRunning    JobStatus = "running"
+	StatusSuccessful JobStatus = "successful"
+	StatusPending    JobStatus = "pending"
+	StatusFailed     JobStatus = "failed"
```
alesstimec, line 23: should there be a `stopping` and `stopped` status, too?

internal/dbmodel/job_tracker.go
```diff
@@ -0,0 +1,87 @@
...
+type JobTrackerEntry struct {
+	// JobID is the unique identifier for the job. This is not to be set manually, please
+	// use the constructor [NewJobTrackerEntry] instead.
+	JobID uuid.UUID `gorm:"type:uuid;primaryKey"`
+	// JobType is the type of the job, e.g., bootstrap, destroy, etc.
+	JobType string `gorm:"type:varchar(128);not null"`
+	// Stop signals whether the job should be stopped. This is set to true when the job is
+	// requested to be stopped, but it does not mean the job has stopped yet.
+	// This can be determined by the Status field.
+	Stop bool `gorm:"not null;default:false"`
```
alesstimec: why is `Stop` separated from `Status`?

internal/dbmodel/job_tracker.go
```diff
@@ -0,0 +1,87 @@
...
+	// Status holds the current status of the job.
+	Status JobStatus `gorm:"type:job_tracker_status;not null;default:'pending'"`
+	// Error holds any error message associated with the job. Not to be set manually
+	// and will only ever be set if the status is [StatusFailed].
+	Error string `gorm:"type:text"`
+
+	CreatedAt time.Time
```
alesstimec, line 43: just include `gorm.Model`, but consider if we want `DeletedAt`

internal/dbmodel/job_tracker.go
```diff
@@ -0,0 +1,87 @@
...
+	UpdatedAt time.Time
+	DeletedAt gorm.DeletedAt `gorm:"index"`
+}
+
+// NewJobTrackerEntry creates a new JobTrackerEntry with the given jobType.
+// The Status is set to [statusPending] by default.
+// And a new JobID is generated automatically.
+func NewJobTrackerEntry(jobType string) *JobTrackerEntry {
+	return &JobTrackerEntry{
+		JobID:   uuid.New(),
```
alesstimec: ugh.. `uuid.New()` may panic..

internal/dbmodel/job_tracker.go
```diff
@@ -0,0 +1,87 @@
...
+
+	j.Error = err.Error()
+	j.Status = StatusFailed
+
+	return nil
+}
+
+// SetRunning marks the job as running.
+func (j *JobTrackerEntry) SetRunning() {
+	j.Error = ""
```
alesstimec, line 77: dd

internal/dbmodel/job_tracker.go
```diff
@@ -0,0 +1,87 @@
...
+	j.Status = StatusRunning
+}
+
+// SetSuccessful marks the job as successful.
+func (j *JobTrackerEntry) SetSuccessful() {
+	j.Error = ""
```
alesstimec, line 83: dd

internal/dbmodel/job_tracker_test.go
```diff
@@ -0,0 +1,81 @@
...
+	c.Assert(db.Create(entry).Error, qt.IsNil)
+
+	entry.Status = "not a real status"
+	c.Assert(db.Save(entry).Error, qt.ErrorMatches, ".*invalid input value for enum job_tracker_status.*")
+}
+
+func (j *jobTrackerSuite) TestJobTracker_StatusesSetCorrectly(c *qt.C) {
+	entry := dbmodel.JobTrackerEntry{}
+
+	entry.SetFailed(errors.New("test error"))
```
alesstimec: check error

alesstimec: also, as it is currently implemented, you should check if setting a nil error returns an error :)

internal/dbmodel/sql/postgres/026_job_tracker.up.sql
ale8k: Gorm model defaults to id as primary key, I want uuid and for it to be named JobId.

internal/db/job_tracker.go
ale8k: Yup, basically impossible but good to catch it. Be hilarious if we ever saw this.

internal/db/job_tracker.go
ale8k: Because they had specific logic to make sure error is set/unset, this is just stopping? I can add a SetStop if you'd like but it wouldn't do a great deal?

internal/dbmodel/job_tracker.go
ale8k: No no, once a job has received stop, it'll set one of these statuses. It'd never say "stopped", it'd likely say "failed" with a reason why. Otherwise it was successful.

internal/dbmodel/job_tracker.go
ale8k: Because stop signals for a job to stop, status is simply the result of what happened when the job stopped.

internal/dbmodel/job_tracker.go
ale8k: Can't. We're not using ID, we're using JobID and UUIDs.

internal/dbmodel/job_tracker.go
ale8k: Very, very unlikely.

internal/dbmodel/job_tracker.go
ale8k: If somebody sets failed, and then sets running accidentally, error will be left set. We don't want that.

internal/db/job_tracker.go
```diff
@@ -0,0 +1,137 @@
...
+		JobID: jobId,
+	}
+	if err := entry.SetFailed(jobErr); err != nil {
+		return errors.E(op, err)
+	}
+
+	return d.updateJobStatus(ctx, entry)
+}
+
+func (d *Database) updateJobStatus(ctx context.Context, entry dbmodel.JobTrackerEntry) (err error) {
```
alesstimec: s/updateJobStatus/updateJob/ - it updates status AND the error

internal/db/job_tracker_test.go
```diff
@@ -0,0 +1,104 @@
...
+
+func (s *dbSuite) TestJobTracker_StopJob(c *qt.C) {
+	ctx := c.Context()
+	err := s.Database.Migrate(ctx)
+	c.Assert(err, qt.IsNil)
+
+	jobId, err := s.Database.AddJob(ctx, "test-job-type")
+	c.Assert(err, qt.IsNil)
+
+	err = s.Database.StopJob(ctx, jobId)
```
alesstimec, line 44: try stopping a job that does not exist

internal/db/job_tracker_test.go
```diff
@@ -0,0 +1,104 @@
...
+func (s *dbSuite) TestJobTracker_StatusSetters(c *qt.C) {
+	ctx := c.Context()
+	err := s.Database.Migrate(ctx)
+	c.Assert(err, qt.IsNil)
+
+	jobId, err := s.Database.AddJob(ctx, "test-job-type")
+	c.Assert(err, qt.IsNil)
+
+	// Pending
+	entry := &dbmodel.JobTrackerEntry{}
```
alesstimec, line 66: try setting status for a job that does not exist

internal/db/job_tracker_test.go
```diff
@@ -0,0 +1,104 @@
...
+
+func (s *dbSuite) TestJobTracker_GetStatus(c *qt.C) {
+	ctx := c.Context()
+	err := s.Database.Migrate(ctx)
+	c.Assert(err, qt.IsNil)
+
+	jobId, err := s.Database.AddJob(ctx, "test-job-type")
+	c.Assert(err, qt.IsNil)
+
+	status, err := s.Database.GetJobStatus(ctx, jobId)
```
alesstimec, line 111: try getting status for a job that does not exist

internal/dbmodel/job_tracker.go
alesstimec: s/Stop/StopSignal/?

internal/dbmodel/job_tracker.go
```diff
@@ -0,0 +1,87 @@
...
+	Stop bool `gorm:"not null;default:false"`
+	// Status holds the current status of the job.
+	Status JobStatus `gorm:"type:job_tracker_status;not null;default:'pending'"`
+	// Error holds any error message associated with the job. Not to be set manually
+	// and will only ever be set if the status is [StatusFailed].
+	Error string `gorm:"type:text"`
+
+	CreatedAt time.Time
+	UpdatedAt time.Time
+	DeletedAt gorm.DeletedAt `gorm:"index"`
```
alesstimec: remove?

internal/dbmodel/job_tracker.go
alesstimec: as unlikely as uuid-s clashing?

alesstimec (approved):
LGTM with a few more comments :)

internal/db/job_tracker.go
kian99: What I mean is, to reduce what is exported from `dbmodel` you can export only the status types like `StatusRunning JobStatus` and delete the `SetX` methods.

This method is already called `SetJobSuccessful`, so there doesn't need to be an exported `entry.SetSuccessful` method. If the method was called `UpdateJob` then it would be useful to have an exported `entry.SetSuccessful` message.

internal/db/job_tracker.go
kian99: But you don't need to update the whole row, just the one column, see https://gorm.io/docs/update.html#Update-single-column
```go
db.Model(&entry).Update("status", entry.Status)
```

internal/dbmodel/job_tracker_test.go
kian99: Not checking the error from this according to the linter.

internal/dbmodel/sql/postgres/026_job_tracker.up.sql
kian99: I think it's okay to hard-delete. Old jobs can be left for as long as we want, functionally the same as doing a soft-delete.

internal/db/job_tracker.go
ale8k: It may be error too, it's a general function. We don't want to update error every time?

internal/db/job_tracker.go
ale8k: The methods on the dbmodel contain business logic to handle ensuring the error is unset. The methods here just wrap updating the db with the correct status. This prevents developer error. We further have an enum as a last catch case to prevent wrong statuses being added. I'm not sure I'm following why we would change it? Perhaps a call later?

internal/dbmodel/sql/postgres/026_job_tracker.up.sql
ale8k: Updating to hard delete.

internal/db/job_tracker.go
kian99: If it's a general function, then isn't it wrong? You're fetching 2 columns but then updating the whole row. You can also update multiple columns https://gorm.io/docs/update.html#Updates-multiple-columns

kian99: I understand you mentioned,
>if a value is zeroed on the model, and is nullable and has no default in the schema, then GORM ignores it during updates. So, specifically selecting them prior to the update fixes this

But that could be prone to errors if we add a field right?

kian99: Resolving, I can see from SQL generated it's just doing an update on the two columns and the docstring for Select says,
>Select specify fields that you want when querying, creating, updating

internal/db/job_tracker.go
kian99: Resolving, as discussed we just want logic on the object not inside the storage method.

kian99 (approved):

PR #1628, opened 2025-06-24
test: enable logging sql by param

ale8k:
## Description

Allows logging sql per jimmtest.PostgresDB. It overrides the environment variable and can be done per test.

## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests

### DIFF
--- a/internal/auth/oauth2_test.go
+++ b/internal/auth/oauth2_test.go
@@ -29,7 +29,7 @@ import (
 
 func setupTestAuthSvc(ctx context.Context, c *qt.C, expiry time.Duration) (*auth.AuthenticationService, *db.Database, sessions.Store, func()) {
 	db := &db.Database{
-		DB: jimmtest.PostgresDB(c, time.Now),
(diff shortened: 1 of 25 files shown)


kian99 (commented):
Do we need this? I.e. logging per test? You can just run a single test with logging turned on

ale8k:
> Do we need this? I.e. logging per test? You can just run a single test with logging turned on

It felt useful to me, just means one less window to switch 😝 in seriousness, if you wanted to keep the logging on for a single test, you can now. So I guess that's kinda/maybe/possibly helpful?

kian99:
I'm inclined to leave it as is, but open to it if it's useful. Keeping logging on for a single test when running all the tests seems like something you'd never do. It would only print the logs if that test failed anyway.

ale8k:
> I'm inclined to leave it as is, but open to it if it's useful. Keeping logging on for a single test when running all the tests seems like something you'd never do. It would only print the logs if that test failed anyway.

Sure thing, closing it and if we ever need it can reopen.

PR #1630, opened 2025-06-25
feat(model): add migration-mode column to model

SimoneDutto:
## Description

migration-mode tracks the state of a model being migrated. It is set during the `Import` facade, and it is checked in our cleanup routine to make sure not to delete migrating models if they are not found in the target controller.

## Note

Set a default for a column does this:
"""
From PostgreSQL 11, adding a column with a constant default value no longer means that each row of the table needs to be updated when the ALTER TABLE statement is executed. Instead, the default value will be returned the next time the row is accessed, and applied when the table is rewritten, making the ALTER TABLE very fast even on large tables.
"""

[doc](https://www.postgresql.org/docs/current/ddl-alter.html#:~:text=5.-,Changing%20a%20Column's%20Default%20Value,default%20for%20future%20INSERT%20commands.)

### DIFF
--- a/internal/db/model.go
+++ b/internal/db/model.go
@@ -1,10 +1,11 @@
-// Copyright 2024 Canonical.
+// Copyright 2025 Canonical.
 
 package db
 
 import (
 	"context"
 
+	"github.com/juju/juju/state"
 	"gorm.io/gorm"
 
 	"github.com/canonical/jimm/v3/internal/dbmodel"
@@ -250,4 +251,44 @@ func (d *Database) CountModelsByController(ctx context.Context, ctl dbmodel.Cont
 		return 0, errors.E(op, dbError(err))
 	}
 	return int(count), nil
+}
+
+// SetMigrationMode sets the migration mode for a model with the specified UUID.
+func (d *Database) SetMigrationMode(ctx context.Context, model *dbmodel.Model, mode state.MigrationMode) (err error) {
+	const op = errors.Op("db.SetMigrationMode")
+	if err := d.ready(); err != nil {
+		return errors.E(op, err)
+	}
+
+	durationObserver := servermon.DurationObserver(servermon.DBQueryDurationHistogram, string(op))
+	defer durationObserver()
+	defer servermon.ErrorCounter(servermon.DBQueryErrorCount, &err, string(op))
+
+	db := d.DB.WithContext(ctx)
+	switch {
+	case model.UUID.Valid:
+		db = db.Where("uuid = ?", model.UUID.String)
+	default:
+		return errors.E(op, "missing uuid", errors.CodeBadRequest)
+	}
+	// Check if the model doesn't exist or the current migration mode is already set to the desired value.
+	var currentModel dbmodel.Model
+	if err := db.First(&currentModel).Error; err != nil {
+		if errors.ErrorCode(err) == errors.CodeNotFound {
+			return errors.E(op, err, "model not found")
+		}
+		return errors.E(op, dbError(err))
+	}
+	if currentModel.MigrationMode == mode {
+		return nil
+	}
+	db = d.DB.WithContext(ctx)
+	if err := db.Where("uuid = ?", model.UUID.String).Model(model).Update("migration_mode", mode).Error; err != nil {
+		if errors.ErrorCode(err) == errors.CodeNotFound {
+			return errors.E(op, err, "model not found")
+		}
+		return errors.E(op, dbError(err))
+	}
+
+	return nil
 }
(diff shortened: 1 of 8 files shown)


internal/db/model.go
```diff
@@ -251,3 +252,35 @@ func (d *Database) CountModelsByController(ctx context.Context, ctl dbmodel.Cont
 	}
 	return int(count), nil
 }
+
+// SetMigrationMode sets the migration mode for a model with the specified UUID.
+func (d *Database) SetMigrationMode(ctx context.Context, model *dbmodel.Model, mode state.MigrationMode) (err error) {
```
kian99, line 257: The signature is a little funny because there will also be a mode inside the `model` object.

internal/db/model_test.go
```diff
@@ -702,3 +702,84 @@ func (s *dbSuite) TestCountModelsByController(c *qt.C) {
...
+			String: modelUUID,
+			Valid:  true,
+		},
+	}
+	err = s.Database.SetMigrationMode(context.Background(), model, state.MigrationModeExporting)
+	c.Assert(err, qt.IsNil)
+
+	err = s.Database.GetModel(context.Background(), model)
+	c.Assert(err, qt.IsNil)
+	c.Check(model.MigrationMode, qt.Equals, state.MigrationModeExporting)
```
kian99: Can you check that other fields of the model weren't modified. Possibly `deepEquals` will help here.

kian99 (commented):
Looks good, just 2 comments.

internal/db/model.go
SimoneDutto: yeah, true but we have this paradigm of using the whole dbmodel even for search, didn't want to break it

internal/db/model_test.go
SimoneDutto: done

kian99 (approved):
lgtm thanks for this

internal/dbmodel/model.go
```diff
@@ -48,6 +49,9 @@ type Model struct {
 
 	// Offers are the ApplicationOffers attached to the model.
 	Offers []ApplicationOffer
+
+	// MigrationMode is the migration mode of the model.
```
ale8k, line 53: Let's have some more detail please aha, I get it is the migration mode from the name 😝 Explain what modes & their purposes.

internal/db/model.go
```diff
@@ -251,3 +252,35 @@ func (d *Database) CountModelsByController(ctx context.Context, ctl dbmodel.Cont
...
+	defer servermon.ErrorCounter(servermon.DBQueryErrorCount, &err, string(op))
+
+	db := d.DB.WithContext(ctx)
+	switch {
+	case model.UUID.Valid:
+		db = db.Where("uuid = ?", model.UUID.String)
+	default:
+		return errors.E(op, "missing uuid", errors.CodeBadRequest)
+	}
+	tx := db.Model(model).Update("migration_mode", mode)
```
ale8k: Just call this result, it's not always a tx.

internal/db/model.go
```diff
@@ -251,3 +252,35 @@ func (d *Database) CountModelsByController(ctx context.Context, ctl dbmodel.Cont
...
+	if err := tx.Error; err != nil {
+		if errors.ErrorCode(err) == errors.CodeNotFound {
+			return errors.E(op, err, "model not found")
+		}
+		return errors.E(op, dbError(err))
+	}
+	if tx.RowsAffected != 1 {
+		return errors.E(op, "model not found", errors.CodeNotFound)
```
ale8k: Or you've set the same mode again? This error is ambiguous

ale8k: Get the model, then you know if it exists. Get the current mode, compare them, and then you know if it is a no-op. Then this error won't be needed and you've covered the two scenarios.

internal/jimm/juju/model_cleanup.go
```diff
@@ -24,6 +25,10 @@ func (j *JujuManager) CleanupNotFoundModels(ctx context.Context) (err error) {
 	defer durationObserver()
 
 	err = j.Database.ForEachModel(ctx, func(m *dbmodel.Model) error {
+		// we don't want to delete models that are in the process of being migrated.
+		if m.MigrationMode != state.MigrationModeNone {
```
ale8k, line 29: This is dangerous I think? What if it is stuck in a migration mode (presuming we miss something in our code or the future)? Then we could be stuck with a model? I think we need a check like:

```go
m.MigrationMode != state.MigrationModeNone && <time hasn't exceeded X>
```

Perhaps another field in addition to migrationMode like `migration_time` and a default cleanup time of like 2 hours or whatever?

ale8k (changes requested):
Think this needs some minor updates

internal/jimm/juju/model_cleanup.go
SimoneDutto: will be addressed in another PR

ale8k (approved):
Thanks for addressing comments, approving.

PR #1631, opened 2025-06-25
feat: bootstrap logs add & query db models and methods

ale8k:
Adds methods to add and query bootstrap logs. 
The logs are added using a constructor with validation in place to prevent redundant values.
The logs are queried by offset, and validation is in place to protect us from performing incorrect queries.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests


internal/db/bootstrap_logs.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/google/uuid"
+
+	"github.com/canonical/jimm/v3/internal/dbmodel"
+	"github.com/canonical/jimm/v3/internal/errors"
+	"github.com/canonical/jimm/v3/internal/servermon"
+)
+
+// AddBootstrapLog creates a bootstrap log entry.
+// It returns an error if the parameters cannot be validated or the database encounters an error.
+func (d *Database) AddBootstrapLog(ctx context.Context, jobId uuid.UUID, lineNumber int, logLine string) (err error) {
```
kian99: I get the feeling that we shouldn't pass in the lineNumber into this function. That would require the layer above to keep track of the line number which seems unnecessary. What about just incrementing the line number based on the previous line in the DB?

internal/db/bootstrap_logs.go
```diff
@@ -0,0 +1,91 @@
...
+		Where("job_id = ?", jobId).
+		Order("line_number ASC")
+
+	var count int64
+	if err := query.Count(&count).Error; err != nil {
+		return loggies, errors.E(op, dbError(err))
+	}
+
+	// Special case to say there aren't any logs, otherwise we check the offset >= count
+	// If we don't check they're both 0, the error below runs which is misleading
```
kian99: Can we rephrase this, I'm finding it very confusing to read.

internal/db/bootstrap_logs.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	durationObserver := servermon.DurationObserver(servermon.DBQueryDurationHistogram, string(op))
+	defer durationObserver()
+	defer servermon.ErrorCounter(servermon.DBQueryErrorCount, &err, string(op))
+
+	// Make sure job exists, if it doesn't, there's no point running the query
+	if err := d.DB.WithContext(ctx).First(&dbmodel.JobTrackerEntry{JobID: jobId}, "job_id = ?", jobId).Error; err != nil {
+		return loggies, errors.E(op, "job not found", dbError(err))
+	}
```
kian99: We can let the DB handle this, no need to do a check that the SQL will handle right?

internal/db/bootstrap_logs.go
```diff
@@ -0,0 +1,91 @@
...
+		Order("line_number ASC")
+
+	var count int64
+	if err := query.Count(&count).Error; err != nil {
+		return loggies, errors.E(op, dbError(err))
+	}
+
+	// Special case to say there aren't any logs, otherwise we check the offset >= count
+	// If we don't check they're both 0, the error below runs which is misleading
+	if int64(offset) == 0 && count == 0 {
```
kian99: Isn't just
```suggestion
	if count == 0 {
```
sufficient?

internal/db/bootstrap_logs.go
```diff
@@ -0,0 +1,91 @@
...
+		return loggies, errors.E(op, "no logs")
+	}
+
+	// Validate the offset isn't greater than the amount of actual logs
```
kian99: In this case we can also just return nothing.

internal/db/bootstrap_logs.go
```diff
@@ -0,0 +1,91 @@
...
+		return errors.E(op, dbError(err))
+	}
+
+	return nil
+}
+
+// QueryBootstrapLog queries for bootstrap logs based on the jobId and offset.
+// It returns an error if the job doesn't exist, there aren't any logs or the offset
+// is greater than the amount of logs,
+func (d *Database) QueryBootstrapLog(ctx context.Context, jobId uuid.UUID, offset int) (loggies []string, err error) {
```
kian99: Can we rename `loggies` :joy:

internal/db/bootstrap_logs_test.go
```diff
@@ -0,0 +1,129 @@
...
+	qt "github.com/frankban/quicktest"
+	"github.com/frankban/quicktest/qtsuite"
+	"github.com/google/uuid"
+
+	"github.com/canonical/jimm/v3/internal/db"
+	"github.com/canonical/jimm/v3/internal/dbmodel"
+	"github.com/canonical/jimm/v3/internal/testutils/jimmtest"
+)
+
+type bootstrapLogsSuite struct {
```
kian99: We don't seem to need a separate test suite here if you want to just re-use `dbSuite`

internal/db/bootstrap_logs_test.go
```diff
@@ -0,0 +1,129 @@
...
+
+	c.Assert(logs2, qt.HasLen, 1)
+	c.Assert(logs2[0].LineNumber, qt.Equals, 0)
+	c.Assert(logs2[0].LogLine, qt.Equals, "Creating Juju controller \"diglett2\" on the-most-amazing-cloud")
+
+	// Expect total length to be 3
+	var count int64
+	err = s.Database.DB.Model(&dbmodel.BootstrapLog{}).Count(&count).Error
+	c.Assert(err, qt.IsNil)
+	c.Assert(count, qt.Equals, int64(3))
```
kian99: You can remove this, it's not testing anything.

internal/db/bootstrap_logs_test.go
```diff
@@ -0,0 +1,129 @@
...
+	_, err = s.Database.QueryBootstrapLog(ctx, jobIdThatDoesntExist, 0)
+	c.Assert(err, qt.ErrorMatches, "job not found")
+
+	// Add job to reference
+	jobId, err := s.Database.AddJob(ctx, "test-job")
+	c.Assert(err, qt.IsNil)
+
+	// Query with no logs
+	_, err = s.Database.QueryBootstrapLog(ctx, jobId, 0)
+	c.Assert(err, qt.ErrorMatches, "no logs")
```
kian99: Instead of an error, what about returning an empty list of logs.

internal/db/bootstrap_logs.go
ale8k: Good thought! Agreed, it'd be nice if it increments. I'll add an incrementing to it.

internal/db/bootstrap_logs.go
ale8k: Sure thing, will do!

internal/db/bootstrap_logs.go
ale8k: Well db will just say log not found, not "job not found" when we do a select. I wanted a nice explicit error?

internal/db/bootstrap_logs.go
ale8k: Oh please let me have it, it sounds so cute 😂

internal/db/bootstrap_logs.go
ale8k: I like the message aha. If we return no logs, the user might not notice an error in their query. So gonna leave as is.

internal/db/bootstrap_logs_test.go
ale8k: Yeah fair.

internal/db/bootstrap_logs_test.go
ale8k: Because we're explicit with the offset >= count, I like being explicit with no logs?

internal/db/bootstrap_logs_test.go
ale8k: Yeah, used dbSuite. I was trying to isolate suites really, but I don't mind really. Removed.

internal/db/bootstrap_logs.go
ale8k: Added.

internal/db/bootstrap_logs.go
```diff
@@ -0,0 +1,96 @@
...
+		Where("job_id = ?", jobId).
+		Order("line_number ASC")
+
+	var count int64
+	if err := query.Count(&count).Error; err != nil {
+		return loggies, errors.E(op, dbError(err))
+	}
+
+	if count == 0 {
+		return loggies, errors.E(op, "no logs")
```
kian99: Here, can we either return `[]string{}, nil` or add an error code to the error, then when we return the result to the client, we don't return with an error, we just send an empty list of logs.

internal/db/bootstrap_logs.go
kian99: Sure

internal/db/bootstrap_logs_test.go
kian99: Made a suggestion about this above.

internal/db/bootstrap_logs.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	// Validate the offset isn't greater than the amount of actual logs
+	if int64(offset) >= count {
+		return loggies, errors.E(op, "offset cannot be greater than or equal to the amount of logs")
+	}
```
kian99: Also, I think this check should go before `if count == 0` otherwise you get different behaviour when offset=N and count == 0 vs when offset=N and count != 0
Additionally, it would be helpful if the error included the current count.

internal/db/bootstrap_logs.go
```diff
@@ -0,0 +1,96 @@
...
+	err = d.DB.WithContext(ctx).
+		Model(&dbmodel.BootstrapLog{}).
+		Where("job_id = ?", jobId).
+		Select("COALESCE(MAX(line_number), 0)").
+		Scan(&currentLineNumber).Error
+	if err != nil {
+		return errors.E(op, "failed to get current line number", err)
+	}
+
+	nextLineNumber := currentLineNumber + 1
```
kian99: Something to keep in mind here is what happens if you call this method at the same time for the same job.
Because the primary key is `(job_id, line_number)` it will fail which is correct.

internal/db/bootstrap_logs.go
ale8k: If I move it above, then when there's 0 logs and offset is 0, it'll say offset cannot be greater or equal to. When instead it should just says no logs because the count is 0?

SimoneDutto (commented):
I'm generally worried we are counting rows, using offsets, and adding rows without being in a transaction and locking the table.

Since we know we are going to bootstrap one controller at the time. I would rather lock the table to add logs.
It shouldn't add overhead because we are not doing bootstrap in parallel and it shouldn't create concurrency issues.
Wdyt?

https://gorm.io/docs/advanced_query.html#Locking

ale8k:
> I'm generally worried we are counting rows, using offsets, and adding rows without being in a transaction and locking the table.
> 
> Since we know we are going to bootstrap one controller at the time. I would rather lock the table to add logs. It shouldn't add overhead because we are not doing bootstrap in parallel and it shouldn't create concurrency issues. Wdyt?
> 
> https://gorm.io/docs/advanced_query.html#Locking

If we lock, it'd be against the where. So only rows with job_id = ?. I'm not sure I see a value in locking the rows when they'll be new and never changed anyway?

ale8k:
I've added an explicit table lock and tx. I couldn't see a way in GORM to lock a table using some existing func. On some reading, others couldn't either, see [here](https://stackoverflow.com/questions/72142999/how-do-i-do-table-locking-in-gormgolang).

internal/db/bootstrap_logs.go
kian99: I found this SO discussion what the best recommend way to handle an offset larger than the amount of items.
See [here](https://stackoverflow.com/questions/11402156/rest-status-code-204-on-paginated-result) and [here](https://stackoverflow.com/questions/32081423/rest-api-pagination-how-to-handle-offset-exceeding-total-count).
The suggestion is to return an HTTP 204 (No Content). Since we're using websockets I'm okay if you return the error or an empty set of results to the client. 
My only suggestion then would be to make the error the same as when `count == 0`.

ale8k:
```
As I stated in the other answer, I would use one of:

HTTP 200 with count=0 and no items in the returned list.
HTTP 400 because invoker asked for an url that is invalid
HTTP 404 because items to return are not found
I think it's safe to return an error (4xx + error info) in this situation because the offset can be exceeded by one of these assumptions:

A coding error
Invoker is not fetching data from begining (no fresh state)
Invoker ignored pagination data (present on each response)
```

Ours would be a coding error with the statement `HTTP 400 because invoker asked for a url that is invalid`. And the suggestion is `(4xx + error info)`. I'll keep the error string there and add a code? Or would you prefer something else?

ale8k:
Added test over the lock. Tracks time between transaction A -> AddBootstrapLog. I've explained in the test docs that unfortunately NOWAIT is not supported in `LOCK [...]`. But, for us I think this is OK. We'd never **not** want to add an incoming log.

internal/db/bootstrap_logs.go
ale8k: Added a code not found to the error.

internal/db/bootstrap_logs.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+func (d *Database) AddBootstrapLog(ctx context.Context, jobId uuid.UUID, logLine string) (err error) {
+	const op = errors.Op("db.AddBootstrapLog")
+
+	if err := d.ready(); err != nil {
+		return errors.E(op, err)
+	}
+
+	return d.Transaction(func(d *Database) error {
```
kian99, line 28: A nit but if we're locking the table, do we also need a transaction? In my head the lock is providing us the same guarantees of the transaction in this case.

internal/db/bootstrap_logs.go
```diff
@@ -0,0 +1,112 @@
...
+		// Lock entire table at start as we're only allowing one bootstrap at a time.
+		if err := lockBootstrapLogs(d); err != nil {
```
kian99: Could you just inline the sql here, I find it makes it easier to read. Especially since this is the only place we use it and the SQL is pretty clear.

internal/dbmodel/bootstrap_logs_test.go
```diff
@@ -0,0 +1,72 @@
...
+	job, err := dbmodel.NewJobTrackerEntry("test-job")
+	c.Assert(err, qt.IsNil)
+	c.Assert(db.Create(job).Error, qt.IsNil)
+
+	// Create a log entry for the job
+	entry, err = dbmodel.NewBootstrapLog(job.JobID, 0, "Creating Juju controller \"aws-controller\" on aws/us-east-1")
+	c.Assert(err, qt.IsNil)
+	c.Assert(db.Create(entry).Error, qt.IsNil)
+
+	// Quickly grab the log in a fresh entry
```
kian99, line 62: As opposed to doing it slowly :eyes:

internal/dbmodel/bootstrap_logs.go
```diff
@@ -0,0 +1,43 @@
...
+	LineNumber int `gorm:"not null;primaryKey"`
+	// LogLine is an actual log line from the bootstrap job.
+	LogLine string `gorm:"type:text;not null"`
+
+	Job JobTrackerEntry `gorm:"constraint:OnDelete:CASCADE;foreignKey:JobID;references:JobID"`
+}
+
+// NewBootstrapLog creates a new BootstrapLog with the given jobId, lineNumber, and logLine.
+// It returns an error if any of the parameters are invalid.
+func NewBootstrapLog(jobId uuid.UUID, lineNumber int, logLine string) (*BootstrapLog, error) {
```
kian99, line 26: dd lineNumber, it's set in the db package

internal/db/bootstrap_logs_test.go
```diff
@@ -0,0 +1,161 @@
...
+			<-finishTransaction
+			return nil
+		})
+
+		c.Assert(err, qt.IsNil)
+	}()
+
+	<-lockAcquired
+
+	// Normal table locks do not support NOWAIT, so this is queue of INSERTS.
```
kian99: I see on [this page](https://www.postgresql.org/docs/current/sql-lock.html) that it seems NOWAIT is supported for table locks, 
>LOCK TABLE obtains a table-level lock, waiting if necessary for any conflicting locks to be released. If NOWAIT is specified, LOCK TABLE does not wait to acquire the desired lock: if it cannot be acquired immediately, the command is aborted and an error is emitted

You could simplify your test by starting with a raw SQL query that locks the table with nowait, then call `db.AddBootstrapLog()` which should fail. As discussed this is just proving to ourselves that these locks work the way we expect.

Although reading my comment again, you'd need to have NOWAIT specific inside of the sql in `db.AddBootstrapLog()`, we don't want that to ensure logs can be queued. So if the suggestion above is not possible I'm favour of deleting the test.
While I'm all for tests that demonstrate to ourselves something works the way we expect (even if it's a library), if it's reliant on time and it could be flakey then I'm more inclined to remove it. Lmk what you think.

ale8k: Ooh. Yeah I'll add that soon, would prefer to see an error. Perhaps I added it incorrectly earlier... and instead of checking further, reasoned that we would likely want to queue it. I'll add NOWAIT for the test but leave our implementation without NOWAIT.

internal/db/bootstrap_logs.go
ale8k: Yeah sure thing. I more testing our exact lock code is run. Should it ever change, we'd see that. But given we're going to NOWAIT in the test and not here, I'm ok with that.

internal/db/bootstrap_logs_test.go
kian99: >I'll add NOWAIT for the test but leave our implementation without NOWAIT.

I edited my comment above, but if you can do that and it works then awesome. If not then no biggie.

ale8k: Yeah I can add it to just our test, not a problem. I think I'm convinced it's worth keeping just to demonstrate to ourselves that it works as intended.

internal/db/bootstrap_logs.go
ale8k: The lock is released at the end of the transaction. Correct me if I'm wrong, but I believe locks are always associated with transactions. 

The docs state:
LOCK TABLE is useless outside a transaction block: the lock would remain held only to the completion of the statement. Therefore PostgreSQL reports an error if LOCK is used outside a transaction block. Use BEGIN and COMMIT (or ROLLBACK) to define a transaction block.

https://www.postgresql.org/docs/current/sql-lock.html#:~:text=However%2C%20if%20the%20view%20has,Clause%20in%20the%20SELECT%20documentation.

ale8k: To add to this, as a little FYI all statements in databases are usually transactions, but they've got an autocommit. So the equivalent I think would be "lock and unlock" immediately without an _explicit_ transaction.

kian99: Cool, TIL.

SimoneDutto (approved):
thanks for the work on the TABLE LOCK!

internal/db/bootstrap_logs.go
```diff
@@ -0,0 +1,112 @@
...
+
+	"github.com/google/uuid"
+
+	"github.com/canonical/jimm/v3/internal/dbmodel"
+	"github.com/canonical/jimm/v3/internal/errors"
+	"github.com/canonical/jimm/v3/internal/servermon"
+)
+
+func lockBootstrapLogs(d *Database) error {
+	return d.DB.Exec("LOCK TABLE bootstrap_logs IN SHARE MODE").Error
```
kian99: Sorry for all the back and forth, this is new stuff for me. But is this mode "SHARE", correct?
I see from [the docs](https://www.postgresql.org/docs/current/explicit-locking.html) 
>SHARE - This mode protects a table against concurrent data changes.
>ACCESS EXCLUSIVE - This mode guarantees that the holder is the only transaction accessing the table in any way.

and further down,
>Only an ACCESS EXCLUSIVE lock blocks a SELECT (without FOR UPDATE/SHARE) statement.

Now since we do a select to get the maximum line number, in theory two connections could read the same line number, then because of the SHARE lock, one gets to write first and then second will get to go next but its write will fail because it will violate the primary key constraint of (job_id, line_number).
Not really a big deal because we expect only one routine to be calling the AddBootstrapLog at a time, but if I'm not wrong, it does mean that the method is still not thread safe.

internal/dbmodel/bootstrap_logs.go
ale8k: How can I remove this? It needs to be set?

kian99: Ah I thought this was being used for the writes, this is for the reads. Can ignore then.

ale8k: This is for writes? It's used to construct a new bootstrap log?

internal/db/bootstrap_logs.go
ale8k: I had to move the sql to a string variable in order to patch it. Resolving.

internal/db/bootstrap_logs.go
ale8k: Added a NOWAIT to the test and used ACCESS EXCLUSIVE to avoid contention.

internal/db/bootstrap_logs.go
```diff
@@ -0,0 +1,123 @@
...
+		}
+		return nil
+	})
+}
+
+// QueryBootstrapLog queries for bootstrap logs based on the jobId and offset.
+//
+// It returns the next offset value to use, and this offset value may be the same
+// as the one initially presented / previously returned. This means no new logs have
+// come in, but they may later, and the client should query again for logs after some time.
```
kian99, line 63: Aha makes sense, so the client will have to do some kind of polling.

internal/db/bootstrap_logs_test.go
```diff
@@ -0,0 +1,158 @@
...
+	newLogs, nextOffsetValue, err = s.Database.QueryBootstrapLog(ctx, jobId, offsetTracker)
+	c.Assert(err, qt.IsNil)
+	// This means no new logs have come in, but they may later, and the client should query again for logs
+	// after some time.
+	c.Assert(newLogs, qt.HasLen, 0)
+	c.Assert(nextOffsetValue, qt.Equals, offsetTracker)
+}
+
+// This test is a behaviour check, that is, we want our lock does indeed reject
+// on an ACCESS EXCLUSIVE mode when inserting new logs.
```
kian99, line 117:
```suggestion
// This test is a behaviour check, that is, we want to validate that holding 
// our EXCLUSIVE mode locks rejects inserting new logs.
```

internal/dbmodel/bootstrap_logs.go
```diff
@@ -0,0 +1,43 @@
...
+type BootstrapLog struct {
+	// JobID is the unique identifier for the job. References a [JobTrackerEntry].
+	JobID uuid.UUID `gorm:"type:uuid;not null;primaryKey"`
+	// LineNumber is the line number of a running bootstrap. It is used to offset the log lines
+	// when fetching logs.
+	LineNumber int `gorm:"not null;primaryKey"`
+	// LogLine is an actual log line from the bootstrap job.
+	LogLine string `gorm:"type:text;not null"`
+
+	Job JobTrackerEntry `gorm:"constraint:OnDelete:CASCADE;foreignKey:JobID;references:JobID"`
```
kian99, line 21: Maybe put this right below `JobID`, makes it similar to our other dbmodels

internal/db/bootstrap_logs_test.go
```diff
@@ -0,0 +1,158 @@
...
+	c.Assert(s.Database.AddBootstrapLog(ctx, jobId, "Creating Juju controller \"diglett\" on the-most-amazing-cloud"), qt.IsNil)
+	c.Assert(s.Database.AddBootstrapLog(ctx, jobId, "Fetching Juju agent binaries"), qt.IsNil)
+
+	newLogs, nextOffsetValue, err = s.Database.QueryBootstrapLog(ctx, jobId, offsetTracker)
+	c.Assert(err, qt.IsNil)
+	offsetTracker = nextOffsetValue
+	collectedLogs = append(collectedLogs, newLogs...)
+
+	c.Assert(s.Database.AddBootstrapLog(ctx, jobId, "Binaries contain gems"), qt.IsNil)
+	c.Assert(s.Database.AddBootstrapLog(ctx, jobId, "Gems appear to be very expensive"), qt.IsNil)
```
kian99, line 99: Lol how'd you come up with these lines

internal/db/bootstrap_logs_test.go
```diff
@@ -0,0 +1,158 @@
...
+	err := s.Database.Migrate(ctx)
+	c.Assert(err, qt.IsNil)
+
+	jobId, err := s.Database.AddJob(ctx, "test-job")
+	c.Assert(err, qt.IsNil)
+
+	finishTransaction := make(chan bool)
+	c.Cleanup(func() {
+		close(finishTransaction)
+	})
```
kian99, line 130: Looks fine at a glance but just ensure this doesn't leak some resources like "tries to close DB but fails because the transaction was still busy"

kian99 (approved):
lgtm thanks for all the work on this! Just a few small bits that are non-blocking

internal/db/bootstrap_logs_test.go
ale8k: Yeah seemed to be ok when running test. Without, the routine is stuck, so I think it is ok.

PR #1632, opened 2025-06-25
fix(bootstrap-lock): fix bootstrap lock by using a table instead of advisory lock

SimoneDutto:
Before we were using a session advisory lock to implement a distributed lock. This has some issues
because you can lock multiple times if you are re-using the same session, and you need to unlock it
from the same connection that locked it. 
So, we've switched to a simpler implementation using a table and a transaction with SELECT FOR UPDATE.

> Updated the tests as well. Please feel free to suggest new tests.

Reference:  https://www.postgresql.org/docs/current/explicit-locking.html#There%20are%20two%20ways%20to%20acquire%20an%20advisory%20lock

I've stress-tested the TestConcurrentLock, here the result:
```console
[11:17:19] ➜  jimm git:(lock-not-locking-for-single-session) ✗ stress -count 200 ./db.test  -check.f bootstrapLockSuite.TestConcurrentLock
5s: 48 runs so far, 0 failures, 12 active
10s: 120 runs so far, 0 failures, 12 active
15s: 180 runs so far, 0 failures, 12 active
16s: 200 runs total, 0 failures
```


.gitignore
```diff
@@ -23,3 +23,5 @@ qa-lxd
 local/traefik/certs/ca.srl
 
 db_schemas
+
+*.test
```
SimoneDutto, line 27: added this to avoid committing compiled test when using `stress`

internal/db/bootstrap_lock.go
```diff
@@ -0,0 +1,66 @@
...
+
+	"gorm.io/gorm/clause"
+
+	"github.com/canonical/jimm/v3/internal/dbmodel"
+)
+
+// LockBootstrap acquires the bootstrap lock for controller bootstrap operations.
+// It returns an error if the lock cannot be acquired.
+func (d *Database) LockBootstrap(ctx context.Context, ttl time.Duration) error {
+	err := d.Transaction(func(tx *Database) error {
```
ale8k, line 21: needs an op code

ale8k, line 21: just return d.Transaction, doesn't need checking again on line 38

internal/db/bootstrap_lock.go
```diff
@@ -0,0 +1,66 @@
...
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// UnlockBootstrap releases the bootstrap lock for controller bootstrap operations.
+// It returns an error if the lock was not held or could not be released.
+func (d *Database) UnlockBootstrap(ctx context.Context) error {
+	err := d.Transaction(func(tx *Database) error {
```
ale8k, line 50: Same here, just return

ale8k: op code

internal/db/bootstrap_lock.go
```diff
@@ -0,0 +1,66 @@
...
+		lock := &dbmodel.BootstrapLock{}
+		if err := tx.DB.First(lock).Error; err != nil {
+			return fmt.Errorf("failed to find bootstrap lock: %w", err)
+		}
+		if !lock.Locked {
+			return errors.New("bootstrap lock is not held")
+		}
+		lock.Locked = false
+		lock.ExpireAt = time.Time{} // Reset the due time
```
kian99:
```suggestion
		lock.ExpireAt = time.Time{} // Reset the expiry time
```

internal/db/bootstrap_lock_test.go
```diff
@@ -0,0 +1,110 @@
...
+					qt.ErrorMatches,
+					"bootstrap lock is already held",
+					qt.Commentf("Expected lock acquisition to fail in second session."),
+				)
+				lockNotAcquiredCounter.Add(1)
+				return
+			}
+			time.Sleep(100 * time.Millisecond) // Simulate some work with the lock
+			lockAcquiredCounter.Add(1)
+			err = s.db.UnlockBootstrap(c.Context())
```
kian99: What about releasing the lock after all the routines are done? Then you avoid any timing issues and don't need the sleep.

internal/db/bootstrap_lock_test.go
```diff
@@ -0,0 +1,110 @@
...
+			c.Assert(err, qt.IsNil, qt.Commentf("Failed to release lock."))
+		}()
+	}
+	wg.Wait()
+	c.Assert(lockAcquiredCounter.Load(), qt.Equals, int32(1), qt.Commentf("Expected only one lock to be acquired concurrently."))
+	c.Assert(lockNotAcquiredCounter.Load(), qt.Equals, int32(9), qt.Commentf("Expected 9 attempts to fail to acquire the lock concurrently."))
+}
+
+func TestAdvisoryLocks(t *testing.T) {
```
kian99: Rename

internal/dbmodel/bootstrap_lock.go
```diff
@@ -0,0 +1,13 @@
...
+
+package dbmodel
+
+import "time"
+
+// BootstrapLock represents a lock for controller bootstrap operations.
+type BootstrapLock struct {
+	ID       uint `gorm:"primarykey"`
+	ExpireAt time.Time
+
```
kian99, line 10: dd empty line?

internal/dbmodel/sql/postgres/027_bootstrap_lock.up.sql
```diff
@@ -0,0 +1,12 @@
+-- create a table to hold the bootstrap lock
+-- the name is plural because GORM requires it, even if we will only have one row.
```
kian99: Could you leave a brief comment about why this approach is superior to Postgres advisory locks in our use case.

kian99 (approved):
Some comments but nothing blocking, thanks for this!

internal/db/bootstrap_lock.go
```diff
@@ -0,0 +1,66 @@
...
+func (d *Database) LockBootstrap(ctx context.Context, ttl time.Duration) error {
+	err := d.Transaction(func(tx *Database) error {
+		lock := &dbmodel.BootstrapLock{}
+		// This is equivalent to a SELECT FOR UPDATE in SQL,
+		// which locks the row for the duration of the transaction.
+		// The other select statements will wait until this lock is released.
+		if err := tx.DB.Clauses(clause.Locking{Strength: "UPDATE"}).First(lock).Error; err != nil {
+			return err
+		}
+		if !lock.Locked || lock.ExpireAt.Before(time.Now()) {
```
ale8k: Could add a method to dbmodel.BootstrapLock `.isLocked()`, it'd read nicer. Our db method shouldn't be processing domain logic.

internal/db/bootstrap_lock.go
```diff
@@ -0,0 +1,66 @@
...
+
+// UnlockBootstrap releases the bootstrap lock for controller bootstrap operations.
+// It returns an error if the lock was not held or could not be released.
+func (d *Database) UnlockBootstrap(ctx context.Context) error {
+	err := d.Transaction(func(tx *Database) error {
+		lock := &dbmodel.BootstrapLock{}
+		if err := tx.DB.First(lock).Error; err != nil {
+			return fmt.Errorf("failed to find bootstrap lock: %w", err)
+		}
+		if !lock.Locked {
```
ale8k, line 55: Or expiry time too?

internal/dbmodel/sql/postgres/027_bootstrap_lock.up.sql
```diff
@@ -0,0 +1,12 @@
...
+-- create a row with id 0 to hold the lock.
+
+
+CREATE TABLE IF NOT EXISTS bootstrap_locks (
+    id BIGSERIAL PRIMARY KEY,
+    expire_at TIMESTAMP WITH TIME ZONE NOT NULL,
```
ale8k: expires_at?

internal/db/bootstrap_lock_test.go
```diff
@@ -0,0 +1,110 @@
...
+				return
+			}
+			time.Sleep(100 * time.Millisecond) // Simulate some work with the lock
+			lockAcquiredCounter.Add(1)
+			err = s.db.UnlockBootstrap(c.Context())
+			c.Assert(err, qt.IsNil, qt.Commentf("Failed to release lock."))
+		}()
+	}
+	wg.Wait()
+	c.Assert(lockAcquiredCounter.Load(), qt.Equals, int32(1), qt.Commentf("Expected only one lock to be acquired concurrently."))
```
ale8k, line 92: Hi @ Copilot

internal/db/bootstrap_lock_test.go
```diff
@@ -0,0 +1,110 @@
...
+
+	"github.com/canonical/jimm/v3/internal/db"
+	"github.com/canonical/jimm/v3/internal/testutils/jimmtest"
+)
+
+type bootstrapLockSuite struct {
+	db *db.Database
+}
+
+func (s *bootstrapLockSuite) Init(c *qt.C) {
```
ale8k: Could use for consistency dbSuite?

internal/db/bootstrap_lock_test.go
```diff
@@ -0,0 +1,110 @@
...
+	gormDB := jimmtest.PostgresDB(c, time.Now)
+	db1 := &db.Database{
+		DB: gormDB,
+	}
+	s.db = db1
+	err := s.db.Migrate(c.Context())
+	c.Assert(err, qt.IsNil)
+}
+
+func (s *bootstrapLockSuite) TestBootstrap_LockAndUnlock(c *qt.C) {
```
ale8k: We don't check expires?

ale8k (changes requested):

internal/db/bootstrap_lock_test.go
SimoneDutto: i prefer this test because it is more similar on how we will use lock/unlock.
The sleep is very short and i've stress test it as well to make sure there shouldn't be any timing issue.

internal/dbmodel/bootstrap_lock.go
SimoneDutto: yeah, wrong place. we usually have a new line to differentiate between "gorm" fields and ours.
changing

internal/db/bootstrap_lock_test.go
kian99: I don't really agree that it's more similar to how we will use lock/unlock. There's already work being done by waiting for all the routines to try get the lock then return. Adding the sleep is just a less consistent way of proving that the lock works.
Even if the sleep is short and it doesn't introduce a timing issue today, we will validate the same behaviour while also being more correct without it.
Ultimately not a big deal but that's my 2c

internal/db/bootstrap_lock.go
SimoneDutto: i prefer this one because `isLocked` is not expressive enough. It should be `isLockedOrExpired` and at this point it is just better to have this condition

internal/db/bootstrap_lock.go
SimoneDutto: if it is Locked == false the expire time shouldn't  be considered, because it has been resetted.

internal/db/bootstrap_lock_test.go
SimoneDutto: yeah, there is a test called `TestLockWithTTL` to test expires

internal/db/bootstrap_lock_test.go
SimoneDutto: fair enough, changed

internal/db/bootstrap_lock.go
ale8k: I don't mean for readability. What I mean is the act of locking, unlocking and checking expiry is not the concern of the db method, but the domain object. It would be a perfect candidate to encapsulate that logic and push towards better DDD patterns. Happy to leave it, but would prefer if we did it (simply because I am an advocate of DDD).

internal/db/bootstrap_lock.go
ale8k: Wouldn't our code be more defensive if we checked against such conditions anyway?

internal/dbmodel/bootstrap_lock.go
```diff
@@ -0,0 +1,13 @@
+// Copyright 2025 Canonical.
+
+package dbmodel
+
+import "time"
+
+// BootstrapLock represents a lock for controller bootstrap operations.
+type BootstrapLock struct {
```
ale8k, line 8: Would really love to see .Lock, .Unlock and .IsLocked methods here.

ale8k (approved):
Approving, but would really love to push the DDD aspects. I've put a request comment into the dbmodel. @kian99 @alesstimec LMK what you guys think too about pushing for these patterns.

internal/dbmodel/bootstrap_lock.go
SimoneDutto: I don't know about creating these methods now, because all of our db logic is in the `db` package. Your proposal would mix it a bit, and we would have some db logic in `dbmodel` as well.

PR #1633, opened 2025-06-26
chore: use the juju client for model related methods

kian99:
## Description

This change started off as a simple one, a continuation of https://github.com/canonical/jimm/pull/1611 where we slowly move our `jujuclient` package to use Juju's Go client so that we don't need to call facades directly and we get nicer types out of the client.
The downside of this change is that I had to replace all usages of `jujuparams.*` with `base.*` where were were using the RPC object in many places (i.e. an object with JSON tags used for marshalling/unmarshalling over the wire). We are now using a different but similar type that simplifies a few rough edges. 

There was also a somewhat unrelated but necessary change in the `internal/rpc` package where I had previously introduced a dependency on `internal/jimm/` that was incorrect and caused cyclic dependencies during this PR. That's now been addressed.

Finally, you'll see new functions in `internal/jujuapi/types.go` that convert from the `base.*` types to the API type. I haven't written any tests for these as our integration tests cover this behavior, but as discussed before, if we have any objections to this approach we can always revisit it. Also, confusingly, the types in Juju's `base` package are not the same types used by Juju's apiserver, this might be worth looking at further. 

Fixes [JUJU-9132](https://warthogs.atlassian.net/browse/JUJU-9132)

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

[JUJU-9132]: https://warthogs.atlassian.net/browse/JUJU-9132?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/jujuapi/types.go
kian99, line 1: This file was renamed and modified, see types.go for what existed previously.

internal/errors/errors.go
```diff
@@ -98,8 +99,9 @@ func E(args ...interface{}) error {
 	// to copy the code from the wrapped error. The interface used to
 	// extract error codes is compatible with both the Error type and juju
 	// API Error types.
-	if ec, ok := e.Err.(interface{ ErrorCode() string }); ok {
-		e.Code = Code(ec.ErrorCode())
+	var errCode interface{ ErrorCode() string }
```
alesstimec: why this change?

internal/jimm/juju/interface.go
```diff
@@ -53,7 +54,7 @@ type API interface {
 	AdoptResources(modelUUID string, controllerVersion version.Number) error
 
 	// ChangeModelCredential replaces cloud credential for a given model with the provided one.
-	ChangeModelCredential(context.Context, names.ModelTag, names.CloudCredentialTag) error
+	ChangeModelCredential(model names.ModelTag, credential names.CloudCredentialTag) error
```
alesstimec: i'm not fond of named arguments in interface definitions.. feels like the godoc above does not fully explain what the method is expected to do

internal/jimm/juju/modelbuilder.go
```diff
@@ -50,30 +52,31 @@ func (b *modelBuilder) Error() error {
 	return b.err
 }
 
-func (b *modelBuilder) jujuModelCreateArgs() (*jujuparams.ModelCreateArgs, error) {
+func (b *modelBuilder) jujuModelCreateArgs() (jujuclient.CreateModelArgs, error) {
```
alesstimec: why this change? i much prefer `return nil, err` to `return something{}, err`

internal/jujuapi/mapper.go
```diff
@@ -0,0 +1,239 @@
...
+
+import (
+	"github.com/juju/juju/api/base"
+	"github.com/juju/juju/cloud"
+	jujuparams "github.com/juju/juju/rpc/params"
+	"github.com/juju/names/v5"
+	"gopkg.in/yaml.v3"
+)
+
+func cloudFromParams(cloudName string, p jujuparams.Cloud) cloud.Cloud {
```
alesstimec, line 17: it would be nice to have some consistency in naming.. `XYToZW` where X is a package, Y a type, Z a package and W a type.. e.g. `paramsCloudToCloudCloud`   wdyt?

alesstimec: or have a type that embeds `cloud.Cloud` and has a `FromParamsCloud` method

internal/jujuclient/dial.go
```diff
@@ -112,8 +110,22 @@ func (d *Dialer) createLoginRequest(ctx context.Context, ctl *dbmodel.Controller
 	return d.createLoginRequest1(ctx, ctl.ResourceTag(), user.ResourceTag(), p)
 }
 
+// toModelAccessString maps relation to a model access string.
```
alesstimec, line 110: hhmmm.. i vaguely remember we have a place for these conversion methods between openfga relations to juju access strings, don't we?

internal/jujuclient/modelmanager.go
```diff
@@ -15,324 +13,104 @@ import (
...
-// from the Create model call. If there is an error returned it will be
-// of type *APIError. CreateModel uses the Create model procedure on the
-// ModelManager facade.
-func (c Connection) CreateModel(ctx context.Context, args *jujuparams.ModelCreateArgs, info *jujuparams.ModelInfo) error {
-	const op = errors.Op("jujuclient.CreateModel")
-	if err := c.Call(ctx, "ModelManager", 9, "", "CreateModel", args, info); err != nil {
-		return errors.E(op, jujuerrors.Cause(err))
-	}
-	return nil
+type CreateModelArgs struct {
```
alesstimec, line 18: godoc

alesstimec: afaics we're introducing a type to avoid excessive number of arguments to the `CreateModel` method?

internal/jujuclient/types.go
```diff
@@ -0,0 +1,105 @@
...
+// them where the Juju client has inconsistent behaviour.
+
+// convertParamsModelInfo converts a params.ModelInfo to a base.ModelInfo.
+// It is copied from the Juju codebase because it is not exported and in some
+// API methods like CreateModel, a base.ModelInfo type is returned, while in
+// calls to ModelInfo a params.ModelInfo type is returned. We want consistent
+// types preferably of the form base.ModelInfo. Until either that inconsitency
+// is fixed or we export the convertParamsModelInfo function, we copy it here
+// here have better consistency in our domain layer.
+func convertParamsModelInfo(modelInfo params.ModelInfo) (base.ModelInfo, error) {
```
alesstimec: again a comment about naming consistency, or having a type that wraps `base.ModelInfo` and has a method for conversion from `params.ModelInfo` and a method that returns the wrapped `base.ModelInfo`

internal/rpc/httpproxy.go
```diff
@@ -18,18 +18,37 @@ import (
...
 	"go.uber.org/zap"
-
-	"github.com/canonical/jimm/v3/internal/jimm/juju"
 )
 
 const (
 	defaultScheme = "https"
 )
 
+type ConnectionDetails struct {
```
alesstimec, line 28: godoc

alesstimec (commented):
1st pass review with a few comments

internal/errors/errors.go
kian99: The error that comes out of the Juju client is wrapped internally due to a `errors.Trace(err)` call where that original `err` has an error code. Then, when do `errors.E(err)` and try to extract an error code, we get nothing because the error with the code is 2 levels deep. So we can try to unwrap the error until we find an error that implement `ErrorCode()` or we reach the end of the list of wrapped errors.

internal/jimm/juju/interface.go
kian99: Will remove the named arguments and update the godoc.

internal/jimm/juju/modelbuilder.go
kian99: Yeah let me change it back.

internal/jujuclient/dial.go
kian99: Yeah inside `internal/jimm/permissions` , and you'll see that's where it was used from initially, but I don't think the jujuclient should be reaching in there for things so I've copied it here. Especially because `internal/jimm/juju` imports this package, so you can get into a dependency cycle very easily (can't remember if this import did cause a cycle but better to avoid importing anything `internal/jimm` from `internal/jujuclient`).

The issue makes me think the conversion of OpenFGA types to Juju strings should happen in some kind of adapter package that like `internal/openfgaToJuju` or at least something along those lines is worth thinking about.

internal/jujuclient/modelmanager.go
kian99: Yeah it's just to simplify the number of arguments

internal/jujuapi/mapper.go
kian99: I hear you, I'll take a look at this

internal/jujuclient/types.go
kian99: This function and name was copied from Juju so maybe this one should stay the same?

internal/jimm/juju/interface.go
kian99: Removed the named arguments where unneccessary (some were useful and I've left them).
Left the docstring since I didn't touch that in this PR.

internal/jujuapi/mapper.go
kian99: Here are some thoughts, we either need to convert from an RPC struct to some type that we use internally or we need to convert to whatever type we use internally to an RPC type - speaking specifically about the `jujuapi` package.

For converting InternalType -> RPC, we can look at `internal/jujuapi/cloud.go` where we've got the API handler `CloudInfo()` which receives a `dbmodel.Cloud` and runs `cloud.ToJujuCloudInfo()` on the dbmodel object converting from the dbmodel type to the RPC type. I don't think that is great because it imports the RPC types into the `dbmodel` package for the purposes of putting the conversion logic in there.

Then for RPC -> InternalType we can look at `AddCloud` in the same file, which calls `cloudFromParams` to convert a `jujuparams.Cloud` to `github.com/juju/juju/cloud.Cloud`.

So ultimately, my idea is that it would be nice if the naming convention for the mapping functions easily indicated we are pushing data down vs surfacing it. 
I.e. if we are converting to an internal type we could name the function `toX` where X is some type. And if we are converting to an RPC type we could name the function `fromX`.
The actual words don't matter to me but at least names that reflect which way the data is going. Another options might be `rpcToX` and `xFromRPC`.

alesstimec: exactly.. some naming convention where i can tell which way we're converting just by looking at the method name

luci1900:
So this is a change to use the existing machinery that juju clients rely on for resolving RPC methods to specific ones in case of multiple versions or similar?

Looks reasonable overall, but I don't know if I understand it enough to approve.

internal/jimm/jimm.go
```diff
@@ -648,3 +649,25 @@ func (j *JIMM) OfferAuthorizer() OfferAuthorizer {
 func (j *JIMM) BootstrapManager() BootstrapManager {
 	return j.bootstrapManager
 }
+
+// DialerAdapter is an adapter that implements the juju.Dialer interface
```
luci1900, line 698: Doesn't jujuclient.Dialer also implement Dial in the same way?

Or is this to avoid the pointer?

internal/jimm/juju/interface.go
```diff
@@ -134,13 +129,13 @@ type API interface {
 	ListApplicationOffers(context.Context, []jujuparams.OfferFilter) ([]jujuparams.ApplicationOfferAdminDetailsV5, error)
 
 	// ListModelSummaries lists models summaries
-	ListModelSummaries(context.Context, jujuparams.ModelSummariesRequest) (jujuparams.ModelSummaryResults, error)
+	ListModelSummaries(jujuparams.ModelSummariesRequest) ([]base.UserModelSummary, error)
 
 	// ModelInfo fetches a model's ModelInfo.
-	ModelInfo(context.Context, *jujuparams.ModelInfo) error
```
luci1900, line 137: No context?

luci1900: Is it passed to the API instance instead?

internal/jimm/juju/modelbuilder.go
```diff
@@ -474,18 +476,7 @@ func (b *modelBuilder) CreateControllerModel() *modelBuilder {
 		return b
 	}
 
-	// Grant JIMM admin access to the model. Note that if this fails,
```
luci1900, line 597: Not needed anymore?

internal/jimm/juju/interface.go
kian99: Yeah it's weird but the context is set during websocket creation, because it doesn't make sense to cancel a single message without cancelling the entire websocket connection. So you don't need to pass the context on the individual calls.
The older methods accept a context but then cancel the whole websocket as well iirc.

internal/jimm/jimm.go
kian99: I'll need to go back and look at this one again.

internal/jimm/juju/modelbuilder.go
kian99: Correct, Juju controllers now trust JWTs signed by JIMM's public key so we don't need to manually adjust permission on Juju.

internal/dbmodel/model_test.go
```diff
@@ -405,9 +406,6 @@ func TestModelFromJujuModelInfo(t *testing.T) {
 			Status:      "running",
 			Message:     "ACTIVE",
 		}},
-		SLA: &jujuparams.ModelSLAInfo{
```
luci1900, line 408: Is this no longer necessary? Or is the zero value actually useful for tests?

internal/jujuclient/types.go
```diff
@@ -0,0 +1,116 @@
...
+// them where the Juju client has inconsistent behaviour.
+
+// convertParamsModelInfo converts a params.ModelInfo to a base.ModelInfo.
+// It is copied from the Juju codebase because it is not exported and has
+// additional logic to parse the migration status. In some API methods like
+// CreateModel, a base.ModelInfo type is returned, while in calls to ModelInfo
+// a params.ModelInfo type is returned. We want consistent types preferably of
+// the form base.ModelInfo. Until the client returns a consistent set of types
+// we copy it here here to have better consistency in our domain layer.
+func convertParamsModelInfo(modelInfo params.ModelInfo) (ModelInfo, error) {
```
kian99, line 26: An important note here. The params.ModelInfo holds a few more fields than base.ModelInfo. The difference is below.
The following fields appear in params.ModelInfo but don't appear in base.ModelInfo (conversely there is nothing missing):
```
	DefaultBase        string `json:"default-base,omitempty"`

	// CloudCredentialValidity contains if model credential is valid, if known.
	CloudCredentialValidity *bool `json:"cloud-credential-validity,omitempty"`

	// SecretBackends contains information about the secret backends
	// currently in use by the model.
	SecretBackends []SecretBackendResult `json:"secret-backends"`

	// Migration contains information about the latest failed or
	// currently-running migration. It'll be nil if there isn't one.
	Migration *ModelMigrationStatus `json:"migration,omitempty"`

	// SLA contains the information about the SLA for the model, if set.
	SLA *ModelSLAInfo `json:"sla"`

	// SupportedFeatures provides information about the set of features
	// supported by this model. The feature set contains both Juju-specific
	// entries (e.g. juju version) and other features that depend on the
	// substrate the model is deployed to.
	SupportedFeatures []SupportedFeature `json:"supported-features,omitempty"`
```
Out of these fields I believe `SLA` is long deprecated and unused, `SupportedFeatures` I'm unsure on its usage, `Migration` is necessary and we've added it to our wrapper, `SecretBackends` seems useful but not currently parsed and `DefaultBase` is also not parsed. So I've gone and checked the output of `juju show-model` after this change via JIMM and via the backing Juju controller. Below is the output from JIMM with the fields from Juju added and marked. Note that the output doesn't include ommitted fields like migration status which is only present in certain circumstances.
```
foo:
  name: jimm-test@canonical.com/foo
  short-name: foo
  model-uuid: fe9cecbf-676e-4592-8c0c-effcaf981400
  model-type: iaas
  controller-uuid: 3217dbc9-8ea9-4381-9e97-01eab0b3f6bb
  controller-name: jimm-dev
  is-controller: false
  owner: jimm-test@canonical.com
  cloud: localhost
  region: localhost
  type: lxd
  life: alive
  status:
    current: available
    since: 7 seconds ago
  users:
    jimm-test@canonical.com:\
      display-name: jimm-test ---- missing
      access: admin
      last-connection: never connected
  sla: unsupported ---- missing
  agent-version: 3.6.12
  credential:
    name: localhost
    owner: jimm-test@canonical.com
    cloud: localhost
    validity-check: invalid ---- backing Juju controller says valid
  supported-features: ---- this entire block is missing
  - name: juju
    description: the version of Juju used by the model
    version: 3.6.12
```
What I'll do is expand our ModelInfo struct to include the missing fields so that we can always return them and add a test that verifies the difference between `juju show-model` against JIMM vs Juju.

internal/jimm/juju/cloudcredential_test.go
```diff
@@ -764,7 +764,7 @@ func TestUpdateCloudCredential(t *testing.T) {
 						}}, nil
 					}
 				},
-				GrantJIMMModelAdmin_: func(_ context.Context, _ names.ModelTag) error {
```
SimoneDutto, line 764: leave the ctx even if it's useless rn because once we migrate to juju 4 client, all of the methods will require ctx

kian99: I was thinking about that, I don't mind adding it back afterwards, wdyt?

internal/jimm/jimm.go
kian99: Right, this was done because the `API` interface (see internal/jimm/juju/interface.go.API) was used as the return value in `internal/jujuclient/dial.go.Dial()`. I.e. returning an interface from `internal/jimm/juju/` rather than its own concrete type.

But with this change, `internal/jimm/juju` is importing `internal/jujuclient` for some of its newly introduced types and we have a circular dependency. It's pretty clear though that there is no need for `jujuclient` to import `internal/jimm/juju`. So I changed the concrete `Dialer` to return its concrete type, `Connection`.

And then finally, we have this factory pattern going on here with the Dialer. See `internal/jimm/juju/interface.go.Dialer` which is an interface that has a method `Dial()` that returns another interface `API`. So to get the compiler to be happy and all the types aligned, we need an adapter for the concrete `Dialer` to be able to return an `API` interface rather than concrete `Connection` type.

internal/jujuclient/types.go
kian99: I'm thinking about doing the above in a follow-up PR.

internal/dbmodel/model_test.go
kian99: Yip SLA is not necessary to return (though this new struct doesn't have it at all). See my other comment here https://github.com/canonical/jimm/pull/1633/changes#r2787386802

internal/jujuclient/types.go
kian99: Alternatively, maybe we should just pass around the rpc type when we call this method instead of the base type.

internal/jimm/juju/cloudcredential_test.go
luci1900: We'll definitely need it, I'd imagine it's simpler to not remove it now.

internal/jujuclient/types.go
luci1900: It doesn't seem like it's a problem to pass the RPC type.

luci1900 (approved):
Mostly mechanical changes, should be fine.

I do think we need to QA this somewhat more thoroughly than other changes, simply because it touches so many areas.

internal/jimm/juju/cloudcredential_test.go
kian99: Yeah let me add them back

internal/jimm/juju/interface.go
```diff
@@ -179,7 +180,7 @@ type API interface {
 	UpdateCredential(context.Context, jujuparams.TaggedCredential) ([]jujuparams.UpdateCredentialModelResult, error)
 
 	// ValidateModelUpgrade validates that a model can be upgraded.
-	ValidateModelUpgrade(context.Context, names.ModelTag, bool) error
+	ValidateModelUpgrade(ctx context.Context, model names.ModelTag, force bool) error
```
SimoneDutto, line 183: why do we need the named args?

internal/jimm/juju/model.go
```diff
@@ -54,28 +55,28 @@ type ModelCreateArgs struct {
 }
 
 // AddModel adds the specified model to JIMM.
-func (j *JujuManager) AddModel(ctx context.Context, user *openfga.User, args *ModelCreateArgs) (_ *jujuparams.ModelInfo, err error) {
+func (j *JujuManager) AddModel(ctx context.Context, user *openfga.User, args *ModelCreateArgs) (base.ModelInfo, error) {
```
SimoneDutto, line 58: why did we have the named return here? I can't see any `defer` something, but double check please

internal/jimm/juju/model.go
```diff
@@ -202,98 +203,96 @@ func (j *JujuManager) addModelPermissions(ctx context.Context, owner *openfga.Us
 // error will have the code CodeNotFound. If the given user does not have
 // access to the model then the returned error will have the code
 // CodeUnauthorized.
-func (j *JujuManager) ModelInfo(ctx context.Context, user *openfga.User, mt names.ModelTag) (*jujuparams.ModelInfo, error) {
+func (j *JujuManager) ModelInfo(ctx context.Context, user *openfga.User, mt names.ModelTag) (jujuclient.ModelInfo, error) {
```
SimoneDutto, line 206: what's the reason of using the struct instead of the pointer?

internal/jimm/juju/model.go
```diff
@@ -302,36 +301,35 @@ func (j *JujuManager) reactToModelInfoSuccess(ctx context.Context, model *dbmode
...
 	}
 	// we collect models belonging to the user and we extract the unique controllers.
 	var uniqueControllers []dbmodel.Controller
 	uniqueControllerMap := make(map[string]struct{}, 0)
-	err := j.ForEachUserModel(ctx, user, func(m *dbmodel.Model, uap jujuparams.UserAccessPermission) error {
+	err := j.ForEachUserModel(ctx, user, func(m *dbmodel.Model, uap string) error {
 		models = append(models, struct {
 			model      *dbmodel.Model
-			userAccess jujuparams.UserAccessPermission
+			userAccess string
```
SimoneDutto, line 332: this was an "enum", converting it to a string isn't dangerous for when we do the permission checks?

internal/jimm/juju/model.go
```diff
@@ -364,33 +362,22 @@ func (j *JujuManager) ListModelSummaries(ctx context.Context, user *openfga.User
...
-	}, nil
+	return modelSummaryResults, nil
 }
 
 // mergeModelInfo replaces fields on the juju model info object with
 // information from JIMM where JIMM specific information should be used.
-func (j *JujuManager) mergeModelInfo(ctx context.Context, user *openfga.User, modelInfo *jujuparams.ModelInfo, jimmModel dbmodel.Model) (*jujuparams.ModelInfo, error) {
-	modelInfo.CloudCredentialTag = jimmModel.CloudCredential.Tag().String()
+func (j *JujuManager) mergeModelInfo(ctx context.Context, user *openfga.User, modelInfo jujuclient.ModelInfo, jimmModel dbmodel.Model) (jujuclient.ModelInfo, error) {
+	modelInfo.CloudCredential = jimmModel.CloudCredential.ResourceTag().Id()
```
SimoneDutto, line 378: so now this is just the cloud credential name instead of the full tag?

internal/jujuapi/types.go
```diff
@@ -89,3 +92,215 @@ func toAddModelArgs(args jujuparams.ModelCreateArgs, authenticatedUser names.Use
...
+		}
+		modelSummaryResults[i] = jujuparams.ModelSummaryResult{
+			Result: &summaryParams,
+		}
+	}
+	return jujuparams.ModelSummaryResults{Results: modelSummaryResults}
+}
+
+func toModelDumpParams(modelDump map[string]any) (string, error) {
+	yamlDump, err := yaml.Marshal(modelDump)
```
SimoneDutto, line 236: yaml.Marshal(nil) is safe, i didn't know that!

internal/jimm/upgrade/upgrade.go
```diff
@@ -105,81 +105,48 @@ func NewUpgradeManager(
 //
 // It returns the cloud and credential to be used for bootstrapping
 // the new controller.
-func (u *upgradeManager) PrepareUpgradeTo(ctx context.Context, modelUUID string, targetVersion version.Number) (jujucloud.Cloud, string, jujucloud.Credential, error) {
+func (u *upgradeManager) PrepareUpgradeTo(ctx context.Context, modelUUID string, targetVersion version.Number) (jujucloud.Cloud, string, *jujucloud.Credential, error) {
```
SimoneDutto: why the pointer? later on you reference this pointer

internal/jujuapi/modelmanager.go
```diff
@@ -299,7 +311,7 @@ func (r *controllerRoot) DumpModelsDB(ctx context.Context, args jujuparams.Entit
 		}
 		results[i].Result, err = r.jimm.JujuManager().DumpModelDB(ctx, r.user, mt)
 		if err != nil {
-			results[i].Error = r.mapError(ctx, errors.E(err))
+			results[i].Error = r.mapError(ctx, err)
```
SimoneDutto, line 314: just thinking out loud.

Are we sure we can remove this `errors.E` because mapError relies on the `Code` to map the error? Yes, because `ErrorCode` is either implemented and this error.E is no-op, or it isn't and the map error wouldn't work regardless. Make sense?

internal/jujuclient/controller.go
```diff
@@ -23,3 +28,20 @@ func (c Connection) ControllerConfig(ctx context.Context) (jujuparams.Controller
 	}
 	return results, nil
 }
+
+// CloudSpec retrieves the cloud spec of the model connected to.
+func (c Connection) CloudSpec(ctx context.Context) (cloudspec.CloudSpec, error) {
```
SimoneDutto, line 33: why do we need this now?

internal/jujuclient/dial.go
SimoneDutto: what do you think about a test to make sure these checks can't drift.

internal/jujuclient/modelmanager_test.go
SimoneDutto, line 1: satisfying

internal/jujuclient/modelmanager.go
```diff
@@ -65,209 +71,52 @@ func (c Connection) ModelInfo(ctx context.Context, info *jujuparams.ModelInfo) e
...
-		}},
-	}
-
-	resp := jujuparams.ModelStatusResults{
-		Results: make([]jujuparams.ModelStatus, 1),
-	}
-	err := c.Call(ctx, "ModelManager", 10, "", "ModelStatus", &args, &resp)
+// ModelStatus retrieves the status of a model from the controller.
+func (c Connection) ModelStatus(ctx context.Context, modelTag names.ModelTag) (base.ModelStatus, error) {
+	statuses, err := modelmanager.NewClient(&c).ModelStatus(modelTag)
```
SimoneDutto, line 138: here `err` is the first error in the `Result` or does it work in the client?
I'm kinda worried because they don't "type" errors when the result is of `Result` and it can't potentially mess with our `NotFound` checks.

I think our `errors.E` was typing the error, so this change you've made could be breaking

internal/jujuclient/types.go
SimoneDutto: mmm, that's not good. Why would this type be less than the rpc type? Should we fix it in juju?

SimoneDutto (commented):
Some questions to be sure everything we're doing here is correct

internal/jimm/juju/interface.go
kian99: I added them where it was unclear what the bool/string were. A names.ModelTag is clear, but `bool, string, bool` was very unclear.

internal/jimm/juju/model.go
kian99: I believe a pointer was used to make the return values "simpler", i.e. return `null`, but there was no reason for a pointer.

internal/jimm/juju/model.go
kian99: Yip, similarly for ownerTag -> Owner and CloudTag -> Cloud

internal/jujuapi/modelmanager.go
kian99: Our errors package since a little while back traverses the error tree for the top-most error with a code/info map

internal/jujuclient/controller.go
kian99: It replaces ControllerModelSummary which was being used to get the controller model's cloud and cloud-credential then we'd fetch the cloud-credential. If/when we remove the cloning of controllers we can remove this and simplify.

internal/jujuclient/modelmanager.go
kian99: It should all work out once we get to the API layer and call `mapError` which will inspect the error with `ErrorCode` until it finds a code.

internal/jimm/upgrade/upgrade.go
kian99: Good catch, fixed to avoid returning a pointer. The CloudSpec() call returned a pointer and the value could be nil if the cloud doesn't require credentials.

internal/jimm/juju/model.go
kian99: So the struct value being a string is okay because that's just used for returning to the user later, the signature for the ForEachUserModel callback could be a better type but it's also only being used here so I believe it's okay.

internal/jujuclient/controller.go
kian99: I left out a bit of info, there was no Juju API equivalent of `ControllerModelSummary`, but CloudSpec() exists and did what we needed.

internal/jujuclient/dial.go
kian99: I've tried to add a test now but the package does blackbox testing so I can't without more changes,

internal/jujuclient/types.go
kian99: To summarise the work done here,
- I tried changing to use the RPC type and that proved challenging because there are functions used on code-paths that are common to both CreateModel and ModelInfo calls. If the one returns a base.ModelInfo (returned by CreateModel) and the other returns a `rpc.ModelInfo`, we create a lot more friction in the code for nothing.
- I discussed making the change in Juju and there is an appetite for it, but I've decided to forego making the change in 3.6 as we'll move to 4.0 soon and it would be a breaking change in 3.6.
- I've added the missing types to our jujuclient package and update our mapping logic in `jujuapi`. Now the response from JIMM and Juju are nearly identical.

internal/jimm/juju/model.go
kian99: Double checked the current code and looks fine without a named return.

SimoneDutto (approved):
ty for the effort

internal/jujuclient/dial.go
alesstimec: perhaps the openfga.Relation type could implement `ModelAccessString` or `ControllerAccessString` as needed to convert them to juju access strings..

internal/jujuclient/modelmanager.go
```diff
@@ -65,209 +99,52 @@ func (c Connection) ModelInfo(ctx context.Context, info *jujuparams.ModelInfo) e
...
 	if err != nil {
-		return errors.E(jujuerrors.Cause(err))
+		return base.ModelStatus{}, err
 	}
-	if resp.Results[0].Error != nil {
-		return errors.E(resp.Results[0].Error)
-	}
-	*status = resp.Results[0]
-	return nil
+	return statuses[0], nil
```
alesstimec, line 145: if `len(statuses) == 0`, we'd get a panic..

alesstimec (approved):

ale8k (approved):

internal/jujuclient/dial.go
kian99: Yeah that could work. I'll keep it in mind but won't add it to this PR.

PR #1634, opened 2025-06-27
feat(model-migration): implement the Import facade

SimoneDutto:
The Import facade is in change of deserializing the model, updating the owner for model and cloud credential, creating the model into JIMM's state and proxing the request to the target controller.


internal/jimm/juju/migrationtarget.go
```diff
@@ -293,3 +354,106 @@ func (j *JujuManager) Activate(ctx context.Context, modelTag names.ModelTag, mig
...
+	}
+	defer api.Close()
+
+	serializedDescrition, err := description.Serialize(modelDescription)
+	if err != nil {
+		return errors.E(op, fmt.Errorf("failed to serialize model description: %w", err))
+	}
+	err = api.Import(serializedDescrition)
+	if err != nil {
+		// TODO: handle migration failure in a cleanup routine.
```
kian99, line 412: I think we're actually okay here, if we return an error to the source controller, I believe it will issue an abort call (not 100% as it might depend how far along into the migration process we are) and in abort we can do the cleanup.

internal/jimm/juju/migrationtarget.go
```diff
@@ -293,3 +354,106 @@ func (j *JujuManager) Activate(ctx context.Context, modelTag names.ModelTag, mig
...
+	}
+	return nil
+}
+
+func (j *JujuManager) importModelFromDescription(ctx context.Context, targetControllerID uint, description description.Model) error {
+	modelName, ok := description.Config()[config.NameKey]
+	if !ok {
+		return fmt.Errorf("model config must contain a %q key", config.NameKey)
+	}
+	// TODO: create the offers in JIMM's state.
```
kian99: I'll create a card for this then we can link the JIRA number

internal/jimm/juju/migrationtarget.go
```diff
@@ -293,3 +354,106 @@ func (j *JujuManager) Activate(ctx context.Context, modelTag names.ModelTag, mig
...
+		UUID: sql.NullString{
+			String: modelUUIDStr,
+			Valid:  true,
+		},
+		Name:              modelNameStr,
+		OwnerIdentityName: description.Owner().Id(),
+		ControllerID:      targetControllerID,
+		CloudCredentialID: cloudCredential.ID,
+		CloudRegionID:     region.ID,
+		Life:              "migrating",
```
kian99: I don't think the Life field should be set to "migrating" but not 100% sure, is that what Juju does?

kian99 (commented):
Looks good, will do a more thorough review on Monday.

internal/jimm/juju/migrationtarget.go
SimoneDutto: yeah, i've put the TODO here so when we do the cleanup card we can check these kinds of things

internal/jimm/juju/migrationtarget.go
SimoneDutto: https://github.com/SimoneDutto/juju/blob/fix-create-model-concurrently/state/migration_import.go#L133-L140

They have something like this to set the Status to `busy`. We don't currently have status. 
I can leave Life empty as well, but i've figured when you do `juju models` you could see the models migrating, which is nice.

kian99: Since Life is a field shared with Juju, I'd prefer it matched Juju's value (better yet we could remove it one day and get the value from Juju when needed).
So either we leave it empty or set it it "active" I guess?

internal/jimm/juju/migrationtarget.go
kian99: Thanks for creating the card, let's link the number https://warthogs.atlassian.net/browse/JUJU-8192

internal/jimm/juju/migrationtarget_test.go
```diff
@@ -550,3 +609,266 @@ func TestLatestLogTime_Success(t *testing.T) {
...
+
+	// Check cloud credential are checked.
+	desc := newModelDescription("bob")
+	desc.SetStatus(description.StatusArgs{Value: "available"})
+	bytes, err := description.Serialize(desc)
+	c.Assert(err, qt.IsNil)
+	err = j.Import(ctx, user, params.SerializedModel{
+		Bytes: bytes,
+	})
+	c.Assert(err, qt.ErrorMatches, "^failed to import model from description.*")
```
kian99: We might not be interested in the exact error message but for the sake of making this test easier to read/verify, can you make the error match check for the part of the error that states that a valid cloud-credential was not found.
And then, I think the actual error message is important to indicate to the user what cloud-credential they'd need to add.

internal/jimm/juju/migrationtarget_test.go
```diff
@@ -550,3 +609,266 @@ func TestLatestLogTime_Success(t *testing.T) {
...
+			c.Check(desc.Owner(), qt.Equals, names.NewUserTag("alice@canonical.com"))
+			c.Check(desc.Users(), qt.HasLen, 0)
+			c.Check(desc.CloudCredential(), qt.Not(qt.IsNil))
+			c.Check(desc.CloudCredential().Name(), qt.Equals, "test-cred")
+			c.Check(desc.CloudCredential().Owner(), qt.Equals, "alice@canonical.com")
+			return nil
+		},
+	}
+
+	j := newTestJujuManager(c, &parameters{
```
kian99, line 660: There's a lot of commonality in these tests. Wdyt about moving the common bits to a `setupMigration()` function and calling that from each test?

internal/jimm/juju/migrationtarget.go
```diff
@@ -293,3 +354,106 @@ func (j *JujuManager) Activate(ctx context.Context, modelTag names.ModelTag, mig
...
+	}
+
+	modelUUID, ok := description.Config()[config.UUIDKey]
+	if !ok {
+		return fmt.Errorf("model config must contain a %q key", config.UUIDKey)
+	}
+	modelUUIDStr, ok := modelUUID.(string)
+	if !ok {
+		return fmt.Errorf("model config %q must be a string", config.UUIDKey)
+	}
```
kian99: You can simplify this into 1 line I think and adjust the error message to "model config must container a string value for key %q"

internal/jimm/juju/migrationtarget.go
```diff
@@ -293,3 +354,106 @@ func (j *JujuManager) Activate(ctx context.Context, modelTag names.ModelTag, mig
...
+	cloudCredential := &dbmodel.CloudCredential{
+		CloudName:         description.CloudCredential().Cloud(),
+		OwnerIdentityName: description.Owner().Id(),
+		Name:              description.CloudCredential().Name(),
+	}
+
+	err := j.Database.GetCloudCredential(ctx, cloudCredential)
+	if err != nil {
+		return err
```
kian99: Let's replace this error message with something like "Failed to find cloud-credential %q", bonus if we check for a not-found error vs general error.

kian99 (approved):
lgtm with some comments

internal/jimm/juju/migrationtarget.go
SimoneDutto: yep, empty might be better, and then `Alive` in `activate`

internal/jimm/juju/migrationtarget_test.go
SimoneDutto: yeah, changed.

internal/jimm/juju/migrationtarget_test.go
SimoneDutto: i don't think you gain much from it, because you pass down the dialer, the env, the user map and return j, the env and the user.

So you don't gain much code saving, and in the future if you need something new from the setup you need to refactor other tests.

On my next PR in this PR if i see i'm re-using these fields as is, i'll create the helper.

internal/jimm/juju/migrationtarget.go
SimoneDutto: you are right!

internal/jimm/juju/migrationtarget.go
SimoneDutto: you already get an explicit message from the db layer.
Ex. cloudcredential \<name\> not found

What are you sggesting here?

kian99: I want the user to see the specific cloud-credential that is not found i.e. "foo/bar/baz"

SimoneDutto: ah ok, you already have that then. You should see cloudcredential \<x\> not found

SimoneDutto: i've put the \<name\> in my first message, but it was hidden by the diamond brackets in markdown

internal/jimm/juju/migrationtarget.go
```diff
@@ -207,8 +212,55 @@ func (j *JujuManager) modifyMigrationInfo(model *migration.ModelInfo, userMappin
...
-	model.ModelDescription.SetUsers(nil) // Clear users with access since JIMM gates access.
+	err := j.modifyModelDescription(model.ModelDescription, userMapping)
+	if err != nil {
+		return errors.E(fmt.Errorf("failed to modify model description: %w", err))
+	}
+	return nil
+}
+
+func (j *JujuManager) modifyModelDescription(modelDescription description.Model, userMapping dbmodel.StringMap) error {
+	// change the owner of the model description if it is a local user
```
ale8k, line 225: Be good to add an op

internal/jimm/juju/migrationtarget.go
```diff
@@ -293,3 +362,97 @@ func (j *JujuManager) Activate(ctx context.Context, modelTag names.ModelTag, mig
...
+	if err != nil {
+		return errors.E(op, fmt.Errorf("failed to deserialize model description: %w", err))
+	}
+	incomingMigration := &dbmodel.IncomingModelMigration{
+		ModelUUID: sql.NullString{
+			String: modelDescription.Tag().Id(),
+			Valid:  true,
+		},
+	}
+	err = j.Database.GetIncomingModelMigration(ctx, incomingMigration)
```
ale8k, line 387: I think we can do the getting of the incoming migration and import model in a transaction here, and split importModelFromDescrption up a bit?

So...

1. GetIncomingModelMigration
2. GetCloudCredential
3. AddModel

Something like (Pseudo):
```go
j.Database.Transaction(func(d *Database) {
     j.Database.GetIncomingModelMigration(ctx, incomingMigration)
     j.updateDescription() // and go doc this
     j.getCloudCredentialFromDescription()
     j.Database.GetCloudCredential(ctx, cloudCredential)
     j.Database.FindRegionByCloudName()
     j.getModelDetailsFromDescription()
     j.Database.AddModel()
})
```

ale8k (changes requested):
I think we should refactor juju manager.Import a bit and wrap it up a transaction. I've left a pseudo example to help.

I also think we need to lock that credentials row when retrieving it,I think a user could in theory just delete the credential whilst a migration is active, and you'd end up with a model created using a credential jimm isn't aware of or some kind of out of sync issue? Maybe I'm over thinking this so open to thoughts.

internal/jimm/juju/migrationtarget.go
SimoneDutto: why? the op is for exported methods, this is just a private method

ale8k: Op is the operation, so any errors here should be associated with that op (something like that should be in our err pkg).

ale8k (approved):
I'll stamp lgtm but I do ask we refactor and use a transaction (using the pseudo example) just in case the attributes are updated, I think it'd be fine but better safe than sorry? I also think it'd be easier to read if structured like I've proposed.

PR #1635, opened 2025-06-27
chore: define CredentialStore interface where it's used

kian99:
## Description
This PR is a minor change to declare an interface where it's used and will be part of removing the shared interface in `internal/jimm/credentials`. I haven't removed the shared interface from all locations but that can be done in a follow-up. The value I found in this was to keep our dependency tree leaner, there were no cycling dependencies or otherwise that this fixed beyond reducing an import.


ale8k (approved):
After some careful consideration, I think I'll approve this... But should any problems arise due to this notable change. They're on you! 😆

SimoneDutto (approved):

PR #1636, opened 2025-06-30
feat: add jaas migrate command

kian99:
## Description
This PR introduces a CLI command to the `jaas` plugin called `migrate` for migrating a model to/from a JIMM controller. There was already a migrate command which migrated a model internally between two controllers connected to JIMM, I have renamed that command to `migrate-internal` since the new `migrate` command follows Juju's semantics.

The PR was larger than anticipated because the CLI command needs to make a call to JIMM to execute a `PrepareModelMigration` call before proceeding to start the migration. In the call to `PrepareModelMigration` JIMM needs to return a token that can be used for authentication during the migration process, that not previously described and so I've added the return of that token in this PR.

Note that the CLI command cannot be integration tested because of limitations with our testing suite and the inability to create multiple Juju controllers so I have opted to do simpler tests for the command we will add more thorough E2E tests in a separate change.

Fixes [JUJU-7943](https://warthogs.atlassian.net/browse/JUJU-7943)

## Engineering checklist
- [ ] Documentation updated
- [x] Covered by unit tests

[JUJU-7943]: https://warthogs.atlassian.net/browse/JUJU-7943?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cmd/jaas/cmd/migrateinternalmodel.go
```diff
@@ -0,0 +1,119 @@
...
+	"github.com/juju/juju/jujuclient"
+
+	"github.com/canonical/jimm/v3/internal/errors"
+	"github.com/canonical/jimm/v3/pkg/api"
+	apiparams "github.com/canonical/jimm/v3/pkg/api/params"
+)
+
+const (
+	migrateInternalModelCommandDoc = `
+The migrate-internal command migates a model, or many models between two controllers
```
SimoneDutto:
```suggestion
The migrate-internal command migrates a model, or many models between two controllers
```

cmd/jaas/cmd/migratemodel.go
```diff
@@ -40,22 +72,25 @@ func NewMigrateModelCommand() cmd.Command {
 	return modelcmd.WrapBase(cmd)
 }
 
-// migrateModelCommand migrates a model.
+// migrateModelCommand migrates a model to/from JAAS.
```
SimoneDutto: from? we don't support exporting models right?

internal/jimm/jujuauth/migrationtokens.go
```diff
@@ -0,0 +1,64 @@
...
+func (s *MigrationTokenGenerator) NewToken(ctx context.Context, tokenArgs MigrationTokenArgs) ([]byte, error) {
+	token, err := s.jwtService.NewJWT(ctx, jimmjwx.JWTParams{
+		Controller: s.jimmUUID,
+		User:       tokenArgs.User,
+		Access: map[string]string{
+			tokenArgs.ModelTag.String(): string(permission.AdminAccess),
+		},
+		ExtraClaims: map[string]any{
+			"migration": true,
+		},
```
SimoneDutto: we don't check this extra claim anywhere in the code right?

internal/jimmjwx/jwt.go
```diff
@@ -150,13 +154,18 @@ func (j *JWTService) NewJWT(ctx context.Context, params JWTParams) ([]byte, erro
 		return nil, err
 	}
 
+	expiry := j.Expiry
+	if params.Expiry != time.Duration(0) {
+		expiry = params.Expiry
+	}
+
```
SimoneDutto, line 160: Can you test the expiry is working?

SimoneDutto (commented):
lgtm in general with two comments

cmd/jaas/cmd/migratemodel.go
kian99: No, but I think we should and I think it would work (maybe with minimal changes) but maybe best if I remove "from" until it's 100% working.

internal/jimm/jujuauth/migrationtokens.go
kian99: Nope, I'll bring this up during a discussion. Something we can use in the future

internal/jimmjwx/jwt.go
kian99: Yip, will do

cmd/jaas/cmd/migratemodel.go
```diff
@@ -3,31 +3,63 @@
...
-
+The migrate command migrates a model to JIMM.
+
+This command is useful to take a model that is already running on a Juju controller
+and migrate it to JIMM. During this process JIMM will modify the details of the model
+to remove any local users with access to the model and replace the model owner with
+an external user i.e. from alice -> alice@canonical.com.
+
+In order to determine the new model owner and to handle any existing application-offers 
+that have already been consumed with local users, you can specify a user mapping file
```
alesstimec: s/can specify/must specify/.. the mapping is not optional

cmd/jaas/cmd/migratemodel.go
```diff
@@ -3,31 +3,63 @@
...
+'''
+
+The mapping must, at a minimum, contain an entry for the model owner.
+
+The user mapping is consulted when relations are periodically validated. I.e. if an offer
+was consumed by user "alice", when JIMM validates the relation it will understand that user
+"alice" has been mapped and checks that "alice@canonical" still has access to the offer.
+Revoking access from "alice@canonical.com" will result in the relation encountering an error.
+
+It may not be possible to know all users that have have consumed offers from a model, so
```
alesstimec: while true, anybody that ever consumed an offer must have at least had `consume` access to the offer. Unless that has been removed, that fact would still be recorded in the serialized model. If access has been removed, than we shouldn't care about it, right, but simply reject access to the offer.

internal/jimm/jujuauth/migrationtokens.go
alesstimec: perhaps remove it in this PR.. let's have a spec on our authorization model changes..

alesstimec (approved):

internal/jimmjwx/jwt.go
kian99: Added a test

cmd/jaas/cmd/migratemodel.go
kian99: Gotcha, then I'll change this line to talk about using `juju juju show-offer <offer-name> --format yaml` to identify which users to map.

cmd/jaas/cmd/migratemodel.go
kian99: Changed

cmd/jaas/cmd/migrateinternalmodel.go
```diff
@@ -0,0 +1,119 @@
...
+	"github.com/juju/juju/jujuclient"
+
+	"github.com/canonical/jimm/v3/internal/errors"
+	"github.com/canonical/jimm/v3/pkg/api"
+	apiparams "github.com/canonical/jimm/v3/pkg/api/params"
+)
+
+const (
+	migrateInternalModelCommandDoc = `
+The migrate-internal command migrates a model, or many models between two controllers
```
ale8k: command migrates model(s) between...

cmd/jaas/cmd/migrateinternalmodel.go
```diff
@@ -0,0 +1,119 @@
...
+in your JAAS system. This performs a model migration, but is named
+"migrate-internal" to avoid confusion with the "migrate" command which migrates
+a model to or from JAAS.
```
ale8k: We can migrate away from JAAS?

cmd/jaas/cmd/migratemodel.go
```diff
@@ -3,31 +3,63 @@
...
+users to external users.
+For example:
+
+my-user-mapping.yaml:
+'''
+alice:alice@canonical.com
+bob:bob@canonical.com
+'''
+
+The mapping must, at a minimum, contain an entry for the model owner.
```
ale8k, line 46: Is this the first entry?

cmd/jaas/cmd/migratemodel.go
```diff
@@ -3,31 +3,63 @@
...
+
+In order to determine the new model owner and to handle any existing application-offers 
+that have already been consumed with local users, you must specify a user mapping file
+with the --user-mapping flag. This should point to a yaml file with a mapping of local 
+users to external users.
+For example:
+
+my-user-mapping.yaml:
+'''
+alice:alice@canonical.com
```
ale8k: Space after ":"

cmd/jaas/cmd/migratemodel.go
```diff
@@ -3,31 +3,63 @@
...
+bob:bob@canonical.com
+'''
+
+The mapping must, at a minimum, contain an entry for the model owner.
+
+The user mapping is consulted when relations are periodically validated. I.e. if an offer
```
ale8k, line 48: Be good to explain that relations are periodically validated, then explain our mapping is consulted. Otherwise the user may just be confused as to what this means.

cmd/jaas/cmd/migratemodel.go
```diff
@@ -3,31 +3,63 @@
...
+was consumed by user "alice", when JIMM validates the relation it will understand that user
+"alice" has been mapped and checks that "alice@canonical" still has access to the offer.
+Revoking access from "alice@canonical.com" will result in the relation encountering an error.
+
+It may not be possible to know all users that have have consumed offers from a model, but using
+"juju show-offer <offer-name> --format yaml" you can see all users that have access to the
```
ale8k, line 54: Could  we automate this? Perhaps a new command which simply takes a local user name and spits out a user mapping? It could iterate over all the offers in the model?

cmd/jaas/cmd/migratemodel.go
```diff
@@ -40,22 +72,25 @@ func NewMigrateModelCommand() cmd.Command {
 	return modelcmd.WrapBase(cmd)
 }
 
-// migrateModelCommand migrates a model.
+// migrateModelCommand migrates a model to JAAS.
```
ale8k: to JAAS from a controller that isn't registered with JAAS.

cmd/jaas/cmd/migratemodel.go
```diff
@@ -40,22 +72,25 @@ func NewMigrateModelCommand() cmd.Command {
...
 }
 
+// Info implements Command.Info.
 func (c *migrateModelCommand) Info() *cmd.Info {
 	return jujucmd.Info(&cmd.Info{
 		Name:     "migrate",
-		Args:     "<controller name> <model uuid> [<model uuid>...]",
-		Purpose:  "Migrate models to the target controller",
+		Args:     "<model-name> <target-controller-name>",
+		Purpose:  "Migrate models to the JIMM targetting a specific controller managed by JIMM.",
```
ale8k: This doesn't read right, "to the JIMM targeting?" It shouldn't care about which instance?

Perhaps just:
`Migrates models from an unmanaged controller to a controller managed by JIMM.` 
Would be OK?

cmd/jaas/cmd/migratemodel.go
```diff
@@ -40,22 +72,25 @@ func NewMigrateModelCommand() cmd.Command {
...
+	userMappingFile   string
 }
 
+// Info implements Command.Info.
 func (c *migrateModelCommand) Info() *cmd.Info {
 	return jujucmd.Info(&cmd.Info{
 		Name:     "migrate",
-		Args:     "<controller name> <model uuid> [<model uuid>...]",
-		Purpose:  "Migrate models to the target controller",
+		Args:     "<model-name> <target-controller-name>",
```
ale8k: I think just jimmController would read better? targetController is confusing me seeing backing vs target

cmd/jaas/cmd/migratemodel.go
```diff
@@ -68,42 +103,57 @@ func (c *migrateModelCommand) SetFlags(f *gnuflag.FlagSet) {
 		"yaml": cmd.FormatYaml,
 		"json": cmd.FormatJson,
 	})
+	f.StringVar(&c.backingController, "backing-controller", "", "Specify the name of the controller that will host the model in JIMM.")
+	f.StringVar(&c.userMappingFile, "user-mapping", "", "Specify a comma-separated user mapping of local users to external users")
```
ale8k, line 107: We only table a map and it is only from file right? Can we also give a comma separated list?

I only ask because I only see us doing:
```go
	content, err := os.ReadFile(c.userMappingFile)
	if err != nil {
		return nil, errors.E(err, "could not read user mapping file")
	}
```

cmd/jaas/cmd/migratemodel.go
```diff
@@ -68,42 +103,57 @@ func (c *migrateModelCommand) SetFlags(f *gnuflag.FlagSet) {
...
 	if len(args) < 2 {
 		return errors.E("Missing controller name and model target arguments")
 	}
-	for i, arg := range args {
-		if i == 0 {
-			c.targetController = arg
-			continue
-		}
-		c.modelTargets = append(c.modelTargets, arg)
+	c.modelUUID = args[0]
```
ale8k: I thought this was model name? The args doc has model name.

cmd/jaas/cmd/migratemodel.go
```diff
@@ -68,42 +103,57 @@ func (c *migrateModelCommand) SetFlags(f *gnuflag.FlagSet) {
...
+	if !names.IsValidModel(c.modelUUID) {
+		return errors.E(fmt.Errorf("Invalid model name %q", c.modelUUID))
+	}
+	c.targetController = args[1]
+	if c.userMappingFile == "" {
+		return errors.E("Missing user mapping file. Please provide a user mapping file with the --user-mapping flag.")
```
ale8k: Why is user mapping a flag when it is mandatory?

cmd/jaas/cmd/migratemodel.go
```diff
@@ -68,42 +103,57 @@ func (c *migrateModelCommand) SetFlags(f *gnuflag.FlagSet) {
...
 	}
 	return nil
 }
 
 // Run implements Command.Run.
 func (c *migrateModelCommand) Run(ctxt *cmd.Context) error {
+
+	// Validate that the current controller exists.
+	// This is the controller where the model currently resides.
 	currentController, err := c.store.CurrentController()
```
ale8k, line 129: Maybe I'm a bit lost but would it be good to check we aren't on JIMM here? And also check current controller != backing controller?

I guess I'm trying to ask shouldn't we cover these scenarios:
currentController = jimmy
targetController = jimmy
backingController = jimmy

currentController = mycontroller
targetController = jimmy
backingController = jimmy

currentController = mycontroller
targetController = jimmy
backingController = mycontroller

I was thinking just for nice error output. Maybe it doesn't matter though.

cmd/jaas/cmd/migratemodel.go
```diff
@@ -114,3 +164,63 @@ func (c *migrateModelCommand) Run(ctxt *cmd.Context) error {
 	}
 	return nil
 }
+
+func (c *migrateModelCommand) prepareMigration() (token []byte, err error) {
+	userMapping, err := c.readUserMappingFile()
```
ale8k: I think we should pass this in and not read the file here. Feels like an argument.

cmd/jaas/cmd/migratemodel.go
```diff
@@ -68,42 +103,57 @@ func (c *migrateModelCommand) SetFlags(f *gnuflag.FlagSet) {
...
+	// Validate that the current controller exists.
+	// This is the controller where the model currently resides.
 	currentController, err := c.store.CurrentController()
 	if err != nil {
 		return errors.E(err, "could not determine controller")
 	}
 
-	apiCaller, err := c.NewAPIRootWithDialOpts(c.store, currentController, "", c.dialOpts)
+	// Contact the target controller (JIMM) to prepare it for migration
+	// and receive a migration token.
```
ale8k: Can just put this doc on prepareMigration?

cmd/jaas/cmd/migratemodel.go
```diff
@@ -68,42 +103,57 @@ func (c *migrateModelCommand) SetFlags(f *gnuflag.FlagSet) {
...
+	token, err := c.prepareMigration()
 	if err != nil {
 		return err
 	}
 
-	client := api.NewClient(apiCaller)
-	specs := []apiparams.MigrateModelInfo{}
-	for _, model := range c.modelTargets {
-		specs = append(specs, apiparams.MigrateModelInfo{TargetModelNameOrUUID: model, TargetController: c.targetController})
+	// Create the migration spec that will be used to initiate the migration.
```
ale8k: Again can put it on getMigrationSpec

cmd/jaas/cmd/migratemodel.go
```diff
@@ -68,42 +103,57 @@ func (c *migrateModelCommand) SetFlags(f *gnuflag.FlagSet) {
...
+	spec, err := c.getMigrationSpec(token)
+	if err != nil {
+		return err
 	}
-	req := apiparams.MigrateModelRequest{Specs: specs}
-	events, err := client.MigrateModel(&req)
+
+	// Dial the target controller and start the migration.
+	apiCaller, err := c.NewAPIRootWithDialOpts(c.store, currentController, "", c.dialOpts)
```
ale8k, line 156: Caller isn't closed

cmd/jaas/cmd/migratemodel.go
```diff
@@ -114,3 +164,63 @@ func (c *migrateModelCommand) Run(ctxt *cmd.Context) error {
...
+	}
+	apiCaller, err := c.NewAPIRootWithDialOpts(c.store, c.targetController, "", c.dialOpts)
+	if err != nil {
+		return nil, err
+	}
+	client := jimmapi.NewClient(apiCaller)
+	response, err := client.PrepareModelMigration(&apiparams.PrepareModelMigrationRequest{
+		BackingControllerName: c.backingController,
+		UserMapping:           userMapping,
+		ModelTag:              names.NewModelTag(c.modelUUID).String(),
```
ale8k: I guess it's safe because it is validated in init. But what do you think about a names package where we have like "New*TagSafe(...)"?

cmd/jaas/cmd/migratemodel.go
```diff
@@ -114,3 +164,63 @@ func (c *migrateModelCommand) Run(ctxt *cmd.Context) error {
 	}
 	return nil
 }
+
+func (c *migrateModelCommand) prepareMigration() (token []byte, err error) {
```
ale8k: Do we need named return values here? You don't appear to be using them?

cmd/jaas/cmd/migratemodel.go
```diff
@@ -114,3 +164,63 @@ func (c *migrateModelCommand) Run(ctxt *cmd.Context) error {
...
+		TargetControllerUUID:  controllerInfo.ControllerUUID,
+		TargetControllerAlias: c.targetController,
+		TargetAddrs:           controllerInfo.APIEndpoints,
+		TargetCACert:          controllerInfo.CACert,
+		ModelUUID:             c.modelUUID,
+		TargetToken:           base64.StdEncoding.EncodeToString(token),
+		// The target user is not needed here, as the user details will be determined
+		// by the contents of the migration token - a JWT token parsed by JIMM.
+		// But Juju requires this field to be set, so we provide the user running the migration.
+		TargetUser: accountDetails.User,
```
ale8k, line 231: We don't use accountDetails other than this, so is it worth running:
```go
accountDetails, err := store.AccountDetails(c.targetController)
```

We could just put like "no-op-user" or something here and not get the account details for jimm?

internal/jimm/juju/jimm.go
```diff
@@ -278,7 +279,7 @@ func (j *JujuManager) PrepareModelMigration(
 	modelUUID string,
 	targetControllerName string,
 	userMapping map[string]string,
-) error {
+) ([]byte, error) {
```
ale8k: Update godoc, to explain it's returning a token too. 

Side thing:
Was also just thinking, might be good to add a claim to the token for the incoming model migration, even just an id would be nice. That way we can scope our tokens per migration, because right now you could be in migration an and b, and swap the tokens which I don't think is right. Feel free to leave this though.

internal/jimm/juju/juju_test.go
```diff
@@ -58,8 +63,23 @@ func newTestJujuManager(c *qt.C, p *parameters) *juju.JujuManager {
...
-		p.Dialer, p.CrossModelQueryTimeout)
+		p.Dialer, p.CrossModelQueryTimeout,
+		&mockMigrationTokenGenerator{})
 	c.Assert(err, qt.IsNil)
 
 	return jujuManager
 }
+
+type mockMigrationTokenGenerator struct {
+	user     string
```
ale8k: It's an interface on the juju manager, so these fields aren't settable anyway?

internal/jimm/jujuauth/migrationtokens.go
```diff
@@ -0,0 +1,61 @@
...
+)
+
+const (
+	migrationExpiry = 3 * time.Hour
+)
+
+// MigrationTokenGenerator is responsible for generating migration tokens.
+// It uses the JWTService to create a token with specific claims for migration purposes.
+// The token includes the user, model tag, JIMM's UUID and an indication that it is for migration.
+type MigrationTokenGenerator struct {
```
ale8k: Calling the interface and the implementation the same name is kinda confusing. Maybe call this MigrationTokenService or something idk.

internal/jimm/jujuauth/migrationtokens.go
```diff
@@ -0,0 +1,61 @@
...
+	"time"
+
+	"github.com/juju/juju/core/permission"
+	"github.com/juju/names/v5"
+
+	"github.com/canonical/jimm/v3/internal/jimmjwx"
+)
+
+const (
+	migrationExpiry = 3 * time.Hour
```
ale8k: Why 3 hours? (Just a curiosity question)

internal/jimm/jujuauth/migrationtokens.go
```diff
@@ -0,0 +1,61 @@
...
+// It uses the JWTService to create a token with specific claims for migration purposes.
+// The token includes the user, model tag, JIMM's UUID and an indication that it is for migration.
+type MigrationTokenGenerator struct {
+	jwtService JWTService
+	jimmUUID   string
+}
+
+// newMigrationTokenGenerator creates a new MigrationTokenGenerator instance.
+// It requires a JWTService to generate the tokens and JIMM's UUID used for the audience claim.
+func newMigrationTokenGenerator(jwtService JWTService, jimmUUID string) MigrationTokenGenerator {
```
ale8k: Kinda funky to have a private constructor but public implementation?

internal/jimm/jujuauth/migrationtokens.go
```diff
@@ -0,0 +1,61 @@
...
+	return MigrationTokenGenerator{
+		jwtService: jwtService,
+		jimmUUID:   jimmUUID,
+	}
+}
+
+// MigrationTokenArgs holds the arguments required to generate a migration token.
```
ale8k: There's only two args? Maybe just change the signature of NewToken to take a user and tag, not sure why we need a whole args here?

internal/jimm/jujuauth/migrationtokens.go
```diff
@@ -0,0 +1,61 @@
...
+// NewToken generates a new migration token with the specified user and model tag.
+// The token is valid for a limited time (defined by migrationExpiry) and includes
+// claims that indicate the user has admin access to the specified model tag as
+// well as an extra claim indicating that this token is for migration purposes.
+func (s *MigrationTokenGenerator) NewToken(ctx context.Context, tokenArgs MigrationTokenArgs) ([]byte, error) {
+	token, err := s.jwtService.NewJWT(ctx, jimmjwx.JWTParams{
+		Controller: s.jimmUUID,
+		User:       tokenArgs.User,
+		Access: map[string]string{
+			tokenArgs.ModelTag.String(): string(permission.AdminAccess),
```
ale8k: This reads openly, it's almost like "yeah any token generated for this model, migrate whatever you want for 3 hours".  I think it might be worth detailing which incoming migration this token is for and then check it later down the line. (Update I mentioned this earlier on the PrepareMigration method, it's the same suggestion)

ale8k (changes requested):

internal/jimm/jujuauth/migrationtokens.go
kian99: The interface is defined elsewhere (separate package) so I don't see the issue. The two things don't know about each other.

internal/jimm/jujuauth/migrationtokens.go
kian99: This was just my initial guess at a reasonable value.

internal/jimm/jujuauth/migrationtokens.go
kian99: We use a factory pattern for the Juju auth related stuff, so there is a public constructor in factory.go

internal/jimm/jujuauth/migrationtokens.go
kian99: Sure, changed.

internal/jimm/jujuauth/migrationtokens.go
kian99: I don't fully understand what you're suggesting.
But the checks we do on the model migration endpoints are just IsUserAdmin. To be able to check claims on the token, we'd need to do what I suggested in our discussion about implementing some kind of interface for access checks, where we either check OpenFGA or we check something else like the user's token claims.
Since we don't model migration permissions in OpenFGA currently, we can't do this.

But on a totally different note, I realise I've put this token generator in the wrong place. These are tokens that we send to Juju, and are signed with our JWKS. This won't work during login, those tokens use a shared secret. I'll need to fix this.

internal/jimm/juju/juju_test.go
kian99: The idea with my TODO above was that we could validate these.
I'll remove them for now.

cmd/jaas/cmd/migratemodel.go
kian99: What do you mean?

cmd/jaas/cmd/migratemodel.go
kian99: I think that's covered by "The user mapping is consulted when relations are periodically validated" and then the example.
What are you suggesting to add?

cmd/jaas/cmd/migratemodel.go
kian99: It should be harmless to run rather and seemed better than creating a no-op user.

cmd/jaas/cmd/migratemodel.go
kian99: Yeah leftover, thanks, removed.

cmd/jaas/cmd/migratemodel.go
kian99: Maybe for another day

cmd/jaas/cmd/migratemodel.go
kian99: Sure

cmd/jaas/cmd/migratemodel.go
kian99: Sure, moved

cmd/jaas/cmd/migratemodel.go
kian99: Changed to `Migrate models to JAAS, targetting the desired managed controller.`

cmd/jaas/cmd/migratemodel.go
kian99: Yeah, changed to `jaas-name`

cmd/jaas/cmd/migratemodel.go
kian99: Yeah possibly, a good candidate for a follow-up.

cmd/jaas/cmd/migratemodel.go
kian99: I aimed to follow the Juju `migrate` command as I'm unsure how this command will be run in the future.
We can make it an argument rather than a flag if that's preferred.

cmd/jaas/cmd/migratemodel.go
kian99: Yeah it's a model name, fixing.

cmd/jaas/cmd/migratemodel.go
kian99: I was thinking we could add a comma separated list in a follow-up if we identify the need.

cmd/jaas/cmd/migrateinternalmodel.go
kian99: Theoretically it could, but without testing it, I'll remove the "from jaas" part.

cmd/jaas/cmd/migratemodel.go
kian99: While I don't think we need all those checks, I'll check the model exists on the source controller. That will ensure we're not switched to JIMM (the model won't exist).

internal/jimm/jujuauth/migrationtokens.go
kian99: Resolving the related comments on this package since it'll be changing.

cmd/jaas/cmd/migratemodel.go
ale8k: Where does the entry for the model owner go? Because we say at minimum it must contain a model owner, but just show an example key value set above?

cmd/jaas/cmd/migratemodel.go
ale8k: I mean it'd be good to explain what periodically validated means.

cmd/jaas/cmd/migratemodel.go
ale8k: The juju migrate command doesn't have a user mapping file? I think if it's mandatory it should be an argument.

cmd/jaas/cmd/migratemodel.go
ale8k: Yeah sure, I think I'd prefer we use a no-op user constant though. There's really no need to use IO for a no-op action.

cmd/jaas/cmd/migratemodel.go
kian99: Bob is the model owner, I'll update the wording to make that clearer.

cmd/jaas/cmd/migratemodel.go
kian99: So I can definitely make it an argument, there's just some bits around it worth discussing. If the Juju `migrate` command needs some indicator to know when to call out to the plugin, it would be clearest to do that via a flag. And fortunately we have the `--backing-controller` flag so if that is set then the Juju CLI calls the plugin. But the `--backing-controller` flag is also required just like the user-mapping. So it's a quite weird I'm sure you'd agree.

ale8k: Ahh... I forgotten the `juju migrate` needs to call out to it... Hmm... Maybe chat on call

ale8k (approved):
LGTM, not sure what else you can do with the flags.

PR #1637, opened 2025-07-01
feat: implements a jobtracker package

ale8k:
## Description
The jobtracker packages provides a means to run a routine on a unit of jimm, and cancel it by a signal.

I missed the GetStopSignal in the previous PR for the job tracker dbmodel & methods, so I've quickly added it here.

For testing, I've put some simple polling in. I've left a comment above one of the tests explaining we can mock this. The benefit of polling is we can see it running in an integration fashion. 

## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests

### DIFF
--- a/internal/db/job_tracker.go
+++ b/internal/db/job_tracker.go
@@ -60,7 +60,7 @@ func (d *Database) GetJobStopSignal(ctx context.Context, jobId uuid.UUID) (stopS
 
 	db := d.DB.WithContext(ctx)
 	var entry dbmodel.JobTrackerEntry
-	if err := db.Select("stop_signal").First(&entry, "job_id = ?", jobId).Error; err != nil {
+	if err := db.Select("stop_signal").Where("job_id = ?", jobId).First(&entry).Error; err != nil {
 		return false, dbError(err)
 	}
 
@@ -164,7 +164,7 @@ func (d *Database) GetJobStatus(ctx context.Context, jobId uuid.UUID) (status db
 
 	db := d.DB.WithContext(ctx)
 	entry := &dbmodel.JobTrackerEntry{}
-	if err := db.First(entry, "job_id = ?", jobId).Error; err != nil {
+	if err := db.Where("job_id = ?", jobId).First(entry).Error; err != nil {
 		return status, dbError(err)
 	}
 
--- a/internal/db/job_tracker.go
+++ b/internal/db/job_tracker.go
@@ -46,6 +46,27 @@ func (d *Database) AddJob(ctx context.Context, jobType string) (jobId uuid.UUID,
 	return jobId, nil
 }
 
+// GetJobStopSignal retrieves the stop signal for a job by its ID.
+// It returns true if the stop signal is set, false otherwise, and an error if the job does not exist or if the query fails.
+func (d *Database) GetJobStopSignal(ctx context.Context, jobId uuid.UUID) (stopSignal bool, err error) {
+	const op = errors.Op("db.GetJobStopSignal")
+	if err := d.ready(); err != nil {
+		return false, errors.E(op, err)
+	}
+
+	durationObserver := servermon.DurationObserver(servermon.DBQueryDurationHistogram, string(op))
+	defer durationObserver()
+	defer servermon.ErrorCounter(servermon.DBQueryErrorCount, &err, string(op))
+
+	db := d.DB.WithContext(ctx)
+	var entry dbmodel.JobTrackerEntry
+	if err := db.Select("stop_signal").First(&entry, "job_id = ?", jobId).Error; err != nil {
+		return false, dbError(err)
+	}
+
+	return entry.StopSignal, nil
+}
+
 // StopJob marks a job as stopped.
 // It returns an error if the job does not exist or if the update fails.
 func (d *Database) StopJob(ctx context.Context, jobId uuid.UUID) (err error) {
--- a/internal/db/job_tracker_test.go
+++ b/internal/db/job_tracker_test.go
@@ -30,6 +30,32 @@ func (s *dbSuite) TestJobTracker_CreateJob(c *qt.C) {
 	c.Assert(entry.Error, qt.Equals, "")
 }
 
+func (s *dbSuite) TestJobTracker_GetJobStopSignal(c *qt.C) {
+	ctx := c.Context()
+	err := s.Database.Migrate(ctx)
+	c.Assert(err, qt.IsNil)
+
+	_, err = s.Database.GetJobStopSignal(ctx, uuid.New())
+	c.Assert(err, qt.ErrorMatches, ".*not found.*")
+
+	jobId, err := s.Database.AddJob(ctx, "test-job-type")
+	c.Assert(err, qt.IsNil)
+
+	// By default, stop signal should be false
+	stopSignal, err := s.Database.GetJobStopSignal(ctx, jobId)
+	c.Assert(err, qt.IsNil)
+	c.Assert(stopSignal, qt.IsFalse)
+
+	// Set stop signal
+	err = s.Database.StopJob(ctx, jobId)
+	c.Assert(err, qt.IsNil)
+
+	// Now stop signal should be true
+	stopSignal, err = s.Database.GetJobStopSignal(ctx, jobId)
+	c.Assert(err, qt.IsNil)
+	c.Assert(stopSignal, qt.IsTrue)
+}
+
 func (s *dbSuite) TestJobTracker_StopJob(c *qt.C) {
 	ctx := c.Context()
 	err := s.Database.Migrate(ctx)
--- a/internal/db/job_tracker_test.go
+++ b/internal/db/job_tracker_test.go
@@ -32,7 +32,10 @@ func (s *dbSuite) TestJobTracker_CreateJob(c *qt.C) {
 
 func (s *dbSuite) TestJobTracker_GetJobStopSignal(c *qt.C) {
 	ctx := c.Context()
-	err := s.Database.Migrate(ctx)
+	_, err := s.Database.GetJobStopSignal(ctx, uuid.New())
+	c.Assert(err, qt.ErrorMatches, "upgrade in progress")
+
+	err = s.Database.Migrate(ctx)
 	c.Assert(err, qt.IsNil)
 
 	_, err = s.Database.GetJobStopSignal(ctx, uuid.New())
--- a/internal/jobtracker/jobtracker.go
+++ b/internal/jobtracker/jobtracker.go
@@ -38,20 +38,20 @@ type Tracker struct {
 // NewJobTracker creates and returns a new Tracker instance using the provided JobTrackerStore and stopInterval.
 // It returns an error if the store is nil or if stopInterval is not greater than zero.
 func NewJobTracker(store Store, stopInterval time.Duration) (*Tracker, error) {
-	tracker := &Tracker{}
 	if store == nil {
-		return tracker, errors.New("store cannot be nil")
+		return nil, errors.New("store cannot be nil")
 	}
 	if stopInterval <= 0 {
-		return tracker, errors.New("stopInterval must be greater than zero")
+		return nil, errors.New("stopInterval must be greater than zero")
 	}
-	tracker.stopInterval = stopInterval
-	tracker.store = store
 
-	return tracker, nil
+	return &Tracker{
+		stopInterval: stopInterval,
+		store:        store,
+	}, nil
 }
 
-// Run runs a new job and reurns the job ID.
+// Run runs a new job and returns the job ID.
 func (j *Tracker) Run(ctx context.Context, jobType string, job func(ctx context.Context) error, maxDuration time.Duration) (uuid.UUID, error) {
 	jobId, err := j.store.AddJob(ctx, jobType)
 	if err != nil {
@@ -63,7 +63,7 @@ func (j *Tracker) Run(ctx context.Context, jobType string, job func(ctx context.
 	return jobId, nil
 }
 
-// handleJob runs a job with a given context, job ID, polling interval, and deadline.
+// handleJob runs a job with a given context, job ID, and deadline.
 // It manages the job's lifecycle, including setting its status in the store, handling retries on store operations,
 // and responding to stop signals or context cancellations. The job is run in a separate goroutine, and its status
 // is updated as running, successful, or failed based on its result or context expiration.
--- a/internal/jobtracker/jobtracker.go
+++ b/internal/jobtracker/jobtracker.go
@@ -56,7 +56,7 @@ func (j *Tracker) Run(ctx context.Context, jobType string, job func(ctx context.
 		return jobId, err
 	}
 
-	// no context is passed to handleJob as this is a background
+	// No context is passed to handleJob as this is a background
 	// job and is not tied to the request context.
 	go j.handleJob(jobId, maxDuration, job)
 
@@ -83,7 +83,6 @@ func (j *Tracker) handleJob(
 }
 
 func (j *Tracker) runJob(ctx context.Context, id uuid.UUID, jobErrCh chan error, job func(context.Context) error) {
-
 	if err := j.store.SetJobRunning(ctx, id); err != nil {
 		jobErrCh <- fmt.Errorf("failed to set job running, job not starting: %w", err)
 		return
--- a/internal/jobtracker/jobtracker.go
+++ b/internal/jobtracker/jobtracker.go
@@ -39,7 +39,7 @@ func NewJobTracker(store Store, stopInterval time.Duration) (*Tracker, error) {
 	if store == nil {
 		return nil, errors.New("store cannot be nil")
 	}
-	if stopInterval <= 0 {
+	if stopInterval <= 4 {
 		return nil, errors.New("stopInterval must be greater than zero")
 	}
 
@@ -50,13 +50,14 @@ func NewJobTracker(store Store, stopInterval time.Duration) (*Tracker, error) {
 }
 
 // Run runs a new job and returns the job ID.
-func (j *Tracker) Run(ctx context.Context, jobType string, job func(ctx context.Context) error, maxDuration time.Duration) (uuid.UUID, error) {
-	jobId, err := j.store.AddJob(ctx, jobType)
+func (j *Tracker) Run(jobType string, job func(ctx context.Context) error, maxDuration time.Duration) (uuid.UUID, error) {
+	parentCtx := context.Background()
+	jobId, err := j.store.AddJob(parentCtx, jobType)
 	if err != nil {
 		return jobId, err
 	}
 
-	go j.handleJob(ctx, jobId, maxDuration, job)
+	go j.handleJob(parentCtx, jobId, maxDuration, job)
 
 	return jobId, nil
 }
@@ -68,18 +69,18 @@ func (j *Tracker) Run(ctx context.Context, jobType string, job func(ctx context.
 // If a stop signal is received or the context is canceled or times out, the job is marked as failed.
 // Store operations are retried up to 5 times with a 30-second delay between attempts in case of transient errors.
 func (j *Tracker) handleJob(
-	ctx context.Context,
+	parentCtx context.Context,
 	id uuid.UUID,
 	maxDuration time.Duration,
 	job func(ctx context.Context) error,
 ) {
-	jobCtx, cancelJob := context.WithTimeout(ctx, maxDuration)
+	jobCtx, cancelJob := context.WithTimeout(parentCtx, maxDuration)
 	defer cancelJob()
 
 	jobErrCh := make(chan error)
 
-	go j.runJob(ctx, jobCtx, id, jobErrCh, job)
-	j.monitorJob(ctx, jobCtx, id, jobErrCh, cancelJob)
+	go j.runJob(parentCtx, jobCtx, id, jobErrCh, job)
+	j.monitorJob(parentCtx, jobCtx, id, jobErrCh, cancelJob)
 }
 
 func (j *Tracker) runJob(ctx, jobCtx context.Context, id uuid.UUID, jobErrCh chan error, job func(context.Context) error) {
@@ -98,12 +99,11 @@ func (j *Tracker) monitorJob(ctx, jobCtx context.Context, id uuid.UUID, jobErrCh
 	for {
 		select {
 		case <-jobCtx.Done():
-			err := jobCtx.Err()
-			if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
-				if err := j.store.SetJobFailed(ctx, id, err); err != nil {
-					zapctx.Error(ctx, "error marking the job as failed", zap.Error(err), zap.String("id", id.String()))
-				}
+			ctxErr := jobCtx.Err()
+			if err := j.store.SetJobFailed(ctx, id, ctxErr); err != nil {
+				zapctx.Error(ctx, "error marking the job as failed", zap.Error(err), zap.String("id", id.String()))
 			}
+
 			return
 		case err := <-jobErrCh:
 			if err != nil {
--- a/internal/jobtracker/jobtracker.go
+++ b/internal/jobtracker/jobtracker.go
@@ -11,8 +11,6 @@ import (
 	"time"
 
 	"github.com/google/uuid"
-	"github.com/juju/clock"
-	"github.com/juju/retry"
 	"github.com/juju/zaputil/zapctx"
 	"go.uber.org/zap"
 )
@@ -89,38 +87,36 @@ func (j *Tracker) runJob(ctx, jobCtx context.Context, id uuid.UUID, jobErrCh cha
 		jobErrCh <- fmt.Errorf("failed to set job running, job not starting: %w", err)
 		return
 	}
-	err := job(jobCtx)
-	jobErrCh <- err
+	jobErrCh <- job(jobCtx)
 }
 
 func (j *Tracker) monitorJob(ctx, jobCtx context.Context, id uuid.UUID, jobErrCh chan error, cancelJob context.CancelFunc) {
 	ticker := time.NewTicker(j.stopInterval)
 	defer ticker.Stop()
 
+	// TODO(ale8k): Add mo
 	for {
 		select {
 		case <-jobCtx.Done():
-			switch err := jobCtx.Err(); err {
-			case context.Canceled:
-				if err := retryCall(func() error { return j.store.SetJobFailed(ctx, id, context.Canceled) }); err != nil {
-					zapctx.Error(ctx, "error marking the job as failed", zap.Error(err), zap.String("id", id.String()))
-				}
-			case context.DeadlineExceeded:
-				if err := retryCall(func() error { return j.store.SetJobFailed(ctx, id, context.DeadlineExceeded) }); err != nil {
+			err := jobCtx.Err()
+			if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
+				if err := j.store.SetJobFailed(ctx, id, err); err != nil {
 					zapctx.Error(ctx, "error marking the job as failed", zap.Error(err), zap.String("id", id.String()))
 				}
 			}
 			return
 		case err := <-jobErrCh:
 			if err != nil {
-				if err := retryCall(func() error { return j.store.SetJobFailed(ctx, id, err) }); err != nil {
+				if err := j.store.SetJobFailed(ctx, id, err); err != nil {
 					zapctx.Error(ctx, "error marking the job as failed", zap.Error(err), zap.String("id", id.String()))
 				}
-			} else {
-				if err := retryCall(func() error { return j.store.SetJobSuccessful(ctx, id) }); err != nil {
-					zapctx.Error(ctx, "error marking the job as successful", zap.Error(err), zap.String("id", id.String()))
-				}
+				return
 			}
+
+			if err := j.store.SetJobSuccessful(ctx, id); err != nil {
+				zapctx.Error(ctx, "error marking the job as successful", zap.Error(err), zap.String("id", id.String()))
+			}
+
 			return
 		case <-ticker.C:
 			shouldStop, err := j.store.GetJobStopSignal(ctx, id)
@@ -135,16 +131,4 @@ func (j *Tracker) monitorJob(ctx, jobCtx context.Context, id uuid.UUID, jobErrCh
 		}
 	}
 
-}
-
-func retryCall(f func() error) error {
-	if err := retry.Call(retry.CallArgs{
-		Attempts: 5,
-		Delay:    time.Second * 30,
-		Func:     f,
-		Clock:    clock.WallClock,
-	}); err != nil {
-		return err
-	}
-	return nil
 }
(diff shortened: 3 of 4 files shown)


internal/jobtracker/jobtracker.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+// Run runs a new job and reurns the job ID.
+func (j *Tracker) Run(ctx context.Context, jobType string, job func(ctx context.Context) error, deadline time.Duration) (uuid.UUID, error) {
+	jobId, err := j.store.AddJob(ctx, jobType)
+	if err != nil {
+		return jobId, err
+	}
+
+	go j.runJob(ctx, jobId, deadline, job)
```
kian99: Need to come back to this after lunch, but one thing that stood out to me, is that I think we want a new context when starting the job. The context associated with the call to `Run` will end once the user's request completes.

ale8k: We create a new one within runJob but use the top level context as a parent. I guess we could just remove the parent if you're worried people might cancel it?

internal/jobtracker/jobtracker.go
```diff
@@ -0,0 +1,151 @@
...
+			// do eventually write the correct status.
+			if err != nil || shouldStop {
+				cancelJob()
+			}
+		}
+	}
+
+}
+
+func retryCall(f func() error) error {
```
SimoneDutto: why are we adding `retry` here? we are just setting something in the database, we don't retry these types of operation anywhere in the code

internal/jobtracker/jobtracker.go
```diff
@@ -0,0 +1,151 @@
...
+			}
+			return
+		case <-ticker.C:
+			shouldStop, err := j.store.GetJobStopSignal(ctx, id)
+
+			// If we fail to get the stop signal for any reason, we do a best
+			// effort (as the db probably has died on us), so we stop the job,
+			// and hope our status setters on context cancellation
+			// do eventually write the correct status.
+			if err != nil || shouldStop {
```
SimoneDutto, line 126: in case of err i would just rely on the timeout.

I can imagine:
- jimm starts the bootstrap
- networking issue and we can't reach postgres 1 single time
- we don't want to cancel the bootstrap, we can just retry next time.
- we will eventually cancel the bootstrap on timeout

internal/jobtracker/jobtracker.go
```diff
@@ -0,0 +1,151 @@
...
+					zapctx.Error(ctx, "failed to set job failed", zap.Error(err))
+				}
+			case context.DeadlineExceeded:
+				if err := retryCall(func() error { return j.store.SetJobFailed(ctx, id, context.DeadlineExceeded) }); err != nil {
+					zapctx.Error(ctx, "failed to set job failed", zap.Error(err))
+				}
+			}
+			return
+		case err := <-jobErrCh:
+			if err != nil {
```
SimoneDutto, line 103: if you cancel the context in case of errors you can simplify your tests by waiting on <- ctx.Done() instead of polling

internal/jobtracker/jobtracker_test.go
```diff
@@ -0,0 +1,189 @@
...
+	var status dbmodel.JobStatus
+	var pollerr error
+	for i := 0; i < 10; i++ {
+		status, pollerr = s.db.GetJobStatus(testCtx, id)
+		c.Assert(pollerr, qt.IsNil)
+		if status == dbmodel.StatusFailed {
+			break
+		}
+		time.Sleep(100 * time.Millisecond)
+	}
```
SimoneDutto: if you cancel the context in case of errors you can remove this code and just <-ctx.Done() i believe

internal/jobtracker/jobtracker_test.go
```diff
@@ -0,0 +1,189 @@
...
```
SimoneDutto: same here, you can just <-ctx.Done instead of polling, right?

internal/jobtracker/jobtracker.go
```diff
@@ -0,0 +1,151 @@
...
+	return jobId, nil
+}
+
+// manageJob runs a job with a given context, job ID, polling interval, and deadline.
+// It manages the job's lifecycle, including setting its status in the store, handling retries on store operations,
+// and responding to stop signals or context cancellations. The job is run in a separate goroutine, and its status
+// is updated as running, successful, or failed based on its result or context expiration.
+// If a stop signal is received or the context is canceled or times out, the job is marked as failed.
+// Store operations are retried up to 5 times with a 30-second delay between attempts in case of transient errors.
+func (j *Tracker) manageJob(
```
kian99: `handleJob` or `startJob` read better to me than "manage". Also not opposed to `runJob`.

internal/jobtracker/jobtracker_test.go
```diff
@@ -0,0 +1,189 @@
...
+	c.Assert(err, qt.IsNil)
+	s.tracker = tracker
+}
+
+// !!! --- For PR reviewers, currently we're polling the job status as we are waiting
+// for the forselect to update. We could use gomock / some mock and capture the call. What would you prefer?
+// Gomock would allow us to see the order of calls and expected params, so perhaps a bit cleaner.
+//
+// Or we leave it polling.
+func (s *jobTrackerSuite) TestJobTracker_Run_JobIsSetFailed_JobError(c *qt.C) {
```
kian99: It's a bit hard to read the test name/intent. What about
```suggestion
func (s *jobTrackerSuite) TestJobTracker_JobWithError(c *qt.C) {
```
Similar sentiment for the other test names.
I think you should even drop the `TestJobTracker` prefix since they're already in the jobTracker suite.

internal/jobtracker/jobtracker_test.go
```diff
@@ -0,0 +1,189 @@
...
+	id, err := s.tracker.Run(testCtx, "test-job-type", aRunnignJob, time.Second*1000)
+	c.Assert(err, qt.IsNil)
+
+	<-jobRunning
+	status, err := s.db.GetJobStatus(testCtx, id)
+	c.Assert(err, qt.IsNil)
+	c.Assert(status, qt.Equals, dbmodel.StatusRunning)
+}
+
+func (s *jobTrackerSuite) TestJobTracker_Run_JobIsSetSuccessful(c *qt.C) {
```
kian99: Like this one could just be `TestJobIsSetSuccessful`

internal/jobtracker/jobtracker_test.go
```diff
@@ -0,0 +1,189 @@
...
+
+	aFastDyingJob := func(ctx context.Context) error {
+		return errors.New("I died really fast")
+	}
+
+	id, err := s.tracker.Run(testCtx, "test-job-type", aFastDyingJob, time.Second*1000)
+	c.Assert(err, qt.IsNil)
+
+	var status dbmodel.JobStatus
+	var pollerr error
```
kian99: What do you think about testing `j.manageJob()` which will return only when the job completes.
If you only want to test the exported method, then I think the polling is fine since you can't know when the job is done.

kian99: Polling for the job to complete could be moved to a helper method as I see its reused.

internal/jobtracker/jobtracker_test.go
kian99: I don't think that should work, because the context passed to Run() shouldn't be the same context used for the job (one is tied to HTTP request and the other isn't).

internal/jobtracker/jobtracker_test.go
```diff
@@ -0,0 +1,189 @@
...
+
+	c.Assert(status, qt.Equals, dbmodel.StatusFailed)
+
+	j := dbmodel.JobTrackerEntry{}
+	var jobErr string
+	c.Assert(s.db.DB.First(&j).Where("job_id = ?", id).Select("error").Scan(&jobErr).Error, qt.IsNil)
+	c.Assert(jobErr, qt.Equals, "context canceled")
+}
+
+func (s *jobTrackerSuite) TestJobTracker_Run_JobIsSetRunning(c *qt.C) {
```
kian99: Technically this test leaks the go routines managing the job. That is worth addressing.
It also looks like this test could be racy, what if the job completes before you get the job status?

internal/jobtracker/jobtracker.go
```diff
@@ -0,0 +1,151 @@
+// Copyright 2025 Canonical.
+
+// Package jobtracker provides a way to run routines in one instance of JIMM and track them in another.
+// That is, their status can be checked and they can be stopped.
+
```
alesstimec: dd

internal/jobtracker/jobtracker.go
```diff
@@ -0,0 +1,151 @@
...
+	"github.com/juju/clock"
+	"github.com/juju/retry"
+	"github.com/juju/zaputil/zapctx"
+	"go.uber.org/zap"
+)
+
+// JobTrackerStore defines the interface for tracking the lifecycle and status of jobs.
+// It provides methods to add a new job, update its status (running, successful, or failed),
+// and retrieve a stop signal for a specific job.
+type JobTrackerStore interface {
```
alesstimec: suggestion: rename this to Store.. so when i have to refer to it from another package, i can do jobtracker.Store.. instead of jobtracker.JobTrackerStore.

internal/jobtracker/jobtracker.go
```diff
@@ -0,0 +1,151 @@
...
+// Tracker manages job tracking operations using a provided JobTrackerStore.
+// It periodically performs tasks based on the specified stopInterval duration.
+type Tracker struct {
+	store        JobTrackerStore
+	stopInterval time.Duration
+}
+
+// NewJobTracker creates and returns a new Tracker instance using the provided JobTrackerStore and stopInterval.
+// It returns an error if the store is nil or if stopInterval is not greater than zero.
+func NewJobTracker(store JobTrackerStore, stopInterval time.Duration) (*Tracker, error) {
```
alesstimec: suggestion: rename to New. This way we use `jobtracker.New` instead of `jobtracker.NewJobTracker`

internal/jobtracker/jobtracker.go
```diff
@@ -0,0 +1,151 @@
...
+	tracker := &Tracker{}
+	if store == nil {
+		return tracker, errors.New("store cannot be nil")
+	}
+	if stopInterval <= 0 {
```
alesstimec: should we use a default value if one is not provided?

internal/jobtracker/jobtracker.go
```diff
@@ -0,0 +1,151 @@
...
+		return tracker, errors.New("stopInterval must be greater than zero")
+	}
+	tracker.stopInterval = stopInterval
+	tracker.store = store
+
+	return tracker, nil
+}
+
+// Run runs a new job and reurns the job ID.
+func (j *Tracker) Run(ctx context.Context, jobType string, job func(ctx context.Context) error, deadline time.Duration) (uuid.UUID, error) {
```
alesstimec: suggestion: rename deadline to maxDuration.. deadline would be something along this lines "no later than by 3pm today"

internal/jobtracker/jobtracker.go
```diff
@@ -0,0 +1,151 @@
...
+// manageJob runs a job with a given context, job ID, polling interval, and deadline.
+// It manages the job's lifecycle, including setting its status in the store, handling retries on store operations,
+// and responding to stop signals or context cancellations. The job is run in a separate goroutine, and its status
+// is updated as running, successful, or failed based on its result or context expiration.
+// If a stop signal is received or the context is canceled or times out, the job is marked as failed.
+// Store operations are retried up to 5 times with a 30-second delay between attempts in case of transient errors.
+func (j *Tracker) manageJob(
+	ctx context.Context,
+	id uuid.UUID,
+	deadline time.Duration,
```
alesstimec: same suggestion to rename deadline

internal/jobtracker/jobtracker.go
alesstimec: or simply `do`

internal/jobtracker/jobtracker.go
```diff
@@ -0,0 +1,151 @@
...
+	j.monitorJob(ctx, jobCtx, id, jobErrCh, cancelJob)
+}
+
+func (j *Tracker) runJob(ctx, jobCtx context.Context, id uuid.UUID, jobErrCh chan error, job func(context.Context) error) {
+	if err := j.store.SetJobRunning(ctx, id); err != nil {
+		jobErrCh <- fmt.Errorf("failed to set job running, job not starting: %w", err)
+		return
+	}
+	err := job(jobCtx)
+	jobErrCh <- err
```
alesstimec: i think you should also cancel the `jobCtx` once the job returns, so that `monitorJob` can stop, too.

internal/jobtracker/jobtracker_test.go
```diff
@@ -0,0 +1,189 @@
...
+	err := db.Migrate(context.Background())
+	c.Assert(err, qt.IsNil)
+
+	s.db = db
+	tracker, err := jobtracker.NewJobTracker(db, time.Second*5)
+	c.Assert(err, qt.IsNil)
+	s.tracker = tracker
+}
+
+// !!! --- For PR reviewers, currently we're polling the job status as we are waiting
```
alesstimec: for PR author - remove this comment and simply comment on your own PR to get the message across to reviewers ;)

internal/jobtracker/jobtracker.go
```diff
@@ -0,0 +1,151 @@
...
+	ticker := time.NewTicker(j.stopInterval)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-jobCtx.Done():
+			switch err := jobCtx.Err(); err {
+			case context.Canceled:
+				if err := retryCall(func() error { return j.store.SetJobFailed(ctx, id, context.Canceled) }); err != nil {
+					zapctx.Error(ctx, "failed to set job failed", zap.Error(err))
```
alesstimec: i had to read the `failed to set job failed` three times to understand what's going on and what failed. maybe `error making the job as failed` and include the job uuid in the message, otherwise it is useless.. same for the message below

internal/jobtracker/jobtracker.go
alesstimec: or have an alternative way to say "this job is done. stop monitoring"

alesstimec: also.. maybe set the job status to "done" once it's done

internal/jobtracker/jobtracker.go
ale8k: If it fails, and the job is stopped, we'd be left with a poor status. I just thought giving it a couple tries (at least) minimises chance of failure? (I.e., a temporary restart?)

internal/jobtracker/jobtracker.go
ale8k: Hard to say, I wouldn't really know what a sensible default is until running it a few times and seeing how it performs in a Q/A.

internal/jobtracker/jobtracker.go
ale8k: I can also imagine the db stays down, a controller is eventually bootstrapped and we have no knowledge that it exists? I think it is better to just stop the job and hope the cleanup is successful.

internal/jobtracker/jobtracker_test.go
ale8k: So we can't do:
```go
	cancelCtx()
	id, err := s.tracker.Run(cancellableTestCtx, "test-job-type", aFastDyingJob, time.Second*1000)
	c.Assert(err, qt.IsNil)
```
Because the AddJob inside of Run will return a context cancelled 😢(otherwise would work I think?). 

And doing:
```go
	id, err := s.tracker.Run(cancellableTestCtx, "test-job-type", aFastDyingJob, time.Second*1000)
	c.Assert(err, qt.IsNil)
        <-jobStarted
	cancelCtx()
```
Can have timing issues between the cancel & status being set to failed I think?

internal/jobtracker/jobtracker_test.go
ale8k: I was kind of just name spacing the tests, but sure will go update names.

internal/jobtracker/jobtracker_test.go
ale8k: Agreed, I'll make helper.

And we could with a mocked store. I think I'd remove the testing of Run then though? But at that point, I'd probably just test Run with a mock?

internal/jobtracker/jobtracker_test.go
ale8k: It can't complete because the context is alive for the duration of the test?

internal/jobtracker/jobtracker.go
ale8k: If I cancel it, it'll could hit the cancelled condition and be returned as failed. We'd be hoping that:
```go
	jobErrCh <- err
        cancelJobCtx() // always run afterwards?
```

internal/jobtracker/jobtracker_test.go
ale8k: Yep fair point 😆

internal/jobtracker/jobtracker.go
ale8k: But I want to set the error from the job error? If I do that then we'd just have context cancelled / deadline and we would need to store the error in a variable outside the loop and put an if condition into the ctx.Done() that checks this has come from an error. That feels less simple?

ale8k:
@alesstimec check out you're happy so I don't miss comments ;)

internal/db/job_tracker_test.go
```diff
@@ -30,6 +30,32 @@ func (s *dbSuite) TestJobTracker_CreateJob(c *qt.C) {
 	c.Assert(entry.Error, qt.Equals, "")
 }
 
+func (s *dbSuite) TestJobTracker_GetJobStopSignal(c *qt.C) {
+	ctx := c.Context()
+	err := s.Database.Migrate(ctx)
+	c.Assert(err, qt.IsNil)
+
+	_, err = s.Database.GetJobStopSignal(ctx, uuid.New())
```
alesstimec, line 41: for completeness, can you also try it before migration?

internal/jobtracker/jobtracker.go
```diff
@@ -0,0 +1,150 @@
...
+	if store == nil {
+		return tracker, errors.New("store cannot be nil")
+	}
+	if stopInterval <= 0 {
+		return tracker, errors.New("stopInterval must be greater than zero")
+	}
+	tracker.stopInterval = stopInterval
+	tracker.store = store
+
+	return tracker, nil
```
alesstimec:
```suggestion
	return &Tracker{
	  stopInterval: stopInterval,
	  store: store,
	}, nil
```

internal/jobtracker/jobtracker.go
```diff
@@ -0,0 +1,150 @@
...
+	store        Store
+	stopInterval time.Duration
+}
+
+// NewJobTracker creates and returns a new Tracker instance using the provided JobTrackerStore and stopInterval.
+// It returns an error if the store is nil or if stopInterval is not greater than zero.
+func NewJobTracker(store Store, stopInterval time.Duration) (*Tracker, error) {
+	tracker := &Tracker{}
+	if store == nil {
+		return tracker, errors.New("store cannot be nil")
```
alesstimec: `return nil, errors.New....`? why return a pointer to an empty tracker?

internal/jobtracker/jobtracker.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+	if stopInterval <= 0 {
+		return tracker, errors.New("stopInterval must be greater than zero")
```
alesstimec: same... `return nil, errors.New...`

internal/jobtracker/jobtracker.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+	tracker.stopInterval = stopInterval
+	tracker.store = store
+
+	return tracker, nil
+}
+
+// Run runs a new job and reurns the job ID.
```
alesstimec: s/reurns/returns/

internal/jobtracker/jobtracker.go
```diff
@@ -0,0 +1,150 @@
...
+	if err != nil {
+		return jobId, err
+	}
+
+	go j.handleJob(ctx, jobId, maxDuration, job)
+
+	return jobId, nil
+}
+
+// handleJob runs a job with a given context, job ID, polling interval, and deadline.
```
alesstimec: i don't see a `polling interval` argument

internal/jobtracker/jobtracker.go
alesstimec: then
`jobErrCh <- job(jobCtx)`

internal/jobtracker/jobtracker.go
```diff
@@ -0,0 +1,150 @@
...
+				}
+			case context.DeadlineExceeded:
+				if err := retryCall(func() error { return j.store.SetJobFailed(ctx, id, context.DeadlineExceeded) }); err != nil {
+					zapctx.Error(ctx, "error marking the job as failed", zap.Error(err), zap.String("id", id.String()))
+				}
+			}
+			return
+		case err := <-jobErrCh:
+			if err != nil {
+				if err := retryCall(func() error { return j.store.SetJobFailed(ctx, id, err) }); err != nil {
```
alesstimec: this makes me think we should simply have
```
j.store.SetJobStatus(err)
```

alesstimec (commented):
A few comments.. And we also discussed that we'd remove the controller if we failed to record its status or fail to add it to JIMM.. How would that fit into this PR?

internal/db/job_tracker_test.go
ale8k: Yup

internal/jobtracker/jobtracker.go
ale8k: Ah I see what you mean

internal/jobtracker/jobtracker.go
ale8k: Yup good spot, it's on the tracker now.

internal/jobtracker/jobtracker.go
ale8k: We're already cancelling it in the defer when we create the context, also feels strange the thing using the context cancels its own context?

internal/jobtracker/jobtracker.go
ale8k: Not sure what you mean? It needs the id?

ale8k:
> A few comments.. And we also discussed that we'd remove the controller if we failed to record its status or fail to add it to JIMM.. How would that fit into this PR?

The job (that we'll write) running the cmd will know when the command completes / fails, at which point it can remove the controller row we've created ahead of time.

internal/db/job_tracker.go
```diff
@@ -46,6 +46,27 @@ func (d *Database) AddJob(ctx context.Context, jobType string) (jobId uuid.UUID,
...
+		return false, errors.E(op, err)
+	}
+
+	durationObserver := servermon.DurationObserver(servermon.DBQueryDurationHistogram, string(op))
+	defer durationObserver()
+	defer servermon.ErrorCounter(servermon.DBQueryErrorCount, &err, string(op))
+
+	db := d.DB.WithContext(ctx)
+	var entry dbmodel.JobTrackerEntry
+	if err := db.Select("stop_signal").First(&entry, "job_id = ?", jobId).Error; err != nil {
```
alesstimec: db.Select("stop_signal").Where("job_id = ?", jobId).First(&entry)

internal/jobtracker/jobtracker.go
alesstimec: what i mean is don't do 
```
err := job(jobCtx)
jobErrCh <- err
```
if you can do 
```
jobErrCh <- job(jobCtx)
```
in a single line..

internal/jobtracker/jobtracker.go
alesstimec: sure sure sure.. but instead of having `SetJobFailed` and `SetJobSuccessful` have only one method that is `SetJobStatus` and pass in the status as `successful` or `failed`..

internal/jobtracker/jobtracker.go
alesstimec: hmm.. you know i'm in favor of doing it simple and add complexity if the need arises..

alesstimec (commented):
A few more comments, but the general shape looks ok

internal/jobtracker/jobtracker.go
```diff
@@ -0,0 +1,134 @@
...
+		return
+	}
+	jobErrCh <- job(jobCtx)
+}
+
+func (j *Tracker) monitorJob(ctx, jobCtx context.Context, id uuid.UUID, jobErrCh chan error, cancelJob context.CancelFunc) {
+	ticker := time.NewTicker(j.stopInterval)
+	defer ticker.Stop()
+
+	// TODO(ale8k): Add mo
```
alesstimec: dd or complete the comment

alesstimec (approved):

alesstimec:
image

internal/jobtracker/jobtracker.go
```diff
@@ -0,0 +1,134 @@
...
+	stopInterval time.Duration
+}
+
+// NewJobTracker creates and returns a new Tracker instance using the provided JobTrackerStore and stopInterval.
+// It returns an error if the store is nil or if stopInterval is not greater than zero.
+func NewJobTracker(store Store, stopInterval time.Duration) (*Tracker, error) {
+	if store == nil {
+		return nil, errors.New("store cannot be nil")
+	}
+	if stopInterval <= 0 {
```
kian99: We should give stopInterval a minimum, how does 5s sound?

internal/jobtracker/jobtracker.go
```diff
@@ -0,0 +1,134 @@
...
+}
+
+// Run runs a new job and returns the job ID.
+func (j *Tracker) Run(ctx context.Context, jobType string, job func(ctx context.Context) error, maxDuration time.Duration) (uuid.UUID, error) {
+	jobId, err := j.store.AddJob(ctx, jobType)
+	if err != nil {
+		return jobId, err
+	}
+
+	go j.handleJob(ctx, jobId, maxDuration, job)
```
kian99: You need to use a new context here. Or better yet, don't pass a context in, create one inside handleJob().
Otherwise the parent context can be cancelled and unintentionally cancel the job.

internal/jobtracker/jobtracker.go
```diff
@@ -0,0 +1,134 @@
...
+	maxDuration time.Duration,
+	job func(ctx context.Context) error,
+) {
+	jobCtx, cancelJob := context.WithTimeout(ctx, maxDuration)
+	defer cancelJob()
+
+	jobErrCh := make(chan error)
+
+	go j.runJob(ctx, jobCtx, id, jobErrCh, job)
+	j.monitorJob(ctx, jobCtx, id, jobErrCh, cancelJob)
```
kian99: Why do we have 2 contexts?

internal/jobtracker/jobtracker.go
```diff
@@ -0,0 +1,134 @@
...
+
+	// TODO(ale8k): Add monitoring for failed status settings.
+	for {
+		select {
+		case <-jobCtx.Done():
+			err := jobCtx.Err()
+			if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
+				if err := j.store.SetJobFailed(ctx, id, err); err != nil {
+					zapctx.Error(ctx, "error marking the job as failed", zap.Error(err), zap.String("id", id.String()))
+				}
```
kian99: If the context was done but it wasn't a cancel or deadlineExceeded, we haven't accounted for timeout errors for example, the job wouldn't have a status?
I thin we should always set the job to failed if the context is done.

internal/jobtracker/jobtracker_test.go
```diff
@@ -0,0 +1,157 @@
...
+	j := dbmodel.JobTrackerEntry{}
+	var jobErr string
+	c.Assert(s.db.DB.First(&j).Where("job_id = ?", id).Select("error").Scan(&jobErr).Error, qt.IsNil)
+	c.Assert(jobErr, qt.Equals, "context deadline exceeded")
+}
+
+func (s *jobTrackerSuite) TestRun_CancelledJob(c *qt.C) {
+	testCtx := c.Context()
+
+	// Create a new tracker, specific to this test.
```
kian99: Why does this test need a new tracker? I presume in order for the cancel to be polled more quickly than the default but a comment explaining that would be nice.

internal/jobtracker/jobtracker_test.go
```diff
@@ -0,0 +1,157 @@
...
+
+	s.pollJob(testCtx, id, c, dbmodel.StatusFailed)
+
+	j := dbmodel.JobTrackerEntry{}
+	var jobErr string
+	c.Assert(s.db.DB.First(&j).Where("job_id = ?", id).Select("error").Scan(&jobErr).Error, qt.IsNil)
+	c.Assert(jobErr, qt.Equals, "context canceled")
+}
+
+func (s *jobTrackerSuite) TestRun_JobSetRunning(c *qt.C) {
```
kian99, line 70: In this test we don't make sure the job finishes before completing the test. We can have the job finish after we've verified the status is running then do a `s.pollJob` right at the end.

internal/jobtracker/jobtracker.go
SimoneDutto: fair enough

SimoneDutto (approved):
lgtm with kian's comments addressed

internal/jobtracker/jobtracker.go
ale8k: Yeah I did think about this but didn't know what number to pluck from the sky 😆 Yeah I'll change to 5s.

internal/jobtracker/jobtracker.go
ale8k: One is the parent context, one is the context for the running job. The parent context is simply used for logging, and the job context is used for routine tracking.

internal/jobtracker/jobtracker.go
ale8k: You can only have canceled or deadline exceeded within a .Done()? (We aren't using .Cause anywhere)

internal/jobtracker/jobtracker.go
kian99: I don't think I'm following why we would need the parent context for logging.
I think you should delete the parent context and just use the one context that is created in `handleJob`

internal/jobtracker/jobtracker.go
kian99: If those are the only reasons then why are we checking for them? Can't we just set the job to failed anyway?

ale8k: Ahhh I get your question - We can't because they're different error reasons and we want the error field to explain exactly what has happened. If it's cancelled for whatever reason, we may want to know explicitly that's what happened. Versus it simply hitting the deadline?

internal/jobtracker/jobtracker.go
ale8k: I'll create one to alleviate the concern of the parent being cancelled. But it wouldn't cancel the job if the parent was cancelled, it'd just prevent us updating the status:
```go
		case <-jobCtx.Done():
			err := jobCtx.Err()
			if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
				if err := j.store.SetJobFailed(ctx, id, err); err != nil {
					zapctx.Error(ctx, "error marking the job as failed", zap.Error(err), zap.String("id", id.String()))
				}
			}
			return
```

ale8k: Created a "parentCtx" in Run(...)

internal/jobtracker/jobtracker_test.go
ale8k: That's fair, will add a check to see it is successful

kian99 (commented):
Can we also get rid of the 2 contexts on 1 function before merging. That doesn't make sense. We don't need a separate context for logging. At least from what I can tell

PR #1638, opened 2025-07-03
feat: `qa-lxd-multipass` scripts

andogq:
## Description

- `qa-lxd-multipass`: Start JIMM in a new multipass VM, with a controller running within an LXD container
- `qa-lxd-multipass-forward`: Port forward `443` (JIMM via Traefik) and `8082` (keycloak) to host using SSH

Both these scripts should work on MacOS and Linux, given that multipass is installed.

WD-23338


cmd/jimmsrv/service/service.go
```diff
@@ -375,7 +375,7 @@ func NewService(ctx context.Context, p Params) (*Service, error) {
 	})
 
 	redirectUrl := p.PublicDNSName + jimmhttp.AuthResourceBasePath + jimmhttp.CallbackEndpoint
-	if !strings.HasPrefix(redirectUrl, "https://") || !strings.HasPrefix(redirectUrl, "http://") {
+	if !strings.HasPrefix(redirectUrl, "https://") && !strings.HasPrefix(redirectUrl, "http://") {
```
ale8k, line 396: Does this need changing? Because this would only apply to when it has no protocol?

local/jimm/qa-lxd-multipass.sh
```diff
@@ -0,0 +1,83 @@
...
+
+echo "To continue the QA environment setup, please login to keycloak."
+echo "The username is: \"jimm-test\" and the password is: \"password\""
+multipass exec --working-directory /home/ubuntu/jimm $VM_NAME -- juju login jimm.localhost -c jimm-dev
+
+cleanup_ssh_forward
+
+echo
+echo "Setting up LXD controller"
+multipass exec --working-directory /home/ubuntu/jimm $VM_NAME -- sudo iptables -I FORWARD -i lxdbr0 -j ACCEPT
```
ale8k: Could use heredoc here:
```
multipass exec dev -- bash << 'EOF'
...
EOF
```

ale8k (approved):
lgtm bar the &&. And possible use herdoc's, will read clearer

cmd/jimmsrv/service/service.go
andogq: I don't need this any more, however I do still believe the logic is wrong.

Previously, `http://localhost:1234` would result in `!false || !true` => `true || false` => `true`, therefore resulting in `https://http://localhost:1234`. Same if it originally begins with `https://`.

This only functions correctly if `p.PublicDNSName` doesn't contain a scheme (`!false || !false` => prepend `https://`).

However, if `p.PublicDNSName` is truely a DNS name, it shouldn't contain a scheme anyway, so potentially none of this even matters and `https` can always be assumed.

Anyway, lmk what you want done with this, happy to just revert it if you don't want to worry about it for now.

SimoneDutto:
@andogq i was trying to qa the two new commands, but i believe they are not added to the `Makefile` file.
Why? If you add them to the makefile just like `qa-microk8s` and `qa-lxd` would be better!

andogq:
Thanks @SimoneDutto. I've added `qa-multipass` which will setup the multipass container (should only need to be run once), and `qa-multipass-forward` which will setup the port forwards (will need to be run anytime the JIMM instance is needed).
Let me know of that works for you.

cmd/jimmsrv/service/service.go
kian99: Good spot, thanks!

kian99 (approved):

SimoneDutto:
> Thanks @SimoneDutto. I've added `qa-multipass` which will setup the multipass container (should only need to be run once), and `qa-multipass-forward` which will setup the port forwards (will need to be run anytime the JIMM instance is needed). Let me know of that works for you.

thanks

SimoneDutto (approved):

PR #1639, opened 2025-07-03
feat(model-migration): cleanup failed or partial model migrations

SimoneDutto:
## Description
In this pr we introduce two cleanups: the first is just to remove from JIMM's state the model created during Import on Abort. The second is a routine that runs periodically and it removes the stale migrations with eventual user mappings and models.

To accomplish this:
- created db method to query all stale model migrations
- created db model to delete user mapping associated to a model_uuid
- add method to juju manager to do the cleanup
- improve the tests by adding the model-migration to our `env`

### DIFF
--- a/cmd/jimmsrv/service/service.go
+++ b/cmd/jimmsrv/service/service.go
@@ -51,7 +51,8 @@ import (
 )
 
 const (
-	localDischargePath = "/macaroons"
+	localDischargePath                        = "/macaroons"
+	INTERVAL_BETWEEN_MODEL_MIGRATIONS_CLEANUP = 10 * time.Minute
 )
 
 // OpenFGAParams holds parameters needed to connect to the OpenFGA server.
@@ -285,6 +286,23 @@ func (s *Service) CleanupNotFoundModels(ctx context.Context, trigger <-chan time
 			}
 		case <-ctx.Done():
 			zapctx.Info(ctx, "exiting dying model cleanup polling")
+			return nil
+		}
+	}
+}
+
+// CleanupPartialModelMigrations triggers every `trigger` time and calls the jimm methods to cleanup partial model migrations.
+func (s *Service) CleanupPartialModelMigrations(ctx context.Context, trigger <-chan time.Time) error {
+	for {
+		select {
+		case <-trigger:
+			err := s.jimm.JujuManager().CleanupPartialModelMigrations(ctx)
+			if err != nil {
+				zapctx.Error(ctx, "partial model migrations cleanup", zap.Error(err))
+				continue
+			}
+		case <-ctx.Done():
+			zapctx.Info(ctx, "exiting partial model migration cleanup polling")
 			return nil
 		}
 	}
@@ -560,6 +578,11 @@ func (s *Service) StartServices(ctx context.Context, svc *service.Service) {
 		// CleanupNotFoundModels cleanup - cleans up all models not found on the respective controller.
 		svc.Go(func() error {
 			return s.CleanupNotFoundModels(ctx, time.NewTicker(time.Minute).C)
+		})
+
+		// CleanupPartialModelMigration cleanup - cleans up all partial model migrations.
+		svc.Go(func() error {
+			return s.CleanupPartialModelMigrations(ctx, time.NewTicker(INTERVAL_BETWEEN_MODEL_MIGRATIONS_CLEANUP).C)
 		})
 	}
 
--- a/internal/jimm/juju/migrationtarget.go
+++ b/internal/jimm/juju/migrationtarget.go
@@ -5,6 +5,7 @@ package juju
 import (
 	"context"
 	"database/sql"
+	goerr "errors"
 	"fmt"
 	"time"
 
@@ -25,8 +26,10 @@ import (
 	"github.com/canonical/jimm/v3/internal/openfga"
 )
 
+const TIMEOUT_PENDING_MIGRATION = 24 * time.Hour
+
 // AbortMigration aborts a model migration with the given model UUID.
-// It does this by calling the Abort methodon the target Juju controller.
+// It does this by calling the Abort method on the target Juju controller.
 // It also deletes the migration record from the database, but does not return an error
 // if the deletion fails, as the migration has already been aborted on the target controller.
 func (j *JujuManager) AbortMigration(ctx context.Context, user *openfga.User, modelUUID string) error {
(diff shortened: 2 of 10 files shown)


internal/jimm/juju/migrationtarget.go
```diff
@@ -60,7 +63,27 @@ func (j *JujuManager) AbortMigration(ctx context.Context, user *openfga.User, mo
...
+			Valid:  true,
+		},
+	}
+	err = j.Database.GetModel(ctx, &model)
+	if err != nil {
+		if errors.ErrorCode(err) == errors.CodeNotFound {
+			return nil
+		}
+		// Don't return an error if we fail to get the model.
+		zapctx.Error(ctx, "failed to get model for abort migration", zap.Error(err), zap.String("modelUUID", modelUUID))
```
kian99, line 80: Why not? To avoid confusing Juju? Ah I see the reasoning lower down :+1: 
Consider moving the explanation up higher

internal/jimm/juju/migrationtarget.go
```diff
@@ -466,3 +489,68 @@ func (j *JujuManager) importModelFromDescription(ctx context.Context, targetCont
...
 }
+
+// CleanupPartialModelMigrations cleans up any partial model migrations that have exceeded the timeout.
+// It deletes the incoming model migration record, deletes the user mappings for the model,
+// and deletes the model record from JIMM's state.
+func (j *JujuManager) CleanupPartialModelMigrations(ctx context.Context) error {
+	const op = errors.Op("jimm.CleanupPartialModelMigrations")
+
+	// Get all incoming model migrations that have exceeded the timeout.
+	migrations, err := j.Database.GetIncomingModelMigrationsCreatedBefore(ctx, time.Now().Add(-TIMEOUT_PENDING_MIGRATION))
```
kian99, line 500: Another option but fine with the current name,
```suggestion
	migrations, err := j.Database.GetIncomingModelMigrationsByCreatedDate(ctx, time.Now().Add(-TIMEOUT_PENDING_MIGRATION))
```

internal/jimm/juju/migrationtarget.go
```diff
@@ -466,3 +489,68 @@ func (j *JujuManager) importModelFromDescription(ctx context.Context, targetCont
...
+		err = j.Database.DeleteUserMappingsByModelUUID(ctx, migration.ModelUUID.String)
+		if err != nil {
+			return errors.E(op, err)
+		}
+
+		// Delete the model record from JIMM's state.
+		// We don't need to check the migration-mode here, because we set the migration-mode to None
+		// in the same transaction as the deletion of the incoming model migration record.
+		// This ensures that there is a dangling incoming model migration record
+		// then the model can't have the migration-mode set to none.
```
kian99: I think there is a missing "if" but even with it, this is very confusing to read and I mostly get it I think. I'm inclined towards just removing the entire comment. We understand that the model hasn't been successfully migrated if we still have an `incomingModelMigration` record (that in itself could replace the comment).
```suggestion
		// We don't need to check the migration-mode here, because we set the migration-mode to None
		// in the same transaction as the deletion of the incoming model migration record.
		// This ensures that if there is a dangling incoming model migration record
		// then the model can't have the migration-mode set to none.
```

kian99 (commented):
Looking good with some comments

cmd/jimmsrv/service/service.go
```diff
@@ -561,6 +578,11 @@ func (s *Service) StartServices(ctx context.Context, svc *service.Service) {
 		svc.Go(func() error {
 			return s.CleanupNotFoundModels(ctx, time.NewTicker(time.Minute).C)
 		})
+
+		// CleanupPartialModelMigration cleanup - cleans up all partial model migrations.
+		svc.Go(func() error {
+			return s.CleanupPartialModelMigrations(ctx, time.NewTicker(time.Minute).C)
```
alesstimec: would we want this period configurable? idk, minute sounds good.. but at least make it a constant

internal/jimm/juju/migrationtarget.go
```diff
@@ -60,7 +63,27 @@ func (j *JujuManager) AbortMigration(ctx context.Context, user *openfga.User, mo
...
 		zapctx.Error(ctx, "failed to delete incoming model migration", zap.Error(err), zap.String("modelUUID", modelUUID))
 	}
-	// TODO(SimoneDutto): delete the model from JIMM's state.
+	model := dbmodel.Model{
+		UUID: sql.NullString{
+			String: modelUUID,
+			Valid:  true,
+		},
+	}
+	err = j.Database.GetModel(ctx, &model)
```
alesstimec, line 75: why do we need to Get the model just to delete it? can't we just attempt to delete the model?

internal/jimm/juju/migrationtarget.go
```diff
@@ -466,3 +489,68 @@ func (j *JujuManager) importModelFromDescription(ctx context.Context, targetCont
...
+		// in the same transaction as the deletion of the incoming model migration record.
+		// This ensures that there is a dangling incoming model migration record
+		// then the model can't have the migration-mode set to none.
+		model := dbmodel.Model{
+			UUID: sql.NullString{
+				String: migration.ModelUUID.String,
+				Valid:  true,
+			},
+		}
+		err = j.Database.GetModel(ctx, &model)
```
alesstimec, line 539: same as the other comment: why do we need to Get just to Delete the model?

SimoneDutto: we don't have a delete by `uuid` method. The default gorm delete is by `id`, so you need to get it first.
Didn't want to add the feature in this PR, but i definetely can.

internal/jimm/juju/migrationtarget.go
SimoneDutto: removed.

cmd/jimmsrv/service/service.go
```diff
@@ -51,7 +51,8 @@ import (
 )
 
 const (
-	localDischargePath = "/macaroons"
+	localDischargePath                        = "/macaroons"
+	INTERVAL_BETWEEN_MODEL_MIGRATIONS_CLEANUP = time.Minute
```
kian99: I suggest making this longer, like 5 or 10 minutes. Since it's cleaning things up that are >24 hours old, a minute seems too frequent.

internal/jimm/juju/migrationtarget_test.go
```diff
@@ -873,3 +846,152 @@ func TestImport_APIFailure(t *testing.T) {
 	c.Assert(err, qt.IsNil)
 	c.Assert(m.MigrationMode, qt.Equals, state.MigrationModeImporting)
 }
+
+// environment for testing cleanup of partial model migrations.
+// It contains:
+// - one valid incoming migration to be kept
+// - one stale incoming migration without user-mapping and a model to be deleted
+// - one stale incoming migration with model to be deleted
+// - one state incoming migration
```
kian99:
```suggestion
// - one stale incoming migration
```

kian99 (approved):

internal/jimm/juju/migrationtarget.go
alesstimec: let's add a method so that we can simplify this.

alesstimec (approved):

PR #1640, opened 2025-07-04
fix: incorrect logic during model migration

kian99:
## Description
This PR fixes 3 issues in the model migration logic:
1. The `Abort` RPC method was calling the `AdoptResources` domain function.
2. In `Activate` the controller-tag field is optional but was always being checked.
3. The `AdoptResources` method is called after `Activate`. By this point we have deleted the `IncomingModelMigration` entry. Ideally we should always look for the `model` in any case since we create it near the start of model migration and only use the `IncomingModelMigration` entry for holding the user-mapping.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

## Test instructions
I've identified these issues during my manual QA. An automated QA script will come in a follow-up.

### DIFF
--- a/internal/jimm/juju/migrationtarget_test.go
+++ b/internal/jimm/juju/migrationtarget_test.go
@@ -309,13 +309,18 @@ func TestAdoptResources_Success(t *testing.T) {
 	c := qt.New(t)
 	ctx := context.Background()
 
+	// Note that adopt resources is called after the model
+	// has been activated so the incoming model migration
+	// row was deleted and the model has been created.
+
 	controllerVersion := version.MustParse("3.2.1")
+	modelUUID := "00000002-0000-0000-0000-000000000001"
 
 	// Validate that the API request to Juju is made with a modified version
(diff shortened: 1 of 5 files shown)


internal/jimm/juju/migrationtarget_test.go
```diff
@@ -297,13 +297,18 @@ func TestAdoptResources_Success(t *testing.T) {
 	c := qt.New(t)
 	ctx := context.Background()
 
+	// Note that adopt resources is called after the model
+	// has been activated so the incoming model migration
+	// row was deleted and the model has been created.
```
SimoneDutto, line 314: Can you duplicate this comment in the godoc of the AdoptResource method?

SimoneDutto (approved):
LGTM

internal/jimm/juju/migrationtarget_test.go
kian99: Will do

alesstimec (approved):

PR #1641, opened 2025-07-04
feat: add support for running juju cmds

ale8k:
## Description
This PR brings in the support to run juju cmds. The new package will contain commands in their own respective files, i.e., `bootstrap.go` The commands will use `runCmdWithOutputRetriever` which returns a channel callers can read from and it is closed once the command completes. It guarantees two conditions:
1. All output has been read from the command.
2. The command has finished.
Should the command error, the final `outputLine` contains an error with the code returned. We could capture the previous line and encapsulate this into the error (as the last line [I think?] is always the error). But for now, I think this should suffice. 
  
## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests

### DIFF
--- a/internal/jujucommands/jujucommands.go
+++ b/internal/jujucommands/jujucommands.go
@@ -46,6 +46,8 @@ type outputLine struct {
 //	}
 //
 // The OutputLine channel is closed once all output from the command has finished an the error has been captured.
+//
+// Lines returned are returned with no newlines.
 func runCmdWithOutputRetriever(store jujuclient.ClientStore, cmdAndArgs string) (<-chan outputLine, error) {
 	cmdReader, cmdWriter := io.Pipe()
 
--- a/internal/jujucommands/jujucommands.go
+++ b/internal/jujucommands/jujucommands.go
@@ -58,20 +58,17 @@ func runCmdWithOutputRetriever(store jujuclient.ClientStore, cmdAndArgs string) 
 	cmdCtx.Stdout = cmdWriter
 
 	outputCh := make(chan outputLine)
-	// We buffer cmdFinishedCh with capacity 1 to avoid a deadlock here.
-	// In runCmd, the writer (cmdWriter) is deferred to close only after runCmd completes.
-	// When sending the final error down cmdFinishedCh, if it's unbuffered,
-	// the send blocks because no goroutine is receiving yet.
-	// This blocks runCmd from completing, so the deferred writer.Close() never runs,
-	// meaning no EOF is sent to the reader, which keeps blocking forever.
-	// By buffering cmdFinishedCh, the send won't block, allowing runCmd to complete,
-	// the writer to close, and the reader to receive EOF and finish reading.
+	// We buffer cmdFinishedCh (capacity 1) to avoid deadlock.
+	// runCmd defers cmdWriter.Close(), which only runs after runCmd finishes.
+	// If cmdFinishedCh were unbuffered, sending the final error would block
+	// (since no goroutine is receiving yet), which prevents runCmd from finishing.
+	// This means cmdWriter.Close() never runs, so the reader never gets EOF and hangs.
+	// With a buffered channel, the send doesn't block, allowing runCmd to finish,
+	// the writer to close, and the reader to eventually see EOF and exit.
 	//
-	// There's no need to worry about it blocking, as we'll only ever send one value down
-	// once the command completes.
-	//
-	// We could alternatively manually call cmdWriter.Close() immediately after running the command.
-	cmdFinishedCh := make(chan error, 1)
+	// This is safe because we only send once, after the command finishes.
+	// Alternatively, we could manually close cmdWriter right after the command.
+	cmdFinishedCh := make(chan int, 1)
 
 	go func() {
 		// Read cmdoutput.
@@ -81,9 +78,9 @@ func runCmdWithOutputRetriever(store jujuclient.ClientStore, cmdAndArgs string) 
 		}
 
 		// Wait for cmd to finish.
-		err := <-cmdFinishedCh
-		if err != nil {
-			outputCh <- outputLine{Err: err}
+		code := <-cmdFinishedCh
+		if code != 0 {
+			outputCh <- outputLine{Err: fmt.Errorf("cmd failed with code: %d", code)}
 		}
 
 		close(outputCh)
@@ -98,14 +95,14 @@ func runCmdWithOutputRetriever(store jujuclient.ClientStore, cmdAndArgs string) 
 // The function splits cmdAndArgs into arguments, constructs a Juju command, and runs it.
 // On non-zero exit code, it sends an error to cmdFinishedCh; otherwise, it signals successful completion.
 //
-// Note, the cmdAndArgs argument expects a fully-qualified command strings. I.e.,
+// Note, the cmdAndArgs argument expects a fully-qualified command string. I.e.,
 //
 //	"bootstrap lxd a --add-model=\"my-initial-model\""
 func runCmd(
 	cmdCtx *cmd.Context,
 	cmdWriter *io.PipeWriter,
 	store jujuclient.ClientStore,
-	cmdFinishedCh chan<- error,
+	cmdFinishedCh chan<- int,
 	cmdAndArgs string,
 ) {
 	defer cmdWriter.Close()
@@ -128,9 +125,5 @@ func runCmd(
 
 	code := cmd.Main(jujuCmd, cmdCtx, strings.Split(cmdAndArgs, " "))
 
-	if code != 0 {
-		cmdFinishedCh <- fmt.Errorf("cmd failed with code %d", code)
-	} else {
-		cmdFinishedCh <- nil
-	}
+	cmdFinishedCh <- code
 }
--- a/internal/jujucommands/jujucommands.go
+++ b/internal/jujucommands/jujucommands.go
@@ -1,0 +1,132 @@
+// Copyright 2025 Canonical.
+
+// Package jujucommands provides functions run juju cmds from a JIMM instance.
+// Each command function is run with its own isolated in-mem store.
+
+package jujucommands
+
+import (
+	"fmt"
+	"io"
+	"strings"
+
+	"github.com/juju/cmd/v3"
+	"github.com/juju/juju/cmd/juju/commands"
+	"github.com/juju/juju/jujuclient"
+	"github.com/mitchellh/go-linereader"
+)
+
+var (
+	whitelist = []string{"help", "bootstrap"}
+)
+
+type outputLine struct {
+	Line string
+	Err  error
+}
+
+// runCmdWithOutputRetriever runs the command given by cmdAndArgs using the provided jujuclient.ClientStore.
+// It returns a channel that streams OutputLine values for each line of command output and any final error.
+//
+// Upon receiving an error in an OutputLine, the line previously streamed is guaranteed to be the error line written by
+// the command.
+// I.e., if you run: "help -b"
+// You will get:
+//
+//	Output line: ERROR option provided but not defined: -b // The error line from juju (Which is a none-erroneous OutputLine)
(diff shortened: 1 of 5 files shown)


internal/jujucommands/jujucommands.go
```diff
@@ -0,0 +1,148 @@
...
+func runCmd(
+	cmdCtx *cmd.Context,
+	cmdWriter *io.PipeWriter,
+	store jujuclient.ClientStore,
+	cmdFinishedCh chan<- error,
+	cmdAndArgs string,
+) {
+	defer cmdWriter.Close()
+
+	jujuCmd := commands.NewJujuCommandWithStore(
```
kian99, line 112: Question on this, why don't we use `jujuCmd.Run(cmdCtx)` to execute the command? It seems that would return the error separately which might be easier to use.
It seems like we can also do `jujuCmd.Init(args)` `jujuCmd.SetFlags(flags)`.

internal/jujucommands/jujucommands.go
```diff
@@ -0,0 +1,148 @@
...
+var (
+	whitelist = []string{"help", "bootstrap"}
+)
+
+type outputLine struct {
+	Line string
+	Err  error
+}
+
+// runCmdWithOutputRetriever runs the command given by cmdAndArgs using the provided jujuclient.ClientStore.
```
kian99, line 27: This command would be more helpful on a function/method that is exported.
Should this function be exported? Or do we expect to have something else wrapping this?

internal/jujucommands/jujucommands.go
```diff
@@ -0,0 +1,148 @@
...
+	// the send blocks because no goroutine is receiving yet.
+	// This blocks runCmd from completing, so the deferred writer.Close() never runs,
+	// meaning no EOF is sent to the reader, which keeps blocking forever.
+	// By buffering cmdFinishedCh, the send won't block, allowing runCmd to complete,
+	// the writer to close, and the reader to receive EOF and finish reading.
+	//
+	// There's no need to worry about it blocking, as we'll only ever send one value down
+	// once the command completes.
+	//
+	// We could alternatively manually call cmdWriter.Close() immediately after running the command.
```
kian99: Reading this block a few times helped it make sense to me. But I'm curious what others think about simplifying it down. Something like,
"Use a buffered channel to avoid blocking `runCmd` when the command completes, which is responsible for closing `cmdWriter` and indicating that no more lines are available.".

internal/jujucommands/jujucommands.go
```diff
@@ -0,0 +1,148 @@
...
+		// SetFlags initializes the flags supported by the command.
+		// func (c *OptionalControllerCommand) SetFlags(f *gnuflag.FlagSet) {
+		// 	c.CommandBase.SetFlags(f)
+		// 	// Embedded commands do not use the --client or --controller options.
+		// 	if !c.Embedded {
+		// 		f.BoolVar(&c.Client, "client", false, "Client operation")
+		// 		f.StringVar(&c.ControllerName, "c", "", "Controller to operate in")
+		// 		f.StringVar(&c.ControllerName, "controller", "", "")
+		// 	}
+		// }
```
kian99: Can we remove this block of code in the comment and just say "refer to ..."

kian99 (commented):
Looks good just some comments

internal/jujucommands/jujucommands.go
```diff
@@ -0,0 +1,148 @@
...
+//	for out := range outputCh {
+//		if out.Err != nil {
+//			fmt.Println("Command finished with error:", out.Err)
+//			break
+//		}
+//		fmt.Println("Output line:", out.Line)
+//	}
+//
+// The OutputLine channel is closed once all output from the command has finished an the error has been captured.
+func runCmdWithOutputRetriever(store jujuclient.ClientStore, cmdAndArgs string) (<-chan outputLine, error) {
```
SimoneDutto, line 51: I have a question:
why do we need a channel here? Couldn't we just return a reader and in the consumer side you decide whether or not read line by line, all together, or anything really?

SimoneDutto (commented):
i have just one question and then LGTM

internal/jujucommands/jujucommands.go
SimoneDutto: agree

internal/jujucommands/jujucommands.go
```diff
@@ -0,0 +1,148 @@
+// Copyright 2025 Canonical.
+
+// Package jujucommands provides functions run juju cmds from a JIMM instance.
+// Each command function is run with its own isolated in-mem store.
+
```
alesstimec: dd

internal/jujucommands/jujucommands.go
alesstimec: agreed.. either use a buffered channel and also do
```
select {
  case outputCh <- outputLine{Line: line}:
  default:
}
```
so, try to write to the channel and just skip if the consumer isn't reading fast enough

internal/jujucommands/jujucommands.go
```diff
@@ -0,0 +1,148 @@
...
+	"strings"
+
+	"github.com/juju/cmd/v3"
+	"github.com/juju/juju/cmd/juju/commands"
+	"github.com/juju/juju/jujuclient"
+	"github.com/mitchellh/go-linereader"
+)
+
+var (
+	whitelist = []string{"help", "bootstrap"}
```
alesstimec, line 19: why are we whitelisting the help command?

internal/jujucommands/jujucommands.go
kian99: I was thinking about this and thought something similar. I think we don't need to run the command from this package, because our job tracker is already starting a go routine to run the job in the background - if we start another go routine here it might be unnecessary.

What if we return a struct that wraps the Juju command and just exposes a `Run()` method that returns an `(io.Reader, error)`, then the consumer can decide to run it in a blocking fashion or run it in a go routine.

internal/jujucommands/jujucommands.go
ale8k: We can, but there's two paths - A silent error and sometimes underlying details which are printed in a user friendly fashion. For example, it tries to bootstrap but cannot find the `PersonalCloudMetadata`, in this case the error would be from os.ReadFile which isn't helpful, it will have a nicely formatted written line just before though to be displayed to the user. So I don't think we need the underlying error for our use case honestly. We just care when the command has finished.

internal/jujucommands/jujucommands.go
ale8k: No no not exported. The comment was more to help understand how to use it. Commands in this package will use this.

internal/jujucommands/jujucommands.go
ale8k: Fair, will update!

internal/jujucommands/jujucommands.go
ale8k: I don't think another routine really matters, that's the strength of golang. 

We shouldn't return interfaces, we can return an implementation though. I'm not sure why you'd only want to read part of the command though... We'd also always be interested in full-lines. Perhaps a call on this as I think there's a misunderstanding possibly.

internal/jujucommands/jujucommands.go
ale8k: For the test aha.

internal/jujucommands/jujucommands.go
kian99: Just a ping on my bit above about simplifying this explanation.

internal/jujucommands/jujucommands.go
```diff
@@ -0,0 +1,136 @@
...
+
+	return outputCh, nil
+}
+
+// runCmd executes a Juju command using the provided context, client store, and command arguments.
+// It writes command output to the given io.PipeWriter and signals completion or error via cmdFinishedCh.
+// The function splits cmdAndArgs into arguments, constructs a Juju command, and runs it.
+// On non-zero exit code, it sends an error to cmdFinishedCh; otherwise, it signals successful completion.
+//
+// Note, the cmdAndArgs argument expects a fully-qualified command strings. I.e.,
```
kian99:
```suggestion
// Note, the cmdAndArgs argument expects a fully-qualified command string. I.e.,
```

internal/jujucommands/jujucommands.go
kian99: Gotcha. And then, should we at least use the `Init()` and `SetFlags` methods on the command so that we are not passing around a string that contains flags and arguments? Or is it preferred to pass around strings and then pass that string to cmd.Main()

internal/jujucommands/jujucommands.go
```diff
@@ -0,0 +1,136 @@
...
+//
+// Upon receiving an error in an OutputLine, the line previously streamed is guaranteed to be the error line written by
+// the command.
+// I.e., if you run: "help -b"
+// You will get:
+//
+//	Output line: ERROR option provided but not defined: -b // The error line from juju (Which is a none-erroneous OutputLine)
+//	Command finished with error: cmd failed with code 2 // The command error code line (Contains an error in the OutputLine)
+//
+// Consumers are expected to simply read from the OutputLine channel and check for an error like so:
```
kian99, line 38: I'd add a section/line to explain that if the consumer is not timely in consuming lines from the output channel, then the command execution will block.

internal/jujucommands/jujucommands_test.go
```diff
@@ -0,0 +1,73 @@
...
+func (s *jujucommandsSuite) TestRunCmdWithOutputRetriever(c *qt.C) {
+	outputCh, err := jujucommands.RunCmdWithOutputRetriever(jujuclient.NewEmbeddedMemStore(), "help")
+	c.Assert(err, qt.IsNil)
+
+	// Use a builder to collect streamed output & test entire string.
+	var b strings.Builder
+
+	for out := range outputCh {
+		c.Assert(out.Err, qt.IsNil)
+		b.WriteString(out.Line + "\n")
```
kian99, line 27: It wasn't documented above that the lines from the output channel will not have a new-line character. I suggest adding that fact to the docstring.

kian99 (approved):

internal/jujucommands/jujucommands.go
ale8k: I've tried to simplify it, will push soon (had AI help me aha)

internal/jujucommands/jujucommands.go
ale8k: As discussed, leaving as is.

internal/jujucommands/jujucommands.go
ale8k: I just checked to see if we could run the command manually without wrapping it in main and have our own registry, then using init & flags ourselves makes sense, but we can't because:
```go
func newBootstrapCommand() cmd.Command {
	command := &bootstrapCommand{}
	command.clock = jujuclock.WallClock
	command.CanClearCurrentModel = true
	return modelcmd.Wrap(command,
		modelcmd.WrapSkipModelFlags,
		modelcmd.WrapSkipDefaultModel,
	)
}
```
The command is private unfortunately. 

I did try setting flags manually using .Set, but because they aren't registered yet we'd have to register them otherwise we get an error of: ` no such option -format`

And... If we add the flag "format" and then run the command, we get a panic of: `option redefinition`.

So we're outta luck.

internal/jujucommands/jujucommands.go
ale8k: I'm not certain it blocks, I'm guessing it does between the cmd I/o is piped and reader would hold it up - but... Without checking I'm not 100% - it could buffer somewhere.

internal/jujucommands/jujucommands.go
ale8k: Resolving... Ping me on element if the rewrite isn't ok. I don't get notification on resolved comments.

SimoneDutto (approved):

PR #1642, opened 2025-07-04
Add bootstrap command

ale8k:
## Description
Implements bootstrap cmd for JIMM.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests
- [x] Covered by QA test 

## Test instructions
Q/A via:
Set lxd to be available on 8443 `lxc config set core.https_address "[::]:8443"`

Run juju add-cloud --client
	Type: lxd
	Name: lxdtest
	Endpoint: https://127.0.0.1:8443 (It should show cert auth is available when set)
	Rest just press enter

```go
// Copyright 2025 Canonical.

package main

import (
	"context"
	"fmt"
	"os"
	"path/filepath"

	jujucloud "github.com/juju/juju/cloud"

	"github.com/canonical/jimm/v3/internal/jujucommands"
)

func main() {
	ctx := context.Background()

	absPath, err := filepath.Abs("./jujudata")
	if err != nil {
		fmt.Fprintf(os.Stderr, "Error getting absolute path: %v\n", err)
		os.Exit(1)
	}
	runner := jujucommands.NewCommandRunner(absPath)
	cmd := jujucommands.NewBootstrapCmd(runner)

	cred := jujucloud.NewCredential(jujucloud.CertificateAuthType, map[string]string{
		"server-cert": ``,
		"client-cert": ``,
		"client-key": ``,
	})

	cloudCred := *jujucloud.NewEmptyCloudCredential()
	// This is the default region's credential (when a credential is added, it is always tied
	// to the provided CloudNameAndRegion) implicitly by our command.
	cloudCred.AuthCredentials["default"] = cred

	p := jujucommands.BootstrapCmdParams{
		CloudNameAndRegion:   "testing-lxd",
		ControllerName:       "my-controller",
		LoginTokenRefreshURL: "http://myurl.com",

		PersonalCloud: jujucloud.Cloud{
			Type: "lxd",
			Name: "testing-lxd",
			AuthTypes: jujucloud.AuthTypes{
				jujucloud.CertificateAuthType,
			},
			Endpoint: "https://127.0.0.1:8443",
			Regions: []jujucloud.Region{
				{
					Name:     "default",
					Endpoint: "https://127.0.0.1:8443",
				},
			},
		},
		CloudCred: cloudCred,
	}
	fmt.Println("Starting")
	retriever, store, cleanup, err := cmd.Run(ctx, p)
	if err != nil {
		panic(err)
	}
	_ = cleanup
	_ = store

	for line := range retriever {
		if line.Err != nil {
			fmt.Println(line.Err)
			os.Exit(1)
		}
		fmt.Println(line.Line)
	}
}

```


internal/jujucommands/bootstrap.go
```diff
@@ -0,0 +1,224 @@
...
+// it returns a closure which cleans up the temporary $JUJU_DATA directory created for the
+// lifetime of this command.
+func RunBootstrapCmd(
+	ctx context.Context,
+	p BootstrapCmdParams,
+	personalCloud jujucloud.Cloud,
+	cred jujucloud.CloudCredential,
+	pubKey []byte,
+	privKey []byte,
+) (<-chan outputLine, jujuclient.ClientStore, func(), error) {
```
alesstimec: perhaps we could define a type to be returned by this function that would have these fields..

internal/jujucommands/jujucommands.go
```diff
@@ -0,0 +1,148 @@
...
+	"strings"
+
+	"github.com/juju/cmd/v3"
+	"github.com/juju/juju/cmd/juju/commands"
+	"github.com/juju/juju/jujuclient"
+	"github.com/mitchellh/go-linereader"
+)
+
+var (
+	whitelist = []string{"help", "bootstrap", "update-public-clouds"}
```
alesstimec: we don't need `help`

internal/jujucommands/jujucommands.go
```diff
@@ -0,0 +1,148 @@
...
+//		if out.Err != nil {
+//			fmt.Println("Command finished with error:", out.Err)
+//			break
+//		}
+//		fmt.Println("Output line:", out.Line)
+//	}
+//
+// The OutputLine channel is closed once all output from the command has finished an the error has been captured.
+func runCmdWithOutputRetriever(store jujuclient.ClientStore, cmdAndArgs string) (<-chan outputLine, error) {
+	// memStore := jujuclient.NewEmbeddedMemStore()
```
alesstimec: dd

internal/jujucommands/jujucommands.go
```diff
@@ -0,0 +1,148 @@
...
+	defer cmdWriter.Close()
+
+	jujuCmd := commands.NewJujuCommandWithStore(
+		cmdCtx,
+		store,
+		nil,       // quiet?
+		"",        // unknown param
+		"",        // no help hint
+		whitelist, // whitelist
+		// We set embedded false because some commands can targets either a controller or client.
```
alesstimec: s/targets/target/

internal/jujucommands/jujucommands.go
```diff
@@ -0,0 +1,148 @@
...
+
+		// Wait for cmd to finish.
+		err := <-cmdFinishedCh
+		if err != nil {
+			outputCh <- outputLine{Err: err}
+		}
+
+		close(outputCh)
+	}()
+	go runCmd(cmdCtx, cmdWriter, store, cmdFinishedCh, cmdAndArgs)
```
alesstimec: are we passing in cmdWriter just so that we can close it? if so, you can simply close it after L86 above when you know the command has finished.

internal/jujucommands/jujucommands_test.go
```diff
@@ -0,0 +1,73 @@
...
+	"github.com/frankban/quicktest/qtsuite"
+	"github.com/juju/juju/jujuclient"
+
+	"github.com/canonical/jimm/v3/internal/jujucommands"
+)
+
+type jujucommandsSuite struct{}
+
+func (s *jujucommandsSuite) TestRunCmdWithOutputRetriever(c *qt.C) {
+	outputCh, err := jujucommands.RunCmdWithOutputRetriever(jujuclient.NewEmbeddedMemStore(), "help")
```
alesstimec: whitelisting the `help` commands seems a bit much just for this test.. you can always run `juju update-public-clouds --help` for this test..

internal/jujucommands/bootstrap.go
```diff
@@ -0,0 +1,224 @@
...
+// BootstrapCmdParams holds the parameters to bootstrap a controller for JIMM.
+type BootstrapCmdParams struct {
+	CloudNameAndRegion   string
+	ControllerName       string
+	AgentVersion         string
+	BootstrapTimeout     int
+	LoginTokenRefreshURL string
+}
+
+func (b BootstrapCmdParams) validate() error {
```
alesstimec: no unit tests for this..

internal/jujucommands/bootstrap.go
```diff
@@ -0,0 +1,224 @@
...
+	}
+
+	if b.LoginTokenRefreshURL == "" {
+		return errors.New("login-token-refresh-url cannot be empty")
+	}
+
+	return nil
+}
+
+func (b BootstrapCmdParams) buildBootstrapCmdStr() string {
```
alesstimec: no unit tests for this

internal/jujucommands/jujucommands.go
```diff
@@ -0,0 +1,148 @@
...
+//	for out := range outputCh {
+//		if out.Err != nil {
+//			fmt.Println("Command finished with error:", out.Err)
+//			break
+//		}
+//		fmt.Println("Output line:", out.Line)
+//	}
+//
+// The OutputLine channel is closed once all output from the command has finished an the error has been captured.
+func runCmdWithOutputRetriever(store jujuclient.ClientStore, cmdAndArgs string) (<-chan outputLine, error) {
```
alesstimec: i would appreciate a unit test for this function.. you could mock the `runCmd` function for the test..

alesstimec (dismissed):

internal/jujucommands/jujucommands.go
ale8k: This was a part of the other PR. I don't see it as worth it in mocking it when it's very minimal to process a help/--help command. Due to delicacy, I prefer seeing a command legitimately run...

internal/jujucommands/jujucommands.go
ale8k: As this was discussed in the other PR, I'm going to resolve it here. Can ping if you still want to check it out?

internal/jujucommands/jujucommands_test.go
ale8k: Again was discussed in other PR, ping if it's still an issue. I don't see it as harmful. I see it as functionally equivalent to having a variable for the sake of a test?

internal/jujucommands/bootstrap.go
ale8k: Was discussed in other PR. Going to close, ping if still an issue and I'll do it in a followup.

internal/jujucommands/bootstrap.go
ale8k: On it!

internal/jujucommands/bootstrap.go
ale8k: On it!

internal/jujucommands/bootstrap.go
```diff
@@ -0,0 +1,218 @@
...
+// BootstrapCmdParams holds the parameters to bootstrap a controller for JIMM.
+type BootstrapCmdParams struct {
+	CloudNameAndRegion   string
+	ControllerName       string
+	AgentVersion         string
+	BootstrapTimeout     int
+	LoginTokenRefreshURL string
+}
+
+func (b BootstrapCmdParams) Validate() error {
```
alesstimec, line 38: godoc, please

internal/jujucommands/bootstrap.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+
+	return nil
+}
+
+func (b BootstrapCmdParams) BuildBootstrapCmdStr() string {
+	var builder strings.Builder
+	builder.WriteString("bootstrap")
+
+	builder.WriteString(fmt.Sprintf(" --login-token-refresh-url=%s", b.LoginTokenRefreshURL))
```
alesstimec:
```suggestion
	builder.WriteString(fmt.Sprintf("bootstrap --login-token-refresh-url=%s", b.LoginTokenRefreshURL))
```
and remove L61... the `login-token-refresh-url` flag will always be there

internal/jujucommands/bootstrap.go
```diff
@@ -0,0 +1,218 @@
...
+		builder.WriteString(fmt.Sprintf(" --agent-version=%s", b.AgentVersion))
+	}
+
+	// Conditionally add bootstrap-timeout if set
+	if b.BootstrapTimeout > 0 {
+		builder.WriteString(fmt.Sprintf(" --config bootstrap-timeout=%d", b.BootstrapTimeout))
+	}
+
+	// Always add controller name & cloud at the end
+	builder.WriteString(fmt.Sprintf(" %s", b.CloudNameAndRegion))
```
alesstimec: add both in one `WriteString`?

internal/jujucommands/bootstrap.go
```diff
@@ -0,0 +1,218 @@
...
+	if b.LoginTokenRefreshURL == "" {
+		return errors.New("login-token-refresh-url cannot be empty")
+	}
+
+	return nil
+}
+
+func (b BootstrapCmdParams) BuildBootstrapCmdStr() string {
+	var builder strings.Builder
+	builder.WriteString("bootstrap")
```
alesstimec: we could use templates instead of `strings.Builder`.. 
```
bootstrap --login-token-refresh-url={{ .loginTokenRefreshURL }} {{ if .agentVersion }} --agent-version={{ .agentVersion }} {{ end }} {{ if.bootstrapTimeout }} --config bootstrap-timeout={{ .bootstrapTimeout }} {{ end }} {{ .cloudNameAndRegion }} {{ .controllerName }}
```

internal/jujucommands/bootstrap.go
```diff
@@ -0,0 +1,213 @@
...
+
+	return args
+}
+
+type bootstrapCmd struct {
+	runner Runner
+}
+
+// NewBootstrapCmd creates a new BootstrapCmd with the specified command runner.
+func NewBootstrapCmd(runner Runner) *bootstrapCmd {
```
SimoneDutto, line 96: Adding JIMM here to check clouds, cloud regions, etc are valid and the user has access to it is for another PR?

internal/jujucommands/bootstrap_test.go
```diff
@@ -0,0 +1,256 @@
...
+	dataDir string
+}
+
+func (r *runnerMock) RunJujuCmd(ctx context.Context, args []string) (<-chan jujucommands.OutputLine, error) {
+	return r.impl(ctx, args)
+}
+
+func (r *runnerMock) JujuDataDir() string {
+	return r.dataDir
+}
```
SimoneDutto: we have gomock in the codebase, so you don't need to create this by hand

internal/jujucommands/bootstrap_test.go
```diff
@@ -0,0 +1,256 @@
...
+		qt.DeepEquals,
+		[]string{
+			"bootstrap",
+			"--config",
+			"login-token-refresh-url=myurl.com",
+			"testregion/testcloud",
+			"my-controller",
+		},
+	)
+}
```
SimoneDutto: a table test here would be much more readable. In this way is not clear what is the struct that is being validated because you set/unset fields on the flight

internal/jujucommands/bootstrap_test.go
```diff
@@ -0,0 +1,256 @@
...
+			// Return chan that has one line inside
+			outputCh := make(chan jujucommands.OutputLine, 1)
+			outputCh <- jujucommands.OutputLine{
+				Line: "testing",
+			}
+			close(outputCh)
+			return outputCh, nil
+		},
+		dataDir: c.TempDir(),
+	}
```
SimoneDutto: gomock

internal/jujucommands/bootstrap.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+
+	dataDir := c.runner.JujuDataDir()
+
+	// This is required because
+	// 		store := jujuclient.NewFileClientStore() Looks for JUJU_DATA env var
+	// And
+	// 		jujucloud.WritePersonalCloudMetadata
+	// TODO: Remove this dependency and write the files manually?
+	if err := os.Setenv("JUJU_DATA", dataDir); err != nil {
```
SimoneDutto: relying on env variable to be set is not a good idea, what if it's unset by another go-routine, you would never know.
Is there a safer way to do it?

internal/jujucommands/bootstrap_test.go
```diff
@@ -0,0 +1,256 @@
...
+	// then we're sure the default credential to be used is the one we provided for our provided cloud.
+	// (Given it also exists in the temp directory)
+	c.Assert(
+		personalCloudCred.AuthCredentials["default"].Attributes(),
+		qt.DeepEquals,
+		cloudCred.AuthCredentials["default"].Attributes(),
+	)
+
+	// This was set to a temp dir until cleanup runs. We can use it to check the file exists.
+	_, err = os.Stat(filepath.Join(os.Getenv("JUJU_DATA"), "clouds.yaml"))
```
SimoneDutto: relying on an env var in tests can bring to flaky situation because one test is setting it, the other is cleaning it up

internal/jujucommands/jujucommands_test.go
SimoneDutto, line 1: can you add a test where you cancel the context and the command is cancelled? Our jobtracker relies on the fact that it is possible to cancel a job by cancelling the context, but we don't have a test for it

internal/jujucommands/bootstrap.go
```diff
@@ -0,0 +1,213 @@
...
+	jujucloud "github.com/juju/juju/cloud"
+	"github.com/juju/juju/jujuclient"
+	_ "github.com/juju/juju/provider/lxd"
+	"github.com/juju/version/v2"
+)
+
+// BootstrapCmdParams holds the parameters to bootstrap a controller for JIMM.
+type BootstrapCmdParams struct {
+	// Arguments to be turned into an actual command str.
+	CloudNameAndRegion   string
```
kian99, line 23: Splitting them into `Cloud` and `Region` would be clearer.

internal/jujucommands/bootstrap.go
```diff
@@ -0,0 +1,213 @@
...
+	ControllerName       string
+	AgentVersion         string
+	BootstrapTimeout     int
+	LoginTokenRefreshURL string
+
+	// Additional args required (like adding credential, cloud, etc.) but JIMM will handle.
+
+	// May be left unset, if set, a personal cloud will be created and used for bootstrap.
+	PersonalCloud jujucloud.Cloud
```
kian99, line 32: What is a personal cloud?

internal/jujucommands/bootstrap.go
```diff
@@ -0,0 +1,213 @@
...
+
+	// Check if we can get the cloud as a public cloud.
+	cloudName, regionName := splitCloudNameAndRegion(p.CloudNameAndRegion)
+	isAPublicCloud, err := isAValidPublicCloud(cloudName, regionName)
+	if err != nil {
+		return nil, nil, nil, err
+	}
+	if !isAPublicCloud {
+		// We presume it is a personal cloud
+		// TODO: Check if credential should be cloudname or include region
```
kian99: This todo seems like something that should be done for this PR?

internal/jujucommands/bootstrap.go
```diff
@@ -0,0 +1,213 @@
...
+	// 		store := jujuclient.NewFileClientStore() Looks for JUJU_DATA env var
+	// And
+	// 		jujucloud.WritePersonalCloudMetadata
+	// TODO: Remove this dependency and write the files manually?
+	if err := os.Setenv("JUJU_DATA", dataDir); err != nil {
+		return nil, nil, nil, fmt.Errorf("failed to set JUJU_DATA env var: %w", err)
+	}
+
+	// Update public clouds
+	// TODO: Make this a command of this package
```
kian99: I'm not following what this is trying to say?

internal/jujucommands/bootstrap.go
```diff
@@ -0,0 +1,213 @@
...
+		// We presume it is a personal cloud
+		// TODO: Check if credential should be cloudname or include region
+		if err := jujucloud.WritePersonalCloudMetadata(map[string]jujucloud.Cloud{
+			cloudName: p.PersonalCloud,
+		}); err != nil {
+			return nil, nil, nil, fmt.Errorf("failed to write personal cloud: %w", err)
+		}
+	}
+
+	// TODO: check if cloudName should include region, presuming not right now
```
kian99: Again, this seems like something to do for this PR?

internal/jujucommands/bootstrap.go
```diff
@@ -0,0 +1,213 @@
...
+	// With the clouds set, credentials updated, we now bootstrap.
+	args := p.BuildBootstrapCmdArgs()
+
+	cleanupTmpJujuData := func() {
+		os.RemoveAll(dataDir)
+		os.Unsetenv("JUJU_DATA")
+	}
+
+	outputRetriever, err := c.runner.RunJujuCmd(ctx, args)
+	return outputRetriever, store, cleanupTmpJujuData, err
```
kian99: Can we change this to check the error first and then just return nil at the end. Just a bit clearer.

internal/jujucommands/bootstrap.go
```diff
@@ -0,0 +1,213 @@
...
+}
+
+// isAValidPublicCloud checks if the cloud name (and possibly region) is a valid
+// public cloud and region. If it is a public cloud without the region specified,
+// just the cloud name is checked. If a region is specified, but it isn't valid,
+// an error is returned.
+//
+// TODO: Is there a better way to do this? (Rather than look up metadata and loop through)
```
kian99: I would just delete this if you haven't found a better way.

internal/jujucommands/bootstrap_test.go
```diff
@@ -0,0 +1,256 @@
...
+
+func (r *runnerMock) RunJujuCmd(ctx context.Context, args []string) (<-chan jujucommands.OutputLine, error) {
+	return r.impl(ctx, args)
+}
+
+func (r *runnerMock) JujuDataDir() string {
+	return r.dataDir
+}
+
+func (s *jujucommandsSuite) TestBootstrapCmdParams_Validate(c *qt.C) {
```
kian99, line 18: Looks like the formatting got messed up in this function. So many new lines

internal/jujucommands/bootstrap_test.go
kian99: Additionally, what about just checking that the resulting list of strings doesn't contain config for the field you zeroed out, rather than checking the entire slice?

internal/jujucommands/bootstrap_test.go
```diff
@@ -0,0 +1,256 @@
...
+	testCtx := c.Context()
+
+	mock := runnerMock{
+		impl: func(ctx context.Context, args []string) (<-chan jujucommands.OutputLine, error) {
+			// Return chan that has one line inside
+			outputCh := make(chan jujucommands.OutputLine, 1)
+			outputCh <- jujucommands.OutputLine{
+				Line: "testing",
+			}
+			close(outputCh)
```
kian99: You could simplify this to,

```suggestion
			outputCh := make(chan jujucommands.OutputLine)
			close(outputCh)
```
The same for the other tests.

internal/jujucommands/jujucommands.go
```diff
@@ -13,21 +13,46 @@ import (
 	"github.com/mitchellh/go-linereader"
 )
 
-type outputLine struct {
+var (
+	cmdPrefix = "juju"
+)
```
kian99: Since the command runner has a prefix field, you can delete this and then you won't need to do patching from tests.

internal/jujucommands/jujucommands.go
```diff
@@ -13,21 +13,46 @@ import (
...
+}
 
 // runJujuCmd runs a juju command with the given command string and JUJU_DATA directory.
 // It returns a channel that will receive output lines from the command's stdout and stderr.
 // The command is run in a separate goroutine, and the context can be used to cancel the command.
-func runJujuCmd(ctx context.Context, args []string, jujuDataDir string) (<-chan outputLine, error) {
-	cmd := exec.CommandContext(ctx, cmdPrefix, args...)
-	cmd.Env = append(cmd.Env, "JUJU_DATA="+jujuDataDir)
+func (b *CommandRunner) RunJujuCmd(ctx context.Context, args []string) (<-chan OutputLine, error) {
+	cmd := exec.CommandContext(ctx, b.prefix, args...)
```
kian99: Looks like the tests might need an update since this is using the struct's value rather than the global var.

internal/jujucommands/bootstrap.go
kian99: It looks like you can use the function `SetJujuXDGDataHome` in `github.com/juju/juju/juju/osenv/home.go` which will set a global variable that is looked up before checking for any env vars. But this will still limit us to 1 bootstrap at a time.

internal/jujucommands/bootstrap.go
ale8k: That's for another PR in the job itself, this is just for running the command.

internal/jujucommands/bootstrap.go
ale8k: We pass cloud/region to the arg... If I split it, I'll have to reconstruct it at some point anyway. I'm happy to split it but not sure if it is worth it?

internal/jujucommands/bootstrap.go
ale8k: I thought the same initially when looking at how Juju bootstraps but tldr;

clouds.yaml -> add-cloud additions (private / personal / user clouds)
public-clouds.yaml -> predefined list of public clouds

Their language is "Personal Cloud" so I went with their language rather than "private" or something similar.

internal/jujucommands/bootstrap.go
ale8k: It is. I need to ask really. Not sure how to test this without my own regions...

internal/jujucommands/bootstrap.go
ale8k: We have a file for bootstrap command. The running of this command should be a command in our command package too. I.e., bootstrap.go, updatepubliccloud.go. It's not a big deal though I don't think and didn't want to make this PR larger than it needed to be.

internal/jujucommands/bootstrap.go
ale8k: I had a look at other instances where they update the store credential, and they name all vars cloud name. 

And in here: cmd/juju/cloud/updatecredential.go L279

They update credential without region and the region param is just validating:
```go
		if c.Region != "" {
			if err := validCloudRegion(aCloud, c.Region); err != nil {
				logger.Errorf("%v", err)
				ctx.Warningf("Region %q is not valid for cloud %v.", c.Region, cloudName)
				erred = true
				continue
			}
		}
```

But... The commands help text for update-credential is:
```
Before credential is updated, the new content is validated. For some providers, 
cloud credentials are region specific. To validate the credential for a non-default region, 
use --region.
```

Maybe I'm misunderstanding a bit but it kinda looks like it just stores against cloud name. 

But if you look at the credentials file, they're region specific. 

Anyone reading this comment - It'd be good if another pair of eyes could look over this

internal/jujucommands/bootstrap_test.go
ale8k: Aha, that's me. It read easier, I'll condense it.

internal/jujucommands/jujucommands.go
ale8k: The structure was using the global var (which is why it worked) but I've removed it now as you requested. I updated signature of NewCmdRunner to take a prefix.

internal/jujucommands/bootstrap.go
ale8k: Ahhh nevermind, it is in the credential itself. So it is indeed cloud name.

internal/jujucommands/bootstrap.go
kian99: Sure, fine to leave.

internal/jujucommands/bootstrap.go
kian99: Ah, I think it's fine to remove the todo then. Or reword it to `move this command to its own file`.

internal/jujucommands/bootstrap_test.go
ale8k: Ah what I didn't push this. Resolved.

internal/jujucommands/jujucommands_test.go
ale8k: Will do.

internal/jujucommands/bootstrap_test.go
```diff
@@ -0,0 +1,248 @@
...
+	// then we're sure the default credential to be used is the one we provided for our provided cloud.
+	// (Given it also exists in the temp directory)
+	c.Assert(
+		personalCloudCred.AuthCredentials["default"].Attributes(),
+		qt.DeepEquals,
+		cloudCred.AuthCredentials["default"].Attributes(),
+	)
+
+	// This was set to a temp dir until cleanup runs. We can use it to check the file exists.
+	_, err = os.Stat(filepath.Join(os.Getenv("JUJU_DATA"), "clouds.yaml"))
```
kian99: I think this line needs to be updated now that we are not using an env var anymore

internal/jujucommands/bootstrap_test.go
```diff
@@ -0,0 +1,248 @@
...
+	personalCloudCred, err := store.CredentialForCloud("aws")
+	c.Assert(err, qt.IsNil)
+	c.Assert(
+		personalCloudCred.AuthCredentials["default"].Attributes(),
+		qt.DeepEquals,
+		cloudCred.AuthCredentials["default"].Attributes(),
+	)
+
+	// Make sure no personal cloud was written.
+	_, err = os.Stat(filepath.Join(os.Getenv("JUJU_DATA"), "clouds.yaml"))
```
kian99: This line too

internal/jujucommands/jujucommands.go
```diff
@@ -13,21 +13,45 @@ import (
...
+type CommandRunner struct {
+	prefix      string
+	jujuDataDir string
+}
+
+// NewCommandRunner creates a new CommandRunner with the specified command prefix.
+//
+// dataDir is the JUJU_DATA directory where juju commands will store their data.
+// It must be an ABSOLUTE path.
+func NewCommandRunner(prefix, dataDir string) *CommandRunner {
```
kian99: Can we rename `prefix` to `command`?

internal/jujucommands/bootstrap.go
ale8k: Checked, it is just name.

internal/jujucommands/jujucommands_test.go
```diff
@@ -64,8 +80,9 @@ func (s *jujucommandsSuite) TestRunCmdWithOutputRetriever_Error(c *qt.C) {
 
 func (s *jujucommandsSuite) TestEnvironmentIsCorrectlySet(c *qt.C) {
 	testCtx := c.Context()
-	c.Patch(jujucommands.CmdPrefix, "env")
-	outputCh, err := jujucommands.RunJujuCmd(testCtx, []string{}, "testing-data-is-set")
+
+	runner := jujucommands.NewCommandRunner("env", "testing-data-is-set")
+	outputCh, err := runner.RunJujuCmd(testCtx, []string{})
```
kian99, line 85: Just for consistency with other files, can you put this at the top please

kian99 (approved):

SimoneDutto (approved):
be 🐄-tious merging this

ale8k:
> be 🐄-tious merging this
@SimoneDutto - I will.
image

PR #1643, opened 2025-07-04
feat: redirect agents to backing controller

kian99:
## Description
This change is in support of model migration. During model migration, agents connect to the new controller to validate they will work after migration has completed. JIMM should not be the address for Juju agents, they should point to the backing controller but the migration flow needs to go through JIMM for it to modify things like the model owner. To handle this, when an agent is trying to login to JIMM we return a redirect to the backing controller.

Both integration and unit tests have been added and JIMM's errors package has also been modified to support all the error fields that are used in Juju's RPC error handling/

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

### DIFF
--- a/internal/jujuapi/websocket_test.go
+++ b/internal/jujuapi/websocket_test.go
@@ -17,6 +17,7 @@ import (
 	"strings"
 
 	"github.com/go-chi/chi/v5"
+	"github.com/juju/errors"
 	"github.com/juju/juju/api"
 	"github.com/juju/juju/api/client/client"
 	"github.com/juju/juju/api/client/keymanager"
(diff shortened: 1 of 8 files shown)


internal/jujuapi/websocket_test.go
```diff
@@ -223,6 +224,44 @@ func (s *apiProxySuite) TestSessionTokenLoginProvider(c *gc.C) {
...
+		Tag:      names.NewUnitTag("ubuntu/1"),
+		ModelTag: s.Model.ResourceTag(),
+		Addrs:    []string{u.Host},
+	}
+	dialOpts := api.DialOpts{
+		InsecureSkipVerify: true,
+	}
+	_, err = api.Open(&info, dialOpts)
+	c.Assert(err, gc.NotNil)
+	_, ok := errors.Cause(err).(*api.RedirectError)
```
alesstimec: could we assert that the redirect data matches what we expect?

internal/rpcproxy/rpcproxy.go
```diff
@@ -35,6 +36,22 @@ const (
 	accessRequiredErrorCode = "access required"
 )
 
+// ControllerDetails holds information about the controller
+// that is being proxied to.
+type ControllerDetails struct {
+	Servers [][]jujuparams.HostPort
```
alesstimec: Addresses?

internal/rpcproxy/rpcproxy.go
```diff
@@ -35,6 +36,22 @@ const (
...
+	CACert  string
+}
+
+// RedirectInfo provides information about the controller
+// that is being proxied to.T his information is useful
+// when we need to redirect a client to that controller
+// instead of proxying the request. This is the case during
+// model migration when receiving requests from agents.
+type RedirectInfo interface {
```
alesstimec: suggestion: RedirectInfoStore? RedirectInfoProvider?

alesstimec (approved):
LGTM with a few suggestions

internal/rpcproxy/rpcproxy.go
kian99: I'll go with `RedirectInfoGetter`

internal/rpcproxy/rpcproxylogin_test.go
```diff
@@ -467,3 +528,34 @@ func (m *mockTokenGenerator) GetUser() names.UserTag {
...
+				Value: "controller-1",
+				Type:  "ipv4",
+			},
+			Port: 17070,
+		}}},
+		CACert: "test-ca-cert",
+	}, nil
+}
+
+func expectedRedirectInfo() map[string]interface{} {
```
SimoneDutto, line 547: how is this working if now in the error we return `Addresses` and not `servers`?

SimoneDutto (commented):
it looks good, just a question about the tests

internal/rpcproxy/rpcproxylogin_test.go
kian99: `servers` is what Juju expects, `Addresses` is just the field for our internal type.
More specifically,
The model proxy talks to the DB and gets a `ControllerDetails` struct with the `Addresses` field populated.
It then populates a `jujuparams.RedirectErrorInfo` which serializes to an object with a field called `servers`.

kian99: I tested changing `servers` to `addresses` and it does fail as expected,
```
              + 		"addresses": []any{
              + 			[]any{
              + 				map[string]any{
              + 					"port":  float64(17070),
              + 					"scope": string(""),
              + 					"type":  string("ipv4"),
              + 					"value": string("controller-1"),
              + 				},
              + 			},
              + 		},
                		"ca-cert": string("test-ca-cert"),
              - 		"servers": []any{
              - 			[]any{
              - 				map[string]any{
              - 					"port":  float64(17070),
              - 					"scope": string(""),
              - 					"type":  string("ipv4"),
              - 					"value": string("controller-1"),
              - 				},
              - 			},
              - 		},
                	},
```

SimoneDutto (approved):
lgtm

PR #1644, opened 2025-07-04
feat(delete-model): delete model by uuid

SimoneDutto:
In this PR we add the ability to delete models by UUID.
Before we needed to GetModel first to be able to delete it via its primary key (`ID`).

### DIFF
--- a/internal/db/model.go
+++ b/internal/db/model.go
@@ -113,7 +113,7 @@ func (d *Database) UpdateModel(ctx context.Context, model *dbmodel.Model) (err e
 	return nil
 }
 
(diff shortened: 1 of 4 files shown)


kian99 (approved):
I don't see a test for when neither model UUID or ID is specified (perhaps one already exists), otherwise lgtm.

SimoneDutto:
> I don't see a test for when neither model UUID or ID is specified (perhaps one already exists), otherwise lgtm.

added

alesstimec (approved):
Nice, thank you.

PR #1645, opened 2025-07-07
feat: improve-prechecks

SimoneDutto:
Check cloud, cloud region and cloud credential during `Prechecks`

### DIFF
--- a/internal/jimm/juju/migrationtarget.go
+++ b/internal/jimm/juju/migrationtarget.go
@@ -142,7 +142,8 @@ func (j *JujuManager) ControllerDetailsForIncomingModel(ctx context.Context, mod
 }
 
 // Prechecks checks that the model can be migrated to the target controller.
-// It does this by calling the method of the same name on the target Juju controller.
+// It does this by checking cloud, cloudregion and cloud credentials exists in JIMM, then
+// calling the method of the same name on the target Juju controller.
 // As part of all model migrations passing through JIMM, it modifies the model description
 // to replace any local user references with their external mapping.
 func (j *JujuManager) Prechecks(ctx context.Context, user *openfga.User, model migration.ModelInfo) error {
@@ -162,6 +163,22 @@ func (j *JujuManager) Prechecks(ctx context.Context, user *openfga.User, model m
 	err = j.modifyMigrationInfo(&model, incomingModel.UserMapping)
 	if err != nil {
 		return errors.E(op, fmt.Errorf("failed to modify migration info: %w", err))
+	}
+
+	_, err = j.Database.FindRegionByCloudName(ctx, model.ModelDescription.CloudCredential().Cloud(), model.ModelDescription.CloudRegion())
+	if err != nil {
+		return errors.E(op, fmt.Errorf("failed to find region for cloud %q: %w", model.ModelDescription.CloudCredential().Cloud(), err))
+	}
+
+	cloudCredential := &dbmodel.CloudCredential{
+		CloudName:         model.ModelDescription.CloudCredential().Cloud(),
+		OwnerIdentityName: model.ModelDescription.Owner().Id(),
+		Name:              model.ModelDescription.CloudCredential().Name(),
+	}
+
+	err = j.Database.GetCloudCredential(ctx, cloudCredential)
+	if err != nil {
+		return errors.E(op, err)
 	}
 
 	api, err := j.dialController(ctx, &incomingModel.TargetController)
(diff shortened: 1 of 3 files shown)


kian99 (approved):

internal/jimm/juju/migrationtarget.go
```diff
@@ -164,6 +165,22 @@ func (j *JujuManager) Prechecks(ctx context.Context, user *openfga.User, model m
...
+		return errors.E(op, fmt.Errorf("failed to find region for cloud %q: %w", model.ModelDescription.CloudCredential().Cloud(), err))
+	}
+
+	cloudCredential := &dbmodel.CloudCredential{
+		CloudName:         model.ModelDescription.CloudCredential().Cloud(),
+		OwnerIdentityName: model.ModelDescription.Owner().Id(),
+		Name:              model.ModelDescription.CloudCredential().Name(),
+	}
+
+	err = j.Database.GetCloudCredential(ctx, cloudCredential)
```
ale8k, line 179: Can merge line 173 here, it doesn't need an assigned variable as it isn't going to be reused. I'd prob godoc too explaining we're checking the cloud credential exists.

internal/jujuapi/migrationtarget_test.go
```diff
@@ -55,7 +61,7 @@ func (s *migrationTargetSuite) TestPrechecks(c *gc.C) {
 	modelDescription := description.NewModel(modelDescriptionArgs)
 	modelDescription.SetStatus(description.StatusArgs{Value: "available"})
 	modelDescription.SetCloudCredential(description.CloudCredentialArgs{
-		Name:  jimmtest.TestCloudName + "/alice@canonical.com/cred",
+		Name:  "cred",
```
ale8k, line 64: Is this right? Should it be just called "cred"?

internal/jimm/juju/migrationtarget_test.go
```diff
@@ -239,11 +243,107 @@ func TestPrechecks_ModifiesModelDescription(t *testing.T) {
...
+	model := newMigrationInfo(modelDescriptionArgs{
+		Owner:               "bob",
+		ModelName:           "test-model-2",
+		CloudName:           "test-not-found",
+		CloudCredentialName: "test-cred",
+		CloudRegionName:     "test-region",
+	})
+
+	err := j.Prechecks(ctx, user, model)
+	c.Assert(err, qt.ErrorMatches, `^failed to find region for cloud "test-not-found".*`)
```
ale8k, line 314: Can we test the rest of this error explicitly? I was expecting to see ""test-not-found": cloud doesn't exist" or something.

ale8k (approved):
Minor comments, lgtm otherwise.

But do explicitly check the error in the test plz, it'll make it easier to read.

I'd also prefer if we did a cloud check explicit first too, and create our own error for when the cloud (not region) is missing. Otherwise we're at the behest of FindRgionByCloudName's error message to determine if the cloud or region wasn't found and we've no way of knowing which, so our test isn't really certain.

internal/jimm/juju/migrationtarget.go
```diff
@@ -164,6 +165,22 @@ func (j *JujuManager) Prechecks(ctx context.Context, user *openfga.User, model m
 		return errors.E(op, fmt.Errorf("failed to modify migration info: %w", err))
 	}
 
+	_, err = j.Database.FindRegionByCloudName(ctx, model.ModelDescription.CloudCredential().Cloud(), model.ModelDescription.CloudRegion())
+	if err != nil {
+		return errors.E(op, fmt.Errorf("failed to find region for cloud %q: %w", model.ModelDescription.CloudCredential().Cloud(), err))
```
alesstimec, line 170: include region in the error message?

alesstimec (approved):

internal/jimm/juju/migrationtarget.go
ale8k: I think this aligns with my comment, just check both individually and not try bundle the two cases into one.

internal/jujuapi/migrationtarget_test.go
SimoneDutto: yes, because the credential tag is then built in the `migration_target.go` function

internal/jimm/juju/migrationtarget.go
SimoneDutto: i like them in a single message, it's clear to have "not found region region-name in Cloud cloud-name".

PR #1647, opened 2025-07-10
chore: add offers and permissions during migration import

kian99:
## Description
This PR makes a few tweaks to the `Import` step of model migration, namely:
- Creates offers in JIMM's database alongside model creation.
- Adds permissions for the new model owner to the model and adds the correct permissions for any application offers.
- Ensures permissions are removed during `Abort`.

I also refactored the logic in `Import` to encapsulate it, and avoid leaking methods that are specific to the `Import` method to the rest of the Juju manager.

Fixes [JUJU-8192](https://warthogs.atlassian.net/browse/JUJU-8192)

## Engineering checklist
- [ ] Documentation updated
- [x] Covered by unit tests

[JUJU-8192]: https://warthogs.atlassian.net/browse/JUJU-8192?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/jimm/juju/migrationtarget.go
+++ b/internal/jimm/juju/migrationtarget.go
@@ -405,16 +405,6 @@ func (j *JujuManager) Activate(ctx context.Context, modelTag names.ModelTag, mig
 	return nil
 }
 
-// modelMigrationImport is a struct that holds the JujuManager and the model description
-// that is being imported. It is used to encapsulate the logic for importing a model
-// from a serialized description.
-// This helps to keep the import logic and associated methods separate from the JujuManager
-// methods.
-type modelMigrationImport struct {
-	*JujuManager
-	modelDescription description.Model
-}
-
 // Import imports a model from a serialized description.
 //   - Checks the incoming model migration record in the database.
 //   - Modifies the model description to replace local user references with their external mapping for owner and
@@ -430,59 +420,41 @@ func (j *JujuManager) Import(ctx context.Context, user *openfga.User, serialized
 		return errors.E(op, fmt.Errorf("failed to deserialize model description: %w", err))
 	}
 
-	importer := modelMigrationImport{
-		JujuManager:      j,
-		modelDescription: modelDescription,
-	}
-	err = importer.migrationImport(ctx, user)
-	if err != nil {
-		return errors.E(op, fmt.Errorf("failed to import resources: %w", err))
-	}
-	return nil
-}
-
-// migrationImport imports resources from the model description into JIMM's state
-// and adds the necessary permissions for the model and application offers
-// before calling the import method on the target Juju controller.
-// Since `import` is a reserved word in Go, we use `migrationImport`.
-func (mmi *modelMigrationImport) migrationImport(ctx context.Context, user *openfga.User) error {
-	const op = errors.Op("jimm.migrationImport")
-
 	incomingMigration := &dbmodel.IncomingModelMigration{
 		ModelUUID: sql.NullString{
-			String: mmi.modelDescription.Tag().Id(),
+			String: modelDescription.Tag().Id(),
 			Valid:  true,
 		},
 	}
 
-	err := mmi.Database.GetIncomingModelMigration(ctx, incomingMigration)
+	err = j.Database.GetIncomingModelMigration(ctx, incomingMigration)
 	if err != nil {
 		return errors.E(op, fmt.Errorf("failed to add incoming model migration: %w", err))
 	}
 
-	err = mmi.modifyModelDescription(mmi.modelDescription, incomingMigration.UserMapping)
+	err = j.modifyModelDescription(modelDescription, incomingMigration.UserMapping)
 	if err != nil {
 		return errors.E(op, fmt.Errorf("failed to modify model description: %w", err))
 	}
 
-	resources, err := mmi.importFromDescription(ctx, incomingMigration.TargetController.ID, mmi.modelDescription)
+	model, offers, err := j.importFromDescription(ctx, incomingMigration.TargetController.ID, modelDescription)
 	if err != nil {
 		return errors.E(op, fmt.Errorf("failed to import model from description: %w", err))
 	}
 
-	err = mmi.addResourcePermissions(ctx, user, resources, incomingMigration.TargetController.ResourceTag())
+	err = j.addModelAndOfferPermissions(ctx, user, model, offers)
 	if err != nil {
 		return errors.E(op, fmt.Errorf("failed to add resource permissions: %w", err))
 	}
 
 	// Call the import method on the target controller to import the model.
-	api, err := mmi.dialController(ctx, &incomingMigration.TargetController)
+	api, err := j.dialController(ctx, &incomingMigration.TargetController)
 	if err != nil {
 		return errors.E(op, fmt.Errorf("failed to dial controller: %w", err))
 	}
 	defer api.Close()
 
-	serializedDescrition, err := description.Serialize(mmi.modelDescription)
+	serializedDescrition, err := description.Serialize(modelDescription)
 	if err != nil {
 		return errors.E(op, fmt.Errorf("failed to serialize model description: %w", err))
 	}
@@ -495,31 +467,26 @@ func (mmi *modelMigrationImport) migrationImport(ctx context.Context, user *open
 	return nil
 }
 
-type importedResources struct {
-	model  dbmodel.Model
-	offers []dbmodel.ApplicationOffer
-}
-
 // importFromDescription imports resources into JIMM's state from a model description.
 // It creates a new model record in the database with the given target controller ID
 // and model description and sets the migration mode to importing.
 // Application offers are created for any offers in the model description.
 // It also ensures that the cloud credential and region are present in the database.
-func (mmi *modelMigrationImport) importFromDescription(ctx context.Context, targetControllerID uint, description description.Model) (*importedResources, error) {
+func (j *JujuManager) importFromDescription(ctx context.Context, targetControllerID uint, description description.Model) (*dbmodel.Model, []*dbmodel.ApplicationOffer, error) {
 	op := errors.Op("jimm.importFromDescription")
 
 	modelNameStr, ok := description.Config()[config.NameKey].(string)
 	if !ok {
-		return nil, errors.E(op, fmt.Errorf("model config must contain a string value for key %q", config.NameKey))
+		return nil, nil, errors.E(op, fmt.Errorf("model config must contain a string value for key %q", config.NameKey))
 	}
 
 	modelUUIDStr, ok := description.Config()[config.UUIDKey].(string)
 	if !ok {
-		return nil, errors.E(op, fmt.Errorf("model config must contain a string value for key %q", config.UUIDKey))
+		return nil, nil, errors.E(op, fmt.Errorf("model config must contain a string value for key %q", config.UUIDKey))
 	}
 
 	if description.CloudCredential() == nil {
-		return nil, errors.E(op, fmt.Errorf("model description must contain a cloud credential"))
+		return nil, nil, errors.E(op, fmt.Errorf("model description must contain a cloud credential"))
 	}
 	cloudCredential := &dbmodel.CloudCredential{
 		CloudName:         description.CloudCredential().Cloud(),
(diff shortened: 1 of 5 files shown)


internal/jimm/juju/migrationtarget.go
```diff
@@ -69,13 +70,17 @@ func (j *JujuManager) AbortMigration(ctx context.Context, user *openfga.User, mo
...
-	// Don't return an error if we fail to delete the model from JIMM's state,
-	// as the migration has already been aborted on the target controller.
+	// Don't return an error if we fail to delete the model/permissions from
+	// JIMM's state, the migration has already been aborted on the target controller.
 	// The model will be cleanup eventually by JIMM's cleanup routine.
 	err = j.Database.DeleteModel(ctx, &model)
 	if err != nil {
 		zapctx.Error(ctx, "failed to delete incoming model migration", zap.Error(err), zap.String("modelUUID", modelUUID))
 	}
+	err = j.OpenFGAClient.RemoveModel(ctx, model.ResourceTag())
```
SimoneDutto, line 80: i think we need this as well in the Cleanup

internal/jimm/juju/migrationtarget.go
```diff
@@ -447,57 +491,115 @@ func (j *JujuManager) Import(ctx context.Context, user *openfga.User, serialized
...
+		err = db.AddModel(ctx, &model)
+		if err != nil {
+			return errors.E(op, fmt.Errorf("failed to add model %q: %w", modelUUIDStr, err))
+		}
+		appAndOffers.model = model
+
+		for _, app := range description.Applications() {
+			for _, offer := range app.Offers() {
+				// construct the offer URL with the same logic as Juju (modelOwner, modelName, offerName, <blank-controller-name>)
+				offerURL := jujucrossmodel.MakeURL(description.Owner().Id(), modelNameStr, offer.OfferName(), "")
```
SimoneDutto, line 534: question: what's the forth argument?

internal/jimm/juju/migrationtarget.go
```diff
@@ -447,57 +491,115 @@ func (j *JujuManager) Import(ctx context.Context, user *openfga.User, serialized
...
+			MigrationMode:     state.MigrationModeImporting,
+		}
+		err = db.AddModel(ctx, &model)
+		if err != nil {
+			return errors.E(op, fmt.Errorf("failed to add model %q: %w", modelUUIDStr, err))
+		}
+		appAndOffers.model = model
+
+		for _, app := range description.Applications() {
+			for _, offer := range app.Offers() {
```
SimoneDutto, line 532: just to make sure i understand:
we create the offer in JIMM and the target juju controller with the external identity, we will then:
- convert the local user to external in the discharge endpoint (already done)
- find a solution to map these offers for terraform as we've discussed yesterday

internal/jimm/juju/migrationtarget.go
```diff
@@ -400,11 +405,22 @@ func (j *JujuManager) Activate(ctx context.Context, modelTag names.ModelTag, mig
 	return nil
 }
 
+// modelMigrationImport is a struct that holds the JujuManager and the model description
+// that is being imported. It is used to encapsulate the logic for importing a model
+// from a serialized description.
+// This helps to keep the import logic and associated methods separate from the JujuManager
+// methods.
+type modelMigrationImport struct {
+	*JujuManager
```
SimoneDutto: i would prefer here to have an more specific interface here to better understand the methods of the juju manager used by the modelMigrationImport.
What do you think?

internal/jimm/juju/migrationtarget.go
kian99: The controller name

internal/jimm/juju/migrationtarget.go
kian99: I don't think so, if there are dangling tuples i.e. tuples that reference a resource that doesn't exist, we have an existing cleanup routine to remove them.

internal/jimm/juju/migrationtarget.go
kian99: >we create the offer in JIMM and the target juju controller with the external identity, we will then:

Yes we will create the offer in JIMM. Juju will create the offer in its handling of the `Import` call (using the model owner as part of the offer URL).

On the other two points,
1. Yes, we will look up the user-mappings for the model in the discharger to convert local users to external users.
2. Yip, we'll have to handle that in https://warthogs.atlassian.net/browse/JUJU-8247

internal/jimm/juju/migrationtarget.go
kian99: I can see that being useful. But it feels like a bit too much abstraction without the need for it right now (beyond some readability). We don't test this piece in isolation, we just test the exported Import method so I'm in favor of not adding an interface for it (at least for now).

internal/jimm/juju/migrationtarget.go
SimoneDutto: fair enough

internal/jimm/juju/migrationtarget.go
SimoneDutto: i don't like this encapsulation too much because:
- it's a little bit of an inception to call a juju manager method from inside a modelMigrationImport's method called itself from juju manager
- i don't like that we use model description from this struct that is being modified as a side effect of calling methods on this struct

But if the other reviewers are fine, i won't be opposed to it

internal/jimm/juju/migrationtarget.go
SimoneDutto: I guess the controller_name must be specified only when you are consuming a cross-controller relations.
Which shouldn't be an issue because here we are just importing offers in JIMM's state

internal/jimm/juju/migrationtarget.go
```diff
@@ -400,11 +405,22 @@ func (j *JujuManager) Activate(ctx context.Context, modelTag names.ModelTag, mig
 	return nil
 }
 
+// modelMigrationImport is a struct that holds the JujuManager and the model description
+// that is being imported. It is used to encapsulate the logic for importing a model
+// from a serialized description.
+// This helps to keep the import logic and associated methods separate from the JujuManager
```
ale8k: If we have to doc that we're making this for encapsulation, then we're likely doing something wrong I think. We shouldn't have to explicitly state we're creating a struct to encapsulate logic.

internal/jimm/juju/migrationtarget.go
```diff
@@ -413,32 +429,60 @@ func (j *JujuManager) Import(ctx context.Context, user *openfga.User, serialized
 	if err != nil {
 		return errors.E(op, fmt.Errorf("failed to deserialize model description: %w", err))
 	}
+
+	importer := modelMigrationImport{
```
ale8k: We hav the struct defined above, and then a JujuManager method, can we keep things in-order please? It makes it easier to follow if we have 

..juju manager method
..juju manager method
..modelMigrationImport struct
..modelMigrationImport method
..modelMigrationImport method

internal/jimm/juju/migrationtarget.go
```diff
@@ -400,11 +405,22 @@ func (j *JujuManager) Activate(ctx context.Context, modelTag names.ModelTag, mig
 	return nil
 }
 
+// modelMigrationImport is a struct that holds the JujuManager and the model description
+// that is being imported. It is used to encapsulate the logic for importing a model
+// from a serialized description.
+// This helps to keep the import logic and associated methods separate from the JujuManager
+// methods.
+type modelMigrationImport struct {
```
ale8k: Needs a constructor to protect us from a nil JujuManager

internal/jimm/juju/migrationtarget.go
ale8k: Also maybe rename it to modelMigrationImporter? We call it "importer" here

internal/jimm/juju/migrationtarget.go
alesstimec: even if we don't remove the model from OpenFGA in this step all tuples referencing it should be removed by the cleanup routine

internal/jimm/juju/migrationtarget.go
```diff
@@ -413,32 +429,60 @@ func (j *JujuManager) Import(ctx context.Context, user *openfga.User, serialized
...
+	if err != nil {
+		return errors.E(op, fmt.Errorf("failed to import resources: %w", err))
+	}
+	return nil
+}
+
+// migrationImport imports resources from the model description into JIMM's state
+// and adds the necessary permissions for the model and application offers
+// before calling the import method on the target Juju controller.
+// Since `import` is a reserved word in Go, we use `migrationImport`.
```
ale8k: ?? You can't call it importMigration?

internal/jimm/juju/migrationtarget.go
```diff
@@ -447,57 +491,115 @@ func (j *JujuManager) Import(ctx context.Context, user *openfga.User, serialized
...
+		return nil, errors.E(op, fmt.Errorf("model description must contain a cloud credential"))
 	}
 	cloudCredential := &dbmodel.CloudCredential{
 		CloudName:         description.CloudCredential().Cloud(),
 		OwnerIdentityName: description.Owner().Id(),
 		Name:              description.CloudCredential().Name(),
 	}
 
-	err := j.Database.GetCloudCredential(ctx, cloudCredential)
+	err := mmi.Database.GetCloudCredential(ctx, cloudCredential)
```
ale8k: We're transitively using the database from the JujuManager with no guarantee the JujuManager nor Database is safe to use. I think I'd much prefer to see two interfaces, one for the JujuManager and one for the Database methods that will be used within the modelMigrationImporter. 

I think this could just be two functions and it'd read clearer, because really at the end of the day the signatures within the importer are:

```go
func (mmi *modelMigrationImport) addResourcePermissions(ctx context.Context, user *openfga.User, resources *importedResources, targetControllerTag names.ControllerTag) error
func (mmi *modelMigrationImport) importFromDescription(ctx context.Context, targetControllerID uint, description description.Model) (*importedResources, error)
```

So I'd be in favour of just two simple functions in this file and even remove the type "importedResources", because without validating (as in to check nil on offers & model is set correctly) it it's no different from returning the two values.

internal/jimm/juju/migrationtarget.go
```diff
@@ -413,32 +429,60 @@ func (j *JujuManager) Import(ctx context.Context, user *openfga.User, serialized
...
 	}
-	err = j.Database.GetIncomingModelMigration(ctx, incomingMigration)
+
+	err := mmi.Database.GetIncomingModelMigration(ctx, incomingMigration)
 	if err != nil {
 		return errors.E(op, fmt.Errorf("failed to add incoming model migration: %w", err))
 	}
-	err = j.modifyModelDescription(modelDescription, incomingMigration.UserMapping)
+
+	err = mmi.modifyModelDescription(mmi.modelDescription, incomingMigration.UserMapping)
```
ale8k: Calling through to JujuManager is confusing here I think (for a moment it looked like this would be a responsibility of the importer)

internal/jimm/juju/migrationtarget.go
```diff
@@ -447,57 +491,115 @@ func (j *JujuManager) Import(ctx context.Context, user *openfga.User, serialized
...
 	})
 	if err != nil {
-		return errors.E(op, fmt.Errorf("failed to add model %q: %w", modelUUIDStr, err))
+		return nil, errors.E(op, fmt.Errorf("failed to import model from description: %w", err))
+	}
+
+	return &appAndOffers, nil
+}
+
+func (mmi *modelMigrationImport) addResourcePermissions(ctx context.Context, user *openfga.User, resources *importedResources, targetControllerTag names.ControllerTag) error {
```
ale8k: For consistency maybe godoc this too? The name doesn't really read so clearly, that is, it doesn't say to me "set owner of model and relate app offers to this model / rename this func

ale8k (commented):
Needs to rereview and have a think... I'm not a fan of having a struct (the importer) with two methods that has the JujuManager inline and then uses those inline calls that are specific to model imports inside its own functions. I think just two funcs would read a lot clearer to be honest. 

TLDR; the separation here is more confusing than helpful.

Code wise everything seems fine to me though.

kian99:
@SimoneDutto @ale8k Let me remove that struct. Will ping you guys when that's done.

internal/jimm/juju/migrationtarget.go
kian99: You can't call it `import` since it's a reseved keyword.

kian99:
@SimoneDutto @ale8k I've removed the modelMigrationImport struct so it should read more clearly now.

SimoneDutto (approved):
lgtm

internal/jimm/juju/migrationtarget.go
```diff
@@ -413,25 +419,35 @@ func (j *JujuManager) Import(ctx context.Context, user *openfga.User, serialized
...
-	err = j.importModelFromDescription(ctx, incomingMigration.TargetController.ID, modelDescription)
+
+	model, offers, err := j.importFromDescription(ctx, incomingMigration.TargetController.ID, modelDescription)
 	if err != nil {
 		return errors.E(op, fmt.Errorf("failed to import model from description: %w", err))
 	}
 
+	err = j.addModelAndOfferPermissions(ctx, user, model, offers)
+	if err != nil {
+		return errors.E(op, fmt.Errorf("failed to add resource permissions: %w", err))
```
alesstimec, line 450: now, this is a bit of a tricky situation.. we've imported the model into JIMM's state, but failed to set permissions. how do we get out of this situation? should we give users the tooling to re-set model and offer permissions? but the problem is that we don't have a record of who should have permission to them..

alesstimec (approved):
LGTM.. but we need to think about the situation where we fail to set permissions to imported model/offers..

internal/jimm/juju/migrationtarget.go
kian99: I think this might be okay, if this fails, the migration will be aborted and in JIMM's Abort handler we remove the model.

PR #1648, opened 2025-07-11
feat: improved integration testing

kian99:
## Description

This PR adds a new `tests` folder in the root project directory that aims to hold all integration tests that can't go into Go source files. I've added a simple README.md to outline the basic structure and test format. Then I've added the first few tests. 

The first two already were already running in our Github action. The first is to deploy a model and application and the second was to verify JIMM's SSH server is running. I've simply moved these to test scripts.

----------------------
**Update:** The model migration test was failing in CI because our integration test uses the snapstore's latest JAAS snap plugin. In this PR we are making a change to the plugin which is needed for the test to pass. So, I've removed that script for now and will add it in a follow-up.

While we could build the jaas plugin from source, I would prefer to consider that in a future change as when I tried that the runner ran out of space so that might cause more CI headaches.

-----------------------

~~In a follow-up PR I will modify our integration test workflow to use the run-tests.sh script.~~ I've done this in this PR to be able to validate the test works.

In addition to this, there are a few minor changes to JIMM to aid debugging like some tweaks to error messages that I encountered when writing the test.

Fixes [JUJU-7952](https://warthogs.atlassian.net/browse/JUJU-7952)

## Engineering checklist
- [ ] Documentation updated
- [ ] Covered by unit tests

[JUJU-7952]: https://warthogs.atlassian.net/browse/JUJU-7952?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/integration-test.yaml
kian99, line 1: For the details to why this file was changed, see the commit description on the commit [ci: tweak integration test workflow](https://github.com/canonical/jimm/pull/1648/commits/cb1b45d49c69d9b688d3074ce619e89c95b5953e)

tests/run-tests.sh
```diff
@@ -0,0 +1,37 @@
...
+set -euo pipefail
+
+TESTS_DIR="tests"
+
+echo $"Running tests in $TESTS_DIR"
+for folder in "$TESTS_DIR"/*/ ; do
+    # Skip if not a directory
+    [ -d "$folder" ] || continue
+    for test_script in "$folder"*; do
+        # Only run regular files that are executable and not directories
```
ale8k, line 17: Ty for comment.

tests/run-tests.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+
+# Run all executable files in all subfolders of the current directory
+# and stop execution if any test fails.
+# This script is intended to be run from the root of the project
+# and will execute all test scripts found in the `tests` directory.
+
+set -euo pipefail
```
ale8k, line 8: Is the pipefail incase any scripts do x | y | z? (As it isn't used right now)

tests/deployapp/deploy-basic-app.sh
```diff
@@ -0,0 +1,17 @@
...
+# before running `juju status` and then destroys the model.
+
+set -euo pipefail
+
+echo "Creating a new model and deploying haproxy"
+juju add-model foo
+juju deploy haproxy
+
+echo "Waiting 5 seconds and then running 'juju status'"
+sleep 5
```
ale8k, line 13: Could juju wait-for on loop until successful?

tests/deployapp/deploy-basic-app.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+# This script creates a model and deploys a basic application to it
+# before running `juju status` and then destroys the model.
+
+set -euo pipefail
```
ale8k, line 6: Not using pipe fail?

local/jimm/setup-controller.sh
```diff
@@ -23,6 +24,14 @@ if [ "${SKIP_BOOTSTRAP:-false}" == true ]; then
   exit 0
 fi
 
+BOOTSTRAP_ARGS=(lxd "${CONTROLLER_NAME}" --config "${CLOUDINIT_FILE}")
+
+if [[ "$SKIP_CONNECT_JIMM" != "true" ]]; then
+  BOOTSTRAP_ARGS+=(--config "login-token-refresh-url=https://jimm.localhost/.well-known/jwks.json")
```
ale8k, line 30: Why? Let's make a new dir in local called juju with setup controller?

.github/workflows/integration-test.yaml
ale8k: Gotcha, but I'm confused about how JIMM is to be started. Why was JIMM started on manual run only?

ale8k: And how would JIMM be started here?

ale8k (approved):
Ty for this, it's exactly what I was thinking!

.github/workflows/integration-test.yaml
kian99: It was started on manual runs and on PRs. It was just a difference in which docker profile was started inside the `.github/actions/test-server` composite action. Now this workflow is just for PRs and doesn't serve two purposes (where the other was never used).

local/jimm/setup-controller.sh
kian99: There's logic in this file that I don't want to duplicate in order for the controller to be able to communicate with JIMM. But if we want to refactor things in a future PR, can do.

tests/deployapp/deploy-basic-app.sh
kian99: I don't follow? `set -o pipefail` sets pipefail. See https://gist.github.com/mohanpedala/1e2ff5661761d3abd0385e8223e16425#set--o-pipefail

tests/deployapp/deploy-basic-app.sh
kian99: Yeah but we don't actually need the app to be successful and waiting for an app to be ready can take a while so this is fine.

tests/run-tests.sh
kian99: Yip. But I think this might only apply to this file (not entirely sure on the bash inheritor semantics).
Good to have in any case.

internal/jujuapi/controllerroot.go
```diff
@@ -60,8 +60,8 @@ func newControllerRoot(j JIMM, p Params, identityId string) *controllerRoot {
 
 	r.AddMethod("Admin", 1, "Login", rpc.Method(unsupportedLogin))
 	r.AddMethod("Admin", 2, "Login", rpc.Method(unsupportedLogin))
-	r.AddMethod("Admin", 3, "Login", rpc.Method(unsupportedLogin))
-	r.AddMethod("Admin", 4, "Login", rpc.Method(unsupportedLogin))
```
SimoneDutto, line 64: why did you change the func just for version 3-4 and not 1-2?

SimoneDutto (approved):
LGTM

internal/jujuapi/controllerroot.go
kian99: Because 1 and 2 are very old and not used so I figure it wasn't needed there.

PR #1649, opened 2025-07-11
chore: allow PrepareModelMigration to be called multiple times

kian99:
## Description
Migrating a model to JAAS requires a client to call PrepareModelMigration first. This change makes it possible to call PrepareModelMigration multiple times and allows the initial user mapping or backing controller to change. JIMM will ensure that if the model migration has progressed up to a certain point that the PrepareModelMigration will return an error instead to ensure the data is consistent during the migration process.

Fixes [JUJU-8249](https://warthogs.atlassian.net/browse/JUJU-8249)

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

[JUJU-8249]: https://warthogs.atlassian.net/browse/JUJU-8249?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/db/modelmigration.go
+++ b/internal/db/modelmigration.go
@@ -66,11 +66,11 @@ func (d *Database) GetIncomingModelMigrationWithLock(ctx context.Context, modelM
 		return errors.E(op, "missing uuid", errors.CodeBadRequest)
 	}
 
+	lockingClause := clause.Locking{Strength: "UPDATE"}
 	if noWait {
-		db = db.Clauses(clause.Locking{Strength: "UPDATE", Options: "NOWAIT"})
-	} else {
-		db = db.Clauses(clause.Locking{Strength: "UPDATE"})
+		lockingClause.Options = "NOWAIT"
 	}
+	db = db.Clauses(lockingClause)
 	if err := db.First(&modelMigration).Error; err != nil {
 		err = dbError(err)
 		if errors.ErrorCode(err) == errors.CodeNotFound {
--- a/internal/db/modelmigration.go
+++ b/internal/db/modelmigration.go
@@ -15,10 +15,8 @@ import (
 
 // AddOrUpdateIncomingModelMigration stores information about an incoming model migration
 // if it does not already exist, or updates it if it does.
-// It must be run within a transaction as it will lock the existing row for updates
-// if the model migration already exists.
 func (d *Database) AddOrUpdateIncomingModelMigration(ctx context.Context, modelMigration *dbmodel.IncomingModelMigration) (err error) {
-	const op = errors.Op("db.AddIncomingModelMigration")
+	const op = errors.Op("db.AddOrUpdateIncomingModelMigration")
 	if err := d.ready(); err != nil {
 		return errors.E(op, err)
 	}
@@ -27,25 +25,29 @@ func (d *Database) AddOrUpdateIncomingModelMigration(ctx context.Context, modelM
 	defer durationObserver()
 	defer servermon.ErrorCounter(servermon.DBQueryErrorCount, &err, string(op))
 
-	lookup := dbmodel.IncomingModelMigration{
-		ModelUUID: modelMigration.ModelUUID,
-	}
-	// Set noWait to true to ensure that if the row is locked by another transaction,
-	// we will return an error immediately instead of waiting.
-	err = d.GetIncomingModelMigrationWithLock(ctx, &lookup, true)
-	if err == nil {
-		// If the model migration already exists, update it.
-		modelMigration.ID = lookup.ID
-	} else if err != nil && errors.ErrorCode(err) != errors.CodeNotFound {
-		return errors.E(op, err)
-	}
+	err = d.Transaction(func(d *Database) error {
+		lookup := dbmodel.IncomingModelMigration{
+			ModelUUID: modelMigration.ModelUUID,
+		}
+		// Set noWait to true to ensure that if the row is locked by another transaction,
+		// we will return an error immediately instead of waiting.
+		err = d.GetIncomingModelMigrationWithLock(ctx, &lookup, true)
+		if err == nil {
+			// If the model migration already exists, update it.
+			modelMigration.ID = lookup.ID
+		} else if err != nil && errors.ErrorCode(err) != errors.CodeNotFound {
+			return errors.E(op, err)
+		}
 
-	db := d.DB.WithContext(ctx)
+		db := d.DB.WithContext(ctx)
 
-	if err := db.Save(modelMigration).Error; err != nil {
-		return errors.E(op, dbError(err))
-	}
-	return nil
+		if err := db.Save(modelMigration).Error; err != nil {
+			return errors.E(op, dbError(err))
+		}
+
+		return nil
+	})
+	return err
 }
 
 // GetIncomingMigrationWithLock retrieves an incoming model migration locking the row for updates.
--- a/internal/db/modelmigration.go
+++ b/internal/db/modelmigration.go
@@ -6,15 +6,18 @@ import (
 	"context"
 	"time"
 
+	"gorm.io/gorm/clause"
+
 	"github.com/canonical/jimm/v3/internal/dbmodel"
 	"github.com/canonical/jimm/v3/internal/errors"
 	"github.com/canonical/jimm/v3/internal/servermon"
 )
 
-// AddIncomingModelMigration stores information about an incoming model migration.
-//   - returns an error with code errors.CodeAlreadyExists if
-//     a migration row with the same model UUID already exists.
-func (d *Database) AddIncomingModelMigration(ctx context.Context, modelMigration *dbmodel.IncomingModelMigration) (err error) {
+// AddOrUpdateIncomingModelMigration stores information about an incoming model migration
+// if it does not already exist, or updates it if it does.
+// It must be run within a transaction as it will lock the existing row for updates
+// if the model migration already exists.
+func (d *Database) AddOrUpdateIncomingModelMigration(ctx context.Context, modelMigration *dbmodel.IncomingModelMigration) (err error) {
 	const op = errors.Op("db.AddIncomingModelMigration")
 	if err := d.ready(); err != nil {
 		return errors.E(op, err)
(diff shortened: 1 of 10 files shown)


internal/db/modelmigration.go
```diff
@@ -6,15 +6,18 @@ import (
...
 
-// AddIncomingModelMigration stores information about an incoming model migration.
-//   - returns an error with code errors.CodeAlreadyExists if
-//     a migration row with the same model UUID already exists.
-func (d *Database) AddIncomingModelMigration(ctx context.Context, modelMigration *dbmodel.IncomingModelMigration) (err error) {
+// AddOrUpdateIncomingModelMigration stores information about an incoming model migration
+// if it does not already exist, or updates it if it does.
+// It must be run within a transaction as it will lock the existing row for updates
+// if the model migration already exists.
+func (d *Database) AddOrUpdateIncomingModelMigration(ctx context.Context, modelMigration *dbmodel.IncomingModelMigration) (err error) {
```
alesstimec, line 18: this allows for arbitrary update of user mapping.. i suppose this is intended, right?

internal/db/modelmigration.go
```diff
@@ -24,14 +27,60 @@ func (d *Database) AddIncomingModelMigration(ctx context.Context, modelMigration
...
+
+	db := d.DB.WithContext(ctx)
+	switch {
+	case modelMigration.ModelUUID.Valid:
+		db = db.Where("model_uuid = ?", modelMigration.ModelUUID.String)
+	default:
+		return errors.E(op, "missing uuid", errors.CodeBadRequest)
+	}
+
+	if noWait {
```
alesstimec, line 72: how about:
```
lockingClause := clause.Locking{ Strength: "UPDATE" }
if noWait {
  lockingClause.Options = "NOWAIT"
}
db = db.Clauses(lockingClause)
```

kind of reduces the indented bits, bit easier on the eyes

internal/db/modelmigration.go
kian99: Yip, that's the idea

internal/db/modelmigration.go
kian99: Done

internal/db/modelmigration_test.go
```diff
@@ -87,22 +101,94 @@ func (s *dbSuite) TestGetModelMigration(c *qt.C) {
...
+	notifyChan := make(chan struct{})
+	noWait := true
+	go func() {
+		// Lock the migration
+		err := s.Database.Transaction(func(d *db.Database) error {
+			lookup := dbmodel.IncomingModelMigration{
+				ModelUUID: migration.ModelUUID,
+			}
+			err := d.GetIncomingModelMigrationWithLock(context.Background(), &lookup, noWait)
+			c.Check(err, qt.Equals, nil)
```
alesstimec, line 126: are we ok using Check in a goroutine?

kian99: Yeah check() is okay as it will mark the test as failed but won't stop the running routine if it fails. Assert should be avoided in a go routine because if it fails the go routine will stop.

internal/jimm/juju/jimm.go
```diff
@@ -287,7 +287,20 @@ func (j *JujuManager) PrepareModelMigration(
 			return err
 		}
 
-		if err := j.Database.AddIncomingModelMigration(ctx, &dbmodel.IncomingModelMigration{
+		// Verify the model doesn't exist (implies a migration is in progress
```
SimoneDutto: the model existing implies the migration is in progress or complete.
Here from the comment it seems the opposite

internal/jimm/juju/jimm_test.go
```diff
@@ -804,3 +805,120 @@ func TestPrepareModelMigration_Success(t *testing.T) {
...
+			close(notifyChan)
+			<-ctx.Done()
+			return nil
+		})
+		c.Check(err, qt.IsNil)
+	}()
+
+	<-notifyChan // Wait for the migration to be locked
+	_, err = j.PrepareModelMigration(ctx, user, fakeModelUUID, targetController.Name, userMapping)
+	c.Assert(err, qt.IsNotNil)
```
SimoneDutto: could you test the specific error here?

SimoneDutto (approved):
just small comments

internal/jimm/juju/jimm.go
kian99: I see the confusion, I'll update the wording.

internal/db/modelmigration.go
```diff
@@ -24,14 +27,60 @@ func (d *Database) AddIncomingModelMigration(ctx context.Context, modelMigration
 	defer durationObserver()
 	defer servermon.ErrorCounter(servermon.DBQueryErrorCount, &err, string(op))
 
+	lookup := dbmodel.IncomingModelMigration{
+		ModelUUID: modelMigration.ModelUUID,
+	}
+	// Set noWait to true to ensure that if the row is locked by another transaction,
+	// we will return an error immediately instead of waiting.
+	err = d.GetIncomingModelMigrationWithLock(ctx, &lookup, true)
```
ale8k: Can we do the adding and updating explicitly please? So in calls to add or update, they'll attempt to get it first? We often do that.

internal/db/modelmigration.go
```diff
@@ -6,15 +6,18 @@ import (
...
-// AddIncomingModelMigration stores information about an incoming model migration.
-//   - returns an error with code errors.CodeAlreadyExists if
-//     a migration row with the same model UUID already exists.
-func (d *Database) AddIncomingModelMigration(ctx context.Context, modelMigration *dbmodel.IncomingModelMigration) (err error) {
+// AddOrUpdateIncomingModelMigration stores information about an incoming model migration
+// if it does not already exist, or updates it if it does.
+// It must be run within a transaction as it will lock the existing row for updates
+// if the model migration already exists.
+func (d *Database) AddOrUpdateIncomingModelMigration(ctx context.Context, modelMigration *dbmodel.IncomingModelMigration) (err error) {
 	const op = errors.Op("db.AddIncomingModelMigration")
```
ale8k: Op

internal/db/modelmigration.go
```diff
@@ -6,15 +6,18 @@ import (
...
 	"github.com/canonical/jimm/v3/internal/servermon"
 )
 
-// AddIncomingModelMigration stores information about an incoming model migration.
-//   - returns an error with code errors.CodeAlreadyExists if
-//     a migration row with the same model UUID already exists.
-func (d *Database) AddIncomingModelMigration(ctx context.Context, modelMigration *dbmodel.IncomingModelMigration) (err error) {
+// AddOrUpdateIncomingModelMigration stores information about an incoming model migration
+// if it does not already exist, or updates it if it does.
+// It must be run within a transaction as it will lock the existing row for updates
```
ale8k: If it must always be a transaction, I think we can just start a transaction in this method. You can nest transactions and they just act as savepoints.

internal/db/modelmigration.go
```diff
@@ -24,14 +27,60 @@ func (d *Database) AddIncomingModelMigration(ctx context.Context, modelMigration
...
+		// If the model migration already exists, update it.
+		modelMigration.ID = lookup.ID
+	} else if err != nil && errors.ErrorCode(err) != errors.CodeNotFound {
+		return errors.E(op, err)
+	}
+
 	db := d.DB.WithContext(ctx)
 
-	if err := db.Create(modelMigration).Error; err != nil {
+	if err := db.Save(modelMigration).Error; err != nil {
```
ale8k: IIRC using db.Model(x).Save(...) can lead to no updated_at field beign set, I'm guessing this is fine?

internal/db/modelmigration_test.go
```diff
@@ -75,7 +89,7 @@ func (s *dbSuite) TestGetModelMigration(c *qt.C) {
 	c.Assert(s.Database.DB.Create(&migration).Error, qt.IsNil)
 
 	lookup := dbmodel.IncomingModelMigration{ModelUUID: migration.ModelUUID}
-	err = s.Database.GetIncomingModelMigration(context.Background(), &lookup)
+	err := s.Database.GetIncomingModelMigration(context.Background(), &lookup)
```
ale8k, line 92: Please check it looks as intended here, we're just checking err but perhaps it actually gotten nothing

internal/db/modelmigration_test.go
```diff
@@ -44,28 +50,36 @@ func (s *dbSuite) TestAddModelMigration(c *qt.C) {
 	c.Assert(dbMigration.ModelUUID, qt.DeepEquals, migration.ModelUUID)
 	c.Assert(dbMigration.TargetControllerID, qt.Equals, controller.ID)
 	c.Assert(dbMigration.UserMapping, qt.DeepEquals, migration.UserMapping)
-
-	migration.ID = 0 // Reset ID to ensure it is not reused
-	err = s.Database.AddIncomingModelMigration(context.Background(), &migration)
-	c.Assert(err, qt.Not(qt.IsNil))
+	c.Assert(dbMigration.UpdatedAt, qt.Not(qt.Equals), time.Time{})
```
ale8k, line 53: Ahhh you checked the updated_at, scratch my other comment. Ty for this!

internal/jimm/juju/migrationtarget.go
```diff
@@ -420,26 +420,41 @@ func (j *JujuManager) Import(ctx context.Context, user *openfga.User, serialized
...
+	err = j.Database.Transaction(func(d *db.Database) error {
+		incomingMigration = &dbmodel.IncomingModelMigration{
+			ModelUUID: sql.NullString{String: modelDescription.Tag().Id(), Valid: true},
+		}
 
-	err = j.Database.GetIncomingModelMigration(ctx, incomingMigration)
-	if err != nil {
-		return errors.E(op, fmt.Errorf("failed to add incoming model migration: %w", err))
-	}
+		// Set noWait to false to allow the transaction to wait for the lock.
```
ale8k, line 438: Why false? Would we ever import the same migration multiple times and want it to queue?

internal/jimm/juju/jimm.go
```diff
@@ -287,7 +287,20 @@ func (j *JujuManager) PrepareModelMigration(
...
 		}
 
-		if err := j.Database.AddIncomingModelMigration(ctx, &dbmodel.IncomingModelMigration{
+		// Verify the model doesn't exist - if it does that means a migration is
+		// in progress or completed or it could also mean the model failed to be removed
+		// during migration ABORT but that problem should be dealt with separately.
+		model := &dbmodel.Model{
+			UUID: sql.NullString{String: modelUUID, Valid: true},
+		}
+		err := j.Database.GetModel(ctx, model)
```
ale8k: Could make a little func wrapping this called:

"CanPrepareModelMigration" or something, then you can put this doc there. Cause really it's just checking that it's safe to Prepare a model migration and can be godoc'd on a func.

ale8k (approved):

ale8k (approved):

ale8k (approved):
LGTM bar some minor things. But can we not have AddOrUpdate as a method. I'd prefer just Add and Update as separate methods (makes the code clearer). Having AddOrUpdate can lead to some funky stuff.

Can we also just put the TX within the Add/Update and not have godoc saying it must be run in a transaction. If it must be run in a transaction, place it in a transaction aha.

internal/jimm/juju/jimm.go
kian99: I think that's the same thing but hides the logic inside a function. Without needing to reuse that function I'll keep this here as is

internal/db/modelmigration.go
kian99: I'm confused on this one, we are getting it first?

internal/jimm/juju/migrationtarget.go
kian99: Not quite, it's simply that if another routine is holding the lock, the import method should wait to grab it rather than error out.

internal/db/modelmigration.go
```diff
@@ -27,25 +25,29 @@ func (d *Database) AddOrUpdateIncomingModelMigration(ctx context.Context, modelM
...
-	// Set noWait to true to ensure that if the row is locked by another transaction,
-	// we will return an error immediately instead of waiting.
-	err = d.GetIncomingModelMigrationWithLock(ctx, &lookup, true)
-	if err == nil {
-		// If the model migration already exists, update it.
-		modelMigration.ID = lookup.ID
-	} else if err != nil && errors.ErrorCode(err) != errors.CodeNotFound {
-		return errors.E(op, err)
-	}
+	err = d.Transaction(func(d *Database) error {
```
ale8k, line 28: Ty!

PR #1650, opened 2025-07-17
feat(bootstrap): implement bootstrap status facade method

SimoneDutto:
### First commit

Implement bootstrap status facade method by creating the bootstrap manager and exposing API to get status and logs of a bootstrap job.

To do so I needed to add:
- db method to GetJob
- jobID in the context for the `job` func
- tests

```terminal
[10:15:34] ➜  jimm git:(bootstrap-show-facade) ✗ stress -count 100 ./bootstrap.test -check.f bootstrapManagerSuite.TestGetBootstrapStatusAndLogs
5s: 48 runs so far, 0 failures, 12 active
9s: 100 runs total, 0 failures
```

### Second commit
Before I wasn't able to stress test TestGetBootstrapStatusAndLogs because during SetupTestOFGAClient
in RemoveStore it does "DELETE FROM authorization_model WHERE store = (SELECT id FROM store WHERE
name = %s"). This returns an error if the subquery returns multiple entries. And this can happen if
we stress test a test because it is setting up the store multiple times with the same name.


internal/jimm/bootstrap/bootstrap_test.go
```diff
@@ -0,0 +1,137 @@
...
+	s.ofgaClient = ofgaClient
+
+	jobtracker, err := jobtracker.NewJobTracker(db, 1*time.Minute)
+	s.jobTracker = jobtracker
+	c.Assert(err, qt.IsNil)
+	s.manager, err = bootstrap.NewBootstrapManager(db, ofgaClient, jobtracker)
+	c.Assert(err, qt.IsNil)
+}
+
+func (s *bootstrapManagerSuite) TestGetBootstrapStatusAndLogs(c *qt.C) {
```
SimoneDutto, line 54: very proud of this test

internal/jimm/bootstrap/bootstrap.go
```diff
@@ -0,0 +1,65 @@
...
+		store:      store,
+		authSvc:    authSvc,
+		jobtracker: jobtracker,
+	}, nil
+}
+
+// GetBootstrapStatusAndLogs retrieves the status and logs of a bootstrap job.
+// It requires the user to be an admin and returns the status, error message, logs,
+// and a watermark for pagination.
+func (b *bootstrapManager) GetBootstrapStatusAndLogs(ctx context.Context, _ *openfga.User, jobId uuid.UUID, offset int) (params.BootstrapStatusResponse, error) {
```
SimoneDutto, line 53: here i have a discussion point:
- we check the user a jimm admin in `jujuapi` pkg
- should we check it again here?
- trust the caller has already checked the user is a jimm admin?

internal/jimm/bootstrap/bootstrap.go
```diff
@@ -0,0 +1,65 @@
...
+
+// JobTracker interface defines the methods required for job tracking.
+type JobTracker interface {
+	// GetJob retrieves a job entry by its ID.
+	GetJob(ctx context.Context, jobId uuid.UUID) (dbmodel.JobTrackerEntry, error)
+}
+
+type bootstrapManager struct {
+	jobtracker JobTracker
+	store      *db.Database
```
kian99: Why is there a job tracker and the store? Is that because the logs need the DB but you want to test with mock jobs?

internal/jimm/bootstrap/bootstrap.go
```diff
@@ -0,0 +1,65 @@
...
+
+	"github.com/canonical/jimm/v3/internal/db"
+	"github.com/canonical/jimm/v3/internal/dbmodel"
+	"github.com/canonical/jimm/v3/internal/errors"
+	"github.com/canonical/jimm/v3/internal/openfga"
+	"github.com/canonical/jimm/v3/pkg/api/params"
+)
+
+// JobTracker interface defines the methods required for job tracking.
+type JobTracker interface {
```
ale8k, line 17: s/JobTracker/JobTrackerStore

internal/jimm/bootstrap/bootstrap.go
```diff
@@ -0,0 +1,65 @@
...
+
+// NewBootstrapManager creates a new BootstrapManager instance.
+func NewBootstrapManager(store *db.Database, authSvc *openfga.OFGAClient, jobtracker JobTracker) (*bootstrapManager, error) {
+	if store == nil {
+		return nil, errors.E("group store cannot be nil")
+	}
+	if authSvc == nil {
+		return nil, errors.E("group authorisation service cannot be nil")
+	}
+	return &bootstrapManager{
```
kian99, line 43: We can check the jobtracker is not nil too.

internal/jimm/bootstrap/bootstrap.go
ale8k: Trust it, it's ok.

internal/jimm/bootstrap/bootstrap.go
ale8k: We shouldn't need the entire db, be good to define an interface here so we can mock it.

internal/jimm/bootstrap/bootstrap_test.go
```diff
@@ -0,0 +1,137 @@
...
+			}
+			write <- struct{}{}
+			<-read
+			return nil
+		},
+		1*time.Minute,
+	)
+	c.Assert(err, qt.IsNil)
+	watermark := 0
+	for i := 0; i < numLogs/batchSize+1; i++ {
```
kian99: Replace i with `batch`?

internal/jimm/bootstrap/bootstrap_test.go
```diff
@@ -0,0 +1,137 @@
...
+		<-write // Wait for the batch of logs to be written.
+		response, err := s.manager.GetBootstrapStatusAndLogs(ctx, s.adminUser, jobId, watermark)
+		c.Assert(err, qt.IsNil)
+		logs := []string{}
+		for j := 0; j < int(math.Min(float64(batchSize), float64(numLogs-i*batchSize))); j++ {
```
kian99: Replace `j` with `logNumber`?

internal/jimm/bootstrap/bootstrap.go
```diff
@@ -0,0 +1,65 @@
...
+// and a watermark for pagination.
+func (b *bootstrapManager) GetBootstrapStatusAndLogs(ctx context.Context, _ *openfga.User, jobId uuid.UUID, offset int) (params.BootstrapStatusResponse, error) {
+	const op = errors.Op("jimm.GetBootstrapStatusAndLogs")
+
+	job, err := b.jobtracker.GetJob(ctx, jobId)
+	if err != nil {
+		return params.BootstrapStatusResponse{}, errors.E(op, "failed to get job status", err)
+	}
+
+	logs, newOffset, err := b.store.QueryBootstrapLog(ctx, jobId, offset)
```
ale8k: Can we call it loggies? (Internal joke I did in the query haha) 😜

internal/jimm/bootstrap/bootstrap_test.go
```diff
@@ -0,0 +1,137 @@
...
+	s.jobTracker = jobtracker
+	c.Assert(err, qt.IsNil)
+	s.manager, err = bootstrap.NewBootstrapManager(db, ofgaClient, jobtracker)
+	c.Assert(err, qt.IsNil)
+}
+
+func (s *bootstrapManagerSuite) TestGetBootstrapStatusAndLogs(c *qt.C) {
+	ctx := c.Context()
+	read := make(chan struct{})
+	defer close(read)
```
ale8k, line 57: Some space between some vars would help readability for me like uhm:
```go
	ctx := c.Context()

	read := make(chan struct{})
    write := make(chan struct{})
	defer close(read)
	defer close(write)

	numLogs := 101
	batchSize := 10
```

But feel free to ignore, and hurt my feelings 😂

internal/jimm/bootstrap/bootstrap_test.go
```diff
@@ -0,0 +1,137 @@
...
+	write := make(chan struct{})
+	defer close(write)
+	numLogs := 101
+	batchSize := 10
+	jobId, err := s.jobTracker.Run(ctx,
```
ale8k, line 64: And like here, we got ctx on 62, but rest of params on new lines. Again hurt my feelings if you want haha.

internal/jimm/bootstrap/bootstrap_test.go
```diff
@@ -0,0 +1,137 @@
...
+
+			for i := range numLogs {
+				if i%batchSize == 0 && i > 0 {
+					write <- struct{}{} // Signal that a batch of logs has been written.
+					<-read              // Wait for the read before writing the next batch.
+				}
+				err := s.db.AddBootstrapLog(ctx, jobId, "bootstrap logs "+fmt.Sprint(rune(i)))
+				c.Check(err, qt.IsNil)
+			}
+			write <- struct{}{}
```
ale8k, line 79: Why do we need this final write and read?

internal/jimm/bootstrap/bootstrap_test.go
```diff
@@ -0,0 +1,137 @@
...
+			<-read
+			return nil
+		},
+		1*time.Minute,
```
ale8k, line 83: I'd probably set this quite a bit shorter as it is your only way to clean up the routine. And additionally, in your job routine we can check the test context isn't done, and if it is we return. Just as a safety precaution to ensure the routine is cleaned up.

internal/jimm/bootstrap/bootstrap_test.go
ale8k: This kinda reads like a cluster of code, please save my feelings 😝

internal/jobtracker/jobtracker_test.go
```diff
@@ -105,6 +105,21 @@ func (s *jobTrackerSuite) TestRun_JobSetSuccessful(c *qt.C) {
 	s.pollJob(testCtx, id, c, dbmodel.StatusSuccessful)
 }
 
+func (s *jobTrackerSuite) TestRun_JobIdSetInContext(c *qt.C) {
+	testCtx := c.Context()
+
+	aJobWithIdInContext := func(ctx context.Context) error {
+		_, ok := jobtracker.JobIdFromContext(ctx)
```
kian99: Minor but I suggest to check that the value is non-zero or something along those lines.

internal/jimm/bootstrap/bootstrap_test.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+func (s *bootstrapManagerSuite) TestGetBootstrapStatusAndLogs(c *qt.C) {
+	ctx := c.Context()
+	read := make(chan struct{})
+	defer close(read)
+	write := make(chan struct{})
+	defer close(write)
+	numLogs := 101
+	batchSize := 10
```
ale8k, line 62: I think a simpler test where we just read 1, then 2, then 1 again would read better than batching.

kian99 (approved):

internal/jimm/bootstrap/bootstrap.go
kian99: Imo it's sufficient as is with the check in `jujuapi`.

internal/jimm/bootstrap/bootstrap_test.go
```diff
@@ -0,0 +1,137 @@
...
+	watermark := 0
+	for i := 0; i < numLogs/batchSize+1; i++ {
+		<-write // Wait for the batch of logs to be written.
+		response, err := s.manager.GetBootstrapStatusAndLogs(ctx, s.adminUser, jobId, watermark)
+		c.Assert(err, qt.IsNil)
+		logs := []string{}
+		for j := 0; j < int(math.Min(float64(batchSize), float64(numLogs-i*batchSize))); j++ {
+			logs = append(logs, "bootstrap logs "+fmt.Sprint(rune(i*batchSize+j)))
+		}
+		c.Check(response.Logs, qt.DeepEquals, logs)
```
ale8k, line 96: We can assert this?

internal/jimm/bootstrap/bootstrap_test.go
```diff
@@ -0,0 +1,137 @@
...
+		c.Assert(response.Status, qt.Equals, string(dbmodel.StatusRunning))
+		watermark = response.Watermark
+		read <- struct{}{} // Signal it has been read.
+	}
+
+	// check last batch is empty.
+	response, err := s.manager.GetBootstrapStatusAndLogs(ctx, s.adminUser, jobId, watermark)
+	c.Assert(response.Status == string(dbmodel.StatusSuccessful) || response.Status == string(dbmodel.StatusRunning), qt.IsTrue)
```
ale8k, line 105: Shouldn't we only wait for it to be successful? Can pinch my polling code from the job tracker test

internal/jujuapi/jimm.go
```diff
@@ -558,3 +561,19 @@ func (r *controllerRoot) PrepareModelMigration(ctx context.Context, args apipara
 
 	return resp, nil
 }
+
+// BootstrapStatus retrieves the status of a bootstrap job and its logs.
```
ale8k: It also returns the next watermark for consumers to use.

internal/jujuapi/jimm.go
```diff
@@ -558,3 +561,19 @@ func (r *controllerRoot) PrepareModelMigration(ctx context.Context, args apipara
...
+func (r *controllerRoot) BootstrapStatus(ctx context.Context, req apiparams.BootstrapStatusRequest) (apiparams.BootstrapStatusResponse, error) {
+	const op = errors.Op("jujuapi.BootstrapStatus")
+
+	if !r.user.JimmAdmin {
```
ale8k, line 570: Now, thinking about this, admin is ok for now, but maybe we should be checking they own the credential too. Admins should never be able to use others credentials right? Or users in general, so that means someone (who might not be the admin) started a bootstrap but the admin can view their logs? Maybe a little discussion point for later as I'm not sure we'd doing permissions right here. I think we should be checking admin + owns credential for this bootstrap (idk how we'd check that here though 😢)

internal/jujuapi/jimm_unit_test.go
```diff
@@ -80,3 +105,48 @@ func (s *jimmSuite) TestPrepareModelMigration_InvalidUserMapping(c *gc.C) {
...
+
+	// Test job not found
+	_, err = root.BootstrapStatus(ctx, params.BootstrapStatusRequest{
+		JobID:     uuid.New().String(),
+		Watermark: 0,
+	})
+	c.Assert(errors.ErrorCode(err), gc.Equals, errors.CodeNotFound)
+
+	// Test unauthorized user
+	root = newTestControllerRoot(jimm, "alice@canonical.com", false)
```
ale8k, line 146: Yeah as I read this it is feeling like it doesn't make sense that an admin can view the progress of a bootstrap that might not be theirs.

pkg/api/params/params.go
```diff
@@ -560,3 +560,23 @@ type PrepareModelMigrationResponse struct {
 	// as it allows the source controller to authenticate with JIMM.
 	Token string `json:"token" yaml:"token"`
 }
+
+// BootstrapStatusRequest holds the request to get the status of a bootstrap job.
+type BootstrapStatusRequest struct {
+	// JobID is the ID of the bootstrap job to get the status for.
+	JobID string `json:"job-id"`
```
ale8k, line 567: Missing yaml tags

internal/jimm/bootstrap/bootstrap.go
SimoneDutto: because the job tracker is just to start, stop, get status for the job.
However to get the bootstrap logs you can't use the jobtracker but you need the database

pkg/api/params/params.go
```diff
@@ -560,3 +560,23 @@ type PrepareModelMigrationResponse struct {
...
+	// Watermark is the line number to start reading logs from.
+	Watermark int `json:"watermark"`
```
ale8k, line 569: This is omittable no? 0 is fine?

ale8k (approved):
Beeeautiful. Thanks for this! Some minor comments (plz do them) but LGTM otherwise!

internal/jujuapi/jimm.go
SimoneDutto: that's the reason why the bootstrap manager has the openfga client and i'm passing the user down to the methods.
So we can later implement some fine-grained checks

internal/jimm/bootstrap/bootstrap.go
SimoneDutto: why? it's not really a store, since it runs a go-routine for you

internal/jimm/bootstrap/bootstrap_test.go
SimoneDutto: left a comment

internal/jimm/bootstrap/bootstrap_test.go
SimoneDutto: Nope, because we are in a go routine

internal/jimm/bootstrap/bootstrap_test.go
SimoneDutto: I dind't want to poll, so i'm just checking the job is either running or successful.
I think it's fine because here we are not testing that the jobtracker is marking the job as successful

pkg/api/params/params.go
SimoneDutto: we would always set this field from the bootstrap manager, so it would never be omitted.

internal/jimm/bootstrap/bootstrap.go
SimoneDutto: for sure!

internal/jimm/bootstrap/bootstrap_test.go
SimoneDutto: isn't always cleanup just by the fact we return from the `job` func?

PR #1651, opened 2025-07-17
chore: add model-migration integration test

kian99:
## Description
This PR is a follow-up to #1648 to add an integration test validating migration of a model from a standalone Juju controller to JAAS.

The script is long and somewhat complicated but in summary it:
- Assumes JIMM is running with a controller called "qa-lxd" attached.
- Creates a new controller that can resolve JIMM's hostname but is not connected to JIMM.
- Creates 2 models, each with an app and creates a CMR (cross-model relation) between them.
- Move the providing model to JIMM.
- Validates the relation still works.
- Tears down both models.
- The controller is kept alive in case it can be re-used by other tests.

The script is idempotent (to a point) so that it can be re-run if stopped early.

Fixes [JUJU-7952](https://warthogs.atlassian.net/browse/JUJU-7952)
## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests

[JUJU-7952]: https://warthogs.atlassian.net/browse/JUJU-7952?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/modelmigration/migrate-providing-model.sh
+++ b/tests/modelmigration/migrate-providing-model.sh
@@ -1,0 +1,207 @@
+#!/bin/bash
+
+# This script assumes that you have JIMM running with a controller
+# named "qa-lxd" attached (can be overriden via BACKING_CONTROLLER_NAME).
+#
+# This script creates a controller called source-controller and 
+# then creates two models: source-model and sink-model, 
+# relates them, and then migrates the source model to JIMM.
+#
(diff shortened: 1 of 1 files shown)


alesstimec (approved):

SimoneDutto (approved):
big and beautiful script

PR #1652, opened 2025-07-17
feat(bootstrap): bootstrap-status cli

SimoneDutto:
## Description
Add bootstrap status cli.

In this PR to test this command i've added:
- JIMMClient interface to create gomock on
- unit test by using the mocked client and mocked stdout


cmd/jaas/cmd/bootstrap_status.go
```diff
@@ -0,0 +1,127 @@
...
+	"github.com/juju/juju/jujuclient"
+
+	"github.com/canonical/jimm/v3/internal/errors"
+	"github.com/canonical/jimm/v3/pkg/api"
+	"github.com/canonical/jimm/v3/pkg/api/params"
+)
+
+const (
+	bootstrapStatusCommandDoc = `
+Displays logs from a bootstrap job.
```
ale8k: For* the job could still be running.

cmd/jaas/cmd/bootstrap_status.go
```diff
@@ -0,0 +1,127 @@
...
+`
+	bootstrapStatusCommandExample = `
+    juju bootstrap-status 2cb433a6-04eb-4ec4-9567-90426d20a004 
+`
+)
+
+// NewbootstrapStatusCommand returns a command to display full model status.
```
ale8k: NewBootstrap... and it's logs not full model status?

cmd/jaas/cmd/bootstrap_status.go
```diff
@@ -0,0 +1,127 @@
...
+func NewBootstrapStatusCommand() cmd.Command {
+	cmd := &bootstrapStatusCommand{
+		store: jujuclient.NewFileClientStore(),
+	}
+
+	return modelcmd.WrapBase(cmd)
+}
+
+// bootstrapStatusCommand displays full
```
ale8k: Full model status aha?

cmd/jaas/cmd/bootstrap_status.go
```diff
@@ -0,0 +1,127 @@
...
+// model status.
+type bootstrapStatusCommand struct {
+	modelcmd.ControllerCommandBase
+
+	store    jujuclient.ClientStore
+	dialOpts *jujuapi.DialOpts
+	jobId    string
+	client   JIMMClient
```
ale8k: Why the interface.go file?

cmd/jaas/cmd/bootstrap_status.go
```diff
@@ -0,0 +1,127 @@
...
+}
+
+func (c *bootstrapStatusCommand) Info() *cmd.Info {
```
ale8k: godoc

cmd/jaas/cmd/bootstrap_status.go
```diff
@@ -0,0 +1,127 @@
...
+	return jujucmd.Info(&cmd.Info{
+		Name:     "bootstrap-status",
+		Args:     "<job uuid>",
+		Purpose:  "Displays full model status",
```
ale8k: Model status?

cmd/jaas/cmd/bootstrap_status.go
```diff
@@ -0,0 +1,127 @@
...
+		Doc:      bootstrapStatusCommandDoc,
+		Examples: bootstrapStatusCommandExample,
+	})
+}
+
+// SetFlags implements Command.SetFlags.
+func (c *bootstrapStatusCommand) SetFlags(f *gnuflag.FlagSet) {
```
ale8k: implements the cmd.Command interface. ?

cmd/jaas/cmd/bootstrap_status.go
```diff
@@ -0,0 +1,127 @@
...
+	apiCaller, err := c.NewAPIRootWithDialOpts(c.store, currentController, "", c.dialOpts)
+	if err != nil {
+		return err
+	}
+
+	c.client = api.NewClient(apiCaller)
+	return nil
+}
+
+// Run implements Command.Run.
```
ale8k: Fix godocs to be consistent when we're just saying it implements X.

cmd/jaas/cmd/bootstrap_status.go
```diff
@@ -0,0 +1,127 @@
...
+func (c *bootstrapStatusCommand) Run(ctxt *cmd.Context) error {
+	watermark := 0
+	for {
+		response, err := c.client.BootstrapStatus(&params.BootstrapStatusRequest{
```
ale8k: Could've used recursion here, missed your opportunity 😜

pkg/api/params/params.go
```diff
@@ -560,3 +560,33 @@ type PrepareModelMigrationResponse struct {
...
+	JobID string `json:"job-id"`
+	// Watermark is the line number to start reading logs from.
+	Watermark int `json:"watermark"`
+}
+
+// JobStatus represents the status of a bootstrap job.
+type JobStatus string
+
+const (
+	StatusRunning    JobStatus = "running"
```
ale8k, line 576: I think these should be set to the dbmodel statuses, as if we ever change them it'll break these.

cmd/jaas/cmd/bootstrap_status.go
```diff
@@ -0,0 +1,127 @@
...
+		}
+		switch response.Status {
+		case params.StatusRunning:
+			for _, log := range response.Logs {
+				_, err = ctxt.Stdout.Write([]byte(log + "\n"))
+				if err != nil {
+					return errors.E(err, "failed to write bootstrap log")
+				}
+			}
+			watermark = response.Watermark
```
ale8k: So we loop indefinitely, but this is under the presumption the job ends. If for any reason the status is never changed from running, this will just run forever? I think we should have a default timeout and allow users to specify?

cmd/jaas/cmd/bootstrap_status_unit_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+	command := &bootstrapStatusCommand{
+		client: s.client,
+		jobId:  "test-job-id",
+	}
+	ctx := &cmd.Context{
+		Context: c.Context(),
+		Stdout:  s.writer,
+	}
+	err := command.Run(ctx)
```
ale8k: It'd be good to test the status cases all get hit and not just successful.

It'd also be good to see the watermark logic is OK too.

And lastly it'd be good to see some logs actually come out of the stdout?

ale8k (changes requested):

ale8k (changes requested):
Reviewed just the CLI side, I don't think it should be allowed to loop forever & some more test cases (like suggested) would be good!

cmd/jaas/cmd/bootstrap_status.go
SimoneDutto: because i can see in the future we could have other commands using this common JIMMClient interface

cmd/jaas/cmd/bootstrap_status.go
SimoneDutto: trueee

pkg/api/params/params.go
SimoneDutto: i think using dbmodel types in `params` is wrong.
To make sure we don't break compatibility i've created tests in `dbmodel` to make sure they are the same.

cmd/jaas/cmd/bootstrap_status.go
SimoneDutto: I think for this command is fine, because you are watching some logs up until it returns because it is successful or failed.
I see this command similar to `tail -f`.

It should be responsability of JIMM to make sure we don't have a `running` job forever.

In addition, users can always just do ctrl+c

cmd/jaas/cmd/bootstrap_status_unit_test.go
SimoneDutto: yeah, sorry was waiting for the other to go in before making changes here!

cmd/jaas/cmd/bootstrap_status.go
```diff
@@ -0,0 +1,136 @@
...
+// NewbootstrapStatusCommand returns a command to display logs for a bootstrap job.
+func NewBootstrapStatusCommand() cmd.Command {
+	cmd := &bootstrapStatusCommand{
+		store: jujuclient.NewFileClientStore(),
+	}
+
+	return modelcmd.WrapBase(cmd)
+}
+
+// bootstrapStatusCommand displays ogs for a bootstrap job.
```
kian99:
```suggestion
// bootstrapStatusCommand displays logs for a bootstrap job.
```

cmd/jaas/cmd/bootstrap_status.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	apiCaller, err := c.NewAPIRootWithDialOpts(c.store, currentController, "", c.dialOpts)
+	if err != nil {
+		return err
+	}
+
+	c.client = api.NewClient(apiCaller)
+
+	c.sleepBetweenGetLogs = sleepBetweenGetLogs
```
kian99: Is it a config value for tests?

cmd/jaas/cmd/bootstrap_status.go
kian99: I was going to suggest adding a timeout to the context and checking at the end of each loop if the context has timed out. But thinking about how this command is similar to something like `journalctl`, if run without arguments it should run and return immediately. If a flag like `--follow` is added we should loop forever.

cmd/jaas/cmd/bootstrap_status.go
```diff
@@ -0,0 +1,136 @@
...
+			return nil
+		case params.StatusPending:
+			_, err := ctxt.Stdout.Write([]byte("Bootstrap job is pending...\n"))
+			if err != nil {
+				return errors.E(err, "failed to write bootstrap pending message")
+			}
+		default:
+			return errors.E("unknown bootstrap job status: %s", response.Status)
+		}
+		time.Sleep(c.sleepBetweenGetLogs)
```
kian99: I need a reminder why the server can't stream logs to the client over the websocket as they become available. I know we have Juju RPC to adhere to but I'm trying to remember why this doesn't work the same way as `juju debug-log`.

pkg/api/params/params.go
kian99: Agreed that we shouldn't use dbmodel types in this package.

kian99 (approved):
lgtm but the behaviour of looping forever should only be done if a flag like `--follow` is used, otherwise the command should immediately return.

cmd/jaas/cmd/bootstrap_status.go
SimoneDutto: Yeah, sort of.

The approach to do unit testing here is fairly similar to how they do it in juju.

We do whitebox testing and manually set private fields.

We could go for `NewBootstrapCommand(cfg config.BootstrapCommandConfig)` to avoid whitebox testing.

I though that since we were introducing gomock we could get closer to juju's way

cmd/jaas/cmd/bootstrap_status.go
SimoneDutto: Yeah, i like this one!

cmd/jaas/cmd/bootstrap_status.go
SimoneDutto: I'm not sure, i've just followed the spec because i thought there were some known limitations around using RPC calls as they are.

I thought we could implement a watcher as well, but is it really worth to introduce the watcher paradigm for logs?

kian99: The debug-log doesn't use a watcher, it just streams logs via a websocket that doesn't conform to Juju RPC and instead the server only sends to the client (no client to server messages are sent). Maybe that's the reason - because for bootstrap logs, you need to send an initial RPC request with the bootstrap ID.

cmd/jaas/cmd/bootstrap_status.go
kian99, line 1: This file should be renamed to `bootstrapstatus` btw.

ale8k: Underscore is fine no?

cmd/jaas/cmd/bootstrap_status.go
ale8k: Agreed!

cmd/jaas/cmd/bootstrap_status.go
ale8k: You're on the money, we wanted to send the ID and watermark the logs into a resp. I do think we could still do it the other way but we voted on doing this.

pkg/api/params/params.go
ale8k: There needs to be something linking it to that status, perhaps a constants package in the future? Fine for now.

ale8k (approved):
LGTM, ty for this!

PR #1653, opened 2025-07-21
chore: add bootstrap start api types

kian99:
## Description
This PR adds new API types for starting a controller bootstrap job.

Note that the file `pkg/api/params/params.go` file contains structs where some include both `yaml` and `json` tags while others only include a `json` tag. I believe only the JSON tags are necessary because Juju RPC only works by unmarshalling JSON, it doesn't offer support for yaml and the yaml tags were added unnecessarily.


SimoneDutto (approved):

kian99:
There are some issues with this that I found when doing the facade method. Closing and will handle this alongside the facade method.

PR #1654, opened 2025-07-21
feat: bootstrap start facade

kian99:
## Description
This PR adds a new facade method to JIMM called `bootstrapStart` for starting a bootstrap job.

There are a few things that are worth discussing that I'll bring up offline, mainly:
- The use of error fields in our RPC structs. We don't need to put error fields in the return types.
- What type we return from the `internal/jimm` layer.
- Where we run validation.

Fixes [JUJU-8149](https://warthogs.atlassian.net/browse/JUJU-8149)

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

[JUJU-8149]: https://warthogs.atlassian.net/browse/JUJU-8149?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/jimm/bootstrap/types.go
+++ b/internal/jimm/bootstrap/types.go
@@ -1,0 +1,40 @@
+// Copyright 2025 Canonical.
+
+package bootstrap
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/canonical/jimm/v3/internal/errors"
+)
+
+// BootstrapParams defines the parameters required for bootstrapping a JIMM controller.
+type BootstrapParams struct {
+	ControllerName string
+	CloudName      string
+	CloudRegion    string
+	AgentVersion   string
+	TimeoutSeconds int
+}
+
+// Validate checks if the BootstrapParams are valid.
+func (p BootstrapParams) validate() error {
+	var msgs []string
+	if p.ControllerName == "" {
+		msgs = append(msgs, "controller name cannot be empty")
+	}
+	if p.CloudName == "" {
+		msgs = append(msgs, "cloud name cannot be empty")
+	}
+	if p.CloudRegion == "" {
+		msgs = append(msgs, "cloud region cannot be empty")
+	}
+	if p.AgentVersion == "" {
+		msgs = append(msgs, "agent version cannot be empty")
+	}
+	if msgs != nil {
+		return errors.E(fmt.Sprintf("invalid bootstrap parameters:\n%s", strings.Join(msgs, "\n")))
+	}
+	return nil
+}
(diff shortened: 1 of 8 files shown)


internal/jimm/bootstrap/types.go
```diff
@@ -0,0 +1,38 @@
...
+package bootstrap
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/canonical/jimm/v3/internal/errors"
+)
+
+type BootstrapParams struct {
```
kian99, line 13: This is essentially a copy of the params type. If we want we can just pass the params type down into `internal/jimm/bootstrap` instead.

internal/jujuapi/jimm.go
```diff
@@ -101,7 +104,8 @@ func init() {
 		// JIMM Model Migrations
 		r.AddMethod("JIMM", 4, "PrepareModelMigration", prepareModelMigration)
 		// JIMM Bootstrap
-		r.AddMethod("JIMM", 4, "BootstrapStatus", rpc.Method(r.BootstrapStatus))
+		r.AddMethod("JIMM", 4, "BootstrapStatus", bootstrapStatus)
+		r.AddMethod("JIMM", 4, "BootstrapStart", bootstrapStart)
```
SimoneDutto, line 108: what do we gain from defining the func instead of inlining it?

SimoneDutto (approved):
LGTM

internal/jujuapi/jimm.go
kian99: Consistency

internal/jimm/bootstrap/types.go
```diff
@@ -0,0 +1,40 @@
...
+	ControllerName string
+	CloudName      string
+	CloudRegion    string
+	AgentVersion   string
+	TimeoutSeconds int
+}
+
+// Validate checks if the BootstrapParams are valid.
+func (p BootstrapParams) validate() error {
+	var msgs []string
```
alesstimec, line 23: while this is a very user friendly approach, the resulting error message might be confusing (as it returns a multi-line message).. i would prefer to just return a simple error message in each of the if statements.. but that's just me.. try try try again, until you succeed :)

internal/jujuapi/jimm.go
alesstimec: at some point we'll have to think about JIMM facade versioning. It's getting a bit confusing now, with version 4 implementing everything SSH related and the bootstrappy bits, where the former might not be usable until, well, who knows when..

alesstimec (approved):
LGTM with a silent *grunt*

internal/jimm/bootstrap/types.go
kian99: Ha, gotta be kind to the users.

PR #1655, opened 2025-07-21
feat(bootstrap): create a binary puller to download juju client binaries to a folder

SimoneDutto:
## Description
Create Puller to download binaries from Launchpad.

The idea is to create a temp file in a specified folder, download the binary and return a struct with the path to the binary and a cleanup func to be called to remove the file.

## QA

Add this test
```go
func (s *pullerSuite) TestPullReal(c *qt.C) {
	jujuSpec := JujuBinarySpec{
		VersionWithMinor: "3.6",
		VersionWithFix:   "3.6.2",
		Os:               "linux",
		Arch:             "amd64",
	}

	puller, err := NewPuller(PullerConfig{})
	c.Assert(err, qt.IsNil)
	_, err = puller.Pull(context.Background(), jujuSpec)
	c.Assert(err, qt.IsNil)
}

```

`ls -lh /tmp` and you should the binary downloaded

### DIFF
--- a/internal/jujuclifetcher/fetcher.go
+++ b/internal/jujuclifetcher/fetcher.go
@@ -1,0 +1,147 @@
+// Copyright 2025 Canonical.
+
+package jujuclifetcher
+
+import (
+	"bytes"
+	"context"
+	"html/template"
+	"io"
+	"net/http"
+	"os"
+	"time"
+
+	"github.com/juju/clock"
+	jujuerrrors "github.com/juju/errors"
+	"github.com/juju/retry"
+	"github.com/juju/zaputil/zapctx"
+	"go.uber.org/zap"
+)
+
+// launchPadURL is the base URL for the Juju binary downloads.
+const launchPadURL = "https://launchpad.net/juju"
+
+// launchPadTemplate is the template for constructing the download URL for Juju binaries.
+const launchPadTemplate = "{{.BaseURL}}/{{.VersionWithMinor}}/{{.VersionWithPatch}}/+download/juju-{{.VersionWithPatch}}-{{.Os}}-{{.Arch}}.tar.xz"
+
+var retryRequestError = jujuerrrors.New("retry request error")
+
+// jujuCLIFetcherConfig holds the configuration for the Juju binary fetcher.
+type jujuCLIFetcherConfig struct {
+	BaseURL string // Base URL for the Juju binary downloads. Example: "https://launchpad.net/juju"
+	Dir     string // Directory to store the downloaded binaries. Defaults to the system's temp directory.
+}
+
+// JujuBinarySpec defines the specifications for a Juju binary to be downloaded.
+type JujuBinarySpec struct {
+	VersionWithMinor string // Version with minor version number, e.g., "3.6"
+	VersionWithPatch string // Version with patch version number, e.g., "3.6.2"
+	Os               string // Operating system, e.g., "linux"
+	Arch             string // Architecture, e.g., "amd64"
+}
+
+type jujuCLIFetcher struct {
+	config   jujuCLIFetcherConfig
+	template template.Template
+}
+
+// NewJujuCLIFetcher creates a new jujuCLIFetcher instance with the provided configuration.
+// If the BaseURL is not provided, it defaults to the launchpad URL.
+func NewJujuCLIFetcher(cfg jujuCLIFetcherConfig) (*jujuCLIFetcher, error) {
+	if cfg.BaseURL == "" {
+		// Default to the launchpad URL if no base URL is provided.
+		cfg.BaseURL = launchPadURL
+	}
+	tmpl, err := template.New("URL").Parse(launchPadTemplate)
+	if err != nil {
+		return nil, err
+	}
+	return &jujuCLIFetcher{
+		config:   cfg,
+		template: *tmpl,
+	}, nil
+}
+
+// Binary represents a downloaded Juju binary.
+// It contains the full path to the binary file.
+// It also provides a method to clean up the binary file.
+type Binary struct {
+	FullPath string
+}
+
+// Remove removes the binary file from the filesystem.
+// It returns an error if the file cannot be removed.
+func (b Binary) Remove() error {
+	return os.Remove(b.FullPath)
+}
+
+// Fetch downloads the Juju binary specified by the JujuBinarySpec.
+// It returns a Binary instance containing the full path to the downloaded binary.
+// If the download fails, it returns an error.
+// It retries the download on server errors or rate limiting.
+// The retry logic uses exponential backoff.
+// The context can be used to cancel the operation.
+func (p jujuCLIFetcher) Fetch(ctx context.Context, spec JujuBinarySpec) (*Binary, error) {
+	var buf bytes.Buffer
+	err := p.template.Execute(&buf, map[string]string{
+		"BaseURL":          p.config.BaseURL,
+		"VersionWithMinor": spec.VersionWithMinor,
+		"VersionWithPatch": spec.VersionWithPatch,
+		"Os":               spec.Os,
+		"Arch":             spec.Arch,
+	})
+	if err != nil {
+		return nil, err
+	}
+	var file *os.File
+	err = retry.Call(retry.CallArgs{
+		Func: func() error {
+			resp, err := http.Get(buf.String())
+			if err != nil {
+				return err
+			}
+			defer resp.Body.Close()
+			// Retry on server errors or rate limiting.
+			if (resp.StatusCode >= 500 && resp.StatusCode < 600) || resp.StatusCode == http.StatusTooManyRequests {
+				return retryRequestError
+			}
+			if resp.StatusCode != http.StatusOK {
+				return jujuerrrors.Errorf("request failed with status %d", resp.StatusCode)
+			}
+			file, err = os.CreateTemp(p.config.Dir, "juju-*")
+			if err != nil {
+				return err
+			}
+			defer file.Close()
+			_, err = io.Copy(file, resp.Body)
+			if err != nil {
+				return err
+			}
+			return nil
+		},
+		IsFatalError: func(err error) bool {
+			return !jujuerrrors.Is(err, retryRequestError)
+		},
+		BackoffFunc: retry.DoubleDelay,
+		Attempts:    10,
+		Delay:       time.Second,
+		Clock:       clock.WallClock,
+		Stop:        ctx.Done(),
+	})
+	if err != nil {
+		if file != nil {
+			// Clean up the file if it was created.
+			// We ignore the error because there is not much we can do about it.
+			errRemove := os.Remove(file.Name())
+			if errRemove != nil {
+				zapctx.Error(ctx, "failed to remove temporary file", zap.Error(errRemove), zap.String("file", file.Name()))
+			}
+		}
+		return nil, err
+	}
+	defer file.Close()
+
+	return &Binary{
+		FullPath: file.Name(),
+	}, nil
+}
--- a/internal/jujuclifetcher/fetcher_test.go
+++ b/internal/jujuclifetcher/fetcher_test.go
@@ -1,0 +1,112 @@
+// Copyright 2025 Canonical.
+
+package jujuclifetcher
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"testing"
+
+	qt "github.com/frankban/quicktest"
+)
+
+func TestFetch(t *testing.T) {
+	c := qt.New(t)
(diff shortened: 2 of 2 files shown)


internal/jujubinarypuller/puller.go
```diff
@@ -0,0 +1,140 @@
...
+}
+
+type puller struct {
+	config   PullerConfig
+	template template.Template
+}
+
+// NewPuller creates a new Puller instance with the provided configuration.
+func NewPuller(cfg PullerConfig) (*puller, error) {
+	if cfg.BaseURL == "" {
```
SimoneDutto: here i've set the default to launchpad URL but if you prefer i can just check it's not empty and export `launchPadURL`

internal/jujubinarypuller/puller.go
```diff
@@ -0,0 +1,140 @@
...
+
+// PullerConfig holds the configuration for the Juju binary puller.
+type PullerConfig struct {
+	BaseURL string // Base URL for the Juju binary downloads. Example: "https://launchpad.net/juju"
+	Dir     string // Directory to store the downloaded binaries. Defaults to the system's temp directory.
+}
+
+type JujuBinarySpec struct {
+	VersionWithMinor string // Version with minor version number, e.g., "3.6"
+	VersionWithFix   string // Version with fix version number, e.g., "3.6.2"
```
kian99: [semver.org](https://semver.org/) calls the final number the "patch" version, I'd opt for that.
```suggestion
	VersionWithPatch   string // Version with patch version number, e.g., "3.6.2"
```

internal/jujubinarypuller/puller.go
```diff
@@ -0,0 +1,140 @@
...
+	template template.Template
+}
+
+// NewPuller creates a new Puller instance with the provided configuration.
+func NewPuller(cfg PullerConfig) (*puller, error) {
+	if cfg.BaseURL == "" {
+		// Default to the launchpad URL if no base URL is provided.
+		cfg.BaseURL = launchPadURL
+	}
+	template, err := template.New("URL").Parse(launchPadTemplate)
```
kian99: The var `template` is aliasing the package name. Seems like it's not conflicting here but would be better to avoid this.

internal/jujubinarypuller/puller.go
```diff
@@ -0,0 +1,140 @@
...
+				return err
+			}
+			return nil
+		},
+		IsFatalError: func(err error) bool {
+			return !jujuerrrors.Is(err, retryRequestError)
+		},
+		BackoffFunc: retry.DoubleDelay,
+		Attempts:    5,
+		Delay:       time.Second,
```
kian99: Can we set a cap? When increasing attempts, we should avoid a very high delay.

internal/jujubinarypuller/puller.go
```diff
@@ -0,0 +1,140 @@
...
+	Arch             string // Architecture, e.g., "amd64"
+}
+
+type puller struct {
+	config   PullerConfig
+	template template.Template
+}
+
+// NewPuller creates a new Puller instance with the provided configuration.
+func NewPuller(cfg PullerConfig) (*puller, error) {
```
kian99: The name is a bit weird. Some other options might be `downloader` or `jujuCLIFetcher` or `binaryFetcher`

internal/jujubinarypuller/puller_test.go
```diff
@@ -0,0 +1,116 @@
...
+
+	file, err := os.Open(binary.FullPath)
+	c.Assert(err, qt.IsNil)
+	content, err := io.ReadAll(file)
+	c.Assert(err, qt.IsNil)
+	c.Assert(string(content), qt.Equals, "test content")
+}
+
+func TestPuller(t *testing.T) {
+	qtsuite.Run(qt.New(t), &pullerSuite{})
```
kian99: This is something we haven't discussed as a team, but opt not to use a suite where possible.
When I discussed this with Francesco he said that suites are useful if you want to run a test with a number of different implementations swapped in. If there is some setup for each test, it's idiomatic to just call the setup function at the start of the test. 

Here we don't have a setup function either. So these can all just be regular testing.T tests and then do `c := qt.New(t)` at the start.

internal/jujubinarypuller/puller.go
kian99: Separately, if the base URL is not launchpad, it's unlikely the launchpad template will still work but there is no association between the base URL and the template.

kian99 (commented):
Almost there for me, just some comments.

internal/jujubinarypuller/puller.go
SimoneDutto: 1sec
2sec
4sec
8sec
16sec

with 5 attempts.
For a total of less than 30 secs. 

I think it's even too short now, maybe with 10 attempts is better

internal/jujubinarypuller/puller_test.go
SimoneDutto: fair point, i though i needed the suite and in the end i didn't

internal/jujubinarypuller/puller.go
SimoneDutto: renamed

internal/jujubinarypuller/puller.go
SimoneDutto: i thought about creating a `type` to choose between launchpad and others, but i think it's a little premature.

SimoneDutto: For now it is defaulting to launchpad template and this accommodates both production code and tests.
Once we want to support different templates we need to add some additional code regardeless

internal/jujuclifetcher/fetcher.go
```diff
@@ -0,0 +1,141 @@
...
+	"github.com/juju/clock"
+	jujuerrrors "github.com/juju/errors"
+	"github.com/juju/retry"
+)
+
+// launchPadURL is the base URL for the Juju binary downloads.
+const launchPadURL = "https://launchpad.net/juju"
+
+// launchPadTemplate is the template for constructing the download URL for Juju binaries.
+const launchPadTemplate = "{{.BaseURL}}/{{.VersionWithMinor}}/{{.VersionWithPatch}}/+download/juju-agents-{{.VersionWithPatch}}-{{.Os}}-{{.Arch}}.tar.xz"
```
alesstimec: emm.. no, not the agents binary.. you want
```
"{{.BaseURL}}/{{.VersionWithMinor}}/{{.VersionWithPatch}}/+download/juju-{{.VersionWithPatch}}-{{.Os}}-{{.Arch}}.tar.xz
```

internal/jujuclifetcher/fetcher.go
```diff
@@ -0,0 +1,141 @@
...
+
+var retryRequestError = jujuerrrors.New("retry request error")
+
+// jujuCLIFetcherConfig holds the configuration for the Juju binary fetcher.
+type jujuCLIFetcherConfig struct {
+	BaseURL string // Base URL for the Juju binary downloads. Example: "https://launchpad.net/juju"
+	Dir     string // Directory to store the downloaded binaries. Defaults to the system's temp directory.
+}
+
+type JujuBinarySpec struct {
```
alesstimec, line 36: godoc, please

internal/jujuclifetcher/fetcher.go
```diff
@@ -0,0 +1,141 @@
...
+// Binary represents a downloaded Juju binary.
+// It contains the full path to the binary file.
+// It also provides a method to clean up the binary file.
+type Binary struct {
+	FullPath string
+}
+
+// Clean removes the binary file from the filesystem.
+// It returns an error if the file cannot be removed.
+func (b Binary) Clean() error {
```
alesstimec:
```suggestion
func (b Binary) Remove() error {
```

internal/jujuclifetcher/fetcher.go
```diff
@@ -0,0 +1,141 @@
...
+		"Os":               spec.Os,
+		"Arch":             spec.Arch,
+	})
+	if err != nil {
+		return nil, err
+	}
+	var file *os.File
+	err = retry.Call(retry.CallArgs{
+		Func: func() error {
+			file, err = os.CreateTemp(p.config.Dir, "juju-agents-*")
```
alesstimec: not agents :)

internal/jujuclifetcher/fetcher.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+type JujuBinarySpec struct {
+	VersionWithMinor string // Version with minor version number, e.g., "3.6"
+	VersionWithPatch string // Version with patch version number, e.g., "3.6.2"
+	Os               string // Operating system, e.g., "linux"
+	Arch             string // Architecture, e.g., "amd64"
+}
+
+type jujuCLIFetcher struct {
```
alesstimec, line 43: what if we make this a binary repository kind of thing.. a struct responsible for downloading binaries on demand, tracking what has been downloaded, acting as kind of a LRU cache for binaries. removing unneeded binaries,etc.

internal/jujuclifetcher/fetcher.go
SimoneDutto: ah, the jira card is wrong then

internal/jujuclifetcher/fetcher.go
SimoneDutto: while i agree it would be nice to have a cache of some sort there are some points to keep in mind:
- the cache must be configurable for number of entries, or max size
- the cache needs to be aware of when it's safe to delete a binary from the cache (for example, there could be some race conditions where a binary is deleted before it's consumed by the job tracker)

With all of that in mind i would address caching in a follow-up if that's ok for you

alesstimec: fine for a follow-up.. but i think we should go for a fixed number of entries (e.g. 2, latest LTS and current stable) in the cache and just do LRU eviction

internal/jujuclifetcher/fetcher.go
```diff
@@ -0,0 +1,142 @@
...
+		BackoffFunc: retry.DoubleDelay,
+		Attempts:    10,
+		Delay:       time.Second,
+		Clock:       clock.WallClock,
+		Stop:        ctx.Done(),
+	})
+	if err != nil {
+		if file != nil {
+			// Clean up the file if it was created.
+			// We ignore the error because there is not much we can do about it.
```
alesstimec, line 134: log the error?

internal/jujuclifetcher/fetcher.go
kian99, line 1: With Ales's suggestion to make this a binary repository, can we rename it in this PR to `jujuCLIStore` or something similar. Wdyt?

internal/jujuclifetcher/fetcher.go
```diff
@@ -0,0 +1,142 @@
...
+			// Retry on server errors or rate limiting.
+			if (resp.StatusCode >= 500 && resp.StatusCode < 600) || resp.StatusCode == http.StatusTooManyRequests {
+				return retryRequestError
+			}
+			if resp.StatusCode != http.StatusOK {
+				return jujuerrrors.Errorf("request failed with status %d", resp.StatusCode)
+			}
+			file, err = os.CreateTemp(p.config.Dir, "juju-*")
+			if err != nil {
+				return err
```
alesstimec, line 113: can we log these errors.. it might be interesting to see if we're failing to create a tempdir/file because we're out of storage space.

alesstimec (approved):

internal/jujuclifetcher/fetcher.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+	return &jujuCLIFetcher{
+		config:   cfg,
+		template: *tmpl,
+	}, nil
+}
+
+// Binary represents a downloaded Juju binary.
+// It contains the full path to the binary file.
+// It also provides a method to clean up the binary file.
```
kian99, line 67: Some of these comments should change either in this PR or a follow-up. Considering we want to cache the files, we discussed that the `Binary` object should have something like a `Done()` method to indicate that a caller has finished using it and the store keeps track of when it is safe to delete a file if it needs to make space for a new one.

kian99 (approved):

internal/jujuclifetcher/fetcher.go
SimoneDutto: yeah makes sense

internal/jujuclifetcher/fetcher.go
SimoneDutto: i believe whoever is calling `Fetch` will log the error

internal/jujuclifetcher/fetcher.go
SimoneDutto: I'll do the rename in the cache PR, just to avoid doing it two times in case another name is better suited

PR #1656, opened 2025-07-22
feat: add bootstrap command to jaas plugin

kian99:
## Description
This PR builds on #1654 and implements a new CLI method on the JAAS CLI to bootstrap a controller by making a facade call.

This PR made some changes to the `bootstrap-status` command, specifically:
- Switched to use `gocheck` instead of `quicktest`. Mixing different testing suites in the same package causes issues when running an individual tests where it also runs all tests from the other suite.
- Extracted the polling logs logic into a separate function that can be called from the bootstrap method.
- Refactored how we create a mock client to be in-line with Juju.

The `bootstrap` command will send a request to JIMM to start the bootstrap job and then print the response (in yaml or json format). In order to then watch the logs it is necessary to no longer use `c.out` as that will try to format the logs in either yaml or json. It would be cleaner to add a message to stderr like "use bootstrap-status <job-id>" to see logs but to keep the bootstrap process as similar to `juju bootstrap` as possible, I've made this compromise.

Fixes [JUJU-8152](https://warthogs.atlassian.net/browse/JUJU-8152)

## Engineering checklist
- [ ] Documentation updated
- [x] Covered by unit tests

[JUJU-8152]: https://warthogs.atlassian.net/browse/JUJU-8152?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


alesstimec (approved):

SimoneDutto (approved):
LGTM

PR #1657, opened 2025-07-22
feat: run commands via os.exec

ale8k:
The function to run commands now uses os.Exec and reads in line exactly like previously. It additionally uses a shell parser to protect us from bad escapes and an environment override to protect us from bad environments.

## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -115,7 +115,7 @@ require (
 	github.com/cespare/xxhash/v2 v2.3.0 // indirect
 	github.com/cjlapao/common-go v0.0.39 // indirect
 	github.com/coreos/go-systemd/v22 v22.5.0 // indirect
-	github.com/creack/pty v1.1.18 // indirect
+	github.com/creack/pty v1.1.24 // indirect
 	github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc // indirect
 	github.com/decred/dcrd/dcrec/secp256k1/v4 v4.2.0 // indirect
 	github.com/distribution/reference v0.5.0 // indirect
@@ -339,15 +339,15 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.4.1 // indirect
 )
 
+require mvdan.cc/sh/v3 v3.12.0
+
 require (
 	cloud.google.com/go/auth v0.16.2 // indirect
 	cloud.google.com/go/auth/oauth2adapt v0.2.8 // indirect
 	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/compute/armcompute/v6 v6.4.0 // indirect
 	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/msi/armmsi v1.2.0 // indirect
 	github.com/anmitsu/go-shlex v0.0.0-20200514113438-38f4b401e2be // indirect
-	github.com/chzyer/readline v1.5.1 // indirect
 	github.com/juju/packaging/v4 v4.0.0 // indirect
-	github.com/juju/terms-client/v2 v2.0.0 // indirect
 	github.com/klauspost/compress v1.18.0 // indirect
 	github.com/moby/patternmatcher v0.6.0 // indirect
 	github.com/moby/sys/sequential v0.6.0 // indirect
--- a/internal/jujucommands/jujucommands.go
+++ b/internal/jujucommands/jujucommands.go
@@ -19,22 +19,21 @@ type outputLine struct {
 	Err  error
 }
 
+var (
+	cmdPrefix = "juju"
+)
+
 // runJujuCmd runs a juju command with the given command string and JUJU_DATA directory.
 // It returns a channel that will receive output lines from the command's stdout and stderr.
 // The command is run in a separate goroutine, and the context can be used to cancel the command.
 func runJujuCmd(ctx context.Context, cmdStr, jujuDataDir string) (<-chan outputLine, error) {
-	env := func(key string) string {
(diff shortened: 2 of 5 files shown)


internal/jujucommands/jujucommands.go
```diff
@@ -5,127 +5,86 @@
...
+	}
+
+	go readLines(linereader.New(stdOut))
+	go readLines(linereader.New(stdErr))
 
+	go func() {
+		// We wait for readers to finished in the case the command exits but
+		// the readers are still processing output. Small chance this could happen,
+		// but this protects us.
+		wg.Wait()
```
kian99, line 69: ~~Shouldn't this go outside the go routine?~~ Nvm, I see what we're trying to do. But either the context.Done() or stdout/stderr on the command close and ensure `readLines` returns. Would like to see a test validating that.

internal/jujucommands/jujucommands.go
```diff
@@ -5,127 +5,86 @@
...
-	// the writer to close, and the reader to eventually see EOF and exit.
-	//
-	// This is safe because we only send once, after the command finishes.
-	// Alternatively, we could manually close cmdWriter right after the command.
-	cmdFinishedCh := make(chan int, 1)
+	if err := cmd.Start(); err != nil {
+		return nil, fmt.Errorf("failed to start command: %w", err)
+	}
+
+	outputCh := make(chan outputLine, 10) // buffered to avoid blocking
```
kian99, line 46: If the consumer gets very slow the command will block once these 10 entries get filled right? Just to make sure I understand.

internal/jujucommands/jujucommands.go
```diff
@@ -5,127 +5,86 @@
...
+			return jujuDataDir
+		}
+		return ""
+	}
+	args, err := shell.Fields(cmdStr, env)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse command: %w", err)
+	}
+
+	cmd := exec.Command("juju", args...)
```
kian99: The env vars are being converted to args here? I would've expected to see `cmd.Env()` being used.

internal/jujucommands/jujucommands.go
```diff
@@ -5,127 +5,86 @@
...
+			select {
+			case outputCh <- outputLine{Line: line}:
+			case <-ctx.Done():
+				return
+			}
 		}
+	}
+
+	go readLines(linereader.New(stdOut))
+	go readLines(linereader.New(stdErr))
```
SimoneDutto: Do we need all of this machinery?
At the end of the day we have the cmd.Output and we can read line by line and publish it to the db right? Why are we pushing into a channel and doing all this complex routines?

I think it's too complex for the use case that we are serving, which is "pushing few KB to some rows to the db".

SimoneDutto: Can we do something like?
writer, reader := io.Pipe()

pass the writer to the cmd.Run and Read from a go routine and push it to the db?
I think you can even `Tee` stderr and stdout

SimoneDutto (commented):
i have a question

internal/jujucommands/jujucommands.go
ale8k: This is just me being paranoid that a read could be slow.

internal/jujucommands/jujucommands.go
ale8k: We can use combined yes, I actually made a mistake here where I was going to split the outputLine and on the stderr populate an error line, but I'd forgotten. It also doesn't make too much sense what I was going to do because again it is under the presumption juju only ever writes one line when it errors. I'll move it over to combined output and a line reader and should we want to tweak this in the future we can.

--- Scratch that, I used the pipes so we can read adhoc. 

If we use output or whatever else, we'll have to wait and we don't want to wait - we want to stream.

As for whether to convert error lines into errors, as you can see I'm not and we're just treating stderr as a line. I'm not sure if we should convert them to errors?

ale8k: As discussed, leaving as is.

SimoneDutto (approved):

internal/jujucommands/jujucommands_test.go
```diff
@@ -68,6 +69,49 @@ For the full list of supported commands run:
...
+		lines = append(lines, out.Line)
+		if out.Err != nil {
+			errs = append(errs, out.Err)
+		}
+	}
+
+	// 0 holds the error line
+	// 1 is the exit status (but we're just filtering them into two slices to see
+	// that the output from errors is how we expect, i.e., the line before is the actual
+	// error message).
```
kian99: This comment needs some changing. Use a full sentence to say "Index 0 holds the error line", etc. And is the bit about index 1 accurate because we're not inspecting anything with index 1 so that's confusing.

go.mod
```diff
@@ -340,14 +340,14 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.4.1 // indirect
 )
 
+require mvdan.cc/sh/v3 v3.12.0
```
kian99, line 342: I think you need to run `go mod tidy`

kian99 (approved):
Tests are failing because the `juju` binary isn't available. Either we'll need to `snap install` it in our test workflow or replace it with a different command for tests.

internal/jujucommands/jujucommands_test.go
ale8k: It's because I'm filtering on lines 81-85... I'll try reword it

PR #1658, opened 2025-07-22
chore: quality of life fixes

kian99:
## Description

This PR contains several small and unrelated quality of life changes/fixes that I found while working through internal model migration testing with the web team.

The first two are related to Vault. When the vault container restarts, all secrets it contained are lost. This can result in confusing scenarios because the Postgres container keeps data until the container is removed. While testing this can lead to tricky issues where Postgres is aware of a controller, but the controller's admin user/password are missing from Vault. Unfortunately there is no good fix for this as we run Vault in "dev" mode for testing and this always stores data in-memory only. To offer an alternative the two changes below make it easier to set the `INSECURE_SECRET_STORAGE` variable to use Postgres for secrets.
- Parse INSECURE_SECRET_STORAGE as a boolean rather than accepting the value to simply be set.
- Explain in the docker compose how to set the INSECURE_SECRET_STORAGE env var.

The next change was spotted due to some missing validation. In the db method `GetController`, if the controller name was empty, `GetController` would return the first controller found. This can lead to unexpected scenarios and has been changed to make the controller name/UUID required.
- Change behavior of the `GetController` db method to require a controller name/uuid.

Finally, this change was simple but important. The `migrate-internal` command was not registered in the jaas plugin's `main.go` file and therefore wasn't available.
- Ensure the migrate-internal command is present in the jaas plugin.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

### DIFF
--- a/local/jimm/qa-lxd-multipass.sh
+++ b/local/jimm/qa-lxd-multipass.sh
@@ -39,7 +39,7 @@ multipass exec $VM_NAME -- bash <<- 'EOF'
 EOF
 
 echo "Setting up JIMM"
-multipass exec --working-directory /home/ubuntu/jimm $VM_NAME -- bash <<- 'EOF'
+multipass exec --working-directory /home/ubuntu/jimm "$VM_NAME" -- bash <<- EOF
     make certs
 
     # Re-copy and update certs (workaround to keep the same generated certs but simply update the VM's certs only)
@@ -49,6 +49,7 @@ multipass exec --working-directory /home/ubuntu/jimm $VM_NAME -- bash <<- 'EOF'
     make version/commit.txt
     make version/version.txt
 
+    export INSECURE_SECRET_STORAGE=${INSECURE_SECRET_STORAGE}
     # TODO(ale8k): Have docker cache images somewhere that can be shared, the compose takes forever otherwise.
     docker compose --profile dev up --wait -d
 EOF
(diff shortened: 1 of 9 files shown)


SimoneDutto (approved):

local/jimm/qa-lxd-multipass.sh
```diff
@@ -39,7 +39,7 @@ multipass exec $VM_NAME -- bash <<- 'EOF'
 EOF
 
 echo "Setting up JIMM"
-multipass exec --working-directory /home/ubuntu/jimm $VM_NAME -- bash <<- 'EOF'
+multipass exec --working-directory /home/ubuntu/jimm "$VM_NAME" -- bash <<- EOF
```
kian99, line 42: This is worth understanding,
>When you use <<- 'EOF' (with single quotes) for a heredoc in bash, shell expansion—including variable substitution—does NOT occur inside the heredoc. Everything is treated as literal text.
>
>If you want variable expansion (e.g., $INSECURE_SECRET_STORAGE) to happen, you should use an unquoted heredoc delimiter: <<- EOF (no quotes). This way, variables from your local shell are expanded before being sent to the remote command.

Courtesy of GPT while I was trying to figure out how to pass env vars through to this heredoc. Tested it with a smaller script and verified it works as expected.

ale8k: Didn't know this, nice.

ale8k (approved):

PR #1659, opened 2025-07-23
feat(bootstrap): implement bootstrap stop facade

SimoneDutto:
## Description
Implement stop bootstrap facade

### DIFF
--- a/internal/jimm/bootstrap/bootstrap.go
+++ b/internal/jimm/bootstrap/bootstrap.go
@@ -18,6 +18,9 @@ import (
 type JobTracker interface {
 	// GetJob retrieves a job entry by its ID.
 	GetJob(ctx context.Context, jobId uuid.UUID) (dbmodel.JobTrackerEntry, error)
+
+	// StopJob stops a job by its ID.
+	StopJob(ctx context.Context, jobId uuid.UUID) error
 }
 
 type Store interface {
@@ -81,4 +84,24 @@ func (b *bootstrapManager) StartBootstrap(ctx context.Context, user *openfga.Use
 	}
 
 	return "", errors.E(op, "not implemented")
+}
+
+// StopBootstrap stops a bootstrap job by its ID.
+func (b *bootstrapManager) StopBootstrap(ctx context.Context, user *openfga.User, jobId uuid.UUID) error {
+	const op = errors.Op("jimm.StopBootstrap")
+
+	if user == nil {
+		return errors.E(op, "user cannot be nil")
+	}
+
+	if jobId == uuid.Nil {
+		return errors.E(op, "job ID cannot be nil")
+	}
+
+	err := b.jobtracker.StopJob(ctx, jobId)
+	if err != nil {
+		return errors.E(op, "failed to stop job", err)
+	}
+
+	return nil
 }
--- a/internal/jobtracker/jobtracker.go
+++ b/internal/jobtracker/jobtracker.go
@@ -18,6 +18,8 @@ import (
 	"github.com/canonical/jimm/v3/internal/errors"
 )
 
+const minimumRefreshIntervalseconds = 5
+
 // JobIdContextKey is a context key for storing job IDs.
 type JobIdContextKey struct{}
 
@@ -27,6 +29,7 @@ type JobIdContextKey struct{}
 type Store interface {
 	AddJob(ctx context.Context, jobType string) (uuid.UUID, error)
 	GetJob(ctx context.Context, job *dbmodel.JobTrackerEntry) (err error)
+	StopJob(ctx context.Context, jobId uuid.UUID) (err error)
 	SetJobRunning(ctx context.Context, jobId uuid.UUID) error
 	SetJobSuccessful(ctx context.Context, jobId uuid.UUID) error
 	SetJobFailed(ctx context.Context, jobId uuid.UUID, jobErr error) error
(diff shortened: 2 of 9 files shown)


internal/jobtracker/jobtracker.go
```diff
@@ -27,32 +27,34 @@ type JobIdContextKey struct{}
...
+	StopJob(ctx context.Context, jobId uuid.UUID) (err error)
 	SetJobRunning(ctx context.Context, jobId uuid.UUID) error
 	SetJobSuccessful(ctx context.Context, jobId uuid.UUID) error
 	SetJobFailed(ctx context.Context, jobId uuid.UUID, jobErr error) error
 	GetJobStopSignal(ctx context.Context, jobId uuid.UUID) (stopSignal bool, err error)
 }
 
 // Tracker manages job tracking operations using a provided JobTrackerStore.
-// It periodically performs tasks based on the specified stopInterval duration.
+// It periodically performs tasks based on the specified refreshInterval duration.
```
SimoneDutto, line 40: rename to refreshInterval to make it clear.

stopInterval seems more like maxDuration

internal/jobtracker/jobtracker.go
```diff
@@ -27,32 +27,34 @@ type JobIdContextKey struct{}
...
+// refreshInterval is the interval between successive checks for job status.
+// It returns an error if the store is nil or if refreshInterval is not greater than zero.
+func NewJobTracker(store Store, refreshInterval time.Duration) (*Tracker, error) {
 	if store == nil {
 		return nil, goerr.New("store cannot be nil")
 	}
-	if stopInterval <= 4 {
-		return nil, goerr.New("stopInterval must be greater than zero")
+	if refreshInterval <= 4 {
+		return nil, goerr.New("refreshInterval must be greater than zero")
```
kian99:
```suggestion
	if refreshInterval <= 4 {
		return nil, goerr.New("refreshInterval minimum value is %d")
```
And make the value a constant and make it `< 5`.

internal/jobtracker/jobtracker.go
```diff
@@ -27,32 +27,34 @@ type JobIdContextKey struct{}
...
+// NewJobTracker creates and returns a new Tracker instance using the provided JobTrackerStore and refreshInterval.
+// refreshInterval is the interval between successive checks for job status.
+// It returns an error if the store is nil or if refreshInterval is not greater than zero.
+func NewJobTracker(store Store, refreshInterval time.Duration) (*Tracker, error) {
 	if store == nil {
 		return nil, goerr.New("store cannot be nil")
 	}
-	if stopInterval <= 4 {
-		return nil, goerr.New("stopInterval must be greater than zero")
+	if refreshInterval <= 4 {
```
kian99: We can compare a duration with an int? This is checking that the value is at least 5 second but I see in the test you've set it to 10 milliseconds which shouldn't be possible.

internal/jobtracker/jobtracker_test.go
```diff
@@ -30,7 +30,7 @@ func (s *jobTrackerSuite) Init(c *qt.C) {
 	c.Assert(err, qt.IsNil)
 
 	s.db = db
-	tracker, err := jobtracker.NewJobTracker(db, time.Second*5)
+	tracker, err := jobtracker.NewJobTracker(db, 10*time.Millisecond)
```
kian99: This shouldn't be possible. The minimum should be 5 seconds.
I'm not sure how you can speed up the test without using a mock clock. This might be a nice scenario for the new synctest feature when it's no longer experimental in Go 1.25.
Until then maybe we can add a method to the job tracker in `export_test.go` that let's you set a new value for the refreshInternal.

SimoneDutto: what's the issue with refreshing every 10 ms? Other than doing a db query is not doing much.
Do you think it something we should guard against? At the end of the day if somebody wants to refresh faster i don't see an issue

internal/jobtracker/jobtracker.go
SimoneDutto: ah yeah, 5 * time.Second is not 5

internal/jobtracker/jobtracker_test.go
kian99: Yeah we didn't want it polling so quickly, that's why the limit was added, that's all.

SimoneDutto: I though the `<= 5` was just a safe guard against a super low value.

```
const (
	Nanosecond  Duration = 1
	Microsecond          = 1000 * Nanosecond
	Millisecond          = 1000 * Microsecond
	Second               = 1000 * Millisecond
	Minute               = 60 * Second
	Hour                 = 60 * Minute
)
```

internal/jobtracker/jobtracker.go
SimoneDutto: done

internal/jobtracker/jobtracker_test.go
SimoneDutto: done

internal/jobtracker/jobtracker.go
```diff
@@ -18,6 +18,8 @@ import (
 	"github.com/canonical/jimm/v3/internal/errors"
 )
 
+const minimumRefreshIntervalseconds = 5
```
kian99, line 21: You can opt to make this `5*time.Second` and then below, since timeDuration implements the `Stringer` interface just use it in the error message with `%s`.

kian99 (approved):

internal/jimm/bootstrap/bootstrap.go
```diff
@@ -82,3 +85,23 @@ func (b *bootstrapManager) StartBootstrap(ctx context.Context, user *openfga.Use
 
 	return "", errors.E(op, "not implemented")
 }
+
+// StopBootstrap stops a bootstrap job by its ID.
+func (b *bootstrapManager) StopBootstrap(ctx context.Context, user *openfga.User, jobId uuid.UUID) error {
+	const op = errors.Op("jimm.StopBootstrap")
+
+	if user == nil {
```
ale8k, line 93: Why do we need the user here? It doesn't appear we're using the user?

ale8k (approved):
Nice! Thanks for thisssssss.

PR #1661, opened 2025-07-24
feat(binary-store): transform the CLI fetcher into a binary store with caching

SimoneDutto:
## Description
Convert jujuCLIFetcher to jujuCLIStore with caching.

The eviction policy is random on `done` elements.
@alesstimec suggested something smarter, like keeping the stable version. But some problems arise:
- we could set the `LATEST_STABLE` as an env var but it would go stale very quickly.
- in Launchpad there is the concept of "Series", but i could find a way to fetch `latest` from a series. Maybe there is using the launchpad sdk, i didn't check.
- I could get the latest from the last GitHub tag but it seems more of an hack. What if a version is tagged in GitHub but not yet in launchpad?
- we will have two latest stable, one for 3.6 and one for 4.0. 
- maybe once we move to GitHub Release it would be easier.

With all that in mind I went for the simpler approach of random eviction. But let me know what you think.


SimoneDutto:
> ## Description
> Convert jujuCLIFetcher to jujuCLIStore with caching.
> 
> The eviction policy is random on `done` elements. @alesstimec suggested something smarter, like keeping the stable version. But some problems arise:
> 
> * we could set the `LATEST_STABLE` as an env var but it would go stale very quickly.
> * in Launchpad there is the concept of "Series", but i could find a way to fetch `latest` from a series. Maybe there is using the launchpad sdk, i didn't check.
> * I could get the latest from the last GitHub tag but it seems more of an hack. What if a version is tagged in GitHub but not yet in launchpad?
> * we will have two latest stable, one for 3.6 and one for 4.0.
> * maybe once we move to GitHub Release it would be easier.
> 
> With all that in mind I went for the simpler approach of random eviction. But let me know what you think.

It is possible to download an xml from here:
https://launchpad.net/juju/4.0/+rdf and the first entry should be the latest

internal/jujuclistore/store.go
```diff
@@ -0,0 +1,210 @@
...
+	Os               string // Operating system, e.g., "linux"
+	Arch             string // Architecture, e.g., "amd64"
+}
+
+type jujuCLIStore struct {
+	config   JujuCLIStoreConfig
+	template template.Template
+
+	entries map[string]*Binary // Map to keep track of downloaded entries
+	lock    sync.Mutex         // Protects the entries map
```
kian99: nit but I've normally seen the lock go above other fields and commented with something like "protects the fields below". But not something that needs to be changed, just sharing.

internal/jujuclistore/store.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+	tmpl, err := template.New("URL").Parse(launchPadTemplate)
+	if err != nil {
+		return nil, err
+	}
+	if cfg.MaxEntries <= 0 {
+		// Default to 2 entries if MaxEntries is not set or is less than or
+		// equal to zero.
+		cfg.MaxEntries = 2
+	}
```
kian99, line 81: Should we set a max as well? Maybe 10?

internal/jujuclistore/store.go
```diff
@@ -0,0 +1,210 @@
...
+	binary = &Binary{
+		FullPath: file.Name(),
+	}
+	p.entries[url] = binary
+	return binary, nil
+}
+
+// freeSpace checks if the entries map has reached the maximum number of entries.
+// If it has, it deletes a random entry from the map.
+// The randomness is guaranteed by the fact iterating over a map in Go does not guarantee order.
```
kian99: That's not a very good assumption imo. The fact that iterating over a map doesn't guarantee an order just means that the implementation of a map in Go today might change and if you rely on a specific ordering, you'd be out of luck. But it *could* be that, for a given Go version, maps are iterated over in the same order each time.
Basically, if you want random, you need to use something offers randomness explicitly. I'd suggest making a list of the binaries that are "done" and then picking a random number in the range of 0-len(list).

internal/jujuclistore/store.go
```diff
@@ -0,0 +1,210 @@
...
+				zapctx.Error(ctx, "failed to remove binary", zap.Error(err), zap.String("path", binary.FullPath))
+			} else {
+				delete(p.entries, key)
+				deletion = true
+				break
+			}
+		}
+	}
+	if !deletion {
+		// This shouldn't happen because we bootstrap a controller at the time, so there should always be at least one
```
kian99:
```suggestion
		// This shouldn't currently happen because we bootstrap one controller at a time, so there should always be at least one
```

internal/jujuclistore/store.go
```diff
@@ -0,0 +1,210 @@
...
+		FullPath: file.Name(),
+	}
+	p.entries[url] = binary
+	return binary, nil
+}
+
+// freeSpace checks if the entries map has reached the maximum number of entries.
+// If it has, it deletes a random entry from the map.
+// The randomness is guaranteed by the fact iterating over a map in Go does not guarantee order.
+func (p *jujuCLIStore) freeSpace(ctx context.Context) error {
```
kian99: `freeSpace` is a misleading name, I would expect that to be checking for space on the disk. Maybe something like `freeEntry` or `checkEntries` or `checkCache` and add a godoc to express the fact that this must be called with the entries map locked.

internal/jujuclistore/store.go
```diff
@@ -0,0 +1,210 @@
...
+// is no longer used.
+type Binary struct {
+	FullPath string
+
+	done atomic.Bool
+}
+
+func (b *Binary) Done() {
+	b.done.Store(true)
+}
```
kian99: This needs some more thought because it implies a binary can only be used once before its marked as done. Then if its re-used, it can be evicted from the cache while a caller has the object.
Also missing godoc

kian99 (commented):
Some changes needed

internal/jujuclistore/store.go
SimoneDutto: truee, i've switched to reference counting and add a test.

internal/jujuclistore/store.go
SimoneDutto: This is a design choice by golang to not guarantee the order and it is stated explicitly:
- https://go.dev/blog/maps#iteration-order
- https://go.dev/doc/go1#iteration

https://stackoverflow.com/questions/55925822/why-are-iterations-over-maps-random/55925880#55925880
i was reading this stack overflow and it seems it is a choice they will not change in the future.

And even if they will it would be a problem because the eviction policy will become then FIFO, not causing any bugs.

kian99: I think this is an important one to clarify. This code is making the same mistake as highlighted in the issue,
> Programmers had begun to rely on the stable iteration order of early versions of Go, which varied between implementations, leading to portability bugs.

They've made it random to avoid users from relying on consistent ordering and now instead, we are relying on the iteration order to be random. We are relying on undefined behavior which we should never do.

The way the code is now, we cannot correctly say whether it is FIFO or random. So we should change it and not rely on undefined behaviour.

SimoneDutto: I don't agree it is the same error.

There is a difference between "relying on A order" (pre 1.0) or "relying on random order". This code is relying on the fact the from go 1.0 there is no order enforced by the Golang runtime.
It's practically impossible they will change this behavior to enforce an order in a new release, for many reasons listed in the stackoverflow post.

However it's easy enough to manually randomize the entries, so i'll do it.

internal/jujuclistore/store.go
```diff
@@ -0,0 +1,223 @@
...
+	BaseURL string // Base URL for the Juju binary downloads. Example: "https://launchpad.net/juju"
+	Dir     string // Directory to store the downloaded binaries. Defaults to the system's temp directory.
+
+	MaxEntries int // Maximum number of entries to keep in the directory. Defaults to 2.
+}
+
+// JujuBinarySpec defines the specifications for a Juju binary to be downloaded.
+type JujuBinarySpec struct {
+	VersionWithMinor string // Version with minor version number, e.g., "3.6"
+	VersionWithPatch string // Version with patch version number, e.g., "3.6.2"
```
kian99: Rather create the version with patch from the version with minor. So either change this field to just a juju.Version where you can grab the major/minor/patch as you need or make them individual fields.

internal/jujuclistore/store.go
```diff
@@ -0,0 +1,223 @@
...
+		"Os":               spec.Os,
+		"Arch":             spec.Arch,
+	})
+	if err != nil {
+		return nil, err
+	}
+	url := buf.String()
+	p.lock.Lock()
+	defer p.lock.Unlock()
+	binary, ok := p.entries[url]
```
kian99, line 142: There's one issue here that is worth leaving a comment in the code for now. If this were to run in a multi-threaded environment, only 1 binary can be downloaded at a time, since this lock is held for the duration of the download. It can be locked for a long period of time.

kian99 (approved):

internal/jujuclistore/store.go
```diff
@@ -0,0 +1,223 @@
...
+// launchPadURL is the base URL for the Juju binary downloads.
+const launchPadURL = "https://launchpad.net/juju"
+
+// launchPadTemplate is the template for constructing the download URL for Juju binaries.
+const launchPadTemplate = "{{.BaseURL}}/{{.VersionWithMinor}}/{{.VersionWithPatch}}/+download/juju-{{.VersionWithPatch}}-{{.Os}}-{{.Arch}}.tar.xz"
+
+var retryRequestError = jujuerrrors.New("retry request error")
+
+// JujuCLIStoreConfig holds the configuration for the Juju binary fetcher.
+type JujuCLIStoreConfig struct {
```
ale8k: Can just call this config, package is called `jujuclistore`

internal/jujuclistore/store.go
```diff
@@ -0,0 +1,223 @@
...
+	"github.com/juju/retry"
+)
+
+// launchPadURL is the base URL for the Juju binary downloads.
+const launchPadURL = "https://launchpad.net/juju"
+
+// launchPadTemplate is the template for constructing the download URL for Juju binaries.
+const launchPadTemplate = "{{.BaseURL}}/{{.VersionWithMinor}}/{{.VersionWithPatch}}/+download/juju-{{.VersionWithPatch}}-{{.Os}}-{{.Arch}}.tar.xz"
+
+var retryRequestError = jujuerrrors.New("retry request error")
```
ale8k: Errrors? 😆  `s/jujuerrrors/jujuerrrrrrrrrrors`

internal/jujuclistore/store.go
```diff
@@ -0,0 +1,223 @@
...
+
+// JujuCLIStoreConfig holds the configuration for the Juju binary fetcher.
+type JujuCLIStoreConfig struct {
+	BaseURL string // Base URL for the Juju binary downloads. Example: "https://launchpad.net/juju"
```
ale8k: Docs should go above the field no?

internal/jujuclistore/store.go
```diff
@@ -0,0 +1,223 @@
...
+	Dir     string // Directory to store the downloaded binaries. Defaults to the system's temp directory.
```
ale8k: Do we want a default? It feels opinionated to have a default for a generic binary downloader?

internal/jujuclistore/store.go
```diff
@@ -0,0 +1,223 @@
...
+
+	MaxEntries int // Maximum number of entries to keep in the directory. Defaults to 2.
```
ale8k: Again I don't think this should be opinionated.

internal/jujuclistore/store.go
ale8k: Agreed, would prefer a juju.Version.

internal/jujuclistore/store.go
```diff
@@ -0,0 +1,223 @@
...
+// It retries on server errors or rate limiting.
+// If the download fails, it returns an error.
+// The file is created in the directory specified in the configuration.
+// The context can be used to cancel the operation.
+func (p *jujuCLIStore) downloadFile(ctx context.Context, downloadUrl string) (*os.File, error) {
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, downloadUrl, nil)
+	if err != nil {
+		return nil, err
+	}
+	resp, err := http.DefaultClient.Do(req)
```
ale8k: Using the default client means we wouldn't set a timeout right and be at the behest of the server? I think we should be creating a client and setting a sensible timeout?

ale8k (approved):
Using the default client means we wouldn't set a timeout right and be at the behest of the server? I think we should be creating a client and setting a sensible timeout?

internal/jujuclistore/store.go
SimoneDutto: errrrrrrrrrrrrrrr

internal/jujuclistore/store.go
SimoneDutto: I think it's ok since for now we support just launchpad, when we will support multiple use cases we can think about removing the default

internal/jujuclistore/store.go
SimoneDutto: i was relying on the context, but i've added a timeout to the client

internal/jujuclistore/store.go
SimoneDutto: done

ale8k:
This looks good to go to me now, thanks for adding the timeout.

PR #1664, opened 2025-07-29
feat: validate user mapping client side

kian99:
## Description
This PR adds client side validation to the user-mapping used as part of migrating a model to JIMM.

We validate that the user mapping contains an entry for all users that have access to the model or access to offers in the model. We also allow a user mapping to be skipped by providing a `null` value.

The tests for the migration command have been updated. When writing the tests I noticed that we don't include a `Close()` method in our JIMM API interface. This also meant we weren't cleanly closing the connection in some of our new commands for bootstrap so I've added the `Close()` method.

Fixes [JUJU-8264](https://warthogs.atlassian.net/browse/JUJU-8264)

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

[JUJU-8264]: https://warthogs.atlassian.net/browse/JUJU-8264?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cmd/jaas/cmd/migratemodel_test.go
```diff
@@ -34,8 +62,211 @@ bob: bob@canonical.com
...
+	s.store.EXPECT().AccountDetails("target-controller").Return(&jjclient.AccountDetails{
+		User: "test-user",
+	}, nil)
+	s.store.EXPECT().ControllerByName("target-controller").Return(&jjclient.ControllerDetails{
+		ControllerUUID: "target-controller-uuid",
+		APIEndpoints:   []string{"endpoint1"},
+		CACert:         "test-ca-cert",
+	}, nil)
+	// 	Got: {93608db4-f1cb-4da5-9926-8233981aef0a true target-controller-uuid target-controller [endpoint1] test-ca-cert test-user  [] migration-token} (controller.MigrationSpec)
+	// Want: {93608db4-f1cb-4da5-9926-8233981aef0a true target-controller-uuid  [endpoint1] test-ca-cert   [] migration-token} (controller.MigrationSpec)
```
SimoneDutto: leftover?

internal/jimm/juju/migrationtarget.go
```diff
@@ -365,6 +367,11 @@ func (j *JujuManager) Activate(ctx context.Context, modelTag names.ModelTag, mig
 	// if user mappings have been created.
 	err = j.Database.Transaction(func(db *db.Database) error {
 		for localUser, externalUser := range modelMigration.UserMapping {
+			if externalUser == "" {
+				// An empty external user indicates the user intentionally wants
+				// to skip mapping the local user to an external user.
+				continue
+			}
```
SimoneDutto, line 374: with this change when it calls the discharge endpoint and ask for the mapping, here https://github.com/SimoneDutto/jimm/blob/jimm-model-deletion/internal/db/usermapping.go#L33-L63 it would return a `not-found`. 

Proposal:
we can set the mapping for skipped users as empty string or null and we return a more specific error like: <user-not-mapped>. Wdyt?

cmd/jaas/cmd/migratemodel.go
```diff
@@ -235,3 +269,96 @@ func (c *migrateModelCommand) getMigrationSpec(token string, modelUUID string) (
...
+		return fmt.Errorf("could not get model info: %v", err)
+	}
+	modelUsers := modelInfo[0].Result.Users
+
+	var missingUserMessages []string
+
+	for _, user := range modelUsers {
+		if user.UserName == ofganames.EveryoneUser {
+			continue
+		}
```
SimoneDutto, line 293: can you add a comment here? Why are we skipping checks for everyone user?

SimoneDutto (commented):
looking good, just a few comments

internal/jimm/juju/migrationtarget.go
kian99: I don't want to add the "null" mapped users to the DB because they won't really be serving a purpose. If the local user is not found in the model mapping we can always return a nice error like `no mapping found for local-user "bob". "bob" has no access to resource ...`

cmd/jaas/cmd/migratemodel.go
kian99: Will add, it's just because the everyone user is a special user that doesn't need to be mapped.

kian99: Thinking about it some more, when we migrate a model to JIMM, we don't bring any of the old permissions so from that perspective it's fine. 
The user mapping is only needed for ensuring existing app-offer access stays working. 

If the `everyone@external` user had `consume` access on an offer, then you consumed it in a model as user `alice`, when you migrate the model to JIMM, the person performing the migration would need to know that `alice` should be included in the user mapping, and you wouldn't be able to see that from `juju show-offer` but there's not much we can do about that.

cmd/jaas/cmd/migratemodel.go
```diff
@@ -43,7 +49,19 @@ alice: alice@canonical.com
 bob: bob@canonical.com
 '''
 
-The mapping must, at a minimum, contain an entry for the model owner.
+The mapping must contain entries for all users that have access to the model and its offers.
+You can use the "juju show-model <model-name>" command to see the users that have access to
+the model.
+You can also use the "juju show-offer <offer-name>" command to see the users that have access
```
ale8k: Reading this (and trying assimilate a new user experience) it isn't immediately clear which you're suggesting here. I think it should be really clear in that it says "get your model users here and get your application offer users here and populate the map"?

cmd/jaas/cmd/migratemodel.go
```diff
@@ -43,7 +49,19 @@ alice: alice@canonical.com
...
+to each offer.
+
+Any users that you do not wish to be mapped must still be included with a null value or empty
+string in place of the external user. This indicates that you are intentionally skipping this
+local user, for example:
+'''
+alice: alice@canonical.com
```
ale8k, line 63: One day I need to meet Alice, she's everywhere 😆

cmd/jaas/cmd/interface.go
```diff
@@ -6,7 +6,9 @@ import "github.com/canonical/jimm/v3/pkg/api/params"
 
 // JIMMClient is an interface that defines the methods required for JIMM client operations.
 type JIMMClient interface {
```
ale8k: We call the juju client interface for migrations "MigrateAPI" but this "JIMMClient", as a nit can we call this "JIMMAPI" too?

cmd/jaas/cmd/migratemodel.go
```diff
@@ -235,3 +269,108 @@ func (c *migrateModelCommand) getMigrationSpec(token string, modelUUID string) (
...
+	if err != nil {
+		return fmt.Errorf("could not get model info: %v", err)
+	}
+	modelUsers := modelInfo[0].Result.Users
+
+	var missingUserMessages []string
+
+	for _, user := range modelUsers {
+		// Skip checks for the "everyone" since the user
+		// shouldn't need to map this special case user.
```
kian99: This comments reads a bit poorly and I can update it in my follow-up.

internal/jimm/juju/migrationtarget.go
SimoneDutto: what i'm trying to get at is:
- let's say that you migrate a model ignoring a user maybe because you forgot to remove the `null`
- you were wrong and that user should have been mapped
- the discharge endpoint is hit and the `not-found` error is logged to the consumer, which is not necessarily the same who did the migration

So my proposal is aimed at:
-  a more precise error, something like `user-not-mapped`. So it's easier to debug problems when some users are ignored.
- when you get the mapping from the CLI, you can see the ignored users.

Maybe we don't need that, let me know

kian99: Yes I think we're saying the same thing, I suggest we can add an error with a nice message for `user-not-mapped` but that we don't need to keep the unmapped user in the database. The fact that we don't find them in the user-mapping table means that they are not mapped. Unless you specifically want us to keep the mapping around but I don't think that's necessary.

cmd/jaas/cmd/migratemodel.go
```diff
@@ -235,3 +269,108 @@ func (c *migrateModelCommand) getMigrationSpec(token string, modelUUID string) (
 		TargetUser: accountDetails.User,
 	}, nil
 }
+
+// validateUserMapping checks that the user mapping contains all necessary users
+// that have access to the model and its offers.
+func (c *migrateModelCommand) validateUserMapping(userMapping map[string]string, modelUUID, modelName string, jujuAPI MigrateAPI) error {
```
ale8k, line 276: Instead of passing the API, we can just put it on the struct no and assign it in Run / Init? It doesn't feel like it should be an argument here.

internal/jimm/juju/migrationtarget.go
kian99: I can add that in a separate PR.

SimoneDutto: yeah, i was referring to keep the mapping to be explicit on the not-found and not-mapped users.

cmd/jaas/cmd/migratemodel.go
```diff
@@ -235,3 +269,108 @@ func (c *migrateModelCommand) getMigrationSpec(token string, modelUUID string) (
...
+	modelInfo, err := jujuAPI.ModelInfo([]names.ModelTag{names.NewModelTag(modelUUID)})
+	if err != nil {
+		return fmt.Errorf("could not get model info: %v", err)
```
ale8k, line 279: %w?

internal/jimm/juju/migrationtarget.go
SimoneDutto: But it is just an idea, if we don't like we can ignore this comment

SimoneDutto (approved):
LGTM

internal/jimm/juju/migrationtarget.go
kian99: I see. I think it's equivalent enough that it wouldn't make a difference. I'm also thinking about keeping the users in the table as valid usernames - keeping an empty external mapping means we also have to check wherever we use the value.

cmd/jaas/cmd/migratemodel.go
```diff
@@ -142,26 +173,29 @@ func (c *migrateModelCommand) Run(ctxt *cmd.Context) error {
...
 	if err != nil {
-		return fmt.Errorf("could not connect to controller %q: %w", currentController, err)
+		return fmt.Errorf("failure preparing migration: %v", err)
 	}
 
-	client := controllerapi.NewClient(apiCaller)
-	defer client.Close()
+	spec, err := c.getMigrationSpec(token, modelInfo.ModelUUID)
+	if err != nil {
+		return fmt.Errorf("could not get migration spec: %v", err)
```
ale8k, line 196: Why are we using %v? I know this error doesn't need to be unwrapped but is there like a specific reason?

cmd/jaas/cmd/migratemodel.go
```diff
@@ -235,3 +269,108 @@ func (c *migrateModelCommand) getMigrationSpec(token string, modelUUID string) (
...
 }
+
+// validateUserMapping checks that the user mapping contains all necessary users
+// that have access to the model and its offers.
+func (c *migrateModelCommand) validateUserMapping(userMapping map[string]string, modelUUID, modelName string, jujuAPI MigrateAPI) error {
+	modelInfo, err := jujuAPI.ModelInfo([]names.ModelTag{names.NewModelTag(modelUUID)})
+	if err != nil {
+		return fmt.Errorf("could not get model info: %v", err)
+	}
+	modelUsers := modelInfo[0].Result.Users
```
ale8k, line 284: My paranoia says len check first - just to be defensive.

cmd/jaas/cmd/migratemodel.go
```diff
@@ -235,3 +269,108 @@ func (c *migrateModelCommand) getMigrationSpec(token string, modelUUID string) (
...
+		}
+	}
+	if len(missingUserMessages) > 0 {
+		return fmt.Errorf("user mapping is missing the following users:\n%s\n", strings.Join(missingUserMessages, "\n"))
+	}
+
+	return nil
+}
+
+func (c *migrateModelCommand) newJujuClient() (MigrateAPI, error) {
```
ale8k, line 330: This feels like it should be inherited rather than on the command, perhaps we wrap modelcmd.ControllerCommandBase.

cmd/jaas/cmd/migratemodel.go
```diff
@@ -235,3 +269,108 @@ func (c *migrateModelCommand) getMigrationSpec(token string, modelUUID string) (
...
+
+	apiCaller, err := c.NewAPIRootWithDialOpts(c.store, currentController, "", nil)
+	if err != nil {
+		return nil, err
+	}
+
+	return jujuMigrateAPI{apiCaller: apiCaller}, nil
+}
+
+type jujuMigrateAPI struct {
```
ale8k, line 346: Lil godoc explaining we're just wrapping apis to help migrate. I know it is private, but we've got the interface consuming its own implementation. Or perhaps a `See [jujuMigrateAPI]` on MigrateAPI.

cmd/jaas/cmd/migratemodel_test.go
```diff
@@ -34,8 +62,209 @@ bob: bob@canonical.com
...
+		TargetControllerUUID:  "target-controller-uuid",
+		TargetControllerAlias: "target-controller",
+		TargetAddrs:           []string{"endpoint1"},
+		TargetCACert:          "test-ca-cert",
+		TargetToken:           "migration-token",
+		TargetUser:            "test-user",
+	}
+	s.migrateClient.EXPECT().InitiateMigration(migrationSpec).Return("migration-id", nil)
+
+	s.writer.EXPECT().Write(gomock.Any()).DoAndReturn(func(b []byte) (int, error) {
```
ale8k, line 119: Ngl, loving use of go mock again.

cmd/jaas/cmd/migratemodel_test.go
```diff
@@ -34,8 +62,209 @@ bob: bob@canonical.com
...
+	c.Assert(err, gc.IsNil)
+}
+
+func (s *migrateModelSuite) TestMigrate_ReadUserMappingFileWithNull(c *gc.C) {
+	userMappingFile, err := os.CreateTemp(c.MkDir(), "")
+	c.Assert(err, gc.IsNil)
+
+	userMappingContent := `
+alice: alice@canonical.com
+bob: null
```
ale8k, line 157: Can we add a test for "" too?

ale8k (approved):
Spot on. Thanks for this - some minor comments.

cmd/jaas/cmd/migratemodel.go
kian99: Exactly because it doesn't need to be unwrapped :D

PR #1665, opened 2025-07-30
feat: validate migration users backend

kian99:
## Description
This PR builds on #1664 and adds validation of the user mapping on the server side. We check that all users who have access to the model or have access to application-offers in the model are included in the user mapping.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

### DIFF
--- a/internal/jimm/juju/migrationtarget.go
+++ b/internal/jimm/juju/migrationtarget.go
@@ -26,6 +26,7 @@ import (
 	"github.com/canonical/jimm/v3/internal/dbmodel"
 	"github.com/canonical/jimm/v3/internal/errors"
 	"github.com/canonical/jimm/v3/internal/openfga"
+	ofganames "github.com/canonical/jimm/v3/internal/openfga/names"
 )
 
 const TIMEOUT_PENDING_MIGRATION = 24 * time.Hour
@@ -212,6 +213,9 @@ func (j *JujuManager) validateUserMapping(modelDescription description.Model, us
 
 	modelUsers := modelDescription.Users()
 	for _, user := range modelUsers {
+		if user.Name().Id() == ofganames.EveryoneUser {
+			continue
+		}
 		if _, ok := userMapping[user.Name().Id()]; !ok {
 			missingUserMessages = append(missingUserMessages, fmt.Sprintf("expected user %q who has %s access to the model", user.Name().Id(), user.Access()))
 		}
@@ -221,6 +225,9 @@ func (j *JujuManager) validateUserMapping(modelDescription description.Model, us
 	for _, app := range apps {
 		for _, offer := range app.Offers() {
 			for user, access := range offer.ACL() {
+				if user == ofganames.EveryoneUser {
+					continue
+				}
 				if _, ok := userMapping[user]; !ok {
 					missingUserMessages = append(missingUserMessages, fmt.Sprintf("expected user %q who has %s access to offer %q", user, access, offer.OfferName()))
 				}
--- a/internal/jimm/juju/migrationtarget.go
+++ b/internal/jimm/juju/migrationtarget.go
@@ -7,6 +7,7 @@ import (
 	"database/sql"
 	goerr "errors"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/juju/description/v9"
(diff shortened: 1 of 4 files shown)


internal/jimm/juju/migrationtarget.go
```diff
@@ -165,7 +167,10 @@ func (j *JujuManager) Prechecks(ctx context.Context, user *openfga.User, model m
 		return errors.E(op, fmt.Errorf("failed to get model migration %q: %w", model.UUID, err))
 	}
 
-	// TODO(Kian): Validate user mapping contains all local users in the model description.
+	err = j.validateUserMapping(model.ModelDescription, incomingModel.UserMapping)
+	if err != nil {
```
ale8k, line 171: Curiosity question: Why not if `err := thing(); err != nil`?

internal/jimm/juju/migrationtarget.go
```diff
@@ -251,6 +290,9 @@ func (j *JujuManager) modifyMigrationInfo(model *migration.ModelInfo, userMappin
 		// This is to ensure that the migration does not proceed with an invalid owner.
 		return errors.E(fmt.Errorf("no external user mapping found for local user %q", model.Owner.Id()))
 	}
+	if !names.IsValidUser(newOwner) {
```
ale8k, line 293: Good catch

internal/jimm/juju/migrationtarget.go
```diff
@@ -201,6 +206,40 @@ func (j *JujuManager) Prechecks(ctx context.Context, user *openfga.User, model m
 	return nil
 }
 
+// validateUserMapping checks that the provided user mapping contains all the users
+// that either have access to the model or have access to any application offers in the model.
+func (j *JujuManager) validateUserMapping(modelDescription description.Model, userMapping dbmodel.StringMap) error {
+	var missingUserMessages []string
+
+	modelUsers := modelDescription.Users()
```
ale8k, line 214: Could a model description have no users (like even no admin user or will it always have admin?)? 

And if:
- it always has admin 
- and the Len is one
- and you cannot delete the admin user (idk if this is possible)

Then we could return early here?

internal/jimm/juju/migrationtarget_test.go
```diff
@@ -224,6 +224,118 @@ func TestControllerDetailsForIncomingModel(t *testing.T) {
...
+		Name:  "test-cred",
+		Cloud: names.NewCloudTag("test"),
+	})
+	appArgs := description.ApplicationArgs{}
+	app := modelDescription.AddApplication(appArgs)
+
+	// Add an offer with an ACL for a user that is not mapped.
+	offerArgs := description.ApplicationOfferArgs{
+		OfferName: "test-offer",
+		ACL:       map[string]string{"jack": "admin"},
```
ale8k, line 329: Alice and Jack - Like the TV program? 😆

internal/jimm/juju/migrationtarget_test.go
```diff
@@ -224,6 +224,118 @@ func TestControllerDetailsForIncomingModel(t *testing.T) {
...
+		Owner:                  names.NewUserTag("bob"),
+		Name:                   "test-model",
+		AgentVersion:           version.MustParse("3.2.1"),
+		ControllerAgentVersion: version.MustParse("3.2.1"),
+		ModelDescription:       modelDescription,
+	}
+	return modelInfo
+}
+
+func TestPreChecks_SkipsEveryoneUser(t *testing.T) {
```
ale8k, line 343: No users test? Although I guess that cannot happen?

ale8k (approved):
LGTM honestly, nothing to change honestly.

internal/jimm/juju/migrationtarget.go
kian99: Ah no reason honestly, if it's a longish statement I usually put it on its own line.

internal/jimm/juju/migrationtarget.go
kian99: It could certainly have no users, even no admin user with access (I think?) but that shouldn't be an issue for this code because then `modelUsers` would be empty and `offer.ACL()` will be empty.
I wouldn't return early because even if no one has access to the model, technically a user could have access to an offer in the model.

internal/jimm/juju/migrationtarget_test.go
kian99: I think no users is possible but shouldn't be an issue.

kian99: I'll add a test for it.

internal/jimm/juju/migrationtarget.go
ale8k: Ahhh yes. Fair point!

alesstimec, line 214: i hope model owner is always included in the returned list of users

alesstimec (approved):
LGTM, although i am not a fan of these multiline error messages..

internal/jimm/juju/migrationtarget.go
kian99: I think technically it doesn't have to be but almost always would be.

kian99: Double checked and if the model owner is missing from the user mapping we return a different error elsewhere so I think it's alright.

PR #1666, opened 2025-07-30
feat(model deletion): delete a model when not-found calling `ModelInfo` or `ModelStatus`

SimoneDutto:
## Description

Previously, when deleting a model via the facade, we only marked it as deleted in JIMM and invoked Juju for deletion, but kept the model in the JIMM database until the cleanup routine ran. If ModelInfo was called during this window, JIMM would forward the request to Juju, which might return "not found" (satisfying Terraform's wait-for-delete), even though JIMM still retained the model. Attempting to re-create a model with the same name before JIMM's cleanup completed would result in an "already-exists" error. This change ensures that JIMM state is fully cleaned up before allowing model recreation.

> ModelInfo and ModelStatus are slightly different. ModelInfo convert `not-found` errors into `unauthorized` (9 years-old commit) https://github.com/juju/juju/blob/3.6/apiserver/facades/client/modelmanager/modelmanager.go#L905. ModelStatus returns a not-found.

FYI: the Juju CLI uses the ModelStatus facade to wait for not found. I need to update my terraform PR to have the same behavior

# QA

`juju add-model test`

`juju destroy-model test` 

(quickly)
`juju add-model test`

Before it was failing with `already-exists` now it doesn't

### DIFF
--- a/internal/jimm/juju/model.go
+++ b/internal/jimm/juju/model.go
@@ -265,6 +265,15 @@ func (j *JujuManager) ModelInfo(ctx context.Context, user *openfga.User, mt name
 		UUID: mt.Id(),
 	}
 	if err := api.ModelInfo(ctx, mi); err != nil {
+		// If the model is not found or code not authorized on the backing controller then
+		// we delete the model from JIMM.
+		// Juju returns CodeAuthorized when the model is not found for this facade.
+		if errors.ErrorCode(err) == errors.CodeNotFound || errors.ErrorCode(err) == errors.CodeUnauthorized {
+			errDelete := j.Database.DeleteModel(ctx, &m)
+			if errDelete != nil {
+				return nil, errors.E(op, errDelete)
+			}
+		}
 		return nil, errors.E(op, err)
 	}
 
@@ -435,7 +444,24 @@ func (j *JujuManager) ModelStatus(ctx context.Context, user *openfga.User, mt na
 	err := j.doModelAdmin(ctx, user, mt, func(m *dbmodel.Model, api API) error {
 		ms.OwnerTag = m.Owner.Tag().String()
 		ms.ModelTag = mt.String()
-		return api.ModelStatus(ctx, &ms)
+		err := api.ModelStatus(ctx, &ms)
+		if err != nil {
+			// If the model is not found on the backing controller then
+			// we delete the model from JIMM.
+			if errors.ErrorCode(err) == errors.CodeNotFound {
+				errDelete := j.Database.DeleteModel(ctx, &dbmodel.Model{
+					UUID: sql.NullString{
+						String: mt.Id(),
+						Valid:  true,
+					},
+				})
+				if errDelete != nil {
+					return errDelete
+				}
+			}
+			return err
+		}
+		return nil
 	})
 	if err != nil {
 		return nil, errors.E(op, err)
--- a/internal/jimm/juju/model.go
+++ b/internal/jimm/juju/model.go
@@ -269,7 +269,7 @@ func (j *JujuManager) ModelInfo(ctx context.Context, user *openfga.User, mt name
 		// we delete the model from JIMM.
 		// Juju returns CodeAuthorized when the model is not found for this facade.
 		if errors.ErrorCode(err) == errors.CodeNotFound || errors.ErrorCode(err) == errors.CodeUnauthorized {
-			errDelete := j.Database.DeleteModel(ctx, &m)
+			errDelete := j.deleteModel(ctx, mt)
 			if errDelete != nil {
 				return nil, errors.E(op, errDelete)
 			}
@@ -449,12 +449,7 @@ func (j *JujuManager) ModelStatus(ctx context.Context, user *openfga.User, mt na
 			// If the model is not found on the backing controller then
 			// we delete the model from JIMM.
 			if errors.ErrorCode(err) == errors.CodeNotFound {
-				errDelete := j.Database.DeleteModel(ctx, &dbmodel.Model{
-					UUID: sql.NullString{
-						String: mt.Id(),
-						Valid:  true,
-					},
-				})
+				errDelete := j.deleteModel(ctx, mt)
 				if errDelete != nil {
 					return errDelete
 				}
@@ -467,6 +462,27 @@ func (j *JujuManager) ModelStatus(ctx context.Context, user *openfga.User, mt na
 		return nil, errors.E(op, err)
 	}
 	return &ms, nil
+}
+
+// deleteModel deletes the model with the given ModelTag from JIMM's database and
+// clean the openfga tuples related to the model.
+func (j *JujuManager) deleteModel(ctx context.Context, mt names.ModelTag) error {
+	err := j.Database.DeleteModel(ctx, &dbmodel.Model{
+		UUID: sql.NullString{
+			String: mt.Id(),
+			Valid:  true,
+		},
+	})
+	if err != nil {
+		zapctx.Error(ctx, "failed to delete model from database", zap.String("model", mt.Id()), zap.Error(err))
+		return err
+	}
+	err = j.OpenFGAClient.RemoveModel(ctx, mt)
+	if err != nil {
+		zapctx.Error(ctx, "failed to remove model from OpenFGA", zap.String("model", mt.Id()), zap.Error(err))
+		return err
+	}
+	return nil
 }
 
 // ForEachUserModel calls the given function once for each model that the
--- a/internal/jimm/juju/model_test.go
+++ b/internal/jimm/juju/model_test.go
@@ -1679,6 +1679,64 @@ func TestModelInfo(t *testing.T) {
 	}
 }
 
+func TestModelInfoNotFound(t *testing.T) {
+	c := qt.New(t)
+
+	j := newTestJujuManager(c, &parameters{
+		Dialer: &jimmtest.Dialer{
+			API: &jimmtest.API{
+				ModelInfo_: func(_ context.Context, mi *jujuparams.ModelInfo) error {
+					return errors.E(errors.CodeNotFound, "model not found")
+				},
+			},
+		},
+	})
+
+	env := jimmtest.ParseEnvironment(c, modelInfoTestEnv)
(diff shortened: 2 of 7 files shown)


internal/jimm/juju/model.go
```diff
@@ -265,6 +265,15 @@ func (j *JujuManager) ModelInfo(ctx context.Context, user *openfga.User, mt name
 		UUID: mt.Id(),
 	}
 	if err := api.ModelInfo(ctx, mi); err != nil {
+		// If the model is not found or code not authorized on the backing controller then
+		// we delete the model from JIMM.
+		// Juju returns CodeAuthorized when the model is not found for this facade.
```
kian99, line 270: I think we can get rid of the CodeUnauthorized condition. I remember there was a comment in the model cleanup routine that old versions of Juju returned this error when a model was not available. But JIMM shouldn't be facing auth errors since we dial the controller as a superuser. Even if we dial as the user making the request, they should have access to the model.

It also feels like it could possibly lead to some unintended behavior deleting the model when we shouldn't.

internal/jimm/juju/model.go
```diff
@@ -265,6 +265,15 @@ func (j *JujuManager) ModelInfo(ctx context.Context, user *openfga.User, mt name
...
+		if errors.ErrorCode(err) == errors.CodeNotFound || errors.ErrorCode(err) == errors.CodeUnauthorized {
+			errDelete := j.Database.DeleteModel(ctx, &m)
+			if errDelete != nil {
+				return nil, errors.E(op, errDelete)
+			}
```
kian99: We also want to delete the permissions in OpenFGA. A function to do both these things that is also used in the cleanup routine would be good.

internal/jimm/juju/model_test.go
```diff
@@ -1679,6 +1679,64 @@ func TestModelInfo(t *testing.T) {
...
+
+	env := jimmtest.ParseEnvironment(c, modelInfoTestEnv)
+	env.PopulateDBAndPermissions(c, j.ResourceTag(), j.Database, j.OpenFGAClient)
+	dbUser, err := dbmodel.NewIdentity("alice@canonical.com")
+	c.Assert(err, qt.IsNil)
+
+	user := openfga.NewUser(dbUser, j.OpenFGAClient)
+
+	_, err = j.ModelInfo(context.Background(), user, names.NewModelTag("00000002-0000-0000-0000-000000000001"))
+	c.Assert(errors.ErrorCode(err), qt.Equals, errors.CodeNotFound)
```
kian99: Calling this twice should result in the same error, i.e. verify it is idempotent.

internal/jimm/juju/model.go
```diff
@@ -265,6 +265,15 @@ func (j *JujuManager) ModelInfo(ctx context.Context, user *openfga.User, mt name
 		UUID: mt.Id(),
 	}
 	if err := api.ModelInfo(ctx, mi); err != nil {
+		// If the model is not found or code not authorized on the backing controller then
+		// we delete the model from JIMM.
+		// Juju returns CodeAuthorized when the model is not found for this facade.
+		if errors.ErrorCode(err) == errors.CodeNotFound || errors.ErrorCode(err) == errors.CodeUnauthorized {
+			errDelete := j.Database.DeleteModel(ctx, &m)
+			if errDelete != nil {
+				return nil, errors.E(op, errDelete)
```
kian99: This might/might not need an extra check where if the error from the DB is NotFound, we return a CodeNotFound to the user.

internal/jujuapi/modelmanager_test.go
```diff
@@ -1086,10 +1084,6 @@ func (s *modelManagerSuite) TestDestroyModel(c *gc.C) {
 	c.Assert(mis[0].Error, gc.Equals, (*jujuparams.Error)(nil))
 	c.Assert(mis[0].Result.Life, gc.Equals, life.Dying)
 
-	// Kill the model.
-	err = s.JIMM.Database.DeleteModel(ctx, s.Model)
-	c.Assert(err, gc.Equals, nil)
```
kian99, line 1091: Why was this removed?

internal/jimm/juju/model.go
SimoneDutto: https://github.com/juju/juju/blob/3.6/apiserver/facades/client/modelmanager/modelmanager.go#L905
as i wrote in the description it seems that the `ModelInfo` facade is mapping a not-found error to a unauthorize error

internal/jimm/juju/model.go
SimoneDutto: AKAIK gorm doesn't return a not found when it doesn't find the record to delete and returns just a `nil` error

internal/jujuapi/modelmanager_test.go
SimoneDutto: it's not needed anymore because the model is deleted from the database `ModelInfo` is called few lines before

internal/jimm/juju/model.go
kian99: Missed that, thanks!

internal/jimm/juju/model.go
SimoneDutto: done

internal/jimm/juju/model_test.go
```diff
@@ -1679,6 +1679,91 @@ func TestModelInfo(t *testing.T) {
...
+	err = j.Database.GetModel(context.Background(), &model)
+	c.Assert(errors.ErrorCode(err), qt.Equals, errors.CodeNotFound)
+
+	_, err = j.ModelInfo(context.Background(), user, mt)
+	c.Assert(errors.ErrorCode(err), qt.Equals, errors.CodeNotFound)
+
+	// Check the openfga tuple is deleted as a consequence of the error
+	ok, err = user.IsModelReader(c.Context(), mt)
+	c.Assert(err, qt.IsNil)
+	c.Assert(ok, qt.IsFalse)
```
kian99, line 1718: nit but can you move this check up to before the second ModelInfo call, otherwise it reads as if it is only expected to deleted after the second call to ModelInfo.

internal/jimm/juju/model_test.go
```diff
@@ -1679,6 +1679,91 @@ func TestModelInfo(t *testing.T) {
...
+	err = j.Database.GetModel(context.Background(), &model)
+	c.Assert(errors.ErrorCode(err), qt.Equals, errors.CodeNotFound)
+
+	_, err = j.ModelStatus(context.Background(), user, mt)
+	c.Assert(errors.ErrorCode(err), qt.Equals, errors.CodeNotFound)
+
+	// Check the openfga tuple is deleted as a consequence of the error
+	ok, err = user.IsModelReader(c.Context(), mt)
+	c.Assert(err, qt.IsNil)
+	c.Assert(ok, qt.IsFalse)
```
kian99, line 1762: Same here.

internal/jimm/juju/model.go
```diff
@@ -435,14 +444,47 @@ func (j *JujuManager) ModelStatus(ctx context.Context, user *openfga.User, mt na
...
+		UUID: sql.NullString{
+			String: mt.Id(),
+			Valid:  true,
+		},
+	})
+	if err != nil {
+		zapctx.Error(ctx, "failed to delete model from database", zap.String("model", mt.Id()), zap.Error(err))
+		return err
+	}
+	err = j.permissionManager.RemoveModel(ctx, mt)
```
kian99: I like the use of the permission manager. It also would've been possible to just do `j.authSvc.RemoveModel()` in here right? Maybe one day we'll remove the auth service from the Juju manager. Edit: I've added some more thoughts on this in another comment.

internal/jimm/permissions/relations.go
```diff
@@ -250,3 +251,15 @@ func (j *permissionManager) parseTuple(ctx context.Context, tuple apiparams.Rela
...
 	return &t, nil
 }
+
+// RemoveModel removes all tuples that contain a model as the target.
+func (j *permissionManager) RemoveModel(ctx context.Context, model names.ModelTag) error {
+	const op = errors.Op("jimm.RemoveModel")
+	if model.Id() == "" {
+		return errors.E(op, errors.CodeBadRequest, "model tag cannot be empty")
+	}
+	if err := j.authSvc.RemoveModel(ctx, model); err != nil {
```
kian99: Reading this I have some thoughts,
- If we keep this pattern, we'll need to think about the roles and responsibilities between this permissions manager and the authSvc.
- If other packages in `internal/jimm` are supposed to use the permissions manager instead of using the `authSvc` directly, what we accomplishing? If we compare it to the database, all the managers have a DB object to go directly to the DB.

I think I would drop this change and just use the authSvc directly in the juju manager to keep things simple. But open to different ideas on this.

SimoneDutto: i was thinking exactly the opposite.
We should drop the openfga client from inside JujuManager, because it is a concrete type and it would be more correct to use the permission manager for what we need from openfga

SimoneDutto: Keep using the openfga client directly is the thing as using the db without the managers

SimoneDutto: As discussed offline. I'm using the openfga client directly because the PermissionManager is an unnecessary abstraction layer that we can potentially remove from the JujuManager in a follow-up

kian99 (approved):

internal/jimm/juju/model.go
```diff
@@ -435,14 +444,47 @@ func (j *JujuManager) ModelStatus(ctx context.Context, user *openfga.User, mt na
...
+		UUID: sql.NullString{
+			String: mt.Id(),
+			Valid:  true,
+		},
+	})
+	if err != nil {
+		zapctx.Error(ctx, "failed to delete model from database", zap.String("model", mt.Id()), zap.Error(err))
+		return err
+	}
+	err = j.OpenFGAClient.RemoveModel(ctx, mt)
```
ale8k, line 480: Much prefer if err := ...; err != nil. Wdyt?

ale8k (approved):
lgtm!

internal/jimm/juju/model.go
SimoneDutto: i prefer the two lines, otherwise to me it seems more important `err!=nil` than the actual method called

PR #1667, opened 2025-08-05
feat: bootstrap job

ale8k:
## Description
Implements the bootstrap job

It isn't QA'd manually, I tried to but it is very difficult to get the essentials of the JIMM environment together. QA will happen once everything is plugged in. I have added extensive unhappy path testing for each scenario to cover our bases as best we can and everything is mocked.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests


kian99 (commented):
This looks like it's in the wrong place. We're accepting the JujuManager, a struct holding business logic into the jobtracker package. Instead, we already have `internal/jimm/bootstrap` where we should write the bootstrap job.

internal/jimm/bootstrap/bootstrap_job.go
```diff
@@ -0,0 +1,247 @@
...
+	"github.com/canonical/jimm/v3/internal/dbmodel"
+	"github.com/canonical/jimm/v3/internal/errors"
+	"github.com/canonical/jimm/v3/internal/jimm/juju"
+	"github.com/canonical/jimm/v3/internal/jobtracker"
+	"github.com/canonical/jimm/v3/internal/jujuclistore"
+	"github.com/canonical/jimm/v3/internal/jujucommands"
+	"github.com/canonical/jimm/v3/internal/openfga"
+)
+
+type BootstrapJobStore interface {
```
alesstimec: godoc

internal/jimm/bootstrap/bootstrap_job.go
```diff
@@ -0,0 +1,247 @@
...
+	LockBootstrap(ctx context.Context, ttl time.Duration) error
+	GetController(ctx context.Context, controller *dbmodel.Controller) (err error)
+	AddBootstrapLog(ctx context.Context, jobId uuid.UUID, logLine string) (err error)
+	UnlockBootstrap(ctx context.Context) error
+}
+
+type BootstrapJobJujuManager interface {
```
alesstimec: godoc

internal/jimm/bootstrap/bootstrap_job.go
```diff
@@ -0,0 +1,247 @@
...
+	AddController(ctx context.Context, user *openfga.User, ctl *dbmodel.Controller, creds juju.ControllerCreds) error
+}
+
+type BootstrapJobBinaryStore interface {
```
alesstimec: godoc

internal/jimm/bootstrap/bootstrap_job.go
alesstimec: also.. i'd name this `JobManager` interface.. we know it's a BootstrapJobXX because it's in the bootstrap package

internal/jimm/bootstrap/bootstrap_job.go
alesstimec: just `BinaryStore` will do

internal/jimm/bootstrap/bootstrap_job.go
kian99: Rename to `Store` since we're already in the bootstrap package.

internal/jimm/bootstrap/bootstrap_job.go
kian99: Rename to `ControllerManager` or `JujuManager`

internal/jimm/bootstrap/bootstrap_job.go
```diff
@@ -0,0 +1,247 @@
...
+	Get(ctx context.Context, spec jujuclistore.JujuBinarySpec) (*jujuclistore.Binary, error)
+}
+
+// TODO: Validate params?
+type BootstrapJobParams struct {
```
alesstimec: proper godoc and name it `Params`

internal/jimm/bootstrap/bootstrap_job.go
kian99: Rename to `BinaryStore`

internal/jimm/bootstrap/bootstrap_job.go
kian99: Just `JobParams` or even just `Params`

internal/jimm/bootstrap/bootstrap_job.go
```diff
@@ -0,0 +1,247 @@
...
+	CLIArch    string
+
+	// User defined command arguments
+
+	CloudNameAndRegion string
+	ControllerName     string
+	AgentVersion       string
+	BootstrapTimeout   int
+	CloudCred          jujucloud.CloudCredential
+	// PersonalCloud is the personally defined cloud. Only necessary if the cloud is not a public
```
alesstimec: what is a `personally defined cloud`?

internal/jimm/bootstrap/bootstrap_job.go
```diff
@@ -0,0 +1,247 @@
...
+
+	CLIVersion string
+	CLIOs      string
+	CLIArch    string
+
+	// User defined command arguments
+
+	CloudNameAndRegion string
+	ControllerName     string
+	AgentVersion       string
```
alesstimec: is this related to `CLIVersion`?

internal/jimm/bootstrap/bootstrap_job.go
```diff
@@ -0,0 +1,247 @@
...
+
+		if err := store.LockBootstrap(ctx, bootstrapLockTTL); err != nil {
+			zapctx.Error(
+				ctx,
+				"failed to acquire bootstrap lock",
+				zap.String("job-id", jobId.String()),
+				zap.String("controller-name", p.ControllerName),
+				zap.Error(err),
+			)
+			return fmt.Errorf("failed to acquire bootstrap lock: %w", err)
```
alesstimec: why aren't we using the `errors` package here and in other places in this function?

internal/jimm/bootstrap/bootstrap_job.go
```diff
@@ -0,0 +1,247 @@
...
+					zap.String("job-id", jobId.String()),
+					zap.String("controller-name", p.ControllerName),
+					zap.Error(err),
+				)
+			}
+		}()
+
+		err := store.GetController(ctx, &dbmodel.Controller{Name: p.ControllerName})
+		if err == nil {
+			return fmt.Errorf("controller %q already exists", p.ControllerName)
```
alesstimec: should be returning an error with `AlreadyExists` code, no?

internal/jimm/bootstrap/bootstrap_job.go
```diff
@@ -0,0 +1,247 @@
...
+	// JIMM Provided command arguments (i.e., ones that must be set by JIMM when bootstrapping).
+
+	LoginTokenRefreshURL string
+}
+
+// BootstrapJob return a [JobFunc] [for use in the [Tracker]] responsible for
+// bootstrapping a controller and adding it to JIMM.
+//
+//nolint:gocognit // The cognit is kinda inevitable here.
+func BootstrapJob(
```
kian99: Let's make this a method on the bootstrap manager, as that should have all the dependencies already.

internal/jimm/bootstrap/bootstrap_job.go
```diff
@@ -0,0 +1,247 @@
...
+) jobtracker.JobFunc {
+	const bootstrapLockTTL = 40 * time.Minute
+
+	return func(ctx context.Context) error {
+		jobId, ok := jobtracker.JobIdFromContext(ctx)
+		if !ok {
+			return fmt.Errorf("failed to get job ID from context")
+		}
+
+		zapctx.Debug(
```
kian99: I'm kind of against every log line taking 6 LOC. Can we  just make it 1 even when there are a number of params we want to log.

internal/jimm/bootstrap/bootstrap_job.go
```diff
@@ -0,0 +1,247 @@
...
+		dbCtrlCreds := juju.ControllerCreds{
+			AdminIdentityName: account.User,
+			AdminPassword:     account.Password,
+		}
+		if err := jujuManager.AddController(
+			ctx,
+			user,
+			&dbCtrl,
+			dbCtrlCreds,
+		); err != nil {
```
kian99: Can we make this 1 line?

internal/jimm/bootstrap/bootstrap_job.go
```diff
@@ -0,0 +1,247 @@
...
+
+		for output := range outputCh {
+			if output.Err != nil {
+				return fmt.Errorf("bootstrap command failed: %w", output.Err)
+			}
+			if writeLogErr := store.AddBootstrapLog(
+				ctx,
+				jobId,
+				output.Line,
+			); writeLogErr != nil {
```
kian99: 1 line

internal/jimm/bootstrap/bootstrap_job.go
```diff
@@ -0,0 +1,247 @@
...
+		if err != nil {
+			return fmt.Errorf("failed to get controller details: %w", err)
+		}
+
+		hps, err := network.ParseProviderHostPorts(ctrlDetails.APIEndpoints...)
+		if err != nil {
+			return fmt.Errorf("failed to parse API endpoints for controller: %w", err)
+		}
+		for i := range hps {
+			// Mark all the unknown scopes public.
```
kian99: Why?

internal/jimm/bootstrap/bootstrap_job.go
```diff
@@ -0,0 +1,247 @@
...
+				hps[i].Scope = network.ScopePublic
+			}
+		}
+
+		dbCtrl := dbmodel.Controller{
+			UUID:          ctrlDetails.ControllerUUID,
+			Name:          p.ControllerName,
+			PublicAddress: ctrlDetails.PublicDNSName,
+			CACertificate: ctrlDetails.CACert,
+			// TLSHostname: // Not needed.
```
kian99: dd

internal/jimm/bootstrap/bootstrap_job.go
```diff
@@ -0,0 +1,247 @@
...
+			return fmt.Errorf("failed to add controller to JIMM: %w", err)
+		}
+
+		return nil
+	}
+}
+
+// BootstrapExecutor holds a wrapper to run a command. It is primarily for testing
+// and when running the bootstrap command the implemention to be used is [DefaultBootstrapExecutor].
+type BootstrapExecutor interface {
```
alesstimec: please move all type definitions to the top of the file

internal/jimm/bootstrap/bootstrap_job.go
```diff
@@ -0,0 +1,247 @@
...
+	RunWrapper(
+		ctx context.Context,
+		binaryPath, jujuDataDir string,
+		params jujucommands.BootstrapCmdParams,
+	) (<-chan jujucommands.OutputLine, jujuclient.ClientStore, func(), error)
+}
+
+type DefaultBootstrapExecutor struct{}
```
alesstimec: godoc, if it needs to be exported

internal/jimm/bootstrap/bootstrap_job_test.go
```diff
@@ -0,0 +1,823 @@
...
+	clientStore := mocks.NewMockClientStore(ctrl)
+
+	i, err := dbmodel.NewIdentity("bob@canonical.com")
+	c.Assert(err, qt.IsNil)
+	user := openfga.NewUser(i, nil)
+
+	return ctrl, store, jujuManager, binaryStore, executor, clientStore, user
+}
+
+func (s *bootstrapManagerSuite) TestBootstrapJob(c *qt.C) {
```
kian99: Not something for this PR but I spoke to Francesco a while back and he mentioned that suites are intended for scenarios where you want to test multiple implementations of something but for just running a setup function, just call `setup()` from the start of each test. I've also noticed the tooling has a problem when a package has a mixture of qt suites and regular testing.T tests, at least in my VSCode I'll click run for a qt suite test and it will also run all the testing.T tests which is super slow and annoying.

internal/jimm/bootstrap/bootstrap_job_test.go
```diff
@@ -0,0 +1,823 @@
...
+	// Mocked in order of execution:
+	cleanupCalled := false // To be asserted after job run - ensures cleanup was run.
+	store.EXPECT().LockBootstrap(gomock.Any(), gomock.Any()).Return(nil).Times(1)
+	store.EXPECT().GetController(
+		gomock.Any(),
+		&dbmodel.Controller{Name: jobParams.ControllerName},
+	).Return(
+		errors.E(errors.CodeNotFound, errors.E("test err")),
+	).Times(1)
+	// TODO: Figure a way to check done is indeed deferred?
```
kian99: Seems like we can address this in this PR

internal/jimm/bootstrap/bootstrap_job_test.go
```diff
@@ -0,0 +1,823 @@
...
+	"github.com/canonical/jimm/v3/internal/jimm/bootstrap/mocks"
+	"github.com/canonical/jimm/v3/internal/jujuclistore"
+	"github.com/canonical/jimm/v3/internal/jujucommands"
+	"github.com/canonical/jimm/v3/internal/openfga"
+)
+
+var (
+	jobParams = bootstrap.BootstrapJobParams{
+		JujuDataDir:          "/path/to/a/juju/data/dir",
+		CLIVersion:           "3.6.9",
```
alesstimec: 3.6.9 has not been released yet :)

internal/jimm/bootstrap/bootstrap_job_test.go
```diff
@@ -0,0 +1,823 @@
...
+	store.EXPECT().AddBootstrapLog(gomock.Any(), gomock.Any(), testOutputLine).Return(nil).Times(1)
+	ctrlDetails := &jujuclient.ControllerDetails{
+		APIEndpoints: []string{
+			"10.0.0.1:17070",
+			"172.0.0.1:17070",
+			"192.0.0.1:17070",
+		},
+		ControllerUUID: "I am actually a uuid, I promise",
+		PublicDNSName:  "I am not a public DNS, I am a private DNS",
+		CACert:         "Very secure CA cert, promise",
```
kian99: Very creative

internal/jimm/bootstrap/bootstrap_job_test.go
```diff
@@ -0,0 +1,823 @@
...
+	c.Assert(err, qt.IsNil)
+	jujuManager.EXPECT().AddController(
+		gomock.Any(),
+		user,
+		&dbmodel.Controller{
+			UUID:          ctrlDetails.ControllerUUID,
+			Name:          jobParams.ControllerName,
+			PublicAddress: ctrlDetails.PublicDNSName,
+			CACertificate: ctrlDetails.CACert,
+			// TLSHostname: // Not needed.
```
kian99: dd

alesstimec (commented):
just a quick scan.. will do a proper review in a bit

internal/jimm/bootstrap/bootstrap_job_test.go
```diff
@@ -0,0 +1,823 @@
...
+		binaryStore,
+		executor,
+		user,
+	)
+
+	id, err := s.jobTracker.Run(
+		testCtx,
+		"test-job-type",
+		job,
+		time.Second*1000,
```
kian99: Should we have a smaller timeout value even though we don't expect it to be hit?

internal/jimm/bootstrap/bootstrap_job_test.go
```diff
@@ -0,0 +1,823 @@
...
+
+	// Mocked in order of execution:
+	cleanupCalled := false // To be asserted after job run - ensures cleanup was run.
+	store.EXPECT().LockBootstrap(gomock.Any(), gomock.Any()).Return(nil).Times(1)
+	store.EXPECT().GetController(
+		gomock.Any(),
+		&dbmodel.Controller{Name: jobParams.ControllerName},
+	).Return(
+		errors.E(errors.CodeNotFound, errors.E("test err")),
+	).Times(1)
```
kian99: Should this still be returning an error in this test?

internal/jimm/bootstrap/bootstrap_job_test.go
```diff
@@ -0,0 +1,823 @@
...
+	defer ctrl.Finish()
+
+	// Mocked in order of execution:
+	store.EXPECT().LockBootstrap(gomock.Any(), gomock.Any()).Return(nil).Times(1)
+	store.EXPECT().GetController(
+		gomock.Any(),
+		&dbmodel.Controller{Name: jobParams.ControllerName},
+	).Return(
+		errors.E(errors.CodeNotFound, errors.E("test err")),
+	).Times(1)
```
kian99: Should this still return an error in this test?

internal/jimm/bootstrap/bootstrap_job_test.go
```diff
@@ -0,0 +1,823 @@
...
```
kian99: Should this still return an error in this test?

internal/jimm/bootstrap/bootstrap_job_test.go
```diff
@@ -0,0 +1,823 @@
...
+	defer ctrl.Finish()
+
+	// Mocked in order of execution:
+	cleanupCalled := false // To be asserted after job run - ensures cleanup was run.
+	store.EXPECT().LockBootstrap(gomock.Any(), gomock.Any()).Return(nil).Times(1)
+	store.EXPECT().GetController(
+		gomock.Any(),
+		&dbmodel.Controller{Name: jobParams.ControllerName},
+	).Return(
+		errors.E(errors.CodeNotFound, errors.E("test err")),
```
kian99: Same question everywhere

internal/jimm/bootstrap/bootstrap_job.go
```diff
@@ -0,0 +1,247 @@
...
+				// This is because the bootstrap process may still succeed, and we
+				// don't want to fail the entire job just because we couldn't log.
+				zapctx.Error(ctx, "failed to write bootstrap log", zap.Error(writeLogErr), zap.String("jobId", jobId.String()))
+			}
+		}
+
+		// We could use .CurrentController, but should bootstrap change their behaviour
+		// to not set the default controller, it would break. As such we're explicitly
+		// getting the controller by name.
+		ctrlDetails, err := clientStore.ControllerByName(p.ControllerName)
```
kian99: If you want to reduce the cognition value, we could split this function in two at roughly this point. One that bootstraps and then another that adds the controller to JIMM.

internal/jimm/bootstrap/bootstrap_job.go
ale8k: It already has a store. I'll remove the bootstrap from all interfaces but keep the prefix of job at the least.

internal/jimm/bootstrap/bootstrap_job.go
ale8k: I'm not a fan of naming implementation and interface names the same. Happy to use JobJujuManager?

internal/jimm/bootstrap/bootstrap_job.go
kian99: Ah, then we can just put these on the existing store interface right?

internal/jimm/bootstrap/bootstrap_job.go
ale8k: No no, this is the agent version passed to the bootstrap command.

internal/jimm/bootstrap/bootstrap_job.go
ale8k: A personally defined cloud is what we often call "private" or "user defined" clouds, in Juju world they call them personal clouds.

internal/jimm/bootstrap/bootstrap_job.go
ale8k: I am really against a single line as I'm prone to missing things in the log. Isn't this just a matter of taste?

internal/jimm/bootstrap/bootstrap_job.go
ale8k: What do you mean?

internal/jimm/bootstrap/bootstrap_job_test.go
ale8k: Haha was me being funny. :P

internal/jimm/bootstrap/bootstrap_job_test.go
ale8k: Don't expect it to be hit honestly. It's entirely mocked.

internal/jimm/bootstrap/bootstrap_job_test.go
ale8k: Yes, it's just confirming the controller doesn't exist before continuing.

internal/jimm/bootstrap/bootstrap_job.go
ale8k: Not quite, because that defines a method the job won't be using. If we're going to decouple things, I'd rather do it properly and define an interface specific to what it is using.

ale8k: Actually, on second thought we'd have to pass two in. Maybe quick call on what you'd prefer.

internal/jimm/bootstrap/bootstrap_job.go
ale8k: Ahhhh I was avoiding it honestly as we've been trying to get away from it.

internal/jimm/bootstrap/bootstrap_job.go
ale8k: I copied the logic within our AddController and decided to leave the comment. You got me curious so I had a little dive, but it's here:
https://github.com/canonical/jimm/pull/716/files#diff-f169c12506d378ae48c42fd8541eaaa2eb904414ba795c40e52c98ff6dc7ac65R126

See jimm.go 

I had more of a think about this, and I think I know why. 

If it was privately scoped, then JIMM wouldn't be able to reach the controller. So... I think this just reinforces the presumption that JIMM only communicates to publicly accessible controllers. 

In theory, the scopes are wider now and we don't really account for anything other than "a public controller", here's the scopes:

```go
	ScopeUnknown      Scope = ""
	ScopePublic       Scope = "public"
	ScopeCloudLocal   Scope = "local-cloud"
	ScopeFanLocal     Scope = "local-fan"
	ScopeMachineLocal Scope = "local-machine"
	ScopeLinkLocal    Scope = "link-local"
```
But I don't see any reason why we couldn't support controllers that aren't publicly accessible and in air-gapped scenarios I think we should correctly mark controllers.

kian99: The place where we use the scope is here https://github.com/canonical/jimm/blob/v3/internal/rpc/dial.go#L78, and if you look inside the `maybeReachable` function we will attempt unknown scopes. This was a recent change in the last couple of years though so I assume before we wouldn't have dialed addresses with scope unknown. Seems okay to leave in the logic changing unknown to public for now though.

ale8k: Ahhh nice!

internal/jimm/bootstrap/bootstrap_job_test.go
ale8k: Yeah true. I tend to like using suites to keep the tests collected together or share a simple param and mhmm as for the tooling problem, I don't face this... It hasn't happened to me (or it did a long time ago?)

internal/jimm/bootstrap/bootstrap.go
```diff
@@ -5,15 +5,30 @@ package bootstrap
...
+	"github.com/canonical/jimm/v3/internal/jujuclistore"
 	"github.com/canonical/jimm/v3/internal/openfga"
 	"github.com/canonical/jimm/v3/pkg/api/params"
 )
 
+// Store defines the store methods required by the manager.
+type Store interface {
+	QueryBootstrapLog(ctx context.Context, jobId uuid.UUID, offset int) (loggies []string, nextOffsetValue int, err error)
+
+	// BootstrapJob store methods:
```
alesstimec, line 40: i don't understand. Are you saying that these should be part of a different interface?

internal/jimm/bootstrap/bootstrap.go
```diff
@@ -23,18 +38,34 @@ type JobTracker interface {
...
+
+	store       Store
+	jobtracker  JobTracker
+	jujuManager JujuManager
+	binaryStore BinaryStore
 }
 
 // NewBootstrapManager creates a new BootstrapManager instance.
-func NewBootstrapManager(store Store, authSvc *openfga.OFGAClient, jobtracker JobTracker) (*bootstrapManager, error) {
+// TODO(ale8k): Remove authSvc later, it isn't used.
```
alesstimec, line 77: why wait?

internal/jimm/bootstrap/bootstrap_job.go
```diff
@@ -0,0 +1,245 @@
...
+	"github.com/canonical/jimm/v3/internal/jimm/juju"
+	"github.com/canonical/jimm/v3/internal/jobtracker"
+	"github.com/canonical/jimm/v3/internal/jujuclistore"
+	"github.com/canonical/jimm/v3/internal/jujucommands"
+	"github.com/canonical/jimm/v3/internal/openfga"
+)
+
+// BootstrapExecutor holds a wrapper to run a command. It is primarily for testing
+// and when running the bootstrap command the implemention to be used is [DefaultBootstrapExecutor].
+type BootstrapExecutor interface {
```
alesstimec: interface should be defined where it is used, not where it is implemented

internal/jimm/bootstrap/bootstrap_job.go
```diff
@@ -0,0 +1,245 @@
...
+	RunWrapper(
+		ctx context.Context,
+		binaryPath, jujuDataDir string,
+		params jujucommands.BootstrapCmdParams,
+	) (<-chan jujucommands.OutputLine, jujuclient.ClientStore, func(), error)
+}
+
+// DefaultBootstrapExecutor defines the default, and expected bootstrap executor
+// to be used in a [bootstrapManager.BootstrapJob]
+type DefaultBootstrapExecutor struct{}
```
alesstimec: you don't need a type for this.. could have
```
var defaultWrapperFunction = func(
	ctx context.Context,
	binaryPath, jujuDataDir string,
	params jujucommands.BootstrapCmdParams,
) (<-chan jujucommands.OutputLine, jujuclient.ClientStore, func(), error)
```

alesstimec: if not, does this need to be exported?

internal/jimm/bootstrap/bootstrap_job.go
```diff
@@ -0,0 +1,245 @@
...
+			"starting bootstrap job",
+			zap.String("job-id", jobId.String()),
+			zap.String("controller-name", p.ControllerName),
+		)
+
+		if err := b.store.LockBootstrap(ctx, bootstrapLockTTL); err != nil {
+			zapctx.Error(
+				ctx,
+				"failed to acquire bootstrap lock",
+				zap.String("job-id", jobId.String()),
```
alesstimec: if we are going to have multi-line log lines, like this, fine.. but let's try and reduce LOC by using
```
ctx = zapctx.WithFields(
  ctx, 
  zap.String("job-id", jobId.String())
  zap.String("controller-name", p.ControllerName)
)
```
once, so that we don't have to repeat it every time we log something.. create a context with all the zap fields you care about and use it..

internal/jimm/bootstrap/bootstrap_job.go
```diff
@@ -0,0 +1,245 @@
...
+					zap.String("job-id", jobId.String()),
+					zap.String("controller-name", p.ControllerName),
+					zap.Error(err),
+				)
+			}
+		}()
+
+		err := b.store.GetController(ctx, &dbmodel.Controller{Name: p.ControllerName})
+		if err == nil {
+			return errors.E(errors.CodeAlreadyExists, fmt.Errorf("controller %q already exists", p.ControllerName))
```
alesstimec: see, now, this is what i don't like.. we're using `fmt.Errorf`, but in this one case we switch to using the `errors` package..

internal/jimm/bootstrap/bootstrap_job.go
```diff
@@ -0,0 +1,245 @@
...
+		}
+		if errors.ErrorCode(err) != errors.CodeNotFound {
```
alesstimec: and here you expect an error produced by out `errors` package, but you return a wrapped error..

internal/jimm/bootstrap/bootstrap_job.go
```diff
@@ -0,0 +1,245 @@
...
+			PersonalCloud:        p.PersonalCloud,
+			CloudCred:            p.CloudCred,
+		},
+	)
+	if err != nil {
+		return fmt.Errorf("failed to run bootstrap command: %w", err)
+	}
+	defer cleanup()
+
+	for output := range outputCh {
```
alesstimec: so here we are waiting for the bootstrap process to finish, right?

alesstimec (commented):
i'm really not fond of error handling in this PR

internal/jimm/bootstrap/bootstrap_job.go
kian99: I don't think we've stopped using it anywhere, personally I've just just done things like errors.E(op, fmt.Errorf()) and I think we should continue to do that.

internal/jimm/bootstrap/bootstrap_job.go
```diff
@@ -0,0 +1,245 @@
...
+					ctx,
+					"failed to unlock bootstrap lock",
+					zap.String("job-id", jobId.String()),
+					zap.String("controller-name", p.ControllerName),
+					zap.Error(err),
+				)
+			}
+		}()
+
+		err := b.store.GetController(ctx, &dbmodel.Controller{Name: p.ControllerName})
```
kian99: This check isn't strong enough to prevent someone from bootstrapping twice with the same controller name. I know our bootstrap is limited to 1 controller at a time, but if that wasn't the case then you can have 2 API calls to bootstrap pass this check. I don't think there's a simple solution though because creating a controller in the DB is not currently designed for mid-bootstrap controllers.

internal/jujuclistore/store.go
```diff
@@ -172,11 +172,16 @@ func (p *jujuCLIStore) Get(ctx context.Context, spec JujuBinarySpec) (*Binary, e
...
 		FullPath: file.Name(),
 	}
-	binary.referenceCount.Store(1)
+	binary.RefCount().Store(1)
 	p.entries[url] = binary
 	return binary, nil
 }
 
+// RefCount returns the current reference count for this binary.
+func (b *Binary) RefCount() *atomic.Int32 {
```
kian99: Why do other packages need the reference count?

internal/jimm/bootstrap/bootstrap_job_test.go
```diff
@@ -0,0 +1,796 @@
...
+}
+
+func (s *bootstrapManagerSuite) TestBootstrapJob(c *qt.C) {
+	testCtx := c.Context()
+
+	binaryPath := "/faketmp/juju"
+	testOutputLine := "test-line"
+
+	binary := &jujuclistore.Binary{FullPath: binaryPath}
+	binary.RefCount().Store(1) // Simulate the Get().
```
kian99: I don't think you should be exposing this detail just to test it. I assume you're doing this to verify that you're calling cli.Done()? Can we mock that instead?

internal/jimm/bootstrap/bootstrap_job.go
ale8k: In some independent packages/files I'm sure we've stopped, quick scan shown:
```
/home/ubuntu/repos/jimm/cmd/jaas/cmd/migratemodel.go
```

And I'm fairly certain there's others.

internal/jimm/bootstrap/bootstrap.go
ale8k: In some ways yes, but for convenience and as discussed with the team I used the store interface on the manager.

internal/jimm/bootstrap/bootstrap.go
ale8k: The PR is 1900 lines nearly. I don't want to add more to it.

internal/jimm/bootstrap/bootstrap_job.go
ale8k: I'm aware - but in this circumstance I had no other way to test this. I also asked the @kian99  on this one about using a function vs interface, and he was in favour of an interface.

internal/jimm/bootstrap/bootstrap_job.go
ale8k: Agreed, read comment above.

internal/jimm/bootstrap/bootstrap_job.go
ale8k: Done.

internal/jimm/bootstrap/bootstrap_job.go
ale8k: I did this under request to be fair. I feel like we shouldn't be using our errors package below the API package as the normal go errors package suffices for our use cases (and we can define our own errors to switch the type on within the API package to set codes for reflection when returning them to a facade method). What would you like me to do here? I'm happy to switch it all to jimm.errors but this kinda feels like we need a team discussion as it seems we don't really have a consensus?

internal/jimm/bootstrap/bootstrap_job.go
ale8k: Well it still extracts the code as it is holding the value. See L277 of the test.

internal/jimm/bootstrap/bootstrap_job.go
ale8k: Yup.

internal/jujuclistore/store.go
ale8k: So I needed a way to test the done is called and I had to reason this one with myself for it to make sense to change this package. I wondered what if you didn't want to cleanup until a certain number of uses, i.e.,  "until refcount == x, then done()" and in this case I think it is OK to expose the refcount. It also doesn't do any harm and enables the testing I required.

internal/jimm/bootstrap/bootstrap_job.go
kian99: You do still have to use our errors package below the API layer because:
- The DB uses it so you need to use it in the business logic if you want to work with the errors from the DB.
- We need to construct errors in the business layer that the API can translate into Juju RPC error messages.

internal/jimm/bootstrap/bootstrap_job_test.go
ale8k: I've commented above explaining my reasoning. I'd argue it isn't an implementation detail and is safe to use in a test.

internal/jimm/bootstrap/bootstrap_job.go
ale8k: Ah yeah in dbError(...). 

I'm happy to switch it over to errors.E but I distinctly remember us wanting to move away from it...

internal/jujuclistore/store.go
kian99: >didn't want to cleanup until a certain number of uses

Then you'd just pass the binary object around until you're done then you call Done().
I don't mind much, but I don't think anyone outside the package should rely on it so I'd prefer to not expose it. It just exists so the clistore package can know when it's safe to cleanup the CLI.

It doesn't do any harm until a package starts using it and then the clistore wants to change its implementation but can't because the refcount is now a part of it's public API.

ale8k: Which is fair enough, we could change the implementation. But pragmatically we won't (I don't feel like this will be used anywhere else anyway?) and this felt like the simplest touch?

internal/jimm/bootstrap/bootstrap_job.go
ale8k: We could do a get and start a transaction, then only one of them can succeed/fail? I just didn't add it because we are limited to 1 controller.

internal/jimm/bootstrap/bootstrap_job.go
kian99: I would argue it's less important to use errors.E in the CLI

internal/jimm/bootstrap/bootstrap_job.go
kian99: I don't think a Get in a transaction is enough here, nothing stopping a second routine from doing the same get and passing since the controller still doesn't exist in the DB.
I'd say it's fine for now but I'd suggest adding a comment about this.

internal/jujuclistore/store.go
kian99: Would prefer for this not to be exposed but up to you.

kian99 (approved):
Looks good bar the discussion on the errors.

internal/jimm/bootstrap/bootstrap_job.go
ale8k: If we did a get, if not exist, create it is OK no?

internal/jujuclistore/store.go
ale8k: I'll leave it if it is OK. Would prefer to move on and test it then figure another way to test this. (Will resolve but comment if you change your mind)

internal/jimm/bootstrap/bootstrap_job.go
ale8k: Yeah agreed, I just more meant there's other places we use std err package and it's kinda inconsistent.

internal/jimm/bootstrap/bootstrap_job.go
alesstimec: IFF there's a good reason to move away from errors.E, i'd be glad to hear it..

internal/jimm/bootstrap/bootstrap_job.go
```diff
@@ -0,0 +1,245 @@
+// Copyright 2025 Canonical.
+
+package bootstrap
```
alesstimec: is there a package godoc for the bootstrap package?

internal/jimm/bootstrap/bootstrap_job.go
luci1900: Looking at the errors.E implementation, I'm surprised it doesn't actually store a stack trace like `pkg/errors` does https://github.com/pkg/errors/blob/master/errors.go#L165

internal/jimm/bootstrap/bootstrap.go
alesstimec: create a JIRA ticket, then

internal/jimm/bootstrap/bootstrap_job.go
ale8k: It also doesn't wrap properly (which is my argument to use it only in API layer). We can fix that though, it just means many tests break.

internal/jimm/bootstrap/bootstrap_job.go
alesstimec: this is not used in this PR, nor is it tested?

internal/jimm/bootstrap/bootstrap_job.go
alesstimec: it is NOT done.. please DO IT :D

internal/jimm/bootstrap/bootstrap_job.go
alesstimec: just create an "empty" controller in the DB... we should have an index on controller name, so it'll fail if somebody tries to bootstrap with the same name, no?

internal/jimm/bootstrap/bootstrap_job.go
alesstimec: what if request timeout expires and we're not done bootstrapping?

alesstimec (changes requested):

internal/jimm/bootstrap/bootstrap_job.go
ale8k: There is now.

internal/jimm/bootstrap/bootstrap_job.go
ale8k: then it'll return on line 186 and cleanup will run.

internal/jimm/bootstrap/bootstrap_job.go
ale8k: That's what I'm saying - I think that'd work no?

internal/jimm/bootstrap/bootstrap_job.go
ale8k: It is tested elsewhere and we'll plug this in in the next PR

internal/jimm/bootstrap/bootstrap.go
ale8k: https://warthogs.atlassian.net/browse/JUJU-8344

internal/jimm/bootstrap/bootstrap_job.go
ale8k: then it'll return on line 186 and cleanup will run.

internal/jimm/bootstrap/bootstrap_job.go
```diff
@@ -0,0 +1,245 @@
...
+	PersonalCloud jujucloud.Cloud
+
+	// JIMM Provided command arguments (i.e., ones that must be set by JIMM when bootstrapping).
+
+	LoginTokenRefreshURL string
+}
+
+// BootstrapJob returns a [jobtracker.JobFunc] [for use in the [jobtracker.Tracker]] responsible for
+// bootstrapping a controller and adding it to JIMM.
+func (b *bootstrapManager) BootstrapJob(
```
alesstimec: please move all bootstrapManager methods to boostrap.go, where this type is defined.. let's keep methods of a type and type definition in the same file unless there's a good reason

internal/jimm/bootstrap/bootstrap_job.go
alesstimec: well, this issue needs to be resolved before we merge this PR

internal/jimm/bootstrap/bootstrap_job.go
alesstimec: my question is: should we abort  the bootstrap job because the request context expired? bootstrapping on maas might take ages and i'm sure it won't be lightning fast on openstact, too

internal/jujuclistore/store.go
alesstimec: you don't need to expose the refcount at all (and i would prefer if you didn't).. instead of 
`defer binary.Done()` you do
```
var done = func( binary jujuclistore.Binary) {
   binary.Done()
}

defer done(binary)
```
and mock the done function in the test..

alesstimec (approved):
LGTM with the refcount changes reverted as discussed

PR #1669, opened 2025-08-08
feat: prevent simultaneous changes to migration mode

kian99:
## Description
This PR lays some groundwork for a follow-up change to make the completion of an internal model migration automatic.
Specifically, this PR does the following:
- Adds a DB method called `SetModelMigrating` that locks the DB row for the model and updates its migration mode ensuring that if a model is already undergoing a migration this cannot be set again.
- Remove the use of `"github.com/juju/juju/state"` for the migrationMode field and create our own.
- Add a new migration mode called "migrate-internal"

Partially fixes [JUJU-7481](https://warthogs.atlassian.net/browse/JUJU-7481)

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

[JUJU-7481]: https://warthogs.atlassian.net/browse/JUJU-7481?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/db/model.go
+++ b/internal/db/model.go
@@ -4,8 +4,10 @@ package db
 
 import (
 	"context"
+	"database/sql"
 
 	"gorm.io/gorm"
+	"gorm.io/gorm/clause"
 
 	"github.com/canonical/jimm/v3/internal/dbmodel"
 	"github.com/canonical/jimm/v3/internal/errors"
@@ -73,6 +75,49 @@ func (d *Database) GetModel(ctx context.Context, model *dbmodel.Model) (err erro
 		return errors.E(op, dbError(err))
 	}
 	return nil
+}
+
+// SetModelMigrating updates the model's migration mode to the provided value.
+// This function will return an error if the model's migration mode is already
+// set to anything besides MigrationModeNone.
+func (d *Database) SetModelMigrating(ctx context.Context, uuid string, migrationMode dbmodel.MigrationMode) (m *dbmodel.Model, err error) {
+	const op = errors.Op("db.SetModelMigrating")
+	if err := d.ready(); err != nil {
+		return nil, errors.E(op, err)
+	}
+
+	durationObserver := servermon.DurationObserver(servermon.DBQueryDurationHistogram, string(op))
+	defer durationObserver()
+	defer servermon.ErrorCounter(servermon.DBQueryErrorCount, &err, string(op))
+
+	db := d.DB.WithContext(ctx)
+	err = db.Transaction(func(tx *gorm.DB) error {
+
+		m = &dbmodel.Model{UUID: sql.NullString{String: uuid, Valid: true}}
+		if err := preloadModel("", tx).Clauses(clause.Locking{Strength: "UPDATE"}).First(m).Error; err != nil {
+			return errors.E(op, err)
+		}
+
+		if m.MigrationMode != dbmodel.MigrationModeNone {
+			return errors.E(op, "model is already migrating", errors.CodeBadRequest)
+		}
+
+		m.MigrationMode = migrationMode
+		if err := tx.Save(m).Error; err != nil {
+			return errors.E(op, err)
+		}
+
+		return nil
+	})
+	if err != nil {
+		err = dbError(err)
+		if errors.ErrorCode(err) == errors.CodeNotFound {
+			return nil, errors.E(op, err, "model not found")
+		}
+		return nil, errors.E(op, err)
+	}
+
+	return m, nil
 }
 
 // GetModelsUsingCredential returns all models that use the specified credentials.
--- a/internal/db/model.go
+++ b/internal/db/model.go
@@ -5,6 +5,7 @@ package db
 import (
 	"context"
 	"database/sql"
+	"fmt"
 
 	"gorm.io/gorm"
 	"gorm.io/gorm/clause"
@@ -77,11 +78,12 @@ func (d *Database) GetModel(ctx context.Context, model *dbmodel.Model) (err erro
 	return nil
 }
 
-// SetModelMigrating updates the model's migration mode to the provided value.
+// SetModelMigrationMode updates the model's migration mode to the provided value.
 // This function will return an error if the model's migration mode is already
-// set to anything besides MigrationModeNone.
-func (d *Database) SetModelMigrating(ctx context.Context, uuid string, migrationMode dbmodel.MigrationMode) (m *dbmodel.Model, err error) {
-	const op = errors.Op("db.SetModelMigrating")
+// set to anything besides MigrationModeNone. Use `UpdateModel` to complete the
+// migration and reset the migration mode to MigrationModeNone.
+func (d *Database) SetModelMigrationMode(ctx context.Context, uuid string, migrationMode dbmodel.MigrationMode) (m *dbmodel.Model, err error) {
+	const op = errors.Op("db.SetModelMigrationMode")
 	if err := d.ready(); err != nil {
 		return nil, errors.E(op, err)
 	}
@@ -92,9 +94,12 @@ func (d *Database) SetModelMigrating(ctx context.Context, uuid string, migration
 
 	db := d.DB.WithContext(ctx)
 	err = db.Transaction(func(tx *gorm.DB) error {
-
 		m = &dbmodel.Model{UUID: sql.NullString{String: uuid, Valid: true}}
 		if err := preloadModel("", tx).Clauses(clause.Locking{Strength: "UPDATE"}).First(m).Error; err != nil {
+			err = dbError(err)
+			if errors.ErrorCode(err) == errors.CodeNotFound {
+				return errors.E(op, fmt.Errorf("model with uuid %q does not exist", uuid))
+			}
 			return errors.E(op, err)
 		}
 
@@ -104,17 +109,13 @@ func (d *Database) SetModelMigrating(ctx context.Context, uuid string, migration
 
 		m.MigrationMode = migrationMode
 		if err := tx.Save(m).Error; err != nil {
-			return errors.E(op, err)
+			return errors.E(op, dbError(err))
 		}
 
 		return nil
 	})
 	if err != nil {
-		err = dbError(err)
-		if errors.ErrorCode(err) == errors.CodeNotFound {
-			return nil, errors.E(op, err, "model not found")
-		}
-		return nil, errors.E(op, err)
+		return nil, err
 	}
 
 	return m, nil
--- a/internal/db/model_test.go
+++ b/internal/db/model_test.go
@@ -767,7 +767,7 @@ func (s *dbSuite) TestCountModelsByController(c *qt.C) {
 	c.Assert(count, qt.Equals, 3)
 }
 
-const testSetModelMigratingEnv = `clouds:
+const testSetModelMigrationModeEnv = `clouds:
 - name: test
   type: test
   regions:
@@ -792,11 +792,11 @@ models:
   controller: test
 `
 
-func (s *dbSuite) TestSetModelMigrating(c *qt.C) {
+func (s *dbSuite) TestSetModelMigrationMode(c *qt.C) {
 	err := s.Database.Migrate(context.Background())
 	c.Assert(err, qt.Equals, nil)
 
-	env := jimmtest.ParseEnvironment(c, testSetModelMigratingEnv)
+	env := jimmtest.ParseEnvironment(c, testSetModelMigrationModeEnv)
 	env.PopulateDB(c, s.Database)
 
 	model := env.Models[0].DBObject(c, s.Database)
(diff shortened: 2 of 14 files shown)


internal/db/model.go
```diff
@@ -75,6 +77,49 @@ func (d *Database) GetModel(ctx context.Context, model *dbmodel.Model) (err erro
 	return nil
 }
 
+// SetModelMigrating updates the model's migration mode to the provided value.
+// This function will return an error if the model's migration mode is already
+// set to anything besides MigrationModeNone.
+func (d *Database) SetModelMigrating(ctx context.Context, uuid string, migrationMode dbmodel.MigrationMode) (m *dbmodel.Model, err error) {
```
alesstimec: i'd suggest naming this `SetModelMigrationMode` or something similar as it takes an arbitrary migration mode

internal/db/model_test.go
```diff
@@ -766,3 +766,69 @@ func (s *dbSuite) TestCountModelsByController(c *qt.C) {
...
+	err := s.Database.Migrate(context.Background())
+	c.Assert(err, qt.Equals, nil)
+
+	env := jimmtest.ParseEnvironment(c, testSetModelMigratingEnv)
+	env.PopulateDB(c, s.Database)
+
+	model := env.Models[0].DBObject(c, s.Database)
+	c.Assert(model.MigrationMode, qt.Equals, dbmodel.MigrationModeNone)
+
+	modelWithMigration, err := s.Database.SetModelMigrating(context.Background(), model.UUID.String, dbmodel.MigrationModeMigrateInternal)
```
alesstimec: perhaps a test case where a model does not exist?

internal/dbmodel/model.go
```diff
@@ -14,6 +14,30 @@ import (
...
+	// exported away from JIMM.
+	MigrationModeExporting = MigrationMode(state.MigrationMode("exporting"))
+
+	// MigrationModeImporting reflects a model that is being imported into a
+	// controller, but is not yet fully active.
+	MigrationModeImporting = MigrationMode(state.MigrationMode("importing"))
+
+	// MigrationModeMovingInternal reflects a model that is being moved internally
+	// within JIMM, e.g. from one controller to another.
+	MigrationModeMigrateInternal = MigrationMode("migrate-internal")
```
alesstimec: the other modes seem to be in present continuous.. so maybe 
```
MigrationModelMigratingInternally = MigrationMode("migrating-internally")
```

alesstimec (approved):
LGTM with a few suggestions

internal/db/model_test.go
```diff
@@ -766,3 +766,77 @@ func (s *dbSuite) TestCountModelsByController(c *qt.C) {
...
+- name: test-1
+  uuid: 00000002-0000-0000-0000-000000000001
+  owner: alice@canonical.com
+  cloud: test
+  region: test-region
+  cloud-credential: test-cred
+  controller: test
+`
+
+func (s *dbSuite) TestSetModelMigrating(c *qt.C) {
```
alesstimec: TestSetModelMigrationMode

internal/db/model_test.go
```diff
@@ -766,3 +766,77 @@ func (s *dbSuite) TestCountModelsByController(c *qt.C) {
...
+	c.Assert(modelWithMigration.MigrationMode, qt.Equals, dbmodel.MigrationModeMigrateInternal)
+
+	// Validate the model in the database has the correct migration mode.
+	var migrationMode dbmodel.MigrationMode
+	err = s.Database.DB.Table("models").Select("migration_mode").Where("uuid = ?", model.UUID).Scan(&migrationMode).Error
+	c.Assert(err, qt.IsNil)
+	c.Assert(migrationMode, qt.Equals, dbmodel.MigrationModeMigrateInternal)
+}
+
+func (s *dbSuite) TestSetModelMigrating_ModelDoesNotExist(c *qt.C) {
```
alesstimec: TestSetModelMigrationMode_ModelDoesNotExist

internal/db/model_test.go
```diff
@@ -766,3 +766,77 @@ func (s *dbSuite) TestCountModelsByController(c *qt.C) {
...
+	err := s.Database.Migrate(context.Background())
+	c.Assert(err, qt.Equals, nil)
+
+	_, err = s.Database.SetModelMigrationMode(context.Background(), "fake-uuid", dbmodel.MigrationModeMigrateInternal)
+	c.Assert(err, qt.ErrorMatches, `model with uuid "fake-uuid" does not exist`)
+}
+
+func (s *dbSuite) TestSetModelMigrating_PreventSetAgain(c *qt.C) {
```
alesstimec: TestSetModelMigrationMode_PreventSetAgain

internal/jimm/juju/controller.go
```diff
@@ -695,15 +701,27 @@ func (j *JujuManager) InitiateMigration(ctx context.Context, user *openfga.User,
...
-		return result, errors.E(op, "failed to retrieve the model from the database", err)
+		return result, errors.E(op, fmt.Errorf("failed to set model as migrating: %v", err))
+	}
+
+	// Until we have better handling for partial failures we try to revert
+	// the model migration mode if we fail after this to avoid inconsistenties.
+	rollbackMigrationMode := func() {
+		model.MigrationMode = dbmodel.MigrationModeNone
+		if updateErr := j.Database.UpdateModel(ctx, model); updateErr != nil {
+			zapctx.Error(ctx, "failed to revert model migration mode after failure initiating migration", zap.Error(updateErr))
```
alesstimec, line 718: in case revert fails, we could provide a command to reset migration mode, right?

alesstimec: and at some point we might want alerts on this...

alesstimec (approved):
LGTM with a few comments

internal/jimm/juju/controller.go
kian99: Yeah we can add a command to fix it manually but ideally the code would handle this better.

internal/db/model.go
```diff
@@ -75,6 +78,49 @@ func (d *Database) GetModel(ctx context.Context, model *dbmodel.Model) (err erro
 	return nil
 }
 
+// SetModelMigrationMode updates the model's migration mode to the provided value.
+// This function will return an error if the model's migration mode is already
+// set to anything besides MigrationModeNone. Use `UpdateModel` to complete the
+// migration and reset the migration mode to MigrationModeNone.
+func (d *Database) SetModelMigrationMode(ctx context.Context, uuid string, migrationMode dbmodel.MigrationMode) (m *dbmodel.Model, err error) {
```
ale8k, line 85: Does this need to return the model? I would've expected it to just set the migration mode and call it a day.

ale8k: Also can we rename it to SetInitialMigrationMode? Or Something? Because it returns an error if it is already set? 

And could we add the methods to complete the migration instead of using UpdateModel just for consistency?

internal/db/model.go
```diff
@@ -75,6 +78,49 @@ func (d *Database) GetModel(ctx context.Context, model *dbmodel.Model) (err erro
...
+		if err := preloadModel("", tx).Clauses(clause.Locking{Strength: "UPDATE"}).First(m).Error; err != nil {
+			err = dbError(err)
+			if errors.ErrorCode(err) == errors.CodeNotFound {
+				return errors.E(op, fmt.Errorf("model with uuid %q does not exist", uuid))
+			}
+			return errors.E(op, err)
+		}
+
+		if m.MigrationMode != dbmodel.MigrationModeNone {
+			return errors.E(op, "model is already migrating", errors.CodeBadRequest)
```
ale8k, line 107: No need for the code, it isn't used.

internal/db/model_test.go
```diff
@@ -766,3 +766,77 @@ func (s *dbSuite) TestCountModelsByController(c *qt.C) {
...
+
+	model := env.Models[0].DBObject(c, s.Database)
+	c.Assert(model.MigrationMode, qt.Equals, dbmodel.MigrationModeNone)
+
+	_, err = s.Database.SetModelMigrationMode(context.Background(), model.UUID.String, dbmodel.MigrationModeMigrateInternal)
+	c.Assert(err, qt.IsNil)
+
+	// Attempt to set the model to migrating again should return an error.
+	_, err = s.Database.SetModelMigrationMode(context.Background(), model.UUID.String, dbmodel.MigrationModeMigrateInternal)
+	c.Assert(err, qt.ErrorMatches, `model is already migrating`)
```
ale8k, line 841: @luci1900 This is an example where the code isn't used, and even if it was, you could have multiple "bad requests" in some scenarios. @alesstimec you can see the you'll need to define many codes if there was more paths / more specific scenarios. Just trying to sway you 😉

internal/dbmodel/model.go
```diff
@@ -14,6 +14,30 @@ import (
 	"github.com/canonical/jimm/v3/internal/errors"
 )
 
+// MigrationMode specifies where the Model is with respect to migration.
+// The modes mirror Juju's state values with some JIMM-specific additions.
+// We use our own type to avoid introducing a dependency on Juju's state
```
ale8k, line 19: Don't need to explain this bit

internal/jimm/juju/controller.go
```diff
@@ -656,12 +656,18 @@ func (j *JujuManager) UpdateMigratedModel(ctx context.Context, user *openfga.Use
...
 }
 
-// InitiateMigration triggers the migration of the specified model to a target controller.
-// externalMigration indicates whether this model is moving to a controller managed by
-// JIMM or not.
+// InitiateMigration triggers the migration of the specified model to a controller
+// that is not managed by JIMM.
 func (j *JujuManager) InitiateMigration(ctx context.Context, user *openfga.User, spec jujuparams.MigrationSpec) (jujuparams.InitiateMigrationResult, error) {
-	const op = errors.Op("jimm.InitiateMigration")
+	internalMigration := false
```
ale8k, line 662: If it is always false, we can just pass false?

internal/jimm/juju/controller.go
```diff
@@ -695,15 +701,27 @@ func (j *JujuManager) InitiateMigration(ctx context.Context, user *openfga.User,
...
-	model.SetTag(mt)
-	err = j.Database.GetModel(ctx, &model)
+	migrationMode := dbmodel.MigrationModeExporting
+	if internalMigration {
+		migrationMode = dbmodel.MigrationModeMigrateInternal
+	}
+	model, err := j.Database.SetModelMigrationMode(ctx, mt.Id(), migrationMode)
 	if err != nil {
-		return result, errors.E(op, "failed to retrieve the model from the database", err)
+		return result, errors.E(op, fmt.Errorf("failed to set model as migrating: %v", err))
```
ale8k, line 710: If we propagated this up, the user would get "bad request" but it's more a "conflict" (like HTTP409 example) I think...

internal/jimm/juju/controller.go
```diff
@@ -695,15 +701,27 @@ func (j *JujuManager) InitiateMigration(ctx context.Context, user *openfga.User,
...
+	}
+
+	// Until we have better handling for partial failures we try to revert
+	// the model migration mode if we fail after this to avoid inconsistenties.
+	rollbackMigrationMode := func() {
```
ale8k, line 715: Call this "setMigrationModeNone"? I actually think we should just have a db method for this to be honest. You're essentially defining one anyway.

ale8k (approved):
I think I'd prefer a specific SetMigrationModeNone(). Some minor comments but LGTM!

internal/db/model.go
kian99: It is, it's returned to the client.

internal/dbmodel/model.go
kian99: I would like to leave it in.

internal/jimm/juju/controller.go
kian99: The point is to make it easier to read.

internal/jimm/juju/controller.go
kian99: I think you helped me spot a mistake on my part, the user wouldn't get the code because I'm not passing the `err` value to errors.E() outside the fmt.Sprintf(), will check and fix that.
I agree though it's not a bad request, will see if there's a better error type.

internal/jimm/juju/controller.go
kian99: Can do, but then you'll see in my next PR that it causes a clash between where do logic. In the next PR I've added methods on `dbmodel` for successful/failure completing migration, if I add db methods that encapsulate this logic it will be worse. I'll share in person.

internal/db/model.go
kian99: But considering that it's incorrect, I'll remove it.

internal/db/model.go
kian99: Will rename to `SetInitialMigrationMode`, as for "add the methods to complete the migration", I think this will be interesting in the next PR and may change this function to something like `getModelWithLock` instead so let's discuss that in the follow-up.

kian99:
@ale8k Addressing your comments in #1670

PR #1670, opened 2025-08-08
feat: automatic completion of internal model migration

kian99:
## Description
This PR aims to make the completion of an internal model migration automatic, it builds on #1669. 

Currently a model migration between two controller managed by JIMM must be completed by the user using the `update-migrated-model` command of the `jaas` plugin. With this change we adjust the method `CleanupNotFoundModels` into `PollModels` since this method was being run periodically to identify when models were deleted and remove them from JIMM's database.

The new `PollModels` method does the same thing as before (requests ModelInfo for all models) with a slight optimisation in that we dial each controller only once instead of dialing the same controller repeatedly (not an issue before since we had the dial cache). Next, the logic is based on the model's migration mode, instead of skipping the model if it was in a migrating state, now we can complete the migration if its state is MigrationInternal.

Rather than creating a separate method to poll for internally migrating models I've chosen to put the logic together with polling for deleted models because of the following reasons:
- `CleanupNotFoundModels` was already called every 1 minute which is adequate for completing internal migrations.
- The polling is doing the same (calling `ModelInfo`).

If we want to split these up so that one can run more frequently than the other we can do that in the future.

Finally, ideally we'd have a real integration test in our bash suite for internal model migration, I'll add this in a follow-up.

Fixes [JUJU-7481](https://warthogs.atlassian.net/browse/JUJU-7481)

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

## Test instructions
Successfully tested this locally,
1. Start JIMM `docker compose up -d --wait`
2. Add controller to JIMM `./local/jimm/setup-controller.sh` and `./local/jimm/add-controller.sh`
3. Add another controller with a different name, `export CONTROLLER_NAME=test-controller` and then repeat step 2.
5. `juju add-model test-1` and `juju show-model` and record the model UUID.
6. Switch to the controllers `qa-lxd` and `test-controller` to see where it landed.
8. `go build ./cmd/jaas`
9. Depending on where the model landed `./jaas migrate-internal <controller> <model-UUID>`
10. Run `juju status` after the migration starts and you should get an `unauthorized` error. Wait 1 minute and then `juju status` should succeed.
11. (optionally) Watch logs with `docker logs -f jimm`.

[JUJU-7481]: https://warthogs.atlassian.net/browse/JUJU-7481?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/db/model_test.go
```diff
@@ -766,3 +766,77 @@ func (s *dbSuite) TestCountModelsByController(c *qt.C) {
...
+- name: test-1
+  uuid: 00000002-0000-0000-0000-000000000001
+  owner: alice@canonical.com
+  cloud: test
+  region: test-region
+  cloud-credential: test-cred
+  controller: test
+`
+
+func (s *dbSuite) TestSetModelMigrating(c *qt.C) {
```
alesstimec: TestSetModelMigrationMode

internal/db/model_test.go
```diff
@@ -766,3 +766,77 @@ func (s *dbSuite) TestCountModelsByController(c *qt.C) {
...
+	c.Assert(modelWithMigration.MigrationMode, qt.Equals, dbmodel.MigrationModeMigrateInternal)
+
+	// Validate the model in the database has the correct migration mode.
+	var migrationMode dbmodel.MigrationMode
+	err = s.Database.DB.Table("models").Select("migration_mode").Where("uuid = ?", model.UUID).Scan(&migrationMode).Error
+	c.Assert(err, qt.IsNil)
+	c.Assert(migrationMode, qt.Equals, dbmodel.MigrationModeMigrateInternal)
+}
+
+func (s *dbSuite) TestSetModelMigrating_ModelDoesNotExist(c *qt.C) {
```
alesstimec: SetModelMigrationMode_ModelDoesNotExist

internal/db/model_test.go
```diff
@@ -766,3 +766,77 @@ func (s *dbSuite) TestCountModelsByController(c *qt.C) {
...
+	err := s.Database.Migrate(context.Background())
+	c.Assert(err, qt.Equals, nil)
+
+	_, err = s.Database.SetModelMigrationMode(context.Background(), "fake-uuid", dbmodel.MigrationModeMigrateInternal)
+	c.Assert(err, qt.ErrorMatches, `model with uuid "fake-uuid" does not exist`)
+}
+
+func (s *dbSuite) TestSetModelMigrating_PreventSetAgain(c *qt.C) {
```
alesstimec: SetModelMigrationMode_PreventSetAgain

internal/jimm/juju/model_poller.go
```diff
@@ -0,0 +1,145 @@
...
+
+	// Step 2: Loop over controllers and process their models
+	// This way we only dial each controller once.
+	for controllerUUID, models := range controllerModels {
+		if len(models) == 0 {
+			continue
+		}
+		api, err := j.dialController(ctx, &models[0].Controller)
+		if err != nil {
+			zapctx.Error(ctx, "cannot dial controller", zap.String("controller", controllerUUID), zap.Error(err))
```
alesstimec, line 48: if somebody bypassed JIMM and killed the controller, we would never clean up those models, right?

alesstimec: how would we handle this situation? we might have to provide some tooling to resolve it..

alesstimec (commented):
LGTM in general, but i'm not too fond of the tight coupling between two orthogonal tasks: model cleanup and finishing model migrations, based on the migration mode of the model..

internal/jimm/juju/model_poller.go
kian99: Yip that's true. Nothing for this PR imo though.

internal/db/model_test.go
```diff
@@ -716,6 +718,44 @@ func (s *dbSuite) TestGetModelsByController(c *qt.C) {
 	c.Assert(foundModelNames, qt.DeepEquals, modelNames)
 }
 
+func (s *dbSuite) TestGetModelForUpdate(c *qt.C) {
+	ctx := context.Background()
```
ale8k: c.Context() here no?

internal/db/model.go
```diff
@@ -48,6 +46,37 @@ func (d *Database) GetModel(ctx context.Context, model *dbmodel.Model) (err erro
...
+	err = d.getModel(ctx, model, false)
+	if err != nil {
+		return errors.E(op, err)
+	}
+	return nil
+}
+
+// GetModelForUpdate retrieves a model and locks the row for updates.
+// This method should be used within a transaction.
+func (d *Database) GetModelForUpdate(ctx context.Context, model *dbmodel.Model) (err error) {
```
ale8k: Can we append Tx to the name? I did that for some that should be used in transactions just to differentiate?

internal/db/model_test.go
```diff
@@ -716,6 +718,44 @@ func (s *dbSuite) TestGetModelsByController(c *qt.C) {
...
+	rowLocked := make(chan struct{})
+	go func() {
+		_ = s.Database.Transaction(func(tx *db.Database) error {
+			model := dbmodel.Model{}
+			model.SetTag(mt)
+			err = tx.GetModelForUpdate(ctx, &model)
+			c.Check(err, qt.IsNil)
+
+			close(rowLocked)
+			<-ctx.Done()
```
ale8k: Just use a context here again for consistency with the rowLocked chan, looks kinda strange using ctx to co-ordinate channels here and above using rowLocked

internal/db/model.go
```diff
@@ -48,6 +46,37 @@ func (d *Database) GetModel(ctx context.Context, model *dbmodel.Model) (err erro
...
+	const op = errors.Op("db.GetModelForUpdate")
```
ale8k: Looking over this a few times... I'm not keen on having a GetModelForUpdate. I would've expected something more like...

```go
func (d *Database) ForUpdate(ctx context.Context) *Database {
	if forUpdate {
		db = db.Clauses(clause.Locking{Strength: "UPDATE"})
	}
}

// And somewhere in the code (pseudo)...
// db.ForUpdate().GetModel().UpdateModel()
```

internal/dbmodel/model.go
```diff
@@ -182,3 +182,29 @@ func (m Model) MergeModelSummaryFromController(modelSummaryFromController *jujup
 	modelSummaryFromController.UserAccess = access
 	return *modelSummaryFromController
 }
+
+// ProcessFailedMigration processes a failed migration by resetting the
+// MigrationMode to MigrationModeNone.
+func (m *Model) ProcessFailedMigration() {
```
ale8k: This is beautiful

internal/jimm/juju/model_poller.go
```diff
@@ -0,0 +1,145 @@
...
+
+	"github.com/canonical/jimm/v3/internal/dbmodel"
+	"github.com/canonical/jimm/v3/internal/errors"
+	"github.com/canonical/jimm/v3/internal/servermon"
+)
+
+// PollModels loops over models, contacting the respective controller
+// and checking, based on the model's migration mode, if the model exists.
+func (j *JujuManager) PollModels(ctx context.Context) (err error) {
+	const op = errors.Op("jimm.CleanupNotFoundModels")
```
ale8k: OpCode

internal/jimm/juju/model_poller.go
```diff
@@ -0,0 +1,145 @@
...
+	zapctx.Info(ctx, string(op))
+	durationObserver := servermon.DurationObserver(servermon.JimmMethodsDurationHistogram, string(op))
+	defer durationObserver()
+
+	// Step 1: Group models by controller
+	controllerModels := make(map[string][]*dbmodel.Model)
+	err = j.Database.ForEachModel(ctx, func(m *dbmodel.Model) error {
```
ale8k, line 31: Could we preload models on the controller here instead of looping over?

internal/jimm/juju/model_poller.go
```diff
@@ -0,0 +1,145 @@
...
+	"github.com/juju/zaputil/zapctx"
+	"go.uber.org/zap"
+
+	"github.com/canonical/jimm/v3/internal/dbmodel"
+	"github.com/canonical/jimm/v3/internal/errors"
+	"github.com/canonical/jimm/v3/internal/servermon"
+)
+
+// PollModels loops over models, contacting the respective controller
+// and checking, based on the model's migration mode, if the model exists.
```
ale8k, line 20: based on the model's migration mode, if the model exists. 
If the model exists in JIMM's database, but not on the controller,
it is deleted from JIMM's database.

internal/jimm/juju/model_poller.go
```diff
@@ -0,0 +1,145 @@
...
+		return
+	}
+	zapctx.Info(ctx, "model successfully migrated to controller", zap.String("model", m.UUID.String), zap.String("controller_name", controller.Name))
+}
+
+// checkModelExists checks if the model exists on the controller.
+// This performs eventual cleanup of models that have been deleted through
+// the API (since the deletion of a model is not immediate) and handles
+// cases where the model was deleted directly on the underlying controller.
+func (j *JujuManager) checkModelExists(ctx context.Context, api API, m *dbmodel.Model) {
```
ale8k: maybeCleanupModel would be a better name

internal/jimm/juju/model_poller.go
```diff
@@ -0,0 +1,145 @@
...
+			}
+		}
+	}
+	return nil
+}
+
+// checkModelMigratedInternal checks if the model has been migrated from
+// one controller managed by JIMM to another controller managed by JIMM.
+func (j *JujuManager) checkModelMigratedInternal(ctx context.Context, api API, m *dbmodel.Model) {
+	const op = errors.Op("jimm.checkModelMoved")
```
ale8k: Op

internal/jimm/juju/model_poller.go
```diff
@@ -0,0 +1,145 @@
...
+	zapctx.Info(ctx, string(op))
```
ale8k: zapctx.WithFields be better to always set model uuid

internal/jimm/juju/model_poller.go
```diff
@@ -0,0 +1,145 @@
...
+	// Check if the model has completed a migration.
+	// If modelInfo returns without an error, it definitely hasn't moved yet.
+	modelInfo := &jujuparams.ModelInfo{UUID: m.UUID.String}
+	err := api.ModelInfo(ctx, modelInfo)
+	if err == nil {
+		// If the migration end time is set, it means the model has
+		// failed to migrate otherwise we'd expect a redirect error.
+		if modelInfo.Migration.End != nil {
+			m.ProcessFailedMigration()
+			err = j.Database.UpdateModel(ctx, m)
```
ale8k: Can we inline these error calls, it'll reduce LoC and for me personally it is easier to read

internal/jimm/juju/model_poller.go
```diff
@@ -0,0 +1,145 @@
...
+			if err != nil {
+				zapctx.Error(ctx, "failed to update model after failed migration", zap.String("model", m.UUID.String), zap.Error(err))
```
ale8k: We can return an error in this function and have a single zapctx.Error in PollModels, there's no reason this couldn't return an error and the caller should define the behaviour of what to do in the event of an error. 

tldr bubble errors up please.

internal/jimm/juju/model_poller.go
```diff
@@ -0,0 +1,145 @@
...
+			}
+		}
+		return
+	}
+
+	// Expect a redirect error if the model successfully migrated.
```
ale8k, line 99: Lines 87 to 106 would be good in a fun called handleRedirectInfo or something similar with a godoc explaining why we're rummaging through the error to get the redirect info. This isn't super immediately clear why it must be done like this from reading and whilst I understand what you're doing here, it definitely looks nuanced and is worth a godoc.

Then it'd be good to pass that redirectInfo to another func called getNewControllerName with another godoc detailing why we use alias and what it really means to us using the alias. I'd expect this to just return a string.

ale8k (changes requested):
Some minor changes, but after that lgtm.

internal/db/model.go
kian99: Yes for sure.

internal/db/model.go
kian99: That looks cool but seems like a very different paradigm to our current setup.

internal/db/model_test.go
kian99: I'll use two channel rather then.

internal/jimm/juju/model_poller.go
kian99: Do you mean you want like `SELECT * CONTROLLERS` instead of this which is `SELECT * MODELS`?

internal/jimm/juju/model_poller.go
kian99: Sure

internal/jimm/juju/model_poller.go
kian99: I'd prefer to keep it inline and it can have an inline comment, it's short enough that keeping it all in one place helps.

internal/jimm/juju/model_poller.go
```diff
@@ -0,0 +1,158 @@
...
+			zapctx.Error(ctx, "cannot dial controller", zap.String("controller", controllerUUID), zap.Error(err))
+			continue
+		}
+		// Depending the model's migration mode, we either:
+		// - Check if the model exists (MigrationModeNone)
+		// - Check if the model has completed internal migration (MigrationModeMigrateInternal)
+		// - Do nothing if the model is in any other migration mode (MigrationModeImporting, MigrationModeExporting)
+		for _, m := range models {
+			ctx := zapctx.WithFields(ctx,
+				zap.String("modelOwner", m.OwnerIdentityName),
```
ale8k: We haven't got much consistency on label keys, I tend to use hyphens/snake and this is camel. What do you think about sticking with "model-owner" etc for the labels?

kian99: Yeah we don't have much consistency, I'll change it to model-owner, but I've seen camel case in some places.

internal/jimm/juju/model_poller.go
ale8k: More like
```go
	d.ForEachController(ctx, func(ccontroller *dbmodel.Controller) error {
		// Models not preloaded though
		controllerModels[ccontroller.Name] = ccontroller.Models
		return nil
	})
```

So additionally something like:
```go
 ForEachController(ctx context.Context, f func(*dbmodel.Controller) error, preloadModels bool)
```

And I think it'd read nicer as it just wants the models, for each controller.

internal/db/model.go
kian99: Added it. I kept the test and it continues to work. I also verified that the additional queries as a result of Preload() don't have a `FOR UPDATE` clause and they don't seem to. Fwiw the SQL queries generated from the test `TestGetModelForUpdate` can be seen here https://pastebin.ubuntu.com/p/Bc2qQ339pZ/

internal/jimm/juju/model_poller.go
kian99: I think it's pretty equivalent so I'll leave this one as is.

internal/db/modifiers.go
```diff
@@ -0,0 +1,24 @@
...
+//
+//		s.Database.Transaction(func(tx *db.Database) error {
+//					model := dbmodel.Model{}
+//					model.SetTag(mt)
+//					err = tx.ForUpdate().GetModel(ctx, &model)
+//					// Do some checks with the model
+//	                err = tx.UpdateModel(ctx, &model)
+//					return nil
+//			})
+func (d *Database) ForUpdate() *Database {
```
alesstimec, line 20: this one is a bit scary, because it's so easy to hold it wrong as it does not limit you to a single select, or a select at all..  but then again, we can always argue it's a "skill  issue"

internal/dbmodel/model.go
alesstimec: s/ProcessFailedMigration/MigrationFailed/ or SetMigrationFailed

internal/dbmodel/model.go
```diff
@@ -182,3 +182,29 @@ func (m Model) MergeModelSummaryFromController(modelSummaryFromController *jujup
...
+	m.MigrationMode = MigrationModeNone
+}
+
+// ProcessSuccessfulInternalMigration processes a successful internal migration
+// by setting the MigrationMode to MigrationModeNone and updating the ControllerID.
+func (m *Model) ProcessSuccessfulInternalMigration(controllerID uint) {
```
alesstimec: SetSuccessfulInternalMigration

internal/jimm/juju/model_poller.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+
+	var redirectInfo params.RedirectErrorInfo
+	err = params.Error{Info: errInfo}.UnmarshalInfo(&redirectInfo)
+	if err != nil {
+		return errors.E(op, fmt.Errorf("cannot unmarshal redirect error info: %w", err))
+	}
+
+	// We expect this controller will be known to JIMM.
+	controller := dbmodel.Controller{Name: redirectInfo.ControllerAlias}
```
alesstimec, line 120: please parse the `redirectInfo.ControllerTag` to get the controller UUID and use that.. i have no idea where this alias comes from and i'd be much more comfortable relying on UUID to identify the controller

internal/jimm/juju/model_poller.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+	zapctx.Info(ctx, "model successfully migrated to controller", zap.String("model", m.UUID.String), zap.String("controller_name", controller.Name))
+	return nil
+}
+
+// maybeCleanupModel checks if the model exists on the controller.
+// This performs eventual cleanup of models that have been deleted through
+// the API (since the deletion of a model is not immediate) and handles
+// cases where the model was deleted directly on the underlying controller.
+func (j *JujuManager) maybeCleanupModel(ctx context.Context, api API, m *dbmodel.Model) error {
```
alesstimec, line 140: cleanupDeletedModels?

alesstimec (approved):
LGTM, with a few comments. please use the controller UUID in the redirect error to identify the target controller

internal/dbmodel/model.go
kian99: Will go with `MigrationFailed`.

internal/dbmodel/model.go
kian99: To match the above, I'll go with `InternalMigrationSuccess`.

internal/jimm/juju/model_poller.go
kian99: As discussed offline, Juju's model manager facade doesn't return the controller tag,
https://github.com/juju/juju/blob/3.6/apiserver/common/modelmanagerinterface.go#L208
But model level facades do return the tag https://github.com/juju/juju/blob/3.6/apiserver/admin.go#L386
In internal/jimm we don't have a mechanism for calling model endpoints (i.e. /mode/uuid/api) so we'll stick with the controller alias.

internal/jimm/juju/model_poller.go
kian99: Went with Alex's suggestion for maybCleanupModel which I think is fine.

ale8k (approved):
lgtm now, ty for changes man!

PR #1671, opened 2025-08-11
feat: automatic model migration integration test

kian99:
## Description

This PR adds an integration test that performs an internal model migration between two controllers managed by JIMM.
~~This PR is a draft pending some ongoing work to make it easier to fetch a model's valid migration targets.~~

~~This change also updates the bash test runner to run up to 3 tests in parallel. While this affects the log output it will ensure out integration tests don't become a bottleneck due to the time spent starting Juju controllers. In the future I would like to see the logs redirected to specific files to each test and for these logs to be uploaded to the action's results page for inspection when a test fails. Ales mentioned using the `testscript` [package](https://pkg.go.dev/github.com/rogpeppe/go-internal/testscript) which would be a good way to still have bash tests but use Go's test runner to provide concurrency and other nice features. I didn't consider this approach so if the additions to the bash style runner are not desirable I can remove those and we can look at a better test runner in the future.~~
This didn't seem to work and it looks like the issue is with the Juju CLI. If one test is switching to a model and another test expects a specific model, that will cause issues. We can opt to always specify the model for any commands. There also looks like there could be issues bootstrapping multiple controllers at once. So this speed is probably best left to a separate PR.

Fixes: [JUJU-8362](https://warthogs.atlassian.net/browse/JUJU-8362)

## Engineering checklist
- [ ] Documentation updated
- [ ] Covered by unit tests

[JUJU-8362]: https://warthogs.atlassian.net/browse/JUJU-8362?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/integration-test.yaml
+++ b/.github/workflows/integration-test.yaml
@@ -12,11 +12,6 @@ jobs:
       - name: Checkout JIMM repo
         uses: actions/checkout@v4
 
-      - name: Setup tmate session
-        uses: canonical/action-tmate@main
-        with:
-          detached: true
-
       - name: Set up Docker Compose
         uses: docker/setup-compose-action@v1
 
--- a/.github/workflows/integration-test.yaml
+++ b/.github/workflows/integration-test.yaml
@@ -12,6 +12,11 @@ jobs:
       - name: Checkout JIMM repo
         uses: actions/checkout@v4
 
+      - name: Setup tmate session
+        uses: canonical/action-tmate@main
+        with:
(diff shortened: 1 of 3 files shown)


luci1900 (approved):
Nice to see the entire flow that my previous PR fits into.

SimoneDutto (approved):

PR #1673, opened 2025-08-11
feat: add ListMigrationTargets endpoint and tests

luci1900:
## Description

This endpoint takes a model as a parameter and returns all controllers that:
- aren't already where the model is running
- support the model's cloud
- are not a version older than the model's

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

### DIFF
--- a/internal/jimm/juju/jimm.go
+++ b/internal/jimm/juju/jimm.go
@@ -352,30 +352,22 @@ func (j *JujuManager) ListMigrationTargets(ctx context.Context, user *openfga.Us
 		return nil, errors.E(op, err)
 	}
 
-	var controllers []dbmodel.Controller
-	err = j.Database.ForEachController(ctx, func(c *dbmodel.Controller) error {
-		candidateVersion, err := version.Parse(c.AgentVersion)
-		if err != nil {
-			return err
-		}
-
-		hasRegion := false
-		for _, i := range c.CloudRegions {
-			if i.CloudRegion.ID == model.CloudRegion.ID {
-				hasRegion = true
-				break
-			}
-		}
-
-		if model.Controller.ID != c.ID &&
-			hasRegion &&
-			currentVersion.Compare(candidateVersion) <= 0 {
-			controllers = append(controllers, *c)
-		}
-		return nil
-	})
+	cloudRegion, err := j.Database.FindRegionByCloudName(ctx, model.CloudRegion.CloudName, model.CloudRegion.Name)
 	if err != nil {
 		return nil, errors.E(op, err)
+	}
+
+	var controllers []dbmodel.Controller
+	for _, ctl := range cloudRegion.Controllers {
+		candidateVersion, err := version.Parse(ctl.Controller.AgentVersion)
+		if err != nil {
+			return nil, errors.E(op, err)
+		}
+
+		if model.Controller.ID != ctl.Controller.ID &&
+			currentVersion.Compare(candidateVersion) <= 0 {
+			controllers = append(controllers, ctl.Controller)
+		}
 	}
 
 	return controllers, nil
--- a/internal/jimm/juju/jimm.go
+++ b/internal/jimm/juju/jimm.go
@@ -11,6 +11,7 @@ import (
 	"github.com/google/uuid"
 	jujuparams "github.com/juju/juju/rpc/params"
 	"github.com/juju/names/v5"
+	"github.com/juju/version/v2"
 	"github.com/juju/zaputil/zapctx"
 	"go.uber.org/zap"
 	"golang.org/x/sync/errgroup"
@@ -328,4 +329,54 @@ func (j *JujuManager) PrepareModelMigration(
 	}
 
 	return migrationToken, nil
+}
+
+// ListMigrationTargets returns the list of juju controllers that the given internal
+// model could be migrated to. This includes controllers that support the model's
+// cloud region and version, but excludes the controller the model is already on.
+func (j *JujuManager) ListMigrationTargets(ctx context.Context, user *openfga.User, modelTag names.ModelTag) ([]dbmodel.Controller, error) {
+	const op = errors.Op("jimm.ListMigrationTargets")
+
+	if !user.JimmAdmin {
+		return nil, errors.E(op, errors.CodeUnauthorized, "unauthorized")
+	}
+
+	var model dbmodel.Model
+	model.SetTag(modelTag)
+	if err := j.Database.GetModel(ctx, &model); err != nil {
+		return nil, errors.E(op, err)
+	}
+
+	currentVersion, err := version.Parse(model.Controller.AgentVersion)
+	if err != nil {
+		return nil, errors.E(op, err)
+	}
+
+	var controllers []dbmodel.Controller
+	err = j.Database.ForEachController(ctx, func(c *dbmodel.Controller) error {
+		candidateVersion, err := version.Parse(c.AgentVersion)
+		if err != nil {
+			return err
+		}
+
+		hasRegion := false
+		for _, i := range c.CloudRegions {
+			if i.CloudRegion.ID == model.CloudRegion.ID {
+				hasRegion = true
+				break
+			}
+		}
+
+		if model.Controller.ID != c.ID &&
+			hasRegion &&
+			currentVersion.Compare(candidateVersion) <= 0 {
+			controllers = append(controllers, *c)
+		}
+		return nil
+	})
+	if err != nil {
+		return nil, errors.E(op, err)
+	}
+
+	return controllers, nil
 }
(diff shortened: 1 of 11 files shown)


internal/jimm/juju/jimm.go
```diff
@@ -328,3 +329,43 @@ func (j *JujuManager) PrepareModelMigration(
...
+	if err := j.Database.GetModel(ctx, &model); err != nil {
+		return nil, errors.E(op, err)
+	}
+
+	var controllers []dbmodel.Controller
+	err := j.Database.ForEachController(ctx, func(c *dbmodel.Controller) error {
+		currentVersion, err := version.Parse(model.Controller.AgentVersion)
+		if err != nil {
+			return err
+		}
```
kian99: We should do this outside the loop since it won't change.

internal/jimm/juju/jimm.go
```diff
@@ -328,3 +329,43 @@ func (j *JujuManager) PrepareModelMigration(
...
+
+		candidateVersion, err := version.Parse(c.AgentVersion)
+		if err != nil {
+			return err
+		}
+
+		if model.Controller.ID != c.ID &&
+			model.Controller.CloudRegion == c.CloudRegion &&
```
kian99: Reading this I thought something was off and reading the docstring for `CloudRegion` I see it says,
>CloudRegion is the name of the cloud region which is hosting this controller.

So this is just the region where the controller is deployed but there is another field `CloudRegions` with docstring,
> CloudRegions is the set of cloud-regions that are available on this controller.

That represents what we want.

internal/jimm/juju/jimm_test.go
```diff
@@ -927,3 +927,100 @@ func TestPrepareMigration_MultipleCalls(t *testing.T) {
...
+		jimmAdmin: true,
+		expectedControllers: []dbmodel.Controller{
+			env.Controller("test3").DBObject(c, j.Database),
+		},
+	}, {
+		about:         "add-model user can not list controllers",
+		user:          env.User("bob@canonical.com").DBObject(c, j.Database),
+		expectedError: "unauthorized",
+	}, {
+		about:         "user withouth access rights cannot list controllers",
```
kian99:
```suggestion
		about:         "user without access rights cannot list controllers",
```

internal/jujuapi/jimm_test.go
```diff
@@ -924,3 +926,62 @@ func (s *jimmSuite) TestPrepareModelMigration(c *gc.C) {
...
+		ctl.Addresses = append(ctl.Addresses, []jujuparams.HostPort{{
+			Address: jujuparams.FromMachineAddress(hp.MachineAddress),
+			Port:    hp.Port(),
+		}})
+	}
+	err := s.JIMM.JujuManager().AddController(context.Background(), s.AdminUser, ctl, ctlCreds)
+	c.Assert(err, gc.Equals, nil)
+
+	// Add model that could migrate to target
+	mt := s.AddModel(
```
kian99: You'll need to call this before you add a new controller otherwise the test becomes non-deterministic in that the logic in `AddModel` will randomly pick a supported controller so just run `AddModel` when there is only 1 controller then add the other. I believe that's why the test is failing in CI with error,
>... mismatch at [0].Name: unequal; obtained "controller-1"; expected "controller-2"

pkg/api/params/params.go
```diff
@@ -561,6 +561,12 @@ type PrepareModelMigrationResponse struct {
 	Token string `json:"token" yaml:"token"`
 }
 
+// ListMigrationTargetsRequest holds the model to query controllers for
```
kian99:
```suggestion
// ListMigrationTargetsRequest holds the model to query controllers that are 
// valid targets for an internal model migration.
```

pkg/api/params/params.go
```diff
@@ -561,6 +561,12 @@ type PrepareModelMigrationResponse struct {
...
+type ListMigrationTargetsRequest struct {
+	// ModelTag holds the tag of the model that could be migrated
```
kian99:
```suggestion
	// ModelTag holds the tag of the model.
```

kian99 (commented):
Looks good, nice work on this.
Mainly just a tweak needed in the logic for determining valid target controllers.

internal/jujuapi/jimm_test.go
luci1900: I see. In my defence, I find non-deterministic deploys terrifying.

pkg/api/params/params.go
luci1900: Ah, lost the rest of this while moving things around I think.

internal/jujuapi/jimm_test.go
kian99: A very valid opinion

pkg/api/params/params.go
luci1900: That seems to add no additional info over the field and type. Is that typical?

internal/jimm/juju/jimm.go
luci1900: I'm not sure I understand the distinction. The controller can run on one region, but manage workloads on several other regions too?

kian99: Exactly that

internal/jimm/juju/jimm.go
```diff
@@ -328,3 +329,51 @@ func (j *JujuManager) PrepareModelMigration(
 
 	return migrationToken, nil
 }
+
+// ListMigrationTargets returns a list of controllers the model could be migrated to
```
alesstimec: also mention that it returns controller of a higher version. and end the godoc with a full stop.

internal/jujuapi/jimm.go
```diff
@@ -574,6 +576,29 @@ func (r *controllerRoot) PrepareModelMigration(ctx context.Context, args apipara
...
+// model could be migrated to.
+func (r *controllerRoot) ListMigrationTargets(ctx context.Context, req apiparams.ListMigrationTargetsRequest) (apiparams.ListControllersResponse, error) {
+	const op = errors.Op("jujuapi.ListMigrationTargets")
+
+	mt, err := names.ParseModelTag(req.ModelTag)
+	if err != nil {
+		return apiparams.ListControllersResponse{}, errors.E(op, err, errors.CodeBadRequest)
+	}
+
+	dbControllers, err := r.jimm.JujuManager().ListMigrationTargets(ctx, r.user, mt)
```
alesstimec, line 590: check the error first, then create the `controllersInfo` slice below

internal/jimm/juju/jimm_test.go
```diff
@@ -927,3 +927,129 @@ func TestPrepareMigration_MultipleCalls(t *testing.T) {
...
+		expectedControllers: []dbmodel.Controller{
+			env.Controller("test3").DBObject(c, j.Database),
+			env.Controller("test5").DBObject(c, j.Database),
+		},
+	}, {
+		about:         "add-model user can not list controllers",
+		user:          env.User("bob@canonical.com").DBObject(c, j.Database),
+		expectedError: "unauthorized",
+	}, {
+		about:         "user without access rights cannot list controllers",
```
alesstimec, line 1047: add test cases for:
- non existing model
- model of the newest possible version such that no controllers are returned

internal/jimm/juju/jimm.go
luci1900: Should I also mention that it won't return the controller that the model is currently already on? And also that the cloud region must match?

internal/jimm/juju/jimm.go
```diff
@@ -328,3 +329,53 @@ func (j *JujuManager) PrepareModelMigration(
...
+
+	if !user.JimmAdmin {
+		return nil, errors.E(op, errors.CodeUnauthorized, "unauthorized")
+	}
+
+	var model dbmodel.Model
+	model.SetTag(modelTag)
+	if err := j.Database.GetModel(ctx, &model); err != nil {
+		return nil, errors.E(op, errors.CodeNotFound, fmt.Sprintf("model %q not found", modelTag))
+	}
```
kian99: Technically this is not always a not found error, you'd need to check if the error code coming from the database is a NotFound error otherwise return the error as is.

kian99 (approved):
LGTM, nice work!

internal/jimm/juju/jimm.go
luci1900: Ah, you're right. We were just looking at this. I'll fix.

luci1900: Indeed `dbError` does the NotFound extraction by itself. I'll leave it as that.

internal/jimm/juju/jimm.go
```diff
@@ -328,3 +329,53 @@ func (j *JujuManager) PrepareModelMigration(
...
+		return nil, errors.E(op, err)
+	}
+
+	currentVersion, err := version.Parse(model.Controller.AgentVersion)
+	if err != nil {
+		return nil, errors.E(op, err)
+	}
+
+	var controllers []dbmodel.Controller
+	err = j.Database.ForEachController(ctx, func(c *dbmodel.Controller) error {
```
alesstimec: i wonder if it would be slightly more efficient to look at the current model, get it's cloud region (FindRegionByCloudName), then loop through the list of controllers for that cloud region.

alesstimec (approved):
LGTM with a slightly more efficient approach proposal (i think)

kian99 (approved):

PR #1675, opened 2025-08-15
fix: replace usages of ScanRows to avoid O(n) queries

luci1900:
## Description

Several `Database` methods were making O(n) number of queries due to their usage of `ScanRows` and separately issuing a get query in a loop. This would've likely led to high latency in pathological cases and thus poor UX.

I've replaced most such usage with `Find`, which is able to use preloads/joins. I've also added preloads where they were missing. This should be no less efficient than the previous method, which was also spilling the query result into a slice before iterating over it.

Affected methods:
- ForAllControllers (previous PR)
- ForEachCloudCredential
- ForEachModel
- ListResources

ForEachAuditLogEntry was not changed to avoid potentially high memory usage.

Fix for https://warthogs.atlassian.net/browse/JUJU-8348

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

### DIFF
--- a/internal/db/model.go
+++ b/internal/db/model.go
@@ -159,29 +159,15 @@ func (d *Database) ForEachModel(ctx context.Context, f func(m *dbmodel.Model) er
 
 	db := d.DB.WithContext(ctx)
 	db = preloadModel("", db)
-	rows, err := db.Model(&dbmodel.Model{}).Rows()
-	if err != nil {
-		return errors.E(op, err)
+
+	var models []dbmodel.Model
+	if err := db.Find(&models).Error; err != nil {
+		return errors.E(op, dbError(err))
 	}
-	defer rows.Close()
-	for rows.Next() {
-		var m dbmodel.Model
-		if err := db.ScanRows(rows, &m); err != nil {
-			return errors.E(op, err)
-		}
-		// ScanRows does not use the preloads added on L141, therefore
-		// we need to fetch each model to load the associated
-		// fields otherwise the only populated fields will be association
-		// IDs.
-		if err := d.GetModel(ctx, &m); err != nil {
-			return errors.E(op, err)
-		}
+	for _, m := range models {
 		if err := f(&m); err != nil {
 			return err
 		}
-	}
-	if err := rows.Err(); err != nil {
-		return errors.E(op, dbError(err))
 	}
 	return nil
 }
--- a/internal/db/resource.go
+++ b/internal/db/resource.go
@@ -86,19 +86,10 @@ func (d *Database) ListResources(ctx context.Context, limit, offset int, namePre
 	if err != nil {
 		return nil, err
 	}
-	rows, err := query.Rows()
-	if err != nil {
-		return nil, err
-	}
-	defer rows.Close()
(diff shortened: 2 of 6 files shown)


internal/db/model.go
```diff
@@ -159,30 +159,16 @@ func (d *Database) ForEachModel(ctx context.Context, f func(m *dbmodel.Model) er
 
 	db := d.DB.WithContext(ctx)
 	db = preloadModel("", db)
-	rows, err := db.Model(&dbmodel.Model{}).Rows()
-	if err != nil {
-		return errors.E(op, err)
+
+	var models []dbmodel.Model
```
ale8k, line 163: This needs the test updating to check for preloads right?

luci1900: Testing is tricky, yes. It's what I'm looking at now. Hence the draft PR.

luci1900: I've added testing for preload in the existing test, which does fail with the old method if I remove the `GetModel` call.

I'm not convinced nothing else depends on the old behaviour, though. I'm trying to think of a  more thorough way to test.

luci1900: It's also surprising that `GetModelsByUUID` was already using `Find` and thus with correct preloads. And it's just next to it.

luci1900: Interesting, all of the `ForEach...` database methods use this `ScanRows` pattern instead of `Find`.

internal/db/resource.go
kian99, line 1: I guess no test changes for this one because we weren't pre-loading anything anyway?

kian99 (approved):
lgtm, thanks for this

internal/db/resource.go
luci1900: Exactly, there are no foreign keys to preload on this db model.

SimoneDutto (commented):
LGTM, but i have one point maybe it is worth discussing.

I agree this approach is preferable for everything but `ForEachAuditLogEntry`, because the other tables are very small.
For example, let's say we are streaming all audit logs to the client, if we use this new approach we would allocate a lot of memory during the query. Using `ScanRows` is less efficient but it allocates one row at the time.

Maybe this is a non-issue since we should always fetch a fixed amount of audit logs,, but I wanted to bring it up regardless.
We tend to use `ForEach<table>` methods for many things and with this approach we need to be careful about memory consumption.

luci1900:
> LGTM, but i have one point maybe it is worth discussing.
> 
> I agree this approach is preferable for everything but `ForEachAuditLogEntry`, because the other tables are very small. For example, let's say we are streaming all audit logs to the client, if we use this new approach we would allocate a lot of memory during the query. Using `ScanRows` is less efficient but it allocates one row at the time.
> 
> Maybe this is a non-issue since we should always fetch a fixed amount of audit logs,, but I wanted to bring it up regardless. We tend to use `ForEach<table>` methods for many things and with this approach we need to be careful about memory consumption.

All of the other methods collect the data into a slice anyway.

But you're right that audit logs don't do that. I can change that one back, it's unlikely to be changed to make O(n) queries anyway.

SimoneDutto (approved):
Ty for the changes

PR #1677, opened 2025-08-19
Fix juju cli store 

ale8k:
## Description
The original implementation assumed the downloaded file with the binary to begin with and this assumption was incorrect. Juju tars and compresses the artefacts alongside another another binary called "juju-metadata".

This pull request has adjusted the cli store slightly to do the following:
1. Decompress the stream (using the xz package - perhaps this isn't the best one looking at it's README)
2. Unpack the tar looking for the "./juju" header for the juju binary file
3. Chmod the binary file 

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

### DIFF
--- a/internal/jujuclistore/store.go
+++ b/internal/jujuclistore/store.go
@@ -3,6 +3,7 @@
 package jujuclistore
 
 import (
+	"archive/tar"
 	"bytes"
 	"context"
 	"fmt"
@@ -11,6 +12,7 @@ import (
 	"math/rand/v2"
 	"net/http"
 	"os"
+	"path/filepath"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -19,6 +21,9 @@ import (
 	jujuerrors "github.com/juju/errors"
 	"github.com/juju/retry"
 	"github.com/juju/version/v2"
+	"github.com/juju/zaputil/zapctx"
+	"github.com/ulikunitz/xz"
+	"go.uber.org/zap"
 )
 
 const timeoutRequestDuration = 5 * time.Minute
@@ -29,7 +34,10 @@ const launchPadURL = "https://launchpad.net/juju"
 // launchPadTemplate is the template for constructing the download URL for Juju binaries.
 const launchPadTemplate = "{{.BaseURL}}/{{.VersionWithMinor}}/{{.VersionWithPatch}}/+download/juju-{{.VersionWithPatch}}-{{.Os}}-{{.Arch}}.tar.xz"
 
-var retryRequestError = jujuerrors.New("retry request error")
+var (
+	retryRequestError error = jujuerrors.New("retry request error")
+	maxExtractSize    int64 = 200 * 1024 * 1024
+)
 
 // Config holds the configuration for the Juju binary fetcher.
 type Config struct {
@@ -122,7 +130,6 @@ func (p *jujuCLIStore) Get(ctx context.Context, spec JujuBinarySpec) (*Binary, e
 	if err != nil {
 		return nil, jujuerrors.Annotatef(err, "invalid version %q", spec.Version)
 	}
-
 	err = p.template.Execute(&buf, map[string]string{
 		"BaseURL":          p.config.BaseURL,
 		"VersionWithMinor": fmt.Sprintf("%d.%d", v.Major, v.Minor),
@@ -134,6 +141,15 @@ func (p *jujuCLIStore) Get(ctx context.Context, spec JujuBinarySpec) (*Binary, e
 		return nil, err
 	}
 	url := buf.String()
+	zapctx.Debug(
+		ctx,
+		"getting Juju binary",
+		zap.String("version", spec.Version),
+		zap.String("os", spec.Os),
+		zap.String("arch", spec.Arch),
+		zap.String("url", url),
+	)
+
 	// worst case the lock will be help for the duration of the download
 	// of the binary. For now it is acceptable because we support bootstrapping
 	// one controller at a time.
@@ -144,11 +160,12 @@ func (p *jujuCLIStore) Get(ctx context.Context, spec JujuBinarySpec) (*Binary, e
 		binary.referenceCount.Add(1) // Increment reference count
 		return binary, nil
 	}
-	err = p.freeEntry(ctx)
+	err = p.freeEntry()
 	if err != nil {
 		return nil, err
 	}
 	var file *os.File
+
 	err = retry.Call(retry.CallArgs{
 		Func: func() error {
 			file, err = p.downloadFile(ctx, url)
@@ -169,6 +186,7 @@ func (p *jujuCLIStore) Get(ctx context.Context, spec JujuBinarySpec) (*Binary, e
 	if err != nil {
 		return nil, err
 	}
+
 	binary = &Binary{
 		FullPath: file.Name(),
 	}
@@ -180,7 +198,7 @@ func (p *jujuCLIStore) Get(ctx context.Context, spec JujuBinarySpec) (*Binary, e
 // freeEntry checks if the entries map has reached the maximum number of entries.
 // If it has, it deletes a random entry from the map.
 // It should be called when the entries map is locked to ensure thread safety.
-func (p *jujuCLIStore) freeEntry(ctx context.Context) error {
+func (p *jujuCLIStore) freeEntry() error {
 	if len(p.entries) < p.config.MaxEntries {
 		return nil
 	}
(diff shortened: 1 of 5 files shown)


internal/jujuclistore/store.go
```diff
@@ -29,7 +34,10 @@ const launchPadURL = "https://launchpad.net/juju"
 // launchPadTemplate is the template for constructing the download URL for Juju binaries.
 const launchPadTemplate = "{{.BaseURL}}/{{.VersionWithMinor}}/{{.VersionWithPatch}}/+download/juju-{{.VersionWithPatch}}-{{.Os}}-{{.Arch}}.tar.xz"
 
-var retryRequestError = jujuerrors.New("retry request error")
+var (
+	retryRequestError error = jujuerrors.New("retry request error")
+	maxExtractSize    int64 = 200 * 1024 * 1024
```
alesstimec, line 39: where did we get this number?

internal/jujuclistore/store.go
```diff
@@ -225,21 +243,72 @@ func (p *jujuCLIStore) downloadFile(ctx context.Context, downloadUrl string) (*o
...
+			continue
+		}
+
+		// Create a temp file for the binary
+		f, err := os.CreateTemp(p.config.Dir, "juju-*")
+		if err != nil {
+			return nil, err
+		}
+
+		if _, err := io.Copy(f, limitedReader); err != nil {
```
alesstimec, line 290: shouldn't we be creating a `limitedReader` at this point (after calling `tarReader.Next()`)?

internal/jujuclistore/store.go
```diff
@@ -225,21 +243,72 @@ func (p *jujuCLIStore) downloadFile(ctx context.Context, downloadUrl string) (*o
...
+			f.Close()
+			os.Remove(f.Name())
+			return nil, err
+		}
+
+		if err := f.Chmod(0755); err != nil {
```
alesstimec, line 296: `700`?

internal/jujuclistore/store.go
```diff
@@ -225,21 +243,72 @@ func (p *jujuCLIStore) downloadFile(ctx context.Context, downloadUrl string) (*o
...
+		binaryFile = f
+		break
+	}
+
+	if binaryFile == nil {
+		return nil, fmt.Errorf("juju binary not found in archive")
 	}
-	return file, nil
+
+	if err := binaryFile.Close(); err != nil {
```
alesstimec, line 309: can't we close the file right after writing to it?

internal/jujuclistore/store.go
```diff
@@ -225,21 +243,72 @@ func (p *jujuCLIStore) downloadFile(ctx context.Context, downloadUrl string) (*o
...
+		return nil, jujuerrors.Annotatef(err, "failed to create xz reader for %s", downloadUrl)
 	}
-	_, err = io.Copy(file, resp.Body)
-	if err != nil {
-		file.Close()
-		os.Remove(file.Name())
-		return nil, err
+	tarReader := tar.NewReader(xzReader)
+	// 200mb - limit download size should malicious actor send large file
+	// and destroy jimm's memory with decompression bomb.
```
kian99: Sounds like the plot to an action movie

internal/jujuclistore/store.go
```diff
@@ -180,7 +198,7 @@ func (p *jujuCLIStore) Get(ctx context.Context, spec JujuBinarySpec) (*Binary, e
 // freeEntry checks if the entries map has reached the maximum number of entries.
 // If it has, it deletes a random entry from the map.
 // It should be called when the entries map is locked to ensure thread safety.
-func (p *jujuCLIStore) freeEntry(ctx context.Context) error {
+func (p *jujuCLIStore) freeEntry() error {
```
kian99, line 201: Imo nothing wrong with having a context on a method even if it's not used.
But no need to change it back.

internal/jujuclistore/store.go
kian99: +1

internal/jujuclistore/store_test.go
```diff
@@ -12,19 +14,120 @@ import (
...
+	jujuSpec := JujuBinarySpec{
+		Version: "3.6.2",
+		Os:      "linux",
+		Arch:    "amd64",
+	}
+
+	// Set an extract size 1 byte smaller than the content size
+	// to trigger the decompression bomb protection
+	c.Patch(&maxExtractSize, int64(9))
+	tenBytesContent := []byte("123456789therestofthiswillbemissing") // We'll expect the 0 to missing
```
kian99: What 0?

internal/jujuclistore/store_test.go
```diff
@@ -41,7 +144,8 @@ func TestStore(t *testing.T) {
 	c.Assert(err, qt.IsNil)
 	content, err := io.ReadAll(file)
 	c.Assert(err, qt.IsNil)
-	c.Assert(string(content), qt.Equals, "test content")
+	// We expect the content to be truncated to the maxExtractSize
+	c.Assert(string(content), qt.Equals, "123456789")
```
kian99, line 148: This seems like it will be hard to debug if there is no error

internal/jujuclistore/store_test.go
```diff
@@ -12,19 +14,120 @@ import (
...
+	c := qt.New(t)
+	jujuSpec := JujuBinarySpec{
+		Version: "3.6.2",
+		Os:      "linux",
+		Arch:    "amd64",
+	}
+
+	// Set an extract size 1 byte smaller than the content size
+	// to trigger the decompression bomb protection
+	c.Patch(&maxExtractSize, int64(9))
```
kian99, line 123: Can we make the maxExtractSize a field we pass in instead? I can't remember for sure but patching might make it hard/impossible to run tests in parallel.

internal/jujuclistore/store.go
```diff
@@ -225,21 +243,72 @@ func (p *jujuCLIStore) downloadFile(ctx context.Context, downloadUrl string) (*o
...
-	_, err = io.Copy(file, resp.Body)
-	if err != nil {
-		file.Close()
-		os.Remove(file.Name())
-		return nil, err
+	tarReader := tar.NewReader(xzReader)
+	// 200mb - limit download size should malicious actor send large file
+	// and destroy jimm's memory with decompression bomb.
+
+	limitedReader := io.LimitReader(tarReader, maxExtractSize)
```
SimoneDutto, line 262: why do you need the limit reader? `io.Copy` will always copy to the `os.File` in chunks, the tar reader i would guess operates in chunks as well, so it shouldn't be able to OOM jimm.

The only problem I could see is disk-wise but not memory wise.

SimoneDutto (commented):
a question

internal/jujuclistore/store.go
ale8k: I checked the size of the CLI (80mb) and went double & rounded to nearest 100ed. It's a "roughly" safe number.

internal/jujuclistore/store.go
ale8k: Do you mean limit just the juju binary and not the entire stream?

internal/jujuclistore/store.go
ale8k: I think it is harmless to leave it readable and executable by others. It also helps QA in future.

internal/jujuclistore/store.go
ale8k: We could but I feel this is clearer because we can see something has run, the file isn't nil, and we're closing it.

internal/jujuclistore/store_test.go
ale8k: My bad I used a 0 originally and it was confusing, so switched to a string.

internal/jujuclistore/store_test.go
ale8k: It's only a small test which is clear in the name that it is limiting read sizes. But I agree it may be confusing.

internal/jujuclistore/store_test.go
ale8k: We're already patching in many places, can we visit that then should we ever want to run in parallel?

internal/jujuclistore/store.go
ale8k: The issue is that A^10 can be compressed much easier than ABC^10, and upon decompression the size can exponentially grow. So, given some malicious attacker knew our tar structure and the rough size of a binary file, they could fill it with a text file full of A's.

internal/jujuclistore/store.go
alesstimec: yes

ale8k: No no we want to limit the entire thing otherwise someone could still just stick random files of duplicated characters and our linter will notice.

internal/jujuclistore/store.go
alesstimec: i think this should be configurable via an env var.

internal/jujuclistore/store.go
SimoneDutto: I'm not sure I understand, do you have an example/some articles about this?

If io.Copy is always copying in chunks, how can we OOM Jimm? Even if the decompressed file is 100x the compressed file, it should always be copied in chunks by io.Copy

alesstimec (approved):
looks ok, but i would propose making the max size configurable.. fine for a folloup, but please create a jira card

internal/jujuclistore/store.go
ale8k: I made you a quick example
```go
func TestDecompressionBombExample(t *testing.T) {
	c := qt.New(t)

	// This is an example of a decompression bomb that is 100 times larger than the original content.
	// The content is "im a decompression bomb" repeated 100 times.
	contentDuplicated := []byte("aaaaaaaaaaaaaaaaaaaa") // 20 bytes of 'a'
	// The decompression bomb will be 100 times larger than the original content.
	bombContent := bytes.Repeat(contentDuplicated, 100)
	c.Assert(bombContent, qt.HasLen, 2000) // The original content is 20 bytes, so the bomb content should be 2000 bytes
	// archive is of size
	bombArchive := makeTarXz(t, "bomb", bombContent)
	c.Assert(len(bombArchive), qt.Equals, 144) // The compressed size is 144

	// Same test again, but using the alphabet up to 20 bytes
	contentAlphabet := []byte("abcdefghijklmnopqrst") // 26 bytes of alphabet
	// The decompression bomb will be 100 times larger than the original content.
	bombContentAlphabet := bytes.Repeat(contentAlphabet, 100)
	c.Assert(bombContentAlphabet, qt.HasLen, 2000) // The original content is 26 bytes, so the bomb content should be 2600 bytes
	// archive is of size
	bombArchiveAlphabet := makeTarXz(t, "bomb", bombContentAlphabet)
	c.Assert(len(bombArchiveAlphabet), qt.Equals, 164) // The compressed size is 164
	

	// ^ This is a compression bomb.
}
```

SimoneDutto: i get the compression bomb part, i'm arguing that's impossible to OOM jimm because you would always process this in chunks, both the decrompression and the copying to the file

ale8k: My math is terrible but AI suggested explained an 80mb file could be compressed into:


Content type | Original | Estimated compressed (xz)
-- | -- | --
Highly repetitive (aaaa) | 80 MB | 5–6 MB
Semi-repetitive (dictionary) | 80 MB | 10–20 MB
Unique random words (high-entropy) | 80 MB | 70–80 MB

ale8k: I'm not worried about OOM? I'm worried about disk?

internal/jujuclistore/store.go
ale8k: I'll make card once merged.

SimoneDutto (approved):
lgtm

kian99:
On the limitReader I left a comment in one of the tests where the test showed that if the file was larger than the limit, it would be truncated instead of returning an error. That will be really hard to debug if the Juju file becomes larger than our limit as the binary will decompress but be corrupt.

The concern around a malicious file is valid, but we're only addressing one small problem by verifying the file size. The malicious file could be a bitcoin miner, it could create a reverse shell, etc. A lot of things worse than growing the disk space.

Ales mentioned a good idea that we can verify the hash of the compressed file we download from Launchpad. 
It's not exactly clear how we would add the hash to JIMM. We could add a `jaas` command to add juju binary filenames to JIMM with user providing the hash values (can also be found on Launchpad).

In any case, for this PR I wouldn't do the hash verification, but I would suggest removing the limit reader as that also solves the problem with `c.Patch`.

ale8k:
> On the limitReader I left a comment in one of the tests where the test showed that if the file was larger than the limit, it would be truncated instead of returning an error. That will be really hard to debug if the Juju file becomes larger than our limit as the binary will decompress but be corrupt.
> 
> The concern around a malicious file is valid, but we're only addressing one small problem by verifying the file size. The malicious file could be a bitcoin miner, it could create a reverse shell, etc. A lot of things worse than growing the disk space.
> 
> Ales mentioned a good idea that we can verify the hash of the compressed file we download from Launchpad. It's not exactly clear how we would add the hash to JIMM. We could add a `jaas` command to add juju binary filenames to JIMM with user providing the hash values (can also be found on Launchpad).
> 
> In any case, for this PR I wouldn't do the hash verification, but I would suggest removing the limit reader as that also solves the problem with `c.Patch`.

Totally agree that some verification would be good. Checksum verification would be appropriate here but yeah agreed we'd need it on Launchpad (Maybe Launchpad already produces checksums??). And it isn't necessarily just a malicious file, it is space in general too and currently we don't have a "maxDiskSizeForBinaryCache" kind of setting. If we face issues using it, I'll go ahead and pop that in. But until then I'd prefer we leave the limit reader?

PR #1678, opened 2025-08-19
Update bootstrap start params

ale8k:
## Description
Based on: #1677 
This PR simply updates the start params so that the subsequent PRs can reference these changes.

## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests

### DIFF
--- a/pkg/api/params/params.go
+++ b/pkg/api/params/params.go
@@ -626,8 +626,8 @@ type BootstrapStartParams struct {
 	// Flags hold modifiers for the bootstrap job.
 	Flags BootstrapFlags `json:"flags"`
 
-	// CLIVersion is the version of the Juju CLI that is being used.
-	CLIVersion string `json:"cli-version"`
+	// ControllerVersion is the version of the controller to be bootstrapped.
+	ControllerVersion string `json:"controller-version"`
 }
 
 // BootstrapStartResponse holds the response for starting
--- a/pkg/api/params/params.go
+++ b/pkg/api/params/params.go
@@ -5,6 +5,7 @@ package params
 import (
 	"time"
 
+	jujucloud "github.com/juju/juju/cloud"
 	jujuparams "github.com/juju/juju/rpc/params"
 )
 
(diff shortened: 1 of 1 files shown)


pkg/api/params/params.go
```diff
@@ -614,10 +615,21 @@ type BootstrapStartParams struct {
 	CloudName string `json:"cloud-name"`
 	// RegionName specifies the target region for the controller.
 	RegionName string `json:"region-name"`
+	// Cloud holds the cloud definition that will be used to bootstrap the controller.
+	Cloud jujuparams.Cloud `json:"cloud,omitempty"`
+	// Credential contains the cloud credential and its tag, this credential will be used against the
+	// the cloud provided to bootstrap the controller.
+	// We're using jujucloud.CloudCredential as there's an API client func using this type to add clouds.
+	// Seen here: api/client/cloud/cloud.go L271.
```
kian99: Referencing the line number is quite fragile as it will likely change and I'm not sure which branch to look at.

SimoneDutto: +1

pkg/api/params/params.go
```diff
@@ -614,10 +615,19 @@ type BootstrapStartParams struct {
...
+	// the cloud provided to bootstrap the controller.
+	Credential jujucloud.CloudCredential `json:"credential"`
+
 	// ControllerName specifies the name of the controller as recorded in JIMM.
 	ControllerName string `json:"controller-name"`
 	// Flags hold modifiers for the bootstrap job.
 	Flags BootstrapFlags `json:"flags"`
+
+	// CLIVersion is the version of the Juju CLI that is being used.
+	CLIVersion string `json:"cli-version"`
```
kian99: The name of this parameter is odd, is it trying to say that this is the version of controller that will be bootstrapped? If so, can we just call it `controller-version`? The fact that we use the CLI is an implementation detail.

ale8k: Can an agent version be different from the CLI version? I haven't seen it personally but if it can't then I can rename yeah.

kian99: `juju bootstrap --help` says,
```
The agent version can be specified a simple numeric version, e.g. 2.2.4.

For example, at the time when 2.3.0, 2.3.1 and 2.3.2 are released and your
agent stream is 'released' (default), then a 2.3.1 client can bootstrap:
    * 2.3.0 controller by running '... bootstrap --agent-version=2.3.0 ...';
    * 2.3.1 controller by running '... bootstrap ...';
    * 2.3.2 controller by running 'bootstrap --auto-upgrade'.
```
But even if it couldn't it makes more sense to rename this to `controller-version` if that's what it's doing.

luci1900: I'd agree on `controller-version`, but saying in the comment that it's implicitly also the CLI version.

alesstimec: +1 on controller version.. the cli version is an implementation detail.. so happens we'll always use CLI of the same version..

kian99 (approved):

alesstimec (approved):

PR #1679, opened 2025-08-19
Update bootstrap start jujuapi method

ale8k:
## Description
Based on: #1678

Updates the params to accept a cloud, a credential and the cli version. You'll notice we use `jujucloud.CloudCredential` within our params to pass the credential. This decision was made because Juju does the same. But, they send a "TaggedCredential", containing the `jujucloud.CloudCredential`. We just need the credential so I've omitted the wrapper. 

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests
- [ ] Covered by integration tests

PR #1680, opened 2025-08-19
Implement StartBootstrap

ale8k:
## Description
Based on: #1679

This PR implements StartBootstrap. It also updates the params to just use CLI version and include the cloud and credential. 

It also fixes the OS set for the binary downloaded and fixes where we didn't log the error line (so the CLI looked like it just failed randomly in Q/A). 

I've intentionally omitted writing the test for StartBootstrap as the vast majority is in BootstrapJob. 

There's a few todos scattered around that I'm going to create cards for following this PR. So please ignore them.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

### DIFF
--- a/internal/jimm/bootstrap/bootstrap.go
+++ b/internal/jimm/bootstrap/bootstrap.go
@@ -7,6 +7,8 @@ package bootstrap
 import (
 	"context"
 	"fmt"
+	"os"
+	"runtime"
 	"time"
 
 	"github.com/google/uuid"
@@ -33,6 +35,11 @@ var (
 	}
 )
 
+const (
+	bootstrapJobType     = "bootstrap"
+	maxBootstrapDuration = 60 * time.Minute
+)
+
 // Store defines the store methods required by the manager.
 type Store interface {
 	QueryBootstrapLog(ctx context.Context, jobId uuid.UUID, offset int) (loggies []string, nextOffsetValue int, err error)
@@ -49,9 +56,10 @@ type Store interface {
 type JobTracker interface {
 	// GetJob retrieves a job entry by its ID.
 	GetJob(ctx context.Context, jobId uuid.UUID) (dbmodel.JobTrackerEntry, error)
-
 	// StopJob stops a job by its ID.
 	StopJob(ctx context.Context, jobId uuid.UUID) error
+	// Run runs a new job and returns the job ID.
+	Run(ctx context.Context, jobType string, job jobtracker.JobFunc, maxDuration time.Duration) (uuid.UUID, error)
 }
 
 // JujuManager defines the juju manager methods required by the job.
@@ -65,38 +73,42 @@ type BinaryStore interface {
 }
 
 type bootstrapManager struct {
-	authSvc *openfga.OFGAClient
-
-	store       Store
-	jobtracker  JobTracker
-	jujuManager JujuManager
-	binaryStore BinaryStore
+	store                     Store
+	tracker                   JobTracker
+	jujuManager               JujuManager
+	binaryStore               BinaryStore
+	jimmWellknownJWKSEndpoint string
 }
 
 // NewBootstrapManager creates a new BootstrapManager instance.
-// TODO(ale8k): Remove authSvc later, it isn't used.
 func NewBootstrapManager(
-	authSvc *openfga.OFGAClient,
 	store Store,
 	jobtracker JobTracker,
 	jujuManager JujuManager,
 	binaryStore BinaryStore,
+	jimmWellknownJWKSEndpoint string,
 ) (*bootstrapManager, error) {
 	if store == nil {
 		return nil, errors.E("store cannot be nil")
 	}
-	if authSvc == nil {
-		return nil, errors.E("authorisation service cannot be nil")
-	}
 	if jobtracker == nil {
 		return nil, errors.E("job tracker cannot be nil")
 	}
+	if jujuManager == nil {
+		return nil, errors.E("juju manager cannot be nil")
+	}
+	if binaryStore == nil {
+		return nil, errors.E("binary store cannot be nil")
+	}
+	if jimmWellknownJWKSEndpoint == "" {
+		return nil, errors.E("jimm well-known JWKs endpoint cannot be empty")
+	}
 	return &bootstrapManager{
-		store:       store,
-		authSvc:     authSvc,
-		jobtracker:  jobtracker,
-		jujuManager: jujuManager,
-		binaryStore: binaryStore,
+		store:                     store,
+		tracker:                   jobtracker,
+		jujuManager:               jujuManager,
+		binaryStore:               binaryStore,
+		jimmWellknownJWKSEndpoint: jimmWellknownJWKSEndpoint,
 	}, nil
 }
 
@@ -106,7 +118,7 @@ func NewBootstrapManager(
 func (b *bootstrapManager) GetBootstrapStatusAndLogs(ctx context.Context, _ *openfga.User, jobId uuid.UUID, offset int) (params.BootstrapStatusResponse, error) {
 	const op = errors.Op("jimm.GetBootstrapStatusAndLogs")
 
-	job, err := b.jobtracker.GetJob(ctx, jobId)
+	job, err := b.tracker.GetJob(ctx, jobId)
 	if err != nil {
 		return params.BootstrapStatusResponse{}, errors.E(op, "failed to get job status", err)
 	}
@@ -123,18 +135,6 @@ func (b *bootstrapManager) GetBootstrapStatusAndLogs(ctx context.Context, _ *ope
 	}, nil
 }
 
-// StartBootstrap starts a bootstrap job with the provided parameters.
-func (b *bootstrapManager) StartBootstrap(ctx context.Context, user *openfga.User, params BootstrapParams) (string, error) {
-	const op = errors.Op("jimm.StartBootstrap")
-
-	err := params.validate()
-	if err != nil {
-		return "", errors.E(op, fmt.Errorf("invalid bootstrap parameters: %v", err))
-	}
-
-	return "", errors.E(op, "not implemented")
-}
-
 // StopBootstrap stops a bootstrap job by its ID.
 func (b *bootstrapManager) StopBootstrap(ctx context.Context, user *openfga.User, jobId uuid.UUID) error {
 	const op = errors.Op("jimm.StopBootstrap")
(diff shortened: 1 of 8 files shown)


internal/jimm/bootstrap/bootstrap.go
```diff
@@ -147,14 +147,58 @@ func (b *bootstrapManager) StopBootstrap(ctx context.Context, user *openfga.User
...
+				CLIVersion: params.CLIVersion,
+				// User defined command arguments
+				CloudNameAndRegion: params.CloudNameAndRegion,
+				ControllerName:     params.ControllerName,
+				AgentVersion:       params.AgentVersion,
+				BootstrapTimeout:   params.BootstrapTimeout,
+				CloudCred:          params.CloudCred,
+				PersonalCloud:      params.PersonalCloud,
+				// JIMM Provided command arguments (i.e., ones that must be set by JIMM when bootstrapping).
+				LoginTokenRefreshURL: b.jimmWellknownJWKSEndpoint, // TODO: Set the correct login token refresh URL
```
kian99, line 188: Feels like this could be split up nicely in 2, one struct for bootstrapArgs and another for commandArgs

internal/jimm/bootstrap/bootstrap.go
```diff
@@ -265,18 +307,26 @@ func (b *bootstrapManager) BootstrapJob(
...
+				Arch: runtime.GOARCH,
 			},
 		)
 		if err != nil {
 			return errors.E(fmt.Errorf("failed to get Juju binary: %w", err))
 		}
-
+		// TODO(ale8k): Is it possible to tell the client to only attempt to get status after binary is downloaded?
+		// Because as it stands, the client will be unncessarily polling (perhaps we could put download progress in the logs?).
+		// That way the status calls wouldn't be unncessary and serve a purpose.
```
kian99: We could have the client print a "Waiting for bootstrap to start" message at the beginning until logs start coming in.

kian99 (commented):
Just reviewed the last commit

internal/jimm/bootstrap/bootstrap.go
ale8k: It does already print a pending job message, think that'd suffice?

internal/jimm/bootstrap/bootstrap.go
ale8k: I agree, I started to do it and realised executor takes a big set of params too. If we did it here it'd be inconsistent from that - so maybe a follow up?

internal/jimm/bootstrap/bootstrap.go
```diff
@@ -265,18 +307,26 @@ func (b *bootstrapManager) BootstrapJob(
 			return errors.E(fmt.Errorf("failed to check if controller exists: %w", err))
 		}
 
+		// TODO(ale8k): When downloading, it takes a while and the CLI has no output. Download progress would be a VERY nice to have here.
```
kian99: Since we have the JIRA card I'd either include the `JUJU-XXXX` number or delete it.

internal/jimm/bootstrap/bootstrap.go
kian99: If that suffices then I'd delete the todo. A single message like `downloading Juju CLI` for the user to understand **what** it's doing (printing this from the server side) would be helpful.

internal/jimm/bootstrap/bootstrap.go
```diff
@@ -339,6 +396,12 @@ func (b *bootstrapManager) runBootstrap(
 	if err != nil {
 		return errors.E(fmt.Errorf("failed to get controller details: %w", err))
 	}
+	// TODO(ale8k): At the moment, we can't reach added k8s controllers because the controller is not reachable.
```
kian99: Add the JIRA ref or delete the TODO.

internal/jimm/bootstrap/bootstrap.go
```diff
@@ -367,6 +430,10 @@ func (b *bootstrapManager) runBootstrap(
 		AdminIdentityName: account.User,
 		AdminPassword:     account.Password,
 	}
+	// TODO(ale8k): Fix this...
```
kian99: Same here

internal/jimm/bootstrap/bootstrap_test.go
```diff
@@ -215,13 +213,18 @@ func (s *bootstrapManagerSuite) TestGetBootstrapStatusAndLogs_JobNotFound(c *qt.
...
-	manager, err := bootstrap.NewBootstrapManager(s.ofgaClient, s.db, s.jobTracker, jujuManager, binaryStore)
+	manager, err := bootstrap.NewBootstrapManager(s.db, s.jobTracker, jujuManager, binaryStore, loginTokenRefreshURLParam)
 	c.Assert(err, qt.IsNil)
 
 	_, err = manager.GetBootstrapStatusAndLogs(ctx, s.adminUser, jobId, 0)
 	c.Assert(err, qt.ErrorMatches, "failed to get job status")
 }
 
+// TODO
+func (s *bootstrapManagerSuite) TestStartBootstrap(c *qt.C) {
```
kian99: Is this for this PR? What's the plan here since we already have all the mocked tests and I was under the impression we wouldn't spin up a real controller in these tests.

internal/jimm/bootstrap/types.go
```diff
@@ -6,33 +6,56 @@ import (
...
 )
 
 // BootstrapParams defines the parameters required for bootstrapping a JIMM controller.
 type BootstrapParams struct {
-	ControllerName string
-	CloudName      string
-	CloudRegion    string
-	AgentVersion   string
-	TimeoutSeconds int
+	CLIVersion string
```
kian99, line 17: Should this be controller version to match the params the user passes in?

internal/jimm/bootstrap/types.go
```diff
@@ -6,33 +6,56 @@ import (
...
+
+	CloudNameAndRegion string
+	ControllerName     string
+	AgentVersion       string
+	BootstrapTimeout   int
+
+	CloudCred jujucloud.CloudCredential
+	// PersonalCloud is the personally defined cloud. Only necessary if the cloud is not a public
+	// cloud.
```
kian99:
```suggestion
	// PersonalCloud is the cloud-definition for a non-public cloud.
```

internal/jimm/jimm.go
```diff
@@ -525,11 +525,11 @@ func New(p Parameters) (*JIMM, error) {
 		return nil, err
 	}
 	bootstrapManager, err := bootstrap.NewBootstrapManager(
-		j.OpenFGAClient,
 		j.Database,
 		jobTracker,
 		j.jujuManager,
 		binaryStore,
+		"fixed-later",
```
kian99, line 532: TODO? Or it's fixed in one of the follow-up PRs?

internal/jujuapi/jimm_unit_test.go
```diff
@@ -201,6 +203,9 @@ func (s *jimmUnitTestSuite) TestBootstrapStart(c *gc.C) {
 			AgentVersion: "1.0.0",
 			Timeout:      3600,
 		},
+		Cloud:             jujuparams.Cloud{},
+		Credential:        jujucloud.CloudCredential{},
```
kian99, line 207: Surely this should be a struct from `jujuparams` as well?

kian99 (approved):

internal/jimm/bootstrap/bootstrap.go
```diff
@@ -265,18 +307,26 @@ func (b *bootstrapManager) BootstrapJob(
...
 			jujuclistore.JujuBinarySpec{
 				Version: p.CLIVersion,
-				Os:      p.CLIVersion,
-				Arch:    p.CLIArch,
+				// This is a best effort. The launchpad URL just so happens to have similar filenames to runtime.GOOS and runtime.GOARCH.
+				// If this ever changes, we should update the binary store to use the correct URL
+				// or we should detect the OS and arch here. We don't want to detect it in the binary store
+				// because the consumer may want to detect it in different ways.
+				Os:   runtime.GOOS,
+				Arch: runtime.GOARCH,
```
SimoneDutto, line 319: i would rather use a value coming from an env, and defaulting to this. Because what if you can't bootstrap a controller because this assumption is wrong? You'd need to wait for a JIMM release

internal/jimm/bootstrap/bootstrap.go
SimoneDutto: I don't think it's an issue for the CLI to poll logs while the binary is downloading

internal/jimm/bootstrap/bootstrap.go
```diff
@@ -147,14 +147,58 @@ func (b *bootstrapManager) StopBootstrap(ctx context.Context, user *openfga.User
...
 
+// StartBootstrap starts a bootstrap job with the provided parameters.
+func (b *bootstrapManager) StartBootstrap(ctx context.Context, user *openfga.User, params BootstrapParams) (string, error) {
+	const op = errors.Op("jimm.StartBootstrap")
+
+	if err := params.validate(); err != nil {
+		return "", errors.E(op, fmt.Errorf("invalid bootstrap parameters: %v", err))
+	}
+
+	temp, err := os.MkdirTemp("", "juju-data-dir")
```
SimoneDutto, line 166: what are the rules around tmp dirs with k8s? Are they deleted when the pod is restarted? Shouldn't the job itself handling the creation/destruction of the tempdir?

SimoneDutto (approved):
Some comments, but LGTM

internal/jimm/bootstrap/bootstrap.go
ale8k: Removed comment

internal/jimm/bootstrap/bootstrap.go
ale8k: Removed

internal/jimm/bootstrap/bootstrap.go
ale8k: Removed

internal/jimm/bootstrap/bootstrap.go
ale8k: RemovedRemoved

internal/jimm/bootstrap/bootstrap_test.go
ale8k: Ah I didn't push the stash - need to update the test too - so it can verify dir is created will do it now and ping back when done.

ale8k: Scratch that, I'd need to mock tracker and it feels a bit unnecessary... We wouldn't be testing so much so I'll give this test a miss. We do have many, many job tests.

internal/jimm/jimm.go
ale8k: Fixed in a follow up here: https://github.com/canonical/jimm/pull/1682

internal/jujuapi/jimm_unit_test.go
ale8k: Yes this was the boo boo I said I made yesterday 🤦 , I'm gonna come and change it to params afterwards as another follow up.

internal/jimm/bootstrap/bootstrap.go
ale8k: Yeah can totally make it an env. Mind if I do it in a follow up and add it to my notes of follow ups? I've got a few for when these next 3 are merged.

SimoneDutto: Create a card and it's fine

ale8k: I won't make card, I'm gonna do a bunch of little minor follow ups today.

internal/jimm/bootstrap/bootstrap.go
ale8k: When I run this it did just indeed create a /tmp dir. I didn't check the bits added so without Q/Aing again I couldn't say. I can check if it's a ramfs mount if you'd like later on when Q/Aing with the proxy fixes? (Made a note to check this)

kian99:
@ale8k There were some points you mentioned would be done in a follow-up e.g.
- https://github.com/canonical/jimm/pull/1680#discussion_r2293286944
- https://github.com/canonical/jimm/pull/1680#discussion_r2293264356

Just making sure those are still kept in mind.

ale8k:
> @ale8k There were some points you mentioned would be done in a follow-up e.g.
> 
> * [Implement StartBootstrap #1680 (comment)](https://github.com/canonical/jimm/pull/1680#discussion_r2293286944)
> * [Implement StartBootstrap #1680 (comment)](https://github.com/canonical/jimm/pull/1680#discussion_r2293264356)
> 
> Just making sure those are still kept in mind.

They are, I've got the following here:

Bootstrap todos:
- Update juju cloud.CloudCredential param to params
- Instead of runtime os / arch, just use env.
- Checksum binary check (url has md5 next to it)
- Check mount type of /tmp dir for job mount | grep /tmp & let Simone know 
- Don’t check built in on cli check it on server
- Check if we can determine the cli version to download for a given controller version?

PR #1681, opened 2025-08-21
Update bootstrap start cli command

ale8k:
## Description
Based on: #1680

This PR updates the CLI to pass the CLI version, prevents built-in clouds, retrieves the cloud and credential, updates the printed line to be human readable only and maps the cloud to the expected param type.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

### DIFF
--- a/cmd/jaas/cmd/bootstrap.go
+++ b/cmd/jaas/cmd/bootstrap.go
@@ -35,13 +35,11 @@ logs are streamed to stdout.
 Use the --detach flag to start the bootstrap job and return immediately,
 printing only the job ID, without waiting for the job to complete.
 
-The final argument, version, denotes the version of the Juju CLI
-to use for the bootstrap behind the scenes when bootstrapping via JIMM.
+The final argument, version, denotes the Juju controller to be bootstrapped.
 `
 	bootstrapExamples = `
-	juju [jaas] bootstrap <cloud[/region]> <controller name> <cli version>
+	juju [jaas] bootstrap <cloud[/region]> <controller name> <controller version>
 	juju [jaas] bootstrap mycloud/region mycontroller 3.6.8
-	juju [jaas] bootstrap mycloud/region mycontroller 3.6.9 --agent-version=3.6.8
 `
 )
 
@@ -53,13 +51,12 @@ type bootstrapCommand struct {
 	store            jujuclient.ClientStore
 	bootstrapAPIFunc func() (JIMMAPI, error)
 
-	cloud          string
-	region         string
-	controllerName string
-	cliVersion     string
-	agentVersion   string
-	timeout        int
-	detach         bool
+	cloud             string
+	region            string
+	controllerName    string
+	controllerVersion string
+	timeout           int
+	detach            bool
 }
 
 // NewBootstrapStartCommand returns a command to start a job
@@ -86,7 +83,7 @@ func (c *bootstrapCommand) Init(args []string) error {
 		return fmt.Errorf("cloud name %q not valid", c.cloud)
 	}
 	c.controllerName = args[1]
-	c.cliVersion = args[2]
+	c.controllerVersion = args[2]
 
 	return nil
 }
@@ -98,7 +95,6 @@ func (c *bootstrapCommand) SetFlags(f *gnuflag.FlagSet) {
 		"yaml": cmd.FormatYaml,
 		"json": cmd.FormatJson,
 	})
-	f.StringVar(&c.agentVersion, "agent-version", "", "The version of the Juju agent to use for the bootstrap.")
 	f.IntVar(&c.timeout, "timeout", 0, "The timeout in seconds for the bootstrap operation.")
 	f.BoolVar(&c.detach, "detach", false, "If set, the command will start the bootstrap job and return immediately with the job ID, without waiting for the job to complete.")
 	// TODO(ale8k): Support passing cloud & cloudcredential files, for now we're looking up clouds and credentials added to the store.
@@ -142,16 +138,15 @@ func (c *bootstrapCommand) Run(ctxt *cmd.Context) error {
 	}
 
 	req := apiparams.BootstrapStartParams{
-		CloudName:      c.cloud,
-		RegionName:     c.region,
-		ControllerName: c.controllerName,
-		CLIVersion:     c.cliVersion,
-		Cloud:          cloudToParams(*bootstrapCloud),
-		Credential:     *bootstrapCredential,
+		CloudName:         c.cloud,
+		RegionName:        c.region,
+		ControllerName:    c.controllerName,
+		ControllerVersion: c.controllerVersion,
+		Cloud:             cloudToParams(*bootstrapCloud),
+		Credential:        *bootstrapCredential,
 
 		Flags: apiparams.BootstrapFlags{
-			AgentVersion: c.agentVersion,
-			Timeout:      c.timeout,
+			Timeout: c.timeout,
 		},
 	}
 
@@ -168,7 +163,7 @@ func (c *bootstrapCommand) Run(ctxt *cmd.Context) error {
 
 	if c.detach {
 		fmt.Printf(`
-Bootstrap job started successfully.
+Bootstrap job started.
 You can track the progress via bootstrap-status with the job ID:
 	juju [jaas] bootstrap-status %s
 
--- a/cmd/jaas/cmd/bootstrap.go
+++ b/cmd/jaas/cmd/bootstrap.go
@@ -8,9 +8,12 @@ import (
 
 	"github.com/juju/cmd/v3"
 	"github.com/juju/gnuflag"
+	jujucloud "github.com/juju/juju/cloud"
 	jujucmd "github.com/juju/juju/cmd"
+	"github.com/juju/juju/cmd/juju/common"
 	"github.com/juju/juju/cmd/modelcmd"
 	"github.com/juju/juju/jujuclient"
+	jujuparams "github.com/juju/juju/rpc/params"
 	"github.com/juju/names/v5"
 
 	jimmAPI "github.com/canonical/jimm/v3/pkg/api"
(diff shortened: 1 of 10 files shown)


cmd/jaas/cmd/bootstrap.go
```diff
@@ -31,10 +34,14 @@ logs are streamed to stdout.
 
 Use the --detach flag to start the bootstrap job and return immediately,
 printing only the job ID, without waiting for the job to complete.
+
+The final argument, version, denotes the version of the Juju CLI
+to use for the bootstrap behind the scenes when bootstrapping via JIMM.
```
kian99: This comment is an implementation detail. 
Also, this diverges the `jaas` command from the `juju` command. What is the difference between the final version string and `--agent-version`?

cmd/jaas/cmd/bootstrap.go
```diff
@@ -107,10 +118,37 @@ func (c *bootstrapCommand) Info() *cmd.Info {
...
+
+	// Check built in clouds like localhost (lxd).
+	builtinClouds, err := common.BuiltInClouds()
+	if err != nil {
+		return err
+	}
+
+	if _, isABuiltinCloud := builtinClouds[c.cloud]; isABuiltinCloud {
+		return fmt.Errorf("bootstrap via JIMM does not support built-in clouds like %q", c.cloud)
+	}
```
kian99: Can't we just do this on the server side?

cmd/jaas/cmd/bootstrap.go
```diff
@@ -107,10 +118,37 @@ func (c *bootstrapCommand) Info() *cmd.Info {
...
+
+	// We use [jujucloud.CloudByName] and not [common.CloudByName] as the JIMM bootstrap
+	// will NOT support builtin clouds (localhost, microk8s, docker-desktop, etc.).
```
kian99, line 117: docker-destkop is a built-in cloud?

cmd/jaas/cmd/bootstrap.go
```diff
@@ -128,22 +166,42 @@ func (c *bootstrapCommand) Run(ctxt *cmd.Context) error {
 		return err
 	}
 
-	err = c.out.Write(ctxt, resp)
-	if err != nil {
-		return err
+	if c.detach {
+		fmt.Printf(`
+Bootstrap job started successfully.
```
kian99: Can you remove the word "successfully", at a glance it could cause confusion that bootstrap was successful.

cmd/jaas/cmd/bootstrap.go
ale8k: Final version string is the CLI version and agent is the agent to be bootstrapped (controller version really, why is why I tried to explain in the other PR I wouldn't call it controller version). Not all CLI versions can bootstrap all agent version, so you'd specify the CLI of juju you'd want to use.

cmd/jaas/cmd/bootstrap.go
ale8k: It's a very minor quick check here but agreed, should server side really... I've put it into my notes for the follow up.

cmd/jaas/cmd/bootstrap.go
ale8k: Aha yeah, well, on my Mac it was detected anyways. It ends up just doing k8s - not docker as in compose or whatever else.

cmd/jaas/cmd/bootstrap.go
kian99: Why can't jimm figure out what version of CLI to use based on the controller version the user wants?

cmd/jaas/cmd/bootstrap.go
kian99: I don't understand, docker-desktop shows up when you run `juju clouds`?

cmd/jaas/cmd/bootstrap.go
kian99: I'm not sure everything can always go in a follow-up.

cmd/jaas/cmd/bootstrap.go
```diff
@@ -31,10 +34,14 @@ logs are streamed to stdout.
...
 Use the --detach flag to start the bootstrap job and return immediately,
 printing only the job ID, without waiting for the job to complete.
+
+The final argument, version, denotes the version of the Juju CLI
+to use for the bootstrap behind the scenes when bootstrapping via JIMM.
 `
 	bootstrapExamples = `
-	juju [jaas] bootstrap mycloud/region mycontroller
-	juju [jaas] bootstrap mycloud/region mycontroller --agent-version=3.6.8
+	juju [jaas] bootstrap <cloud[/region]> <controller name> <cli version>
```
alesstimec: s/cli version/controller version/

cmd/jaas/cmd/bootstrap.go
```diff
@@ -78,6 +86,7 @@ func (c *bootstrapCommand) Init(args []string) error {
 		return fmt.Errorf("cloud name %q not valid", c.cloud)
 	}
 	c.controllerName = args[1]
+	c.cliVersion = args[2]
```
alesstimec: controller version?

cmd/jaas/cmd/bootstrap.go
ale8k: In fairness, I haven't checked. Perhaps we can? But I'd need to go investigate on how we'd go about that. I've added it to my notes.

cmd/jaas/cmd/bootstrap.go
ale8k: Yup. When I got the Mac, I originally used docker desktop and attempted to do things locally, and to my surprise docker-desktop was there as a k8s cloud!

cmd/jaas/cmd/bootstrap.go
ale8k: Yup.

cmd/jaas/cmd/bootstrap.go
ale8k: Well I more meant this is a cli related PR and didn't want to do it in this PR, but I can. I'll do it soon.

cmd/jaas/cmd/bootstrap.go
ale8k: I asked in the development channel, the answer was:

A cli version of major.minor can bootstrap any point release agent version. That is, 3.6.8 can bootstrap any 3.6.X. On the same vein, a 3.6.8 does indeed default to a 3.6.8 agent. So, if the user requests a 3.6.8 CLI, they will get a 3.6.8 agent version. This is the path of least resistance for us I think.

On that note, we could determine if a user has requested a 3.6.X, and the store simply checks that a binary of 3.6 exists, ignoring the minor, and populating the correct agent version.

But generally speaking, I think option one is fine and two may be an over-improvement. Wdyt?

kian99: Yes option 1 is fine. So you'll remove the final positional argument and require `--agent-version` to be set? Or keep the positional argument and throw an error if `--agent-version` is set? I think the former is better to align with the Juju CLI somewhat.

And do we agree that we should remove mention about the Juju CLI behind the scenes in the help text?

ale8k: Yes let's remove mention of juju cli for user facing code. I.e., within CLI code and help text.

And yeah, we'll drop agent version and just let the CLI they've implicitly requested do its thing?

kian99: >let the CLI they've implicitly requested do its thing

I'm not following you on this bit. Are you suggesting the command will be `juju [jaas] bootstrap <cloud name>[/region] <controller name> <juju version>` or use `--agent-version`?

ale8k: Just juju version without --agent-version. Wdyt?

kian99: It's okay for me. It just means the jaas bootstrap command won't match the Juju CLI as closely.

cmd/jaas/cmd/bootstrap.go
```diff
@@ -49,6 +56,7 @@ type bootstrapCommand struct {
 	cloud          string
 	region         string
 	controllerName string
+	cliVersion     string
```
alesstimec: controllerVersion

cmd/jaas/cmd/bootstrap.go
ale8k: I've began to strip out agent version in place of controller version as an argument and not a flag.

cmd/jaas/cmd/bootstrap.go
```diff
@@ -67,8 +72,8 @@ func NewBootstrapStartCommand() cmd.Command {
 
 // Init implements modelcmd.Command.
 func (c *bootstrapCommand) Init(args []string) error {
-	if len(args) < 2 {
```
luci1900, line 70: Is this a breaking change then? Does it matter?

cmd/jaas/cmd/bootstrap.go
luci1900: FWIW I think monorepos do get this right: make all related changes to every component in one commit, so it's easy to revert if any of it goes wrong.

I wouldn't mind changes to a different component in the same repo.

ale8k: Sure, I'm happy to move this check over then...

cmd/jaas/cmd/bootstrap.go
ale8k: It's an unreleased feature, so should be ok.

luci1900 (approved):
Afaict it looks fine.

internal/jimm/bootstrap/bootstrap.go
```diff
@@ -393,6 +392,12 @@ func (b *bootstrapManager) runBootstrap(
 	if err != nil {
 		return errors.E(fmt.Errorf("failed to get controller details: %w", err))
 	}
+	// TODO(ale8k): At the moment, we can't reach added k8s controllers because the controller is not reachable.
+	// The CLI spins up a kube-proxy to reach it and we could do the same. The controller details of said controller
+	// may contain proxy details and those proxy details can be used to launch a proxy to contact the controller.
+	// Accessed like so: [ctrlDetails.Proxy.Proxier.Start].
+	//
+	// As such, AddController fails.
```
kian99: You can put [JUJU-8436](https://warthogs.atlassian.net/browse/JUJU-8436) here which is the card Simone made after our conversation today. And the todo probably deserves rewording.

internal/jimm/bootstrap/bootstrap_test.go
```diff
@@ -221,6 +221,11 @@ func (s *bootstrapManagerSuite) TestGetBootstrapStatusAndLogs_JobNotFound(c *qt.
 	c.Assert(err, qt.ErrorMatches, "failed to get job status")
 }
 
+// TODO
+func (s *bootstrapManagerSuite) TestStartBootstrap(c *qt.C) {
+
+}
```
kian99: Remove

kian99 (approved):

PR #1682, opened 2025-08-21
Plugin login token refresh url for bootstrap

ale8k:
## Description
Based on: #1681 

Plugs in the login token refresh url. Rather than compose it from the public dns, well known path and jwks endpoint. 

Users may rewrite this with a proxy in front, and if we did configure it such that somehow we limited them it would be troublesome. The scheme is also NOT optional. After reasoning with these two conditions I felt it was better and easier to create an entirely new env var. 

## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests

### DIFF
--- a/docker-compose.common.yaml
+++ b/docker-compose.common.yaml
@@ -87,6 +87,8 @@ services:
         kZT1O7501gvSxsHDNB/X39/AyrRp9nzakhkna2FIPIGdBSsOA84EPCQong/+Kzq9OniZFN
         v8touvVeAGefsAAAAPZHV0dG9zQGR1dHRvc3BjAQIDBA==
         -----END OPENSSH PRIVATE KEY-----
+      # To Q/A this, use any reachable address for this compose JIMM. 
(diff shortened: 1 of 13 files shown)


docker-compose.common.yaml
```diff
@@ -86,6 +86,8 @@ services:
         kZT1O7501gvSxsHDNB/X39/AyrRp9nzakhkna2FIPIGdBSsOA84EPCQong/+Kzq9OniZFN
         v8touvVeAGefsAAAAPZHV0dG9zQGR1dHRvc3BjAQIDBA==
         -----END OPENSSH PRIVATE KEY-----
+      # To Q/A this, use any reachable address for this compose JIMM. 
+      JIMM_BOOTSTRAP_LOGIN_TOKEN_REFRESH_URL: "http://10.0.0.0:17070/.well-known/jwks.json"
```
kian99: What about `https://jimm.localhost/...`

kian99 (approved):

SimoneDutto (approved):

luci1900 (approved):

PR #1683, opened 2025-08-21
feat: complete migrations when model info is called

SimoneDutto:
In this pr we propose to finish the migration if we receive a redirect error during a ModelInfo call.

### DIFF
--- a/internal/jimm/juju/model.go
+++ b/internal/jimm/juju/model.go
@@ -274,20 +274,29 @@ func (j *JujuManager) ModelInfo(ctx context.Context, user *openfga.User, mt name
 func (j *JujuManager) modelInfo(ctx context.Context, model *dbmodel.Model, api API) (*jujuparams.ModelInfo, error) {
 	modelInfo := &jujuparams.ModelInfo{UUID: model.UUID.String}
 	errFromAPI := api.ModelInfo(ctx, modelInfo)
-	var err error
+
+	if errFromAPI == nil {
+		return j.reactToModelInfoSuccess(ctx, model, modelInfo)
+	} else {
+		return j.reactToModelInfoError(ctx, errFromAPI, model, modelInfo)
+	}
+}
+
+func (j *JujuManager) reactToModelInfoError(ctx context.Context, errFromAPI error, model *dbmodel.Model, modelInfo *jujuparams.ModelInfo) (*jujuparams.ModelInfo, error) {
 	switch model.MigrationMode {
 	case dbmodel.MigrationModeNone:
-		err = j.maybeCleanupModel(ctx, errFromAPI, model)
+		err := j.maybeCleanupModel(ctx, errFromAPI, model)
 		if err != nil {
-			return nil, err
+			zapctx.Error(ctx, "error cleaning model", zap.Error(err))
+			return nil, errors.E("internal server error")
 		}
-		// we need to return both the modelInfo and the error, because
-		// the error from the API must be propagated to the caller.
-		return modelInfo, errFromAPI
+		// propagate the error to the caller.
+		return nil, errFromAPI
 	case dbmodel.MigrationModeMigrateInternal:
-		err = j.checkModelMigratedInternal(ctx, errFromAPI, model, errFromAPI == nil && modelInfo.Migration.End != nil)
+		err := j.checkModelMigratedInternal(ctx, errFromAPI, model)
 		if err != nil {
-			return nil, err
+			zapctx.Error(ctx, "error checking model migration", zap.Error(err))
+			return nil, errors.E("internal server error")
 		}
 		// If the model has been migrated internally, we call api.ModelInfo again
 		// to get the updated model information from the new controller.
@@ -301,9 +310,34 @@ func (j *JujuManager) modelInfo(ctx context.Context, model *dbmodel.Model, api A
 		defer api.Close()
 		errFromAPI := api.ModelInfo(ctx, modelInfo)
 		return modelInfo, errFromAPI
+	// If the migration mode is exporting or importing, we return the error as is.
+	case dbmodel.MigrationModeExporting, dbmodel.MigrationModeImporting:
+		return nil, errFromAPI
 	default:
-		return nil, errors.E("unknown migration mode: " + string(model.MigrationMode))
+		return nil, errors.E("model in unsupported migration mode")
 	}
+
+}
+
+func (j *JujuManager) reactToModelInfoSuccess(ctx context.Context, model *dbmodel.Model, modelInfo *jujuparams.ModelInfo) (*jujuparams.ModelInfo, error) {
+	switch model.MigrationMode {
+	case dbmodel.MigrationModeNone, dbmodel.MigrationModeExporting, dbmodel.MigrationModeImporting:
+		return modelInfo, nil
+	case dbmodel.MigrationModeMigrateInternal:
+		// If the migration end time is set, it means the model has
+		// failed to migrate otherwise we'd expect a redirect error.
+		if modelInfo.Migration.End != nil {
+			model.MigrationFailed()
+			if err := j.Database.UpdateModel(ctx, model); err != nil {
+				return nil, errors.E(fmt.Errorf("failed to update model after failed migration: %w", err))
+			}
+			return modelInfo, nil
+		}
+		return modelInfo, nil
+	default:
+		return nil, errors.E("model in unsupported migration mode")
+	}
+
 }
 
 // modelSummariesMap is a safe map to add records concurrently because the access is guarded by a Mutex.
--- a/internal/jimm/juju/model.go
+++ b/internal/jimm/juju/model.go
@@ -261,23 +261,49 @@ func (j *JujuManager) ModelInfo(ctx context.Context, user *openfga.User, mt name
 	}
 	defer api.Close()
 
-	mi := &jujuparams.ModelInfo{
-		UUID: mt.Id(),
-	}
-	if err := api.ModelInfo(ctx, mi); err != nil {
-		// If the model is not found or code not authorized on the backing controller then
-		// we delete the model from JIMM.
-		// Juju returns CodeAuthorized when the model is not found for this facade.
-		if errors.ErrorCode(err) == errors.CodeNotFound || errors.ErrorCode(err) == errors.CodeUnauthorized {
-			errDelete := j.deleteModel(ctx, mt)
-			if errDelete != nil {
-				return nil, errors.E(op, errDelete)
-			}
-		}
+	modelInfo, err := j.modelInfo(ctx, &m, api)
+	if err != nil {
 		return nil, errors.E(op, err)
 	}
 
-	return j.mergeModelInfo(ctx, user, mi, m)
+	return j.mergeModelInfo(ctx, user, modelInfo, m)
+}
+
+// modelInfo retrieves the model information from the controller and reacts
+// to the error to update JIMM's state.
+func (j *JujuManager) modelInfo(ctx context.Context, model *dbmodel.Model, api API) (*jujuparams.ModelInfo, error) {
+	modelInfo := &jujuparams.ModelInfo{UUID: model.UUID.String}
+	errFromAPI := api.ModelInfo(ctx, modelInfo)
+	var err error
+	switch model.MigrationMode {
+	case dbmodel.MigrationModeNone:
+		err = j.maybeCleanupModel(ctx, errFromAPI, model)
+		if err != nil {
+			return nil, err
+		}
+		// we need to return both the modelInfo and the error, because
+		// the error from the API must be propagated to the caller.
+		return modelInfo, errFromAPI
+	case dbmodel.MigrationModeMigrateInternal:
+		err = j.checkModelMigratedInternal(ctx, errFromAPI, model, errFromAPI == nil && modelInfo.Migration.End != nil)
+		if err != nil {
+			return nil, err
+		}
+		// If the model has been migrated internally, we call api.ModelInfo again
+		// to get the updated model information from the new controller.
+		if err := j.Database.GetModel(ctx, model); err != nil {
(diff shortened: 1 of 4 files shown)


internal/jimm/juju/model.go
```diff
@@ -261,23 +261,49 @@ func (j *JujuManager) ModelInfo(ctx context.Context, user *openfga.User, mt name
...
+	return j.mergeModelInfo(ctx, user, modelInfo, m)
+}
+
+// modelInfo retrieves the model information from the controller and reacts
+// to the error to update JIMM's state.
+func (j *JujuManager) modelInfo(ctx context.Context, model *dbmodel.Model, api API) (*jujuparams.ModelInfo, error) {
+	modelInfo := &jujuparams.ModelInfo{UUID: model.UUID.String}
+	errFromAPI := api.ModelInfo(ctx, modelInfo)
+	var err error
+	switch model.MigrationMode {
```
SimoneDutto, line 286: The reason why this `switch` is replicated here and in the model poller is because we want to threat the error in two different ways.

In the model poller we don't care about the error returned by `maybeCleanupModel`, `checkModelMigratedInternal`, because we're just logging it.
During the ModelInfo facade we have a specific behavior:
- for not-found errors we want to sync jimm's state *and* return the not-found error
- for the redirect we want to retry to fetch model info with the updated controller info and not return any error to the client.

internal/jimm/juju/model.go
```diff
@@ -261,23 +261,49 @@ func (j *JujuManager) ModelInfo(ctx context.Context, user *openfga.User, mt name
...
+		if err := j.Database.GetModel(ctx, model); err != nil {
+			return nil, err
+		}
+		api, err := j.dial(ctx, &model.Controller, names.ModelTag{}, nil)
+		if err != nil {
+			return nil, err
+		}
+		defer api.Close()
+		errFromAPI := api.ModelInfo(ctx, modelInfo)
+		return modelInfo, errFromAPI
```
SimoneDutto, line 312: This might not be the best, because if the facade returns another error we are passing it down to the client, but it should be very unlikely because it means that a newly migrated model is also deleted in a very short window.

Another solution could have been to call `j.modelInfo` recursevely, but i feel like it's not the best.
I'm open to hear your opinion

internal/jimm/juju/model.go
```diff
@@ -261,23 +261,49 @@ func (j *JujuManager) ModelInfo(ctx context.Context, user *openfga.User, mt name
...
 
-	return j.mergeModelInfo(ctx, user, mi, m)
+	return j.mergeModelInfo(ctx, user, modelInfo, m)
+}
+
+// modelInfo retrieves the model information from the controller and reacts
+// to the error to update JIMM's state.
+func (j *JujuManager) modelInfo(ctx context.Context, model *dbmodel.Model, api API) (*jujuparams.ModelInfo, error) {
+	modelInfo := &jujuparams.ModelInfo{UUID: model.UUID.String}
+	errFromAPI := api.ModelInfo(ctx, modelInfo)
```
kian99, line 276: If the error is nil we can return early.

internal/jimm/juju/model.go
```diff
@@ -261,23 +261,49 @@ func (j *JujuManager) ModelInfo(ctx context.Context, user *openfga.User, mt name
...
+	var err error
+	switch model.MigrationMode {
+	case dbmodel.MigrationModeNone:
+		err = j.maybeCleanupModel(ctx, errFromAPI, model)
+		if err != nil {
+			return nil, err
+		}
+		// we need to return both the modelInfo and the error, because
+		// the error from the API must be propagated to the caller.
+		return modelInfo, errFromAPI
```
kian99: Just return the controller error.
```suggestion
		return nil, errFromAPI
```

internal/jimm/juju/model.go
```diff
@@ -261,23 +261,49 @@ func (j *JujuManager) ModelInfo(ctx context.Context, user *openfga.User, mt name
...
+func (j *JujuManager) modelInfo(ctx context.Context, model *dbmodel.Model, api API) (*jujuparams.ModelInfo, error) {
+	modelInfo := &jujuparams.ModelInfo{UUID: model.UUID.String}
+	errFromAPI := api.ModelInfo(ctx, modelInfo)
+	var err error
+	switch model.MigrationMode {
+	case dbmodel.MigrationModeNone:
+		err = j.maybeCleanupModel(ctx, errFromAPI, model)
+		if err != nil {
+			return nil, err
+		}
```
kian99: Mask the error here. If we fail to delete the model it would be weird to see `failed to delete model` in a call to `ModelInfo`. Not sure what message is appropriate though.
```suggestion
		if err != nil {
			return nil, errors.E(err, "internal server error")
		}
```

internal/jimm/juju/model_poller.go
```diff
@@ -77,17 +78,12 @@ func (j *JujuManager) PollModels(ctx context.Context) (err error) {
...
-	// Check if the model has completed a migration.
-	// If modelInfo returns without an error, it definitely hasn't moved yet.
-	modelInfo := &jujuparams.ModelInfo{UUID: m.UUID.String}
-	err := api.ModelInfo(ctx, modelInfo)
-	if err == nil {
+	if errFromAPI == nil {
 		// If the migration end time is set, it means the model has
 		// failed to migrate otherwise we'd expect a redirect error.
-		if modelInfo.Migration.End != nil {
+		if migrationEndSet {
```
kian99: You can check this without even checking the error. If it is set the model definitely didn't finish migration.

internal/jimm/juju/model.go
kian99: This is perfect imo because if the controller is returning an error we should be sending that back to the client in scenarios where we don't want to modify it like this one.

internal/jimm/juju/model_poller.go
kian99, line 1: To simplify things, consider passing just the error code into this helper functions.

internal/jimm/juju/model_poller.go
```diff
@@ -99,35 +95,35 @@ func (j *JujuManager) checkModelMigratedInternal(ctx context.Context, api API, m
...
 	}
 
 	var redirectInfo params.RedirectErrorInfo
-	err = params.Error{Info: errInfo}.UnmarshalInfo(&redirectInfo)
-	if err != nil {
-		return errors.E(op, fmt.Errorf("cannot unmarshal redirect error info: %w", err))
+	errFromAPI = params.Error{Info: errInfo}.UnmarshalInfo(&redirectInfo)
+	if errFromAPI != nil {
+		return errors.E(op, fmt.Errorf("cannot unmarshal redirect error info: %w", errFromAPI))
 	}
```
kian99: This doesn't seem right, we're overwriting `errFromAPI`. Why not keep this as just `err`

internal/jimm/juju/model_poller.go
SimoneDutto: i need to check the error because if the error is not nil this could be `nil`

internal/jimm/juju/model.go
SimoneDutto: no i can't because checkModelMigratedInternal has some behavior around the error being nil and Migration.End being populated

internal/jimm/juju/model.go
SimoneDutto: No, we need to return modelInfo as well. other wise this func will always return nil even when successuful

internal/jimm/juju/model_poller.go
SimoneDutto: right, wrong refactor

internal/jimm/juju/model_poller.go
SimoneDutto: right, wrong refactor

internal/jimm/juju/model.go
SimoneDutto: i would rather mask it a level higher, like so:
```go
modelInfo, err := j.modelInfo(ctx, &m, api)
if err != nil {
zapctx.Error(ctx, "failed to get model info", zap.Error(err), zap.String("model", m.UUID.String))
return nil, errors.E(op, "internal server error")
}
```

internal/jimm/juju/model.go
kian99: What if we handle the non-error case separately then? It seems like it would simplify the readability.

internal/jimm/juju/model.go
SimoneDutto: sike, we can't do that bucause we need to preserve the error code

kian99: But you can't mask it higher because if the controller returned a normal error you want to display that? We know for sure we want to mask the error from `maybeCleanupModel`.

SimoneDutto: we can do what you've suggested but it would be one of the very first place we do error masking.
Usually we don't do it, right?

internal/jimm/juju/model.go
SimoneDutto: I don't think it does, because you would need to check `err == nil && migrationmode and migration.end != nil` just to return early

internal/jimm/juju/model.go
luci1900: Ideally we'd wrap the "internal" error and let it bubble up, relying on an upper level not letting the client see more than they're allowed.

I think it may make most sense to wrap it in `internal server error` and log where it happens.

SimoneDutto: i'm ok with starting now, but we don't usually do it

internal/jimm/juju/model_poller.go
SimoneDutto: Can't do that because in the checkmigration method we are casting the error to retrieve the redirect info
```
// Parse the redirect error to get the new controller details.
errInfo := errors.ErrorInfo(errFromAPI)
```

internal/jimm/juju/model.go
kian99: I think the reason we don't normally mask the error is that we don't call these cleanup routines from within a handler that does something else.

internal/jimm/juju/model_poller.go
kian99: Mmm fair

internal/jimm/juju/model.go
kian99: Yeah but the way the code is now it is super gnarly

internal/jimm/juju/model.go
SimoneDutto: masking done

internal/jimm/juju/model.go
SimoneDutto: i've made the code a little bit better by having the early return for the modeInfo not found at the top of the `case` stm

internal/jimm/juju/model.go
kian99: This doesn't make much sense to me. From `PollModels` we should just call this new `modelInfo` info avoid the switch in `PollModels`. Even if that means the migrated models will end up making an extra modelInfo call, that's okay.

internal/jimm/juju/model.go
```diff
@@ -261,23 +261,53 @@ func (j *JujuManager) ModelInfo(ctx context.Context, user *openfga.User, mt name
...
+		// propagate the error to the caller.
+		return nil, errFromAPI
+	case dbmodel.MigrationModeMigrateInternal:
+		err = j.checkModelMigratedInternal(ctx, errFromAPI, model, errFromAPI == nil && modelInfo.Migration.End != nil)
+		if err != nil {
+			zapctx.Error(ctx, "error checking model migration", zap.Error(err))
+			return nil, errors.E("internal server error")
+		}
+		// If the model has been migrated internally, we call api.ModelInfo again
```
kian99, line 301: What if we just return an error with a code like `ErrModelMigrated` that the ModelInfo() call can use to repeat the call to `modelInfo` and the poller can ignore?

internal/jimm/juju/model.go
SimoneDutto: do you think it's fine to do the extra modelInfo call? I didn't want to add this overhead to the poller

internal/jimm/juju/model.go
```diff
@@ -261,23 +261,56 @@ func (j *JujuManager) ModelInfo(ctx context.Context, user *openfga.User, mt name
...
+	var err error
+	switch model.MigrationMode {
+	case dbmodel.MigrationModeNone:
+		if errFromAPI == nil {
+			return modelInfo, nil
+		}
+		err = j.maybeCleanupModel(ctx, errFromAPI, model)
+		if err != nil {
+			zapctx.Error(ctx, "error checking model migration", zap.Error(err))
+			return nil, errors.E("internal server error")
```
alesstimec, line 291: `return nil, errFromAPI`

the `err` resulting from `maybeCleanupModel` is a side effect of internal logic..

internal/jimm/juju/model.go
alesstimec: what i don't understand why we have to do this. If the model migration failed, we expect the model to still be on the same controller it was before the migration..

internal/jimm/juju/model.go
```diff
@@ -261,23 +261,86 @@ func (j *JujuManager) ModelInfo(ctx context.Context, user *openfga.User, mt name
...
+			model.MigrationFailed()
+			if err := j.Database.UpdateModel(ctx, model); err != nil {
+				return nil, errors.E(fmt.Errorf("failed to update model after failed migration: %w", err))
 			}
+			return modelInfo, nil
 		}
-		return nil, errors.E(op, err)
+		return modelInfo, nil
+	// If the migration mode is exporting or importing, we return the model info as is.
+	case dbmodel.MigrationModeExporting, dbmodel.MigrationModeImporting:
```
kian99: I would combine this case with the `MigrationModeNone` case.

kian99 (approved):
Nice! Ty for this

alesstimec (approved):

PR #1684, opened 2025-08-22
feat: fix integration tests and add juju-dashboard to docker compose

kian99:
## Description
This PR fixes the integration test workflow (see below) and adds the juju-dashboard to our docker compose. Now `docker compose --profile dev up` will also bring up a dashboard container that you can use to interact with JIMM.

The dashboard doesn't currently support being configured via environment variables as it is normally operated by the charm but I have discussed with the web team the idea that we would configure the dashboard by only supplying some environment variables.

This work is a precursor to running the dashboards integration tests with JIMM periodically/on release.

--------------------

Additionally, the integration test action started failing recently and it seems to be related to memory issues as when I tried to debug the issue by SSH'ing into the runner, the SSH connection is closed once we reach the stage where we compile the `jaas` CLI plugin. By switching to a Canonical self-hosted runner, the integration test now works successfully.

Partially fixes [JUJU-7937](https://warthogs.atlassian.net/browse/JUJU-7937)

[JUJU-7937]: https://warthogs.atlassian.net/browse/JUJU-7937?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/local/dashboard/nginx.conf
+++ b/local/dashboard/nginx.conf
@@ -1,0 +1,31 @@
+# The following is the Nginx configuration for the Juju Dashboard.
+# In the future we'd like to remove this file and supply only necessary
+# configuration via env vars.
+
+map $http_upgrade $connection_upgrade {
+    default upgrade;
+    '' close;
+}
+
+server {
+    listen 8080 default_server;
+    listen [::]:8080 default_server;
+
+    root /srv;
+    index index.html;
+
+    server_name _;
+
+    location / {
+        try_files /index.html =404;
+    }
+
+    location /config.js {
+        try_files /config.js =404;
+    }
+
+    location /assets {
+        try_files $uri =404; 
+    }
+
+}
--- a/.github/workflows/integration-test.yaml
+++ b/.github/workflows/integration-test.yaml
@@ -12,6 +12,11 @@ jobs:
       - name: Checkout JIMM repo
         uses: actions/checkout@v4
 
+      - name: Setup tmate session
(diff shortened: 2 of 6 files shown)


local/dashboard/nginx.conf
SimoneDutto, line 1: why do we need nginx for the docker compose? We use traefik, shouldn't be enough?

SimoneDutto (approved):
other than the question LGTM

local/dashboard/nginx.conf
kian99: The Juju dashboard uses nginx in their container to serve static files. 

The reason we need a copy of the nginx config is something we can hopefully improve with the dashboard team.

It becomes a little clearer if you take a look at their container's entrypoint [here](https://github.com/canonical/juju-dashboard/blob/main/entrypoint) you can see it's just copying a base template nginx.conf file and it mentions that the template is normally filled up the charm. We need that base template to be a a full file since we don't have charm logic here.

Running their image as a k8s charm will override the entrypoint with pebble and the charm code contains the nginx conf template, fills it in, then does basically what the entrypoint does. 

Eventually we'd be able to get rid of the nginx file and configure the nginx setup via environment variables, but I think we'd need to change the docker image + entrypoint + charm. So until then, we can just keep a copy of the nginx conf.

alesstimec (approved):

PR #1687, opened 2025-08-26
feat: add controller config to jimm's db

SimoneDutto:
## Description
Add controller-proxy to jimm's db, add crud methods and ToProxier method.
I've done the schema and the code to match as much as possible juju's code.

Only interesting bit is that the `ToProxier` takes a factory interface, in this way we can build test with our custom factory returning a mock proxy. In production we will most likely use the default factory coming from juju.

### DIFF
--- a/internal/db/controllerproxy.go
+++ b/internal/db/controllerproxy.go
@@ -1,0 +1,48 @@
+package db
+
+import (
+	"context"
+
+	"github.com/canonical/jimm/v3/internal/dbmodel"
+	"github.com/canonical/jimm/v3/internal/errors"
+	"github.com/canonical/jimm/v3/internal/servermon"
+)
+
+// GetControllerProxy retrieves the proxy configuration for the specified controller ID.
+func (d *Database) GetControllerProxy(ctx context.Context, controllerID uint) (proxy *dbmodel.ControllerProxy, err error) {
+	const op = errors.Op("db.GetControllerProxy")
+	if err := d.ready(); err != nil {
+		return nil, errors.E(op, err)
+	}
+
+	durationObserver := servermon.DurationObserver(servermon.DBQueryDurationHistogram, string(op))
+	defer durationObserver()
+	defer servermon.ErrorCounter(servermon.DBQueryErrorCount, &err, string(op))
+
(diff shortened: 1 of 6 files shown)


kian99 (approved):
Nice, this is slick

internal/db/controllerproxy.go
```diff
@@ -0,0 +1,48 @@
...
+
+	proxy = &dbmodel.ControllerProxy{}
+	if err := db.Where("controller_id = ?", controllerID).First(proxy).Error; err != nil {
+		return nil, errors.E(op, dbError(err))
+	}
+	return proxy, nil
+}
+
+// AddControllerProxy adds a new proxy configuration for the specified controller ID.
+func (d *Database) AddControllerProxy(ctx context.Context, proxy *dbmodel.ControllerProxy) (err error) {
```
alesstimec: what if `proxy == nil`?

alesstimec: should we validate that `proxy.ControllerId` is specified, `Type` or `Config` not empty?

internal/dbmodel/controllerproxy.go
```diff
@@ -0,0 +1,44 @@
...
+	CreatedAt time.Time
+	UpdatedAt time.Time
+
+	ControllerId uint
+
+	Type   string
+	Config Map
+}
+
+type ProxyFactory interface {
```
alesstimec, line 27: godoc, please

internal/db/controllerproxy.go
SimoneDutto: if controllerId is not specified you'll get an error from postgresql saying the foreign key doesn't exist.

For `Type` and `Config` they are defined as `type TEXT NOT NULL,
    config JSONB NOT NULL,`

internal/dbmodel/controllerproxy.go
SimoneDutto: fixed

internal/db/controllerproxy.go
SimoneDutto: for `proxy == nil` you are right, for create method i can just use the struct and not the pointer

alesstimec (approved):

PR #1688, opened 2025-08-27
feat: add jaas command to list migration targets

kian99:
## Description
This change adds a new command to the `jaas` plugin that utilizes the newly introduced `list-migration-targets` method to return a filtered list of controller that are valid migration targets for a model.

Fixes [JUJU-8388](https://warthogs.atlassian.net/browse/JUJU-8388)

## Engineering checklist
- [ ] Documentation updated
- [x] Covered by unit tests

[JUJU-8388]: https://warthogs.atlassian.net/browse/JUJU-8388?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cmd/jaas/cmd/bootstrap_unit_test.go
```diff
@@ -18,15 +18,15 @@ import (
 )
 
 type bootstrapCmdSuite struct {
-	client *mocks.MockJIMMClient
+	client *mocks.MockJIMMAPI
```
kian99, line 21: The reason for this change is that the interface was renamed in a previous PR to `JIMMAPI` from `JIMMClient` but because the interface definition itself wasn't changed the mocks weren't updated. I've updated the interface definition in this PR so I've had to update the mock and usages of it.

luci1900 (approved):
LGTM

cmd/jaas/cmd/listmigrationtargets.go
```diff
@@ -0,0 +1,127 @@
...
+
+This is useful to obtain a filtered list of controllers that meets the following
+criteria:
+- The controller is not the current controller for the model.
+- The controller can deploy to the the same cloud/region as the current controller.
+- The controller is running a compatible Juju version i.e. newer than or equal to
+  the current controller.
+`
+	listMigrationTargetsExamples = `
+	juju list-migration-targets bb933db6-b151-417f-9a62-e3e80b4ebc9b
```
SimoneDutto, line 32: i could be wrong, but this could be one of the few juju command taking an uuid instead of a name. Wouldn't be better to support `name` or `owner/name`? To make the UX more consistent?

SimoneDutto (commented):
just one question

cmd/jaas/cmd/listmigrationtargets.go
kian99: The model migration one takes a name or uuid. To take a name it would be best to make that change to the backend and then modify the CLI. I think that would be a nice upgrade that we can do in the future.

SimoneDutto (approved):

PR #1689, opened 2025-08-27
feat: support controller's proxy for multi storage

SimoneDutto:
Since proxy-config can contain credentials I've moved the proxy-config to vault.
To support that we need to implement the methods to storage proxy config for:
- CredentialStore
- Database
- InMemoryStore

An implementation note: the struct implementing `ToProxier` has been moved one layer up, to `internal/jimm/juju` to make sure the struct logic is not tight to the storage of the proxy config.

### DIFF
--- a/internal/db/controllerproxy.go
+++ b/internal/db/controllerproxy.go
@@ -1,36 +1,61 @@
+// Copyright 2025 Canonical.
+
 package db
 
 import (
 	"context"
+	"encoding/json"
 
 	"github.com/canonical/jimm/v3/internal/dbmodel"
 	"github.com/canonical/jimm/v3/internal/errors"
 	"github.com/canonical/jimm/v3/internal/servermon"
+	"github.com/juju/zaputil/zapctx"
+	"go.uber.org/zap"
 )
 
+const (
+	// These constant are used to create the appropriate identifiers for controller proxy data.
+	controllerProxySecretKind = "controllerProxy"
+)
+
+type controllerProxy struct {
+	Type   string                 `json:"type"`
+	Config map[string]interface{} `json:"config"`
+}
+
 // GetControllerProxy retrieves the proxy configuration for the specified controller ID.
-func (d *Database) GetControllerProxy(ctx context.Context, controllerID uint) (proxy *dbmodel.ControllerProxy, err error) {
-	const op = errors.Op("db.GetControllerProxy")
+func (d *Database) GetControllerProxy(ctx context.Context, controllerName string) (_ string, _ map[string]interface{}, err error) {
+	const op = errors.Op("database.GetControllerProxy")
+
 	if err := d.ready(); err != nil {
-		return nil, errors.E(op, err)
+		return "", nil, errors.E(op, err)
 	}
 
 	durationObserver := servermon.DurationObserver(servermon.DBQueryDurationHistogram, string(op))
 	defer durationObserver()
 	defer servermon.ErrorCounter(servermon.DBQueryErrorCount, &err, string(op))
 
-	db := d.DB.WithContext(ctx)
-
-	proxy = &dbmodel.ControllerProxy{}
-	if err := db.Where("controller_id = ?", controllerID).First(proxy).Error; err != nil {
-		return nil, errors.E(op, dbError(err))
+	secret := dbmodel.NewSecret(controllerProxySecretKind, controllerName, nil)
+	err = d.GetSecret(ctx, &secret)
+	if err != nil {
+		zapctx.Error(ctx, "failed to get secret data", zap.Error(err))
+		return "", nil, errors.E(op, err)
 	}
-	return proxy, nil
+	// secretData is stored as a JSON object in the database.
+	// So we need to unmarshal and extract the relevant fields.
+	var secretData controllerProxy
+	err = json.Unmarshal(secret.Data, &secretData)
+	if err != nil {
+		zapctx.Error(ctx, "failed to unmarshal secret data", zap.Error(err))
+		return "", nil, errors.E(op, err)
+	}
+	return secretData.Type, secretData.Config, nil
 }
 
-// AddControllerProxy adds a new proxy configuration for the specified controller ID.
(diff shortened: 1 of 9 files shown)


internal/jimm/juju/controllerproxy.go
SimoneDutto, line 1: this is just moved from `dbmodel` to here. Don't need to review

internal/jimm/juju/controllerproxy_test.go
SimoneDutto, line 1: this is just moved from `dbmodel` to here. Don't need to review

internal/jimm/juju/controllerproxy.go
kian99, line 1: I suspect this might need to move again. Iirc our controller dial logic is inside some `rpc` package, that will need this so it might be better to go into the `rpc` package but let's see.

internal/db/controllerproxy.go
```diff
@@ -1,36 +1,76 @@
...
+		return "", nil, errors.E(op, err)
 	}
-	return proxy, nil
+	var secretData map[string]interface{}
+	err = json.Unmarshal(secret.Data, &secretData)
+	if err != nil {
+		zapctx.Error(ctx, "failed to unmarshal secret data", zap.Error(err))
+		return "", nil, errors.E(op, err)
+	}
+	typeI, ok := secretData[typeKey]
```
kian99: Can you put a comment above this line to explain what we're doing regarding extracting the secret data?
Maybe it's just because it's been a while since I looked at this logic but it's a bit confusing.

internal/db/controllerproxy.go
```diff
@@ -1,36 +1,76 @@
...
+	"github.com/juju/zaputil/zapctx"
+	"go.uber.org/zap"
+)
+
+const (
+	// These constant are used to create the appropriate identifiers for controller proxy data.
+	controllerProxyKind = "controllerProxy"
+	controllerProxyTag  = "controllerProxy"
+	typeKey             = "type"
+	configKey           = "config"
```
kian99: These consts deserve better names so we know that they're for the controller proxy secret storage and avoid collisions.

internal/db/controllerproxy.go
```diff
@@ -39,10 +79,35 @@ func (d *Database) AddControllerProxy(ctx context.Context, proxy dbmodel.Control
...
+	defer durationObserver()
+	defer servermon.ErrorCounter(servermon.DBQueryErrorCount, &err, string(op))
 
-	if err := db.Create(&proxy).Error; err != nil {
-		return errors.E(op, dbError(err))
+	secret := dbmodel.NewSecret(controllerProxyKind, controllerName, nil)
+	err = d.DeleteSecret(ctx, &secret)
+	if err != nil {
+		zapctx.Error(ctx, "failed to cleanup controller proxy", zap.Error(err))
+		return errors.E(op, err, "failed to cleanup controller proxy")
```
kian99:
```suggestion
		return errors.E(op, err, "failed to delete controller proxy")
```

internal/db/controllerproxy.go
```diff
@@ -1,36 +1,76 @@
...
+	}
+	configI, ok := secretData[configKey]
+	if !ok {
+		return "", nil, errors.E(op, "missing config")
+	}
+	configMap, ok := configI.(map[string]interface{})
+	if !ok {
+		return "", nil, errors.E(op, "config is not a map")
+	}
+	return typeStr, configMap, nil
```
kian99: typeStr is the proxyType right? Maybe we can change the name

kian99 (commented):
Almost lgtm

internal/jimm/juju/controllerproxy.go
SimoneDutto: yeah maybe that's the case, but for now i would leave it here with the tests and everything, so when we move it it has tests

internal/db/controllerproxy.go
```diff
@@ -39,10 +77,35 @@ func (d *Database) AddControllerProxy(ctx context.Context, proxy dbmodel.Control
 	defer durationObserver()
 	defer servermon.ErrorCounter(servermon.DBQueryErrorCount, &err, string(op))
 
-	db := d.DB.WithContext(ctx)
+	secretData := make(map[string]interface{})
```
kian99: What if you made this a struct in this package with fields then you can share the logic between then Put and Get

internal/testutils/jimmtest/store.go
```diff
@@ -189,3 +190,75 @@ func (s *InMemoryCredentialStore) PutJWKSExpiry(ctx context.Context, expiry time
 
 	return nil
 }
+
+// controllerProxyConfig holds proxy configuration for a controller.
+type controllerProxyConfig struct {
+	proxyType string
+	config    map[string]interface{}
```
luci1900, line 197: Does this have to be untyped?

SimoneDutto: yeah, because the proxier factory takes an untyped config map to create the Proxier based on the type

internal/db/controllerproxy.go
SimoneDutto: yeah! done it

docker-compose.yaml
```diff
@@ -138,19 +138,19 @@ services:
...
-    labels:
-      traefik.enable: true
-      traefik.http.routers.juju-dashboard.rule: Host(`juju-dashboard.localhost`)
-      traefik.http.routers.juju-dashboard.entrypoints: websecure
-      traefik.http.routers.juju-dashboard.tls: true
-      traefik.http.services.juju-dashboard.loadbalancer.server.port: 8080
-    volumes:
-      - ./local/dashboard/config.js:/srv/configs/config.js
-      - ./local/dashboard/nginx.conf:/srv/configs/nginx.conf
+  # juju-dashboard:
```
kian99: Was this for testing?

kian99 (approved):
Nice, ty for this!

docker-compose.yaml
SimoneDutto: my stupid notebook is slower than my `git yeet`

luci1900 (approved):

PR #1690, opened 2025-08-28
feat: add auth style param and lower keycloak poll interval

kian99:
## Description
This PR lowers the polling interval between JIMM and Keycloak in our dev environment used during login via the device-flow.

The polling interval is a parameter sent by the OIDC provider to the client so we only need to change the Keycloak config in order for JIMM to start polling more frequently.

Unfortunately when I tested this, I still couldn't login any faster via the CLI and after some debugging I noticed we were still receiving "slow_down" errors from the oidc provider which causes the poll interval to increase by 5 seconds each time we get this error, see [here](https://cs.opensource.google/go/x/oauth2/+/master:deviceauth.go;l=185). The cause for this turns out to be because of how different oauth providers expect credentials and in the x/oauth2 library there is a useful comment [here](https://cs.opensource.google/go/x/oauth2/+/master:internal/token.go;l=231;drc=014cf778b444f29c82ececa4f3ec1f6fe3db3eaf), c/p below,
>// If we get an error, assume the server wants the
// clientID & clientSecret in a different form.
// See https://code.google.com/p/goauth2/issues/detail?id=31 for background.
// In summary:
// - Reddit only accepts client secret in the Authorization header
// - Dropbox accepts either it in URL param or Auth header, but not both.
// - Google only accepts URL param (not spec compliant?), not Auth header
// - Stripe only accepts client secret in Auth header with Bearer method, not Basic
//
// We used to maintain a big table in this code of all the sites and which way
// they went, but maintaining it didn't scale & got annoying.
// So just try both ways.

So whenever we poll the oidc server and the user has not yet completed login, we do it twice, receive a slow_down error and then end up increasing our polling interval. I've added a simple fix to allow us to define the auth style via config. If a valid auth style is defined we will use that, otherwise we fall back to the auto-detection logic.

Now `juju login` is speedy once you've completed your login at the identity provider.

Fixes [JUJU-8427](https://warthogs.atlassian.net/browse/JUJU-8427)

## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests

## Test instructions
I haven't added any tests but you can validate this by starting jimm and running `juju login`. Before the fix this would take up to 6 seconds, after the fix it is consistently <1s.

[JUJU-8427]: https://warthogs.atlassian.net/browse/JUJU-8427?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


local/keycloak/jimm-realm.json
```diff
@@ -24,7 +24,7 @@
   "actionTokenGeneratedByAdminLifespan": 43200,
   "actionTokenGeneratedByUserLifespan": 300,
   "oauth2DeviceCodeLifespan": 600,
-  "oauth2DevicePollingInterval": 5,
+  "oauth2DevicePollingInterval": 1,
```
luci1900, line 27: Could this be even lower? A whole second for retries can add up.

kian99: Unfortunately it doesn't seem like it. The value is an integer so I can't get it any lower.

luci1900 (approved):
I like enum-style env vars over booleans.

SimoneDutto (approved):
very interesting

PR #1692, opened 2025-08-29
feat: handle model migrations from the model proxy

kian99:
## Description
This PR adds some handling for models undergoing a migration from within the model-proxy.

Initially the first commit demonstrates my attempt at completing a model migration when we get back the expected error from Juju. I.e. we know a model is migrating, Juju returns the expected error, the model-proxy calls the appropriate logic to complete the migration, the model proxy returns the original error to the user. On the user's next websocket connection, the new controller will be dialed. Unfortunately the model proxy can't change the controller it is connected to after it has been instantiated.

The above approach works but very quickly becomes over-complicated imo. In order for the model-proxy to call the correct business logic to complete the migration we have to jump through several hoops. We already do similar in places where it is absolutely necessary but I don't think it is here.

So what I've done instead is to revert the initial changes and gone with a smaller approach where, if a model is migrating and we get the expected error back from Juju, return a masked error with the message "model is finishing migration, please retry later". My thinking is that the model will soon be updated either via the poller or calls to `model-info` and we avoid over-complicating the model proxy.

All this now does is change the error you get when you run `juju status` and the model has been migrated from `error: invalid entity name or password (unauthorized)` to `error: model is finishing migration, please retry later (model migrating)`

I haven't added any tests yet as I'd like to gather feedback on the approach. 

Fixes [JUJU-8365](https://warthogs.atlassian.net/browse/JUJU-8365)

## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests

[JUJU-8365]: https://warthogs.atlassian.net/browse/JUJU-8365?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/rpcproxy/rpcproxy.go
+++ b/internal/rpcproxy/rpcproxy.go
@@ -34,6 +34,7 @@ import (
 
 const (
 	accessRequiredErrorCode = "access required"
+	unauthorizedErrorCode   = "unauthorized"
 )
 
 // ControllerDetails holds information about the controller
@@ -94,6 +95,7 @@ type WebsocketConnectionWithMetadata struct {
 	ControllerUUID string
 	ModelName      string
 	ModelUUID      string
+	MigrationMode  dbmodel.MigrationMode
 }
 
 // LoginService represents the LoginService interface used by the proxy.
@@ -297,6 +299,7 @@ type modelProxy struct {
 	loginService            LoginService
 	modelName               string
 	modelUUID               string
+	modelMigrationMode      dbmodel.MigrationMode
 	conversationId          string
 	authenticatedIdentityID string
 	redirectInfo            RedirectInfoGetter
@@ -475,6 +478,7 @@ func (p *clientProxy) makeControllerConnection(ctx context.Context) error {
 		p.msgs.controllerUUID = connWithMetadata.ControllerUUID
 		p.modelName = connWithMetadata.ModelName
 		p.modelUUID = connWithMetadata.ModelUUID
+		p.modelMigrationMode = connWithMetadata.MigrationMode
 		p.dst = &writeLockConn{conn: connWithMetadata.Conn}
 		controllerToClient := controllerProxy{
 			modelProxy: modelProxy{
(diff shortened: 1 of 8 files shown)


luci1900:
In general, I think it's reasonable to tell users to retry later as long as we're fairly certain this state won't be permanent and they can in fact retry after some reasonable "later".

Whether we do this or not, it's an opportunity to incorrectly race over data.

SimoneDutto:
At some point we might add the `reactToModelProxyError`, which would delete a model from jimm's state if it's `not-found`. But i guess it would be much more difficult to do it for the retry error because it would require a `re-dial` of the new controller.
And the code didn't seem to support this rework very easily. So I agree with @kian99 here and we should just return an "retry-later"

internal/rpcproxy/rpcproxy.go
```diff
@@ -553,6 +540,54 @@ func (p *controllerProxy) start(ctx context.Context) error {
...
+func (p *controllerProxy) processControllerErrors(ctx context.Context, msg *message) bool {
+
+	// Check if the model is migrating. When it has completed its migration, we will receive
+	// an unauthorized error from the controller and want to mask that error and inform
+	// client to try again as JIMM will eventually update the model's controller details
+	// in its database. See internal/jimm/juju/model_poller.go.
+	modelMigrating := p.modelMigrationMode == dbmodel.MigrationModeMigrateInternal || p.modelMigrationMode == dbmodel.MigrationModeExporting
+	if modelMigrating && msg.ErrorCode == unauthorizedErrorCode {
+		msg.ErrorCode = "model migrating"
+		msg.Error = "model is finishing migration, please retry later"
```
SimoneDutto, line 556: is a "." needed here?

SimoneDutto (commented):
With some tests it's all good to me.

An idea could be to call `modelinfo` (or create a model called `syncModel`) within jimm on the model to force the sync, so we are sure that the "retry later" is sooner rather than later

internal/rpcproxy/rpcproxy.go
kian99: I don't think we normally use full stops in the error messages.

internal/rpcproxy/rpcproxy.go
```diff
@@ -553,6 +541,54 @@ func (p *controllerProxy) start(ctx context.Context) error {
...
+// processControllerErrors checks for errors in the message from the controller
+// and decides how to handle them. It returns true if the message should be
+// returned to the client, false if it should not.
+func (p *controllerProxy) processControllerErrors(ctx context.Context, msg *message) bool {
+
+	// Check if the model is migrating. When it has completed its migration, we will receive
+	// an unauthorized error from the controller and we want to mask that error and inform
+	// clients to try again as JIMM will eventually update the model's controller.
+	// See internal/jimm/juju/model_poller.go.
+	modelMigrating := p.modelMigrationMode == dbmodel.MigrationModeMigrateInternal || p.modelMigrationMode == dbmodel.MigrationModeExporting
```
luci1900, line 553: I'm struggling to trace where this field comes from. How does this know that a migration is in fact taking place?

kian99: Yeah this one is definitely not the clearest. When the model proxy is instantiated it receives a set of dependencies, one of those is the `createControllerConn` which is just a callback that will create a websocket connection to the desired controller. It also returns some metadata about the destination like the model UUID, etc.

Then inside this file there are 2 important routines running, one forwards messages from the client -> controller (clientProxy struct) and the other from controller -> client (controllerProxy struct). There is a base struct call `modelProxy` that both of the aforementioned structs inherit from.

And then to finally answer your question, the `p.modelMigrationMode` is set [here](https://github.com/canonical/jimm/pull/1692/files#diff-b9913f07170c61665158f33bb1372a219490dc7d967db39758b6ca8ea401fa12R492) when we call `createControllerConn` and receive the metdata, to which I've added the migration mode.

luci1900: Right, so it knows after having asked a controller over a websocket.

Should be safe, then.

luci1900 (approved):

SimoneDutto (approved):

PR #1693, opened 2025-08-29
feat: use jujuparams and not jujucloud

ale8k:
## Description
In my original PR, we used jujucloud.CloudCredential (because it was easier), it enabled us to send a set of credentials and not worry about which the user has picked as juju will iterate over them. Now, we're explicitly sending a single credential and taking the first. 

It is picking the default credential unless otherwise specified by a new --credential flag.

This also needs Q/A'ing before merging.  So do not merge this until it is QA'd. Once #1682 is merged, this can be rebased with v3 and QA'd by updating the env to point at JIMM.

## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests

### DIFF
--- a/internal/jujuapi/jimm.go
+++ b/internal/jujuapi/jimm.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/google/uuid"
+	"github.com/juju/juju/cloud"
 	"github.com/juju/juju/cmd/juju/common"
 	"github.com/juju/juju/core/network"
 	jujuparams "github.com/juju/juju/rpc/params"
@@ -643,6 +644,14 @@ func (r *controllerRoot) BootstrapStart(ctx context.Context, req apiparams.Boots
 		cloudNameAndRegion = fmt.Sprintf("%s/%s", req.CloudName, req.RegionName)
 	}
 
+	in := cloud.NewCredential(
+		cloud.AuthType(req.Credential.AuthType),
+		req.Credential.Attributes,
+	)
+
+	cloudCred := cloud.NewEmptyCloudCredential()
+	cloudCred.AuthCredentials["default"] = in
+
 	params := bootstrap.BootstrapParams{
 		CLIVersion: "3.6.8",
 
@@ -651,7 +660,7 @@ func (r *controllerRoot) BootstrapStart(ctx context.Context, req apiparams.Boots
 		BootstrapTimeout:   req.Flags.Timeout,
 
 		PersonalCloud: cloudFromParams(req.CloudName, req.Cloud),
-		CloudCred:     req.Credential,
+		CloudCred:     *cloudCred,
 	}
 
 	jobID, err := r.jimm.BootstrapManager().StartBootstrap(ctx, r.user, params)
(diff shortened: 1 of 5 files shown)


kian99 (commented):
I would just be explicit here, have the user pick the credential they want to use so command becomes something like `juju jaas bootstrap <other-position-args> --credential foo`. I would also make it a flag rather than an argument because too many arguments becomes very hard to read.

ale8k:
> I would just be explicit here, have the user pick the credential they want to use so command becomes something like `juju jaas bootstrap <other-position-args> --credential foo`. I would also make it a flag rather than an argument because too many arguments becomes very hard to read.

We can default to first and also add the flag - maybe that's best? I'll update this PR with the flag


I've updated it to just use default and not first (which I think is maybe right?) and also added a --credential flag.

internal/jujuapi/jimm.go
```diff
@@ -643,6 +644,14 @@ func (r *controllerRoot) BootstrapStart(ctx context.Context, req apiparams.Boots
 		cloudNameAndRegion = fmt.Sprintf("%s/%s", req.CloudName, req.RegionName)
 	}
 
+	in := cloud.NewCredential(
+		cloud.AuthType(req.Credential.AuthType),
+		req.Credential.Attributes,
+	)
+
+	cloudCred := cloud.NewEmptyCloudCredential()
+	cloudCred.AuthCredentials["default"] = in
```
ale8k, line 653: This "might" be wrong, me and Joe did check it out about using the term default and not setting DefaultCredential explicitly and it seems ok, but QA should discover if this is wrong and the string explicitly needs setting.

luci1900 (approved):
I like explicit.

internal/jujuapi/jimm.go
SimoneDutto, line 653: can we extract "default" to a const of some sort?

SimoneDutto (commented):
when we say "it might need QA" why aren't we doing the qa before proposing the PR? Otherwise we need to merge this out of trust and then we need another pr to fix it

Maybe there is something blocking the QA, i dont' know

internal/jujuapi/jimm.go
ale8k: Juju doesn't use a const :(. You can also name it whatever you like as long as the default cred string is set. I can't see in the code what it does though if it isn't set, so I need to investigate that (otherwise I'm just gonna explicitly set it to default and create a const on our side.

kian99: I'd also like to see some more investigation before we land stuff that we only think might work. But considering it's still a wip, I'll aprove.

kian99 (approved):

SimoneDutto (approved):

ale8k:
This is proving to be difficult, going to close it and come back later.

kian99:
> This is proving to be difficult, going to close it and come back later.

Do we have a follow-up Jira card for this?

ale8k:
> > This is proving to be difficult, going to close it and come back later.
> 
> Do we have a follow-up Jira card for this?

I'm making one now.

ale8k:
Here: https://warthogs.atlassian.net/browse/JUJU-8531

PR #1694, opened 2025-09-03
Add GoReleaser support

luci1900:
## Description

Add GoReleaser support, including GitHub action.

## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests

### DIFF
--- a/.github/workflows/release-binaries.yaml
+++ b/.github/workflows/release-binaries.yaml
@@ -1,0 +1,35 @@
+name: Release binaries
+
+on:
+  push:
+    tags:
+      - 'v*'
+
+# GitHub considers creating releases and uploading assets as writing contents.
+permissions:
+  contents: write
+
+jobs:
+  goreleaser:
+    runs-on: ubuntu-22.04
+    environment: Release Environment
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          # Allow goreleaser to access older tag information.
+          fetch-depth: 0
+
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version-file: 'go.mod'
+          cache: true
+  
+      - name: Run GoReleaser
+        uses: goreleaser/goreleaser-action@v6
+        with:
+          args: release --clean
+        env:
+          # GitHub sets the GITHUB_TOKEN secret automatically.
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -1,0 +1,49 @@
+version: 2
+
+builds:
+  - id: jimmsrv
+    main: ./cmd/jimmsrv
+    mod_timestamp: '{{ .CommitTimestamp }}'
+    flags:
+      # Avoid binary paths being absolute builder paths
+      - -trimpath
+    ldflags:
+      - '-X main.version={{.Version}} -X main.commit={{.Commit}}'
+    env:
+      - CGO_ENABLED=0
+    goos:
+      - linux
+      - darwin
+    goarch:
+      - amd64
+      - arm64
+    binary: 'jimmsrv_v{{ .Version }}'
+
+  - id: jaas
+    main: ./cmd/jaas
+    mod_timestamp: '{{ .CommitTimestamp }}'
+    flags:
+      - -trimpath
+    ldflags:
+      - '-X main.version={{.Version}} -X main.commit={{.Commit}}'
+    env:
+      - CGO_ENABLED=0
+    goos:
+      - linux
+      - darwin
+    goarch:
+      - amd64
+      - arm64
+    binary: 'jaas_v{{ .Version }}'
+
+archives:
+  - formats: [tar.gz]
+    # this name template makes the OS and Arch compatible with the results of `uname`.
+    name_template: '{{ .ProjectName }}_{{ .Version }}_{{ .Os }}_{{ .Arch }}'
+
+changelog:
+  sort: asc
+  use: github
+
+release:
+  prerelease: auto
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,4 @@ db_schemas
 
 *.test
 __debug_bin*
+dist/


.github/workflows/release-binaries.yaml
```diff
@@ -0,0 +1,43 @@
...
+        uses: actions/setup-go@v5
+        with:
+          go-version-file: 'go.mod'
+          cache: true
+
+      - name: Import GPG key
+        id: import_gpg
+        uses: crazy-max/ghaction-import-gpg@01dd5d3ca463c7f10f7f4f7b4f177225ac661ee4 # => v5
+        with:
+          gpg_private_key: ${{ secrets.GPG_PRIVATE_KEY }}
```
kian99: I assume we'll need to create these secret to sign our releases.
Also wondering if we should rather do the signing in a follow-up.

.goreleaser.yaml
```diff
@@ -0,0 +1,62 @@
+version: 2
+
+before:
+  hooks:
+    - go mod tidy
```
kian99: Is there a need for this? If we don't trigger an error if there was a change it doesn't help us I believe. And I'd rather the build fail if the go.mod is out of date.

.goreleaser.yaml
```diff
@@ -0,0 +1,62 @@
...
+
+builds:
+  - id: jimmsrv
+    main: ./cmd/jimmsrv
+    mod_timestamp: '{{ .CommitTimestamp }}'
+    flags:
+      - -trimpath
+    ldflags:
+      - '-s -w -X main.version={{.Version}} -X main.commit={{.Commit}}'
```
kian99: Could you leave some comments explaining the need for timpath and the -s and -w flags.

.goreleaser.yaml
```diff
@@ -0,0 +1,62 @@
...
+      - linux
+      # - darwin
+      # - windows
+    goarch:
+      - amd64
+      # - arm64
+    ignore:
+    # TODO: disabled => juju/utils symlink_windows.go:54:8: undefined: createSymbolicLink
+    - goos: windows
+      goarch: arm64
```
kian99: Since we're only building for linux anyway, do we need this exclusion?

.goreleaser.yaml
```diff
@@ -0,0 +1,62 @@
...
+      - '-s -w -X main.version={{.Version}} -X main.commit={{.Commit}}'
+    env:
+      - CGO_ENABLED=0
+    goos:
+      - linux
+      # - darwin
+      # - windows
+    goarch:
+      - amd64
+      # - arm64
```
kian99: I'd add in arm

.goreleaser.yaml
```diff
@@ -0,0 +1,62 @@
...
+    ignore:
+    # TODO: disabled => juju/utils symlink_windows.go:54:8: undefined: createSymbolicLink
+    - goos: windows
+      goarch: arm64
+    binary: 'jimmsrv_v{{ .Version }}'
+
+  - id: jaas
```
kian99, line 22: Same comments as above for jaas

.goreleaser.yaml
```diff
@@ -0,0 +1,62 @@
...
+    ignore:
+    # TODO: disabled => juju/utils symlink_windows.go:54:8: undefined: createSymbolicLink
+    - goos: windows
+      goarch: arm64
+    binary: 'jaas_v{{ .Version }}'
+
+archives:
+  - formats: [tar.gz]
+    # this name template makes the OS and Arch compatible with the results of `uname`.
+    name_template: '{{ .ProjectName }}_{{ .Version }}_{{ .Os }}_{{ .Arch }}'
```
kian99, line 42: Do we need a special format? I guess it is because we're putting both binaries into a single archive so we need to provide a name?

.goreleaser.yaml
luci1900: Ah, that's debugging I forgot to undo.

.goreleaser.yaml
luci1900: Exactly, I didn't see a different way to do it.

.github/workflows/release-binaries.yaml
luci1900: If we aren't likely to use the binaries anyway, I'd happily remove the signing.

.goreleaser.yaml
luci1900: Not really a need, no. It's just common. I'll remove it.

go.mod
```diff
@@ -339,6 +339,8 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.4.1 // indirect
 )
 
+require github.com/ulikunitz/xz v0.5.12
```
SimoneDutto: do we need this?

.goreleaser.yaml
```diff
@@ -0,0 +1,58 @@
...
+    flags:
+      - -trimpath
+    ldflags:
+      - '-s -w -X main.version={{.Version}} -X main.commit={{.Commit}}'
+    env:
+      - CGO_ENABLED=0
+    goos:
+      - linux
+      - darwin
+      - windows
```
SimoneDutto: do we care about windows? i say no

.goreleaser.yaml
```diff
@@ -0,0 +1,58 @@
...
+  - formats: [tar.gz]
+    # this name template makes the OS and Arch compatible with the results of `uname`.
+    name_template: '{{ .ProjectName }}_{{ .Version }}_{{ .Os }}_{{ .Arch }}'
+
+changelog:
+  sort: asc
+  use: github
+
+release:
+  prerelease: auto
```
SimoneDutto, line 49: what's the purpose of this?

SimoneDutto (commented):
some comments, have you tried the releaser on your fork?

.goreleaser.yaml
luci1900: I copied them from https://github.com/juju/terraform-provider-juju/blob/main/.goreleaser.yml

trimpath is to avoid builder-absolute paths.

Not sure about the other two.

.goreleaser.yaml
luci1900: Technically we can build for Windows, just not ARM.

go.mod
luci1900: That's just the result of running `go mod tidy`. I'll revert.

.goreleaser.yaml
luci1900: To avoid something like `-rc1` being marked as a release. We have it in the terraform provider https://github.com/juju/terraform-provider-juju/blob/main/.goreleaser.yml#L54

SimoneDutto: yeah, i've added this but it wasn't working properly with hashicorp. Let's hope it work with ghcr

luci1900:
> some comments, have you tried the releaser on your fork?

Yes. https://github.com/luci1900/jimm/actions/runs/17431309462/job/49490164111 released https://github.com/luci1900/jimm/releases/tag/v56

SimoneDutto:
> > some comments, have you tried the releaser on your fork?
> 
> Yes. https://github.com/luci1900/jimm/actions/runs/17431309462/job/49490164111 released https://github.com/luci1900/jimm/releases/tag/v56

niceee

SimoneDutto (approved):
lgtm

.goreleaser.yaml
luci1900: Seems to be this

`  -w    disable DWARF generation`
`  -s    disable symbol table`

.goreleaser.yaml
luci1900: Hmm. I'm not sure how I'd test whether it does work. I'll try something on my fork.

luci1900: Seems to work https://github.com/luci1900/jimm/releases/tag/v62-rc1

.goreleaser.yaml
```diff
@@ -0,0 +1,50 @@
...
+builds:
+  - id: jimmsrv
+    main: ./cmd/jimmsrv
+    mod_timestamp: '{{ .CommitTimestamp }}'
+    flags:
+      # Avoid paths being absolute on the builder
+      - -trimpath
+    ldflags:
+      # -s disables symbol table, -w disables DWARF generation
+      - '-s -w -X main.version={{.Version}} -X main.commit={{.Commit}}'
```
kian99: I'm inclined to say we don't want to build with -s and -w after reading https://stackoverflow.com/questions/29599209/how-to-build-a-release-version-binary-in-go.
But we don't plan on using these releases for our production deployments either so it's not necessarily a problem.

kian99 (approved):

.goreleaser.yaml
luci1900: I'm happy to remove them. I hadn't used them in my previous job, just copied them from the terraform repo.

PR #1695, opened 2025-09-05
feat: security events

SimoneDutto:
## Description
In this PR we add security events for JIMM:
he work is split for different commits to help with the review:
1. refactor the production logger to resemble more the logs for SSLDC
2. add the security logger utils methods
3. login events
4. jimm admin events
5. unauthorized events

Example how the logs look like:

```console
{"level":"error","datetime":"2025-09-05T07:04:17.455Z","caller":"logger/securityevent_logger.go:42","msg":"authz_fail:jimm-test43cc8c@canonical.com","appid":"jimm","event":"authz_fail:jimm-test43cc8c@canonical.com","description":"unauthorized access attempt, error: unauthorized","severity":"CRITICAL"}
```

## Some notes
1.
At the moment the logs are just zapctx logs, in the future (if needed) we could add an interface for logger and log these security events where we want. I didn't do it right now, because these SSDLC specs change a lot from one to another. But feel free to suggest the change if you think it's necessary.

2.
Not really sure what to test here, these are just logs, if our `unauthorized`, `login` logic is right we don't have anything to test.

3.
rebac unauthorized access would be in a follow-up.

### DIFF
--- a/cmd/jimmsrv/service/service.go
+++ b/cmd/jimmsrv/service/service.go
@@ -6,7 +6,6 @@ package service
 
 import (
 	"context"
-	goerr "errors"
 	"fmt"
 	"net/http"
 	"net/url"
@@ -747,7 +746,7 @@ func newOpenFGAClient(ctx context.Context, p OpenFGAParams) (*openfga.OFGAClient
 // controller.
 func ensureControllerAdministrators(ctx context.Context, client *openfga.OFGAClient, controllerUUID string, admins []string) error {
 	controller := names.NewControllerTag(controllerUUID)
-	errs := make([]error, 0, len(admins))
+	tuples := []openfga.Tuple{}
 	for _, username := range admins {
 		userTag := names.NewUserTag(username)
 		i, err := dbmodel.NewIdentity(userTag.Id())
@@ -760,21 +759,20 @@ func ensureControllerAdministrators(ctx context.Context, client *openfga.OFGACli
 			return errors.E(err)
 		}
 		if !isAdmin {
-			tuple := openfga.Tuple{
+			tuples = append(tuples, openfga.Tuple{
 				Object:   ofganames.ConvertTag(userTag),
 				Relation: ofganames.AdministratorRelation,
 				Target:   ofganames.ConvertTag(controller),
-			}
-			err := client.AddRelation(ctx, tuple)
-			if err != nil {
-				errs = append(errs, err)
-				continue
-			}
-			logger.LogGrantJimmAdmin(ctx, username)
+			})
 		}
 	}
-	if len(errs) > 0 {
-		return goerr.Join(errs...)
+	if len(tuples) == 0 {
+		return nil
 	}
+	err := client.AddRelation(ctx, tuples...)
+	if err != nil {
+		return err
+	}
+	logger.LogGrantJimmAdmins(ctx, admins)
 	return nil
 }
--- a/cmd/jimmsrv/service/service.go
+++ b/cmd/jimmsrv/service/service.go
@@ -6,6 +6,7 @@ package service
 
 import (
 	"context"
+	goerr "errors"
 	"fmt"
 	"net/http"
 	"net/url"
@@ -746,7 +747,7 @@ func newOpenFGAClient(ctx context.Context, p OpenFGAParams) (*openfga.OFGAClient
 // controller.
 func ensureControllerAdministrators(ctx context.Context, client *openfga.OFGAClient, controllerUUID string, admins []string) error {
 	controller := names.NewControllerTag(controllerUUID)
-	tuples := []openfga.Tuple{}
+	errs := make([]error, 0, len(admins))
 	for _, username := range admins {
 		userTag := names.NewUserTag(username)
 		i, err := dbmodel.NewIdentity(userTag.Id())
@@ -759,15 +760,21 @@ func ensureControllerAdministrators(ctx context.Context, client *openfga.OFGACli
 			return errors.E(err)
 		}
 		if !isAdmin {
-			tuples = append(tuples, openfga.Tuple{
+			tuple := openfga.Tuple{
 				Object:   ofganames.ConvertTag(userTag),
 				Relation: ofganames.AdministratorRelation,
 				Target:   ofganames.ConvertTag(controller),
-			})
+			}
+			err := client.AddRelation(ctx, tuple)
+			if err != nil {
+				errs = append(errs, err)
+				continue
+			}
+			logger.LogGrantJimmAdmin(ctx, username)
 		}
 	}
-	if len(tuples) == 0 {
-		return nil
+	if len(errs) > 0 {
+		return goerr.Join(errs...)
 	}
-	return client.AddRelation(ctx, tuples...)
+	return nil
 }
--- a/internal/jimm/login/login.go
+++ b/internal/jimm/login/login.go
@@ -15,6 +15,7 @@ import (
 	"github.com/canonical/jimm/v3/internal/db"
 	"github.com/canonical/jimm/v3/internal/dbmodel"
 	"github.com/canonical/jimm/v3/internal/errors"
+	"github.com/canonical/jimm/v3/internal/logger"
 	"github.com/canonical/jimm/v3/internal/openfga"
 	jimmnames "github.com/canonical/jimm/v3/pkg/names"
 )
@@ -161,10 +162,16 @@ func (j *loginManager) LoginClientCredentials(ctx context.Context, clientID stri
 
 	err = j.oAuthAuthenticator.VerifyClientCredentials(ctx, clientID, clientSecret)
 	if err != nil {
+		logger.LogFailedLogin(ctx, clientIdWithDomain)
 		return nil, errors.E(op, err)
 	}
-
-	return j.UserLogin(ctx, clientIdWithDomain)
+	user, err := j.UserLogin(ctx, clientIdWithDomain)
+	if err != nil {
+		logger.LogFailedLogin(ctx, clientIdWithDomain)
+		return nil, errors.E(op, err)
+	}
+	logger.LogSuccessfulLogin(ctx, clientIdWithDomain)
+	return user, nil
 }
 
 // LoginWithSessionToken verifies a user's session token before the user is logged in.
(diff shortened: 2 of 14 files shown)


cmd/jimmsrv/service/service.go
```diff
@@ -733,7 +734,7 @@ func newOpenFGAClient(ctx context.Context, p OpenFGAParams) (*openfga.OFGAClient
 // controller.
 func ensureControllerAdministrators(ctx context.Context, client *openfga.OFGAClient, controllerUUID string, admins []string) error {
```
SimoneDutto: this change is to make sure we add jimm's admin one by one so we can add the security event when it happens.

internal/jujuapi/websocket.go
```diff
@@ -125,14 +127,21 @@ func serveRoot(ctx context.Context, root root, logger auditLogger, wsConn *webso
 	<-conn.Dead()
 }
 
-// mapError maps JIMM errors to errors suitable for use with the juju API.
-func mapError(err error) *jujuparams.Error {
+// r.mapError maps JIMM errors to errors suitable for use with the juju API.
+func (r *controllerRoot) mapError(ctx context.Context, err error) *jujuparams.Error {
```
SimoneDutto, line 125: great suggestion by @alesstimec, so we can access the `user`. Thanks <3

cmd/jimmsrv/service/service.go
```diff
@@ -746,15 +747,21 @@ func ensureControllerAdministrators(ctx context.Context, client *openfga.OFGACli
...
+			tuple := openfga.Tuple{
 				Object:   ofganames.ConvertTag(userTag),
 				Relation: ofganames.AdministratorRelation,
 				Target:   ofganames.ConvertTag(controller),
-			})
+			}
+			err := client.AddRelation(ctx, tuple)
+			if err != nil {
+				errs = append(errs, err)
+				continue
```
alesstimec: i would not do this change.. it means that some (or all) users were not granted admin permissions, which would leave JIMM in an inconsistent state.. i'd rather fail all, no?

internal/jimm/login/login.go
```diff
@@ -161,22 +162,35 @@ func (j *loginManager) LoginClientCredentials(ctx context.Context, clientID stri
 
 	err = j.oAuthAuthenticator.VerifyClientCredentials(ctx, clientID, clientSecret)
 	if err != nil {
+		logger.LogFailedLogin(ctx, clientIdWithDomain)
```
alesstimec, line 165: i'd solve this with a deferred function.. and in that function you log based on the returned error..

internal/jimm/login/login.go
```diff
@@ -161,22 +162,35 @@ func (j *loginManager) LoginClientCredentials(ctx context.Context, clientID stri
...
+	logger.LogSuccessfulLogin(ctx, clientIdWithDomain)
+	return user, nil
 }
 
 // LoginWithSessionToken verifies a user's session token before the user is logged in.
 func (j *loginManager) LoginWithSessionToken(ctx context.Context, sessionToken string) (*openfga.User, error) {
 	const op = errors.Op("jimm.LoginWithSessionToken")
 	jwtToken, err := j.oAuthAuthenticator.VerifySessionToken(sessionToken)
 	if err != nil {
+		logger.LogFailedLogin(ctx, "unkown session token")
```
alesstimec: same here.. deferred function

cmd/jimmsrv/service/service.go
luci1900: Presumably the extra round-trips aren't important in this case, as we won't have hundreds of admins?

luci1900 (approved):
Looks good. I don't see a good way to test this, either.

alesstimec (changes requested):
requires a bit of fixing.. and i would suggest using the defer pattern to log success/failure based on returned error..

internal/jimm/login/login.go
SimoneDutto: i've tried it but i don't like it and i'll explain why.

Using the defer pattern suggest that we should always log a `FailedLogin` in case of error.
But that can be false. 
An example:
```go
func (j *loginManager) LoginWithSessionCookie(ctx context.Context, identityID string) (*openfga.User, error) {
	const op = errors.Op("jimm.LoginWithSessionCookie")
	if identityID == "" {
		return nil, errors.E(op, "missing cookie identity")
	}
	user, err := j.UserLogin(ctx, identityID)
	if err != nil {
		logger.LogFailedLogin(ctx, identityID)
		return nil, errors.E(op, err)
	}
	logger.LogSuccessfulLogin(ctx, identityID)
	return user, nil
}

```

here when the identityId is null we return an error, but this isn't necessarly a FailedLogin event, it could very well be just a `400 bad request`.

So having the defer is wrong for some methods now, and it could wrong in the future as well if we change some logic in the Login methods.

What do you think?

internal/jimm/login/login.go
SimoneDutto: https://github.com/canonical/jimm/pull/1695#discussion_r2324594414

cmd/jimmsrv/service/service.go
```diff
@@ -746,15 +747,21 @@ func ensureControllerAdministrators(ctx context.Context, client *openfga.OFGACli
...
+			}
+			logger.LogGrantJimmAdmin(ctx, username)
 		}
 	}
-	if len(tuples) == 0 {
-		return nil
+	if len(errs) > 0 {
+		return goerr.Join(errs...)
 	}
-	return client.AddRelation(ctx, tuples...)
```
kian99, line 772: Could we just do the logging here if the request returns without an error?

cmd/jimmsrv/service/service.go
SimoneDutto: mmm, yeah. My thinking was that maybe you've put a wrong user in and you don't want JIMM not to start.
But it might be out of scope for this PR

SimoneDutto: i'll revert this change

internal/jimm/login/login.go
alesstimec: well.. the method in your example is called `LoginWithSessionCookie`.. If somebody called it without a cookie, i'd say that is a failed login attempt..

internal/rpcproxy/rpcproxy.go
```diff
@@ -307,11 +308,18 @@ type modelProxy struct {
...
+func (p *modelProxy) sendError(ctx context.Context, socket *writeLockConn, req *message, err error) {
 	if req == nil {
 		// If there was no message to error on, just return.
 		return
 	}
+	if errors.ErrorCode(err) == errors.CodeUnauthorized {
+		logger.LogUnauthorizedAccess(
+			ctx,
+			p.tokenGen.GetUser().String(),
+			fmt.Sprintf("unauthorized access to model proxy for model %s", p.modelUUID),
```
kian99: Slight change in wording and I'd log the error alongside the message, wdyt?
```suggestion
			fmt.Sprintf("unauthorized access in model proxy for model %s:", p.modelUUID),
```

internal/jimm/login/login.go
SimoneDutto: in theory yes, but it's an empty user so the security event will just be useless right?

Basically I don't like the idea of implying that each and every error is a failed login when we call this method.

@luci1900 do you have an opinion on this?

cmd/jimmsrv/service/service.go
SimoneDutto: yeah, i've reverted this change

internal/jimm/login/login.go
SimoneDutto: or @kian99

luci1900: If you defer it, I think you'd have to check what the error is in the deferred closure.

internal/rpcproxy/rpcproxy.go
SimoneDutto: i wouldn't put the error along with the description.
We could make it as an `Extra` params if needed

internal/jimm/login/login.go
kian99: In this case I don't think you'd check the error in the closure because I don't know what specific error you'd be looking for.
I'm okay with always doing a `LogFailedLogin` in these methods.  That means we'd emit a log for events like,
- DB is down
- A bug where we incorrectly reject requests
- Client sends invalid fields like an invalid service account tag.

I'm alright with a "failed login" log in all those cases.

SimoneDutto: another example why i don't like the defer is this one:
```go
func (j *loginManager) LoginWithSessionToken(ctx context.Context, sessionToken string) (_ *openfga.User, err error) {
	const op = errors.Op("jimm.LoginWithSessionToken")
	jwtToken, err := j.oAuthAuthenticator.VerifySessionToken(sessionToken)
	if err != nil {
		logger.LogFailedLogin(ctx, "unkown session token")
		return nil, errors.E(op, err)
	}

	email := jwtToken.Subject()
	user, err := j.UserLogin(ctx, email)
	if err != nil {
		logger.LogFailedLogin(ctx, email)
		return nil, errors.E(op, err)
	}
	logger.LogSuccessfulLogin(ctx, email)
	return user, nil
}
```

using the `defer` here would prevent/make harder to log different identifiers. In the first error we don't have the email, in the second we have it

internal/jimm/login/login.go
```diff
@@ -190,7 +204,13 @@ func (j *loginManager) LoginWithSessionCookie(ctx context.Context, identityID st
 	if identityID == "" {
 		return nil, errors.E(op, "missing cookie identity")
```
alesstimec, line 205: isn't this also a failed login attempt?

internal/logger/securityevent_logger.go
```diff
@@ -0,0 +1,82 @@
...
+	critical eventLevel = "CRITICAL"
+)
+
+type securityEvent struct {
+	Event       string
+	Description string
+	Severity    eventLevel
+}
+
+func (s securityEvent) ToZapFields() []zapcore.Field {
```
alesstimec: godoc, please

alesstimec (approved):

internal/logger/securityevent_logger.go
SimoneDutto: i'll make it private.

internal/jimm/login/login.go
SimoneDutto: mmm, i don't know. It wouldn't be useful to log a security event for an empty cookie, i think

PR #1696, opened 2025-09-05
feat: expose k8s svc options

ale8k:
## Description
This PR exposes the k8s service options. It additionally adds a fixed hostname (which would in theory break controllers behind proxies, but with JIMM in front bootstrapping, I guess you'd never have a proxy in front of your controller anymore...?)

## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests

## Test instructions
Can be tested via:
1. Pulling your microk8s config out and placing it into ~/.kube/config 
2. Switching to the new context
3. Enabling metallb on microk8s 
4. Running `./jaas bootstrap testk8scloud test16 3.6.8 --controller-service-type=loadbalancer`, a free IP of your NAT will be consumed by the LB (you should see it in the "contacting") and a controller will be bootstrapped.

### DIFF
--- a/cmd/jaas/cmd/bootstrap.go
+++ b/cmd/jaas/cmd/bootstrap.go
@@ -43,13 +43,13 @@ service if it knows it is supported by the cluster. This is performed by
 interrogating the cluster for a well known managed deployment such as microk8s,
 GKE or EKS.
 
-When bootstrapping to a k8s cluster Juju does not recognise, there's no
+When bootstrapping to a Kubernetes cluster Juju does not recognise, there's no
 guarantee a load balancer is available, so Juju defaults to a controller
 service type of ClusterIP. In the case of bootstrapping via JIMM, this will 
 not work unless JIMM is deployed within the same cluster. There are three bootstrap
 options available to tell Juju how to set up the controller service. Part of
 the solution may require a load balancer for the cluster to be set up manually
-first, or perhaps an external k8s service via a FQDN will be used
+first, or perhaps an external Kubernetes service via a FQDN will be used
 (this is a cluster specific implementation decision which Juju needs to be
 informed about so it can set things up correctly). The three relevant bootstrap
 options are (see list of bootstrap config items below for a full explanation):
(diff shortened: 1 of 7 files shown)


SimoneDutto (approved):

kian99 (commented):
Looks good just one comment

SimoneDutto:
> Looks good just one comment

where is the comment though? Ahah

cmd/jaas/cmd/bootstrap.go
```diff
@@ -35,6 +35,31 @@ Use the --detach flag to start the bootstrap job and return immediately,
 printing only the job ID, without waiting for the job to complete.
 
 The final argument, version, denotes the Juju controller to be bootstrapped.
+
+Bootstrapping to a k8s cluster requires that the service set up to handle
```
kian99, line 39: Some comments on the wording,
- I'd use `Kubernetes` in place of k8s since it's user-facing.
- Capitalise things like microk8s to `MicroK8s`.
- I would clarify the first instance of "service" to explain that we are referring to a Kubernetes service.
- On the bit about "interrogating the cluster for a well known managed deployment", can we link to some Juju docs on this as well. It's news to me and it's also not clear what flavours of K8s that Juju handles "nicely".
- Let's add a 1-liner to the `bootstrapExamples` for an example of bootstraping to a K8s cluster with service-type=loadbalancer.

ale8k: This document is a direct copy for the Juju bootstrap doc (with some JIMM adjustments from me). Still want me to change it?

I'll add the example though for sure for both cases.

alesstimec: +1 on kian's suggestions

alesstimec (approved):
LGTM with Kian's comment addressed

cmd/jaas/cmd/bootstrap.go
ale8k: Done all, but the interrogating part... AI seems to think Juju is hitting GET /api/v1/namespaces/juju/services/juju-controller. I'll leave explaining this part out because I don't really understand either, I do know that LoadBalancers can be plugged into K8S but as for what is "well known" I'm not sure.

PR #1697, opened 2025-09-05
feat: security events for rebac

SimoneDutto:
## Description
In this PR we make use of the new release of `github.com/canonical/rebac-admin-ui-handlers v0.2.0` to log security events when we return unauthorized error from the rebac admin.

Release: https://github.com/canonical/rebac-admin-ui-handlers/pull/32

It looks like this:
```console
{"event": "authz_fail:jimm-test@canonical.com", "description": "attempted GET /v1/entitlements  in the ReBAC admin API", "severity": "CRITICAL"}
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -27,7 +27,7 @@ require (
 	github.com/antonlindstrom/pgstore v0.0.0-20220421113606-e3a6e3fed12a
 	github.com/canonical/go-service v1.0.0
 	github.com/canonical/ofga v0.10.0
-	github.com/canonical/rebac-admin-ui-handlers v0.1.2
+	github.com/canonical/rebac-admin-ui-handlers v0.2.0
 	github.com/coreos/go-oidc/v3 v3.11.0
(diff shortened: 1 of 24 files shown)


go.mod
```diff
@@ -339,6 +339,8 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.4.1 // indirect
 )
 
+require github.com/ulikunitz/xz v0.5.12
```
SimoneDutto, line 342: result of `go mod tidy`

kian99 (approved):
This is a much nicer way of handling things than in our rpc handler.

SimoneDutto:
> This is a much nicer way of handling things than in our rpc handler.

Yeah, this is nicer and more precise

go.mod
luci1900: Maybe worth keeping that in a separate commit.

luci1900 (approved):

go.mod
SimoneDutto: i usually do squash merge regardless, and it makes sense it's there because I've added the new dependency in the context of this pr

PR #1698, opened 2025-09-08
feat(security-events): add system security events

SimoneDutto:
## Description

Add security events for startup, shutdown and disable monitoring.

> The log level has been set to at least WARNING for all events. This is because we use the default logger to record security events, and we recommend configuring the log level to at least WARN to ensure that these events are captured. Setting the level to INFO could result in excessive logging, which may not be suitable for production environments. 
Keep in mind the only security event set to INFO would be successful login.

### DIFF
--- a/cmd/jimmsrv/main.go
+++ b/cmd/jimmsrv/main.go
@@ -45,6 +45,7 @@ func start(ctx context.Context, s *service.Service) error {
 	logLevel := os.Getenv("JIMM_LOG_LEVEL")
 	logDevMode, _ := strconv.ParseBool(os.Getenv("JIMM_LOG_DEV_MODE"))
 	logger.SetupLogger(ctx, logLevel, logDevMode)
+	logger.LogJimmStartup(ctx)
 	zapctx.Info(ctx, "jimm info",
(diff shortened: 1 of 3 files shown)


luci1900 (approved):

cmd/jimmsrv/main.go
```diff
@@ -45,6 +45,7 @@ func start(ctx context.Context, s *service.Service) error {
 	logLevel := os.Getenv("JIMM_LOG_LEVEL")
 	logDevMode, _ := strconv.ParseBool(os.Getenv("JIMM_LOG_DEV_MODE"))
 	logger.SetupLogger(ctx, logLevel, logDevMode)
+	logger.LogJimmStartup(ctx)
 	zapctx.Info(ctx, "jimm info",
```
kian99, line 49: Can you move this `jimm info` into LogJimmStartup please, seems like a good place for it.

kian99: Ah but I see it's just for a single event so nvm this then.

kian99 (approved):

PR #1699, opened 2025-09-08
feat: add controller cleanup on failed bootstrap

kian99:
## Description
This PR attempts to address the issue that the bootstrap process succeeds but then JIMM fails to register the controller. Some possible reasons this could happen are:
- Networking issues
- A bug in JIMM
- Changes to Juju

In these cases we can either allow the user to manually add the new controller to JIMM or attempt to clean it up. In this PR I've made the default behavior to clean up the controller.

We attempt clean-up by invoking the Juju CLI's `destroy-controller` method. There is a better chance the CLI can destroy the controller although JIMM can't reach it since the CLI was just able to bootstrap and connect to the controller. If we also fail to clean-up, we issue a final error that bootstrap and clean-up failed and user intervention is required to clean-up the dangling resources.

Most of the changes are found in mocks and tests. The tests in `internal/jimm/bootstrap/bootstrap_test.go` are especially long but I've kept their style with some refactoring to incorporate the changes.

Fixes [JUJU-8415](https://warthogs.atlassian.net/browse/JUJU-8415)

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

[JUJU-8415]: https://warthogs.atlassian.net/browse/JUJU-8415?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/jimm/bootstrap/mocks/jujuclientstore.go
kian99, line 1: Note that this file's increase is due to the use of the `-typed` flag when generating the mock.

internal/jimm/bootstrap/bootstrap.go
```diff
@@ -198,23 +211,41 @@ func (b *bootstrapManager) StartBootstrap(ctx context.Context, user *openfga.Use
...
+
+	r := jujucommands.NewCommandRunner(c.binaryPath, c.jujuDataDir)
+	return jujucommands.NewDestroyControllerCmd(r).Run(ctx, p)
 }
 
-// DefaultBootstrapExecutor defines the default, and expected bootstrap executor
+// commandFactory provides a concrete implementation of [CommandFactory]
 // to be used in a [bootstrapManager.BootstrapJob]
-type DefaultBootstrapExecutor struct{}
+type commandFactory struct{}
```
kian99, line 242: For @ale8k, I realise this struct has the same name as the interface it implements. I think that's fine and would prefer not to change it.

internal/jimm/bootstrap/bootstrap.go
```diff
@@ -362,42 +395,47 @@ func (b *bootstrapManager) runBootstrap(
...
-			}
-			return errors.E(fmt.Errorf("bootstrap command failed: %w", output.Err))
+	err = b.consumeCommandOutput(ctx, outputCh, jobId)
+	if err != nil {
+		return err
+	}
+
+	// controllerCleanup is a helper function to cleanup the controller if we fail at any point
+	// after the bootstrap, avoiding orphaned controllers in the cloud.
+	controllerCleanup := func(err error, controllerDetails *jujuclient.ControllerDetails) error {
```
kian99, line 434: This is a closure because it felt more appropriate than a separate method.
I could also use `defer func()` but then I would need to make the error return value a named parameter in order to access it from the deferred function (I think?) so this felt simpler but can definitely change to `defer ...` if that's preferred.

internal/jimm/bootstrap/bootstrap_test.go
```diff
@@ -79,28 +79,36 @@ func assertJobError(c *qt.C, s *bootstrapManagerSuite, id uuid.UUID, errStr stri
...
+		clientStore:    bootstrapmocks.NewMockClientStore(ctrl),
+		executor:       bootstrapmocks.NewMockJujuCommands(ctrl),
+	}
 
 	i, err := dbmodel.NewIdentity("bob@canonical.com")
 	c.Assert(err, qt.IsNil)
 	user := openfga.NewUser(i, nil)
 
-	return ctrl, store, jujuManager, binaryStore, executor, clientStore, user
+	return ctrl, m, user
```
kian99, line 111: In future I suggest we return a struct when we have a lot of values to return.

internal/jimm/bootstrap/bootstrap_test.go
```diff
@@ -235,22 +243,22 @@ func (s *bootstrapManagerSuite) TestBootstrapJob(c *qt.C) {
...
+	ctrl, mocks, user := setupTest(c)
 	defer ctrl.Finish()
 
-	manager, err := bootstrap.NewBootstrapManager(store, s.jobTracker, jujuManager, binaryStore, loginTokenRefreshURLParam)
+	manager, err := bootstrap.NewBootstrapManager(mocks.store, s.jobTracker, mocks.jujuManager, mocks.binaryStore, loginTokenRefreshURLParam)
 	c.Assert(err, qt.IsNil)
 
 	// Mocked in order of execution:
 	cleanupCalled := false // To be asserted after job run - ensures cleanup was run.
-	store.EXPECT().LockBootstrap(gomock.Any(), gomock.Any()).Return(nil).Times(1)
```
kian99, line 246: In Juju the standard is to drop usage of `Times(1)` because the default is to expect the call only once unless otherwise specified. So I'd like to stick with that.

internal/jimm/bootstrap/bootstrap.go
```diff
@@ -319,7 +350,9 @@ func (b *bootstrapManager) BootstrapJob(
 			return errors.E(fmt.Errorf("failed to check if controller exists: %w", err))
 		}
 
-		// TODO(ale8k): When downloading, it takes a while and the CLI has no output. Download progress would be a VERY nice to have here.
+		_ = b.store.AddBootstrapLog(ctx, jobId,
```
SimoneDutto: nice

internal/jimm/bootstrap/bootstrap.go
```diff
@@ -379,42 +412,47 @@ func (b *bootstrapManager) runBootstrap(
...
+			controllerDetailsStr = string(res)
 		}
+
+		zapctx.Error(ctx, "failed to cleanup controller after failing to add it to JIMM", zap.Error(err))
+
+		return errors.E(fmt.Errorf("error post-bootstrap: %w\n"+
+			"automatic cleanup of the controller also failed: %w\n"+
+			"WARNING: manual attach to JIMM or cleanup of the controller is required.\n"+
+			"Controller details:\n%s", err, cleanupErr, controllerDetailsStr))
+
```
SimoneDutto: should we log this into the bootstrap log as well? I guess it would be nice to receive this one in the CLI without checking for JIMM's logs

internal/jimm/bootstrap/bootstrap_test.go
```diff
@@ -281,15 +287,17 @@ func (s *bootstrapManagerSuite) TestBootstrapJob(c *qt.C) {
...
 		},
 		nil,
-	).Times(1)
+	)
+	mocks.commandFactory.EXPECT().New(binaryPath, jobParams.JujuDataDir).
+		Return(mocks.executor)
 	// We don't know the jobid to expect it yet. I did test by moving this line below the call, and it does
 	// pass, but it'd be racey between the starting of the job routine and the EXPECT.
-	store.EXPECT().AddBootstrapLog(gomock.Any(), gomock.Any(), testOutputLine).Return(nil).Times(1)
+	mocks.store.EXPECT().AddBootstrapLog(gomock.Any(), gomock.Any(), gomock.Any()).Return(nil).AnyTimes()
```
SimoneDutto, line 300: i guess here it has become `AnyTimes` because you've added the log about downloading the CLI

internal/jimm/bootstrap/bootstrap.go
SimoneDutto: or maybe we do log the `JobError` in the CLI when the job is done. I'm not sure.

SimoneDutto: If we don't, we should

internal/jimm/bootstrap/bootstrap.go
```diff
@@ -379,42 +412,47 @@ func (b *bootstrapManager) runBootstrap(
...
+			res, _ := yaml.Marshal(controllerDetails)
+			controllerDetailsStr = string(res)
 		}
+
+		zapctx.Error(ctx, "failed to cleanup controller after failing to add it to JIMM", zap.Error(err))
+
+		return errors.E(fmt.Errorf("error post-bootstrap: %w\n"+
+			"automatic cleanup of the controller also failed: %w\n"+
+			"WARNING: manual attach to JIMM or cleanup of the controller is required.\n"+
+			"Controller details:\n%s", err, cleanupErr, controllerDetailsStr))
```
SimoneDutto: I would be more explicit warning users there might be dangling controllers. Maybe something like:

```go
return errors.E(fmt.Errorf(
    "error post-bootstrap: %w\n"+
    "automatic cleanup of the controller also failed: %w\n"+
    "WARNING: resources associated with the controller may remain dangling in your environment.\n"+
    "Manual intervention is required: either attach to JIMM or manually clean up the controller to avoid resource leaks.\n"+
    "Controller details:\n%s",
    err, cleanupErr, controllerDetailsStr),
)
```

internal/jimm/bootstrap/bootstrap_test.go
```diff
@@ -964,15 +1145,20 @@ func (s *bootstrapManagerSuite) TestBootstrapJob_JujuManagerFailsToAddController
...
 
 	pollJob(c, s, id, dbmodel.StatusFailed)
-	assertJobError(c, s, id, "run bootstrap failed: failed to add controller to JIMM: add controller test error")
+	assertJobError(c,
+		s,
+		id,
+		"run bootstrap failed: error post-bootstrap: failed to add controller to JIMM: add controller test error\n"+
+			"the controller has been automatically destroyed")
 	c.Assert(cleanupCalled, qt.IsTrue)
 }
```
SimoneDutto, line 1162: can you add another test where we fail the bootstrap and we fail to cleanup the controller?

internal/jujucommands/destroycontroller.go
```diff
@@ -0,0 +1,65 @@
...
+
+	return nil
+}
+
+// BuildCmdArgs builds the arguments for the command.
+func (b DestroyControllerCmdParams) BuildCmdArgs() []string {
+	var args []string
+	args = append(args, "destroy-controller")
+	args = append(args, b.ControllerName)
+	args = append(args, "--no-prompt")
```
SimoneDutto, line 33: should we pass also `--destroy-all-models` just in case?

internal/jujucommands/destroycontroller.go
```diff
@@ -0,0 +1,65 @@
...
+	"context"
+	"errors"
+	"fmt"
+
+	"github.com/juju/juju/juju/osenv"
+	_ "github.com/juju/juju/provider/lxd"
+)
+
+// DestroyControllerCmdParams holds the parameters to tear-down a controller.
+type DestroyControllerCmdParams struct {
```
SimoneDutto, line 15: would it be better to call `kill-controller` instead or try with `kill-controller` if this one fails?

reading the doc for the command it seems more appropriate
```
If the API server is unreachable, the machines of the controller model will be
destroyed through the cloud provisioner.  If there are additional machines,
including machines within models, these machines will not be destroyed
and will never be reconnected to the Juju controller being destroyed.
```

internal/jimm/bootstrap/bootstrap.go
```diff
@@ -379,42 +412,47 @@ func (b *bootstrapManager) runBootstrap(
...
-			); writeLogErr != nil {
-				zapctx.Error(ctx, "failed to write bootstrap error", zap.Error(writeLogErr), zap.String("jobId", jobId.String()))
-			}
-			return errors.E(fmt.Errorf("bootstrap command failed: %w", output.Err))
+	err = b.consumeCommandOutput(ctx, outputCh, jobId)
+	if err != nil {
+		return err
+	}
+
+	// controllerCleanup is a helper function to cleanup the controller if we fail at any point
```
luci1900, line 432: Is it possible that we get an intermittent network failure / timeout and are in fact incorrect about bootstrap having failed?

kian99: If we've reached this point, then the actual bootstrap succeeded, and we are now adding the controller to JIMM (that just means we are saving the controller details to jimm's database and verifying that jimm can connect to the controller).

If we fail after this point it could be due to networking issues (jimm can't reach the controller - that wouldn't imply something like firewall rules because the Juju CLI is running from the same environment, but it could mean the CLI is connecting in a different way than jimm particularly relevant for Kubernetes controllers).
It could also be a bug in jimm, etc.

internal/jimm/bootstrap/bootstrap.go
kian99: I'll add the line about the danging resources.

internal/jimm/bootstrap/bootstrap.go
kian99: I'm pretty sure the error comes back to the CLI so you'll get it.

kian99: But I'll verify that and reply here.

internal/jujucommands/destroycontroller.go
kian99: I'm not 100% sure myself. Since the Juju CLI was able to finish bootstrap, it should be able to reach the API server. So I don't have a strong lean one way or the other.

internal/jujucommands/destroycontroller.go
kian99: Shouldn't be needed since no models should exist yet. So I'd leave it as is unless it's an issue.

internal/jimm/bootstrap/bootstrap.go
luci1900: Considering it wraps every return out of the function, it may make more sense as a defer. I don't feel strongly about it.

luci1900 (approved):
This seems quite complex and a little scary, but I don't see anything obviously wrong with it.

internal/jimm/bootstrap/bootstrap.go
SimoneDutto: I prefer explicit code, doing too much in the defer feels wrong

internal/jimm/bootstrap/bootstrap_test.go
kian99: Correct

internal/jimm/bootstrap/bootstrap_test.go
kian99: That is the new test I added, `TestBootstrapJob_CleanupControllerFailure`. The other tests fail and the _succeed_ at controller cleanup.

SimoneDutto: i was looking for a test where we check for the error: `automatic cleanup of the controller also failed` and i couldn't find it.

kian99: Ah good catch, I had added 2 tests and then deleted the one that was the same as all the others, but must have mixed things up. I'll fix that soon.

internal/jimm/bootstrap/bootstrap.go
kian99: Yeah the error does get sent back to the client.

SimoneDutto (approved):

internal/jimm/bootstrap/bootstrap.go
ale8k: Yeah that's ok.

internal/jimm/bootstrap/bootstrap.go
```diff
@@ -72,6 +73,18 @@ type BinaryStore interface {
...
 
+// JujuCommands defines the Juju CLI methods that the bootstrap job requires.
+type JujuCommands interface {
+	Bootstrap(ctx context.Context, p jujucommands.BootstrapCmdParams) (<-chan jujucommands.OutputLine, jujuclient.ClientStore, func(), error)
+	DestroyController(ctx context.Context, p jujucommands.DestroyControllerCmdParams) (<-chan jujucommands.OutputLine, error)
+}
+
+// CommandFactory is a wrapper for mocking Juju commands, with a concrete
+// implementation in [commandFactory].
+type CommandFactory interface {
```
ale8k, line 84: Much prefer this, nice.

internal/jimm/bootstrap/bootstrap.go
```diff
@@ -203,23 +216,41 @@ func (b *bootstrapManager) StartBootstrap(ctx context.Context, user *openfga.Use
...
+	return jujucommands.NewDestroyControllerCmd(r).Run(ctx, p)
 }
 
-// DefaultBootstrapExecutor defines the default, and expected bootstrap executor
+// commandFactory provides a concrete implementation of [CommandFactory]
 // to be used in a [bootstrapManager.BootstrapJob]
-type DefaultBootstrapExecutor struct{}
+type commandFactory struct{}
+
+func (h commandFactory) New(binaryPath, jujuDataDir string) JujuCommands {
```
ale8k, line 245: Godoc

internal/jimm/bootstrap/bootstrap.go
ale8k: Why not check the error? Presumably if we can't contact db right now we shouldn't continue?

internal/jimm/bootstrap/bootstrap.go
ale8k: This seems fine to me honestly. I think I would prefer it in its own function but it is ok to leave it.

internal/jimm/bootstrap/bootstrap.go
```diff
@@ -434,24 +475,58 @@ func (b *bootstrapManager) runBootstrap(
...
+
+	err = b.consumeCommandOutput(ctx, outputCh, jobID)
+	if err != nil {
+		return err
 	}
 
 	return nil
 }
+
+func (b *bootstrapManager) consumeCommandOutput(ctx context.Context, outputCh <-chan jujucommands.OutputLine, jobId uuid.UUID) error {
```
ale8k, line 526: writeCommandLogs*?

internal/jujucommands/destroycontroller.go
```diff
@@ -0,0 +1,65 @@
...
+func (b DestroyControllerCmdParams) Validate() error {
+	if b.ControllerName == "" {
+		return errors.New("controller name cannot be empty")
+	}
+
+	return nil
+}
+
+// BuildCmdArgs builds the arguments for the command.
+func (b DestroyControllerCmdParams) BuildCmdArgs() []string {
```
ale8k, line 29: This will be particularly useful for when we want to add the destroy controller job. Nice work!

internal/jimm/bootstrap/bootstrap.go
```diff
@@ -379,42 +412,50 @@ func (b *bootstrapManager) runBootstrap(
...
-			// This is because the bootstrap process may still succeed, and we
-			// don't want to fail the entire job just because we couldn't log.
-			zapctx.Error(ctx, "failed to write bootstrap log", zap.Error(writeLogErr), zap.String("jobId", jobId.String()))
+		var controllerDetailsStr string
+		if controllerDetails != nil {
+			res, _ := yaml.Marshal(controllerDetails)
+			controllerDetailsStr = string(res)
 		}
+
+		zapctx.Error(ctx, "failed to cleanup controller after failing to add it to JIMM", zap.Error(err))
```
ale8k: this is the consume command output error? Not cleanupError?

ale8k (approved):
Beeaaaautiful PR, nice! Some super minor comments. I think some more logging too would be nice.

internal/jimm/bootstrap/bootstrap.go
```diff
@@ -72,6 +73,18 @@ type BinaryStore interface {
...
+// JujuCommands defines the Juju CLI methods that the bootstrap job requires.
+type JujuCommands interface {
+	Bootstrap(ctx context.Context, p jujucommands.BootstrapCmdParams) (<-chan jujucommands.OutputLine, jujuclient.ClientStore, func(), error)
+	DestroyController(ctx context.Context, p jujucommands.DestroyControllerCmdParams) (<-chan jujucommands.OutputLine, error)
+}
+
+// CommandFactory is a wrapper for mocking Juju commands, with a concrete
+// implementation in [commandFactory].
+type CommandFactory interface {
+	New(binaryPath, jujuDataDir string) JujuCommands
```
alesstimec, line 85: could also do 
```
type CommandFactory func(binaryPath, jujuDataDir string) JujuCommands
```

alesstimec (approved):
LGTM

PR #1700, opened 2025-09-08
feat: enable bootstrap stop

ale8k:
## Description
Enables bootstrap stop. It also fixes the following:
1. We were cancelling the context on the readers, but we want the readers to continue when the context is cancelled, the process should close those channels.
2. The default Cancel func of CommandContext is a kill signal, it now uses an interrupt (which is what juju expects for cleanup)
3. When the job context is cancelled, we need to continue writing logs from stderr, but because the same context was used to write logs to the store, we never saw the error logs. It now uses an independent store context.
4. There was a small chance (which I did indeed hit funnily enough) between the bootstrapping on the controller and the addition to our store that we don't add the controller because the context was cancelled.

## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests

### DIFF
--- a/internal/jimm/bootstrap/bootstrap.go
+++ b/internal/jimm/bootstrap/bootstrap.go
@@ -348,6 +348,11 @@ func (b *bootstrapManager) BootstrapJob(
 // runBootstrap wraps the logic of running a controller bootstrap for JIMM into
 // a self-contained function. It is expected, and only expected to be run from within
 // the [bootstrapManager.BootstrapJob].
+//
+// The jobCtx is expected to be the context of the job, and as such will be cancelled
+// when the job is stopped or cancelled. The jobCtx is NOT expected to be used for
+// any store operations, or other operations that should continue even if the job
+// is cancelled.
 func (b *bootstrapManager) runBootstrap(
 	jobCtx context.Context,
 	p JobParams,
--- a/internal/jimm/bootstrap/bootstrap.go
+++ b/internal/jimm/bootstrap/bootstrap.go
@@ -270,16 +270,14 @@ func (b *bootstrapManager) BootstrapJob(
 	executor BootstrapExecutor,
 	user *openfga.User,
 ) jobtracker.JobFunc {
-	const bootstrapLockTTL = 40 * time.Minute
-
-	return func(ctx context.Context) error {
-		jobId, ok := jobtracker.JobIdFromContext(ctx)
+	return func(jobCtx context.Context) error {
+		jobId, ok := jobtracker.JobIdFromContext(jobCtx)
 		if !ok {
 			return fmt.Errorf("failed to get job ID from context")
 		}
 
 		logCtx := zapctx.WithFields(
-			ctx,
+			jobCtx,
 			zap.String("job-id", jobId.String()),
 			zap.String("controller-name", p.ControllerName),
 		)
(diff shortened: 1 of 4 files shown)


internal/jimm/bootstrap/bootstrap.go
```diff
@@ -321,7 +321,7 @@ func (b *bootstrapManager) BootstrapJob(
 
 		// TODO(ale8k): When downloading, it takes a while and the CLI has no output. Download progress would be a VERY nice to have here.
```
SimoneDutto, line 322: remove this TODO in favor of a card. In any case i believe we've discuss we were not gonna do it.

internal/jimm/bootstrap/bootstrap.go
```diff
@@ -444,8 +456,13 @@ func (b *bootstrapManager) runBootstrap(
...
 	// So... Find a nice way to remove controllers if the add fail / clean up dangling ones later.
 	// Once fixed, this can be tested by just making the controller addr something that doesn't exist, and we can test the solution.
+
+	// We use store context here as there's a small chance that somebody
+	// could cancel the context between the controller being successfully bootstrapped
+	// and adding the controller, in which case we'd be left with a bootstrapped
+	// controller that JIMM isn't aware of.
 	if err := b.jujuManager.AddController(
-		ctx,
+		storeCtx,
```
SimoneDutto, line 470: even adding the controller use the `storeCtx`?

SimoneDutto (commented):
Since we have all the mocks, we should be able to write some tests to show that the failing bootstrap still manages to write logs. Right?

internal/jimm/bootstrap/bootstrap.go
kian99: There already is a card I believe. And I've handled it in #1699.

internal/jimm/bootstrap/bootstrap.go
```diff
@@ -270,16 +270,14 @@ func (b *bootstrapManager) BootstrapJob(
 	executor BootstrapExecutor,
 	user *openfga.User,
 ) jobtracker.JobFunc {
-	const bootstrapLockTTL = 40 * time.Minute
-
-	return func(ctx context.Context) error {
-		jobId, ok := jobtracker.JobIdFromContext(ctx)
+	return func(jobCtx context.Context) error {
```
kian99, line 273: Let's update the godoc for `BootstrapJob()` to explain the behavior of jobCtx and when it shouldn't be used.

kian99 (approved):

internal/jimm/bootstrap/bootstrap.go
ale8k: This is that weird issue where you could in theory end up with a successfully bootstrapped controller but no way to add it to jimm.

SimoneDutto:
@ale8k can you add some tests like @kian99 did in its last pr?

ale8k:
> @ale8k can you add some tests like @kian99 did in its last pr?

Perhaps I should merge after @kian99 then? As he's already set up the mocks?

internal/jimm/bootstrap/bootstrap.go
ale8k: Done

kian99:
@ale8k Let's merge this so long?

kian99:
Ah you need another approval nvm.

SimoneDutto (approved):

PR #1701, opened 2025-09-09
feat: add logging for user creation and updates

luci1900:
## Description

Add logging for user creation and updates, matching https://library.canonical.com/corporate-policies/information-security-policies/ssdlc/ssdlc---security-event-logging#%EE%B0%823.4-user-related-events-%5Buser%5D-9 as much as possible.

For user creation, we do not have a "creator" user to log, only the created user.

For user updates, the RBAC assumptions in the spec don't really match openfga so we only log relations made directly from users.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

### DIFF
--- a/internal/db/identity.go
+++ b/internal/db/identity.go
@@ -1,4 +1,4 @@
-// Copyright 2024 Canonical.
+// Copyright 2025 Canonical.
 
 package db
 
@@ -7,6 +7,7 @@ import (
 
 	"github.com/canonical/jimm/v3/internal/dbmodel"
 	"github.com/canonical/jimm/v3/internal/errors"
+	"github.com/canonical/jimm/v3/internal/logger"
 	"github.com/canonical/jimm/v3/internal/servermon"
 )
 
@@ -36,8 +37,12 @@ func (d *Database) GetIdentity(ctx context.Context, u *dbmodel.Identity) (err er
 	defer servermon.ErrorCounter(servermon.DBQueryErrorCount, &err, string(op))
 
 	db := d.DB.WithContext(ctx)
-	if err := db.Where("name = ?", u.Name).FirstOrCreate(&u).Error; err != nil {
+	result := db.Where("name = ?", u.Name).FirstOrCreate(&u)
+	if result.Error != nil {
 		return errors.E(op, err)
+	}
+	if result.RowsAffected == 1 {
+		logger.LogUserCreated(ctx, u.Name)
 	}
 	return nil
 }
--- a/internal/db/identity_test.go
+++ b/internal/db/identity_test.go
@@ -8,6 +8,9 @@ import (
 	"testing"
 
 	qt "github.com/frankban/quicktest"
+	"github.com/juju/zaputil/zapctx"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zaptest/observer"
 
 	"github.com/canonical/jimm/v3/internal/db"
 	"github.com/canonical/jimm/v3/internal/dbmodel"
@@ -60,6 +63,31 @@ func (s *dbSuite) TestGetIdentity(c *qt.C) {
 	err = s.Database.GetIdentity(ctx, u4)
 	c.Assert(err, qt.IsNil)
 	c.Check(u4, qt.DeepEquals, u3)
+}
+
+func (s *dbSuite) TestIdentityUserCreatedLogging(c *qt.C) {
+	ctx := context.Background()
+
+	err := s.Database.Migrate(ctx)
+	c.Assert(err, qt.IsNil)
+
+	core, logs := observer.New(zap.InfoLevel)
+	ctx = zapctx.WithLogger(ctx, zap.New(core))
+
+	i, err := dbmodel.NewIdentity("bob")
+	c.Assert(err, qt.IsNil)
+	err = s.Database.GetIdentity(ctx, i)
+	c.Assert(err, qt.IsNil)
+	// Logging creation
+	c.Assert(logs.Len(), qt.Equals, 1)
+	c.Assert(logs.All()[0].Message, qt.Equals, "user_created:bob")
+
+	i, err = dbmodel.NewIdentity("bob")
+	c.Assert(err, qt.IsNil)
+	err = s.Database.GetIdentity(ctx, i)
+	c.Assert(err, qt.IsNil)
+	// No new creation, no extra logging
+	c.Assert(logs.Len(), qt.Equals, 1)
 }
 
 func TestUpdateIdentityUnconfiguredDatabase(t *testing.T) {
--- a/internal/jimm/permissions/relations.go
+++ b/internal/jimm/permissions/relations.go
@@ -12,6 +12,7 @@ import (
 	"github.com/canonical/jimm/v3/internal/common/pagination"
 	"github.com/canonical/jimm/v3/internal/db"
 	"github.com/canonical/jimm/v3/internal/errors"
+	"github.com/canonical/jimm/v3/internal/logger"
 	"github.com/canonical/jimm/v3/internal/openfga"
 	ofganames "github.com/canonical/jimm/v3/internal/openfga/names"
 	apiparams "github.com/canonical/jimm/v3/pkg/api/params"
@@ -32,6 +33,7 @@ func (j *permissionManager) AddRelation(ctx context.Context, user *openfga.User,
 	if err != nil {
 		return errors.E(op, errors.CodeOpenFGARequestFailed, err)
 	}
+	j.logUserUpdates(ctx, user, parsedTuples, true)
 	return nil
 }
 
@@ -50,6 +52,7 @@ func (j *permissionManager) RemoveRelation(ctx context.Context, user *openfga.Us
 	if err != nil {
 		return errors.E(op, errors.CodeOpenFGARequestFailed, err)
 	}
+	j.logUserUpdates(ctx, user, parsedTuples, false)
 	return nil
 }
 
@@ -249,4 +252,14 @@ func (j *permissionManager) parseTuple(ctx context.Context, tuple apiparams.Rela
 	}
 
 	return &t, nil
+}
+
+// logUserUpdates logs tuple relation changes if they are for a user.
+// This should be the closest equivalent of logging an RBAC role in our Zanzibar-style openfga graph.
+func (j *permissionManager) logUserUpdates(ctx context.Context, user *openfga.User, tuples []openfga.Tuple, isAddition bool) {
+	for _, tuple := range tuples {
+		if tuple.Object.Kind.String() == openfga.UserType.String() {
+			logger.LogUserUpdated(ctx, user.Name, tuple.Object.ID, tuple.Relation.String(), tuple.Target.ID, isAddition)
+		}
+	}
 }
--- a/internal/logger/securityevent_logger.go
+++ b/internal/logger/securityevent_logger.go
@@ -79,6 +79,28 @@ func LogGrantJimmAdmins(ctx context.Context, identityIds []string) {
 	})
 }
 
+// LogUserCreated logs the creation of a user, implicitly an identity
+func LogUserCreated(ctx context.Context, identityId string) {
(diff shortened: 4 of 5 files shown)


internal/logger/securityevent_logger.go
```diff
@@ -79,6 +79,28 @@ func LogGrantJimmAdmins(ctx context.Context, identityIds []string) {
...
+
+// LogUserUpdated
+func LogUserUpdated(ctx context.Context, adminID string, userID string, relation string, target string, isAddition bool) {
+	action := "adding"
+	if !isAddition {
+		action = "removing"
+	}
+	logSecurityEvent(ctx, securityEvent{
+		Event:       fmt.Sprintf("user_updated:%s,%s,%s:%s", adminID, userID, relation, target), // ?
+		Description: fmt.Sprintf("User %s updated %s by %s relation %s for object %s", adminID, userID, action, relation, target),
```
ale8k: Not related to this PR, but some  history to help with understanding our OpenFGA language.

When we first started using it, the language was (in their library)...
User -> Relation -> Object

What it really meant was
User/Object -> Relation -> User/Object

We agreed this was confusing and eventually settled on

Object -> Relation -> Target Object.

ale8k: In "relation" to this PR 😆 - I'm not sure how we could know a user updated a relation because jimm always connects to a store. A user could add a relation / remove / update between two objects, but probably not themselves? So I think this should maybe read...

User X updated relation <type> for objects <a> and <b>

But maybe I'm missing context on the security events and misunderstanding. (Why is this bolded? lol)

internal/db/identity.go
```diff
@@ -36,9 +37,13 @@ func (d *Database) GetIdentity(ctx context.Context, u *dbmodel.Identity) (err er
 	defer servermon.ErrorCounter(servermon.DBQueryErrorCount, &err, string(op))
 
 	db := d.DB.WithContext(ctx)
-	if err := db.Where("name = ?", u.Name).FirstOrCreate(&u).Error; err != nil {
+	result := db.Where("name = ?", u.Name).FirstOrCreate(&u)
+	if result.Error != nil {
 		return errors.E(op, err)
 	}
+	if result.RowsAffected == 1 {
```
ale8k, line 44: If I put an existing user into this, would we get rows affect 1?

luci1900: We'd get `0`, since no row was inserted. At least according to gorm docs, none of this is tested yet.

internal/logger/securityevent_logger.go
luci1900: There's two users here https://github.com/canonical/jimm/blob/e1237fa8b71bf8b87e752ce3257d6c6926f60925/internal/jimm/permissions/relations.go#L25

The admin we check before we allow the call and also the object in the tuple. They can be different and (I assume) they often are.

This matches the example at https://library.canonical.com/corporate-policies/information-security-policies/ssdlc/ssdlc---security-event-logging#%EE%B0%823.4-user-related-events-[user]-9

It's not entirely clear to me that this is what I should log, though.

internal/db/identity.go
ale8k: In normal sql driver (which gorm should be using underneath I think) it's uhh add/delete/update I think, and I wonder if we'd be setting the name to the same thing so it counts as an update. I think it would?

internal/logger/securityevent_logger.go
ale8k: It's not super clear to me either tbh.  

And ah ok so the admin stuff is our "user 0" and in the end is mapped to a tuple anyways.

In our config we have initial admins, these are then created as superusers of jimm in openfga. We do a simple in memory check because they can't be changed unless jimm is restarted.

You would however only ever have one user doing the call though.

internal/logger/securityevent_logger.go
```diff
@@ -79,6 +79,28 @@ func LogGrantJimmAdmins(ctx context.Context, identityIds []string) {
...
+}
+
+// LogUserUpdated
+func LogUserUpdated(ctx context.Context, adminID string, userID string, relation string, target string, isAddition bool) {
+	action := "adding"
+	if !isAddition {
+		action = "removing"
+	}
+	logSecurityEvent(ctx, securityEvent{
+		Event:       fmt.Sprintf("user_updated:%s,%s,%s:%s", adminID, userID, relation, target), // ?
```
alesstimec: is this format correct?
the spec isn't really clear on this.. i would suggest something along these lines
```
user_updated:<admin user>,<user id>,<add/remove>,<relation>:<target>
```

luci1900: Yeah, I wasn't sure what would make sense. The spec is so unclear that I suspect no one cares much about these particular log messages.

internal/logger/securityevent_logger.go
luci1900: Had a long discussion with Alex, ended up logging only relation changes to users.

internal/logger/securityevent_logger.go
luci1900: I ended up also filtering by whether it's a relation added on a user. That should most closely match the RBAC semantics in the spec.

internal/db/identity.go
luci1900: I checked, gorm does indeed return 0 when it was a read and 1 on insert.

internal/logger/securityevent_logger.go
```diff
@@ -79,6 +79,28 @@ func LogGrantJimmAdmins(ctx context.Context, identityIds []string) {
 	})
 }
 
+// LogUserCreated
+func LogUserCreated(ctx context.Context, identityId string) {
```
ale8k, line 83: be good to test this

internal/jimm/permissions/relations.go
```diff
@@ -50,6 +52,7 @@ func (j *permissionManager) RemoveRelation(ctx context.Context, user *openfga.Us
 	if err != nil {
 		return errors.E(op, errors.CodeOpenFGARequestFailed, err)
 	}
+	j.logUserUpdates(ctx, user, parsedTuples, false)
```
ale8k, line 55: and this, we can pass a mock

internal/db/identity.go
ale8k, line 44: I'd put the logger on the *Database, then all can access it

kian99: The function just calls `zapctx.Warn` or `zapctx.Error`, there's nothing to put on the db object.

internal/jimm/permissions/relations.go
```diff
@@ -250,3 +253,13 @@ func (j *permissionManager) parseTuple(ctx context.Context, tuple apiparams.Rela
 
 	return &t, nil
 }
+
+// logUserUpdates logs tuple relation changes if they are for a user.
+// This should be the closest equivalent of logging an RBAC role in our Zanzibar-style openfga graph
+func (j *permissionManager) logUserUpdates(ctx context.Context, user *openfga.User, tuples []openfga.Tuple, isAddition bool) {
+	for _, tuple := range tuples {
+		if tuple.Object.Kind.String() == openfga.UserType.String() {
+			logger.LogUserUpdated(ctx, user.Name, tuple.Object.ID, tuple.Relation.String(), tuple.Target.ID, isAddition)
```
kian99, line 262: The only problem with this is that you wouldn't see when a user is assigned a role or assigned to a group. So it would be hard from these logs to discern when a user when granted a permission. I' suggest we also log events where the `Object` is a group or role.

luci1900: Shouldn't that be for the target being a user?

Giving a group access to an object would map to a role being given access to something in the spec's RBAC semantics, which I'm not sure would be logged.

kian99: Re-reading my initial message it is already confusing me lol. I was just thinking we should log when a group/role is granted access to something for traceability, i.e. if a user is a member/assigned to that group/role we can see their new permissions.

But going back to the ssdlc spec, I agree it wouldn't be logged so leaving it as is seems fine.

kian99 (approved):

internal/jimm/permissions/relations.go
luci1900: Ah, so adding a user to a group would put the group in the target position.

So you mean logging "changes" to groups or roles. That's not really in the spec though, I'm not sure how we'd log it.

kian99: Yeah exactly. To log it we can do the same thing we're doing in this PR but instead of `if tuple.Object.Kind.String() == openfga.UserType.String()` we could use a switch on the object's "kind". But not necessary for this PR.

luci1900: I think ultimately this spec is for an SRE/similar to have a quick overview across projects they aren't familiar with. Further investigation would have to be project-specific and in our case would presumably involve openfga's tuple logs.

internal/db/identity_test.go
```diff
@@ -62,6 +65,31 @@ func (s *dbSuite) TestGetIdentity(c *qt.C) {
 	c.Check(u4, qt.DeepEquals, u3)
 }
 
+func (s *dbSuite) TestIdentityUserCreatedLogging(c *qt.C) {
+	ctx := context.Background()
+
+	err := s.Database.Migrate(ctx)
+	c.Assert(err, qt.IsNil)
+
+	core, logs := observer.New(zap.InfoLevel)
```
luci1900, line 74: @SimoneDutto This seems to work pretty well.

internal/jimm/permissions/relations.go
luci1900: Added tests.

kian99 (commented):
Nice tests

internal/db/identity_test.go
ale8k, line 74: Snazzy

internal/jimm/permissions/relations.go
```diff
@@ -250,3 +253,13 @@ func (j *permissionManager) parseTuple(ctx context.Context, tuple apiparams.Rela
 
 	return &t, nil
 }
+
+// logUserUpdates logs tuple relation changes if they are for a user.
+// This should be the closest equivalent of logging an RBAC role in our Zanzibar-style openfga graph
```
ale8k: End comment in a period ".".

internal/logger/securityevent_logger.go
```diff
@@ -79,6 +79,28 @@ func LogGrantJimmAdmins(ctx context.Context, identityIds []string) {
...
+	logSecurityEvent(ctx, securityEvent{
+		Event:       fmt.Sprintf("user_created:%s", identityId),
+		Description: fmt.Sprintf("User %s was created.", identityId),
+		Severity:    warning,
+	})
+}
+
+// LogUserUpdated
```
ale8k: godoc missing

internal/logger/securityevent_logger.go
```diff
@@ -79,6 +79,28 @@ func LogGrantJimmAdmins(ctx context.Context, identityIds []string) {
 	})
 }
 
+// LogUserCreated
```
ale8k: godoc missing

internal/db/identity_test.go
```diff
@@ -62,6 +65,31 @@ func (s *dbSuite) TestGetIdentity(c *qt.C) {
...
+	ctx = zapctx.WithLogger(ctx, zap.New(core))
+
+	i, err := dbmodel.NewIdentity("bob")
+	c.Assert(err, qt.IsNil)
+	err = s.Database.GetIdentity(ctx, i)
+	c.Assert(err, qt.IsNil)
+	// Logging creation
+	c.Assert(logs.Len(), qt.Equals, 1)
+	c.Assert(logs.All()[0].Message, qt.Contains, "user_created:bob")
```
ale8k: Why contains? Why not just equals the exact message we expect?

ale8k (approved):
LGTM! Some minor comments

internal/db/identity_test.go
luci1900: I wasn't initially matching the user, no other reason.

PR #1702, opened 2025-09-10
feat: bootstrap integration test

ale8k:
## Description
Adds an integration test flow for the JIMM bootstrap. There's some side changes like setting the login token refresh & IP discovery for metallb & introduction of microk8s.

This is kinda disk intensive and we need constraints after some investigation with Kian, I'll be making subsequent cards to enable this.

At the moment it is only K8S, but I think without having a public cloud to test against this is "probably" good enough because k8s or lxd both represent "personal" clouds, and we'd be testing the same logic honestly. 

## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests


.github/actions/test-server/action.yaml
```diff
@@ -82,14 +91,18 @@ runs:
       working-directory: ${{ github.action_path }}/../../..
       env:
         JIMM_VERSION: ${{ inputs.jimm-version }}
+        JIMM_BOOTSTRAP_LOGIN_TOKEN_REFRESH_URL: "http://${{ env.MACHINE_IP }}:17070/.well-known/jwks.json"
 
     - name: Start server based on development version
       if: ${{ inputs.jimm-version == 'dev' }}
       working-directory: ${{ github.action_path }}/../../..
-      run: make dev-env
```
kian99: Why the switch?

.github/actions/test-server/action.yaml
kian99, line 1: This action is used by anyone that wants to start a test JIMM instance and is currently used by Juju and the juju-dashboard repos so I'm just scrutinizing it extra.

scripts/enable_metal_lb_microk8s.sh
kian99, line 1: Why is this file so complicated? The docs don't mention needing to do all this https://microk8s.io/docs/addon-metallb

kian99: Why don't we just choose a suitable address range like `10.12.13.0/24` allowing Microk8s to pick from any address in that range. We can even make the cidr configurable via an env var to allow the caller to set their own.

tests/bootstrap/bootstrap-k8s.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+
+# This script creates a model and deploys a basic application to it
+# before running `juju status` and then destroys the model.
+
+set -euo pipefail
+source "local/jimm/detect-jaas.sh"
+
+
```
kian99: extra new-line

kian99 (commented):
Ty for this. I'd suggest squashing all the commits into one.

ale8k:
> Ty for this. I'd suggest squashing all the commits into one.

Already done :P

.github/actions/test-server/action.yaml
ale8k: Had an env issue and wanted to eliminate any extra bits (like make). I do think it looks cleaner having both "docker compose" or both "make" and I think seeing the exact compose command is clearer. Happy to switch it back but I may make another make target for the test server (unless we already have one?)

.github/actions/test-server/action.yaml
ale8k: Yeah not a problem. In any case, no one can bootstrap without this env so it is needed.

scripts/enable_metal_lb_microk8s.sh
ale8k: So... When using cidr I had a weird issue where it was able to pick the gateway address literally (as in 10.12.13.0), and of course changing it to 2->255 worked, but sometimes it wouldn't. Hence the script which is just checking for 5 ips should we ever want 5 controllers.

.github/actions/test-server/action.yaml
kian99: Let's switch it back, make dev-env acts as the one place to do any pre-checks before starting the compose.

scripts/enable_metal_lb_microk8s.sh
kian99: Can you replace the script and just supply a range then like `10.12.13.50-10.12.13.100`. The script feels hard to maintain since it's not very straightforward.

SimoneDutto: I agree with Kian here. There must be a solution not involving building this complex script. I can't believe this is the solution just to enable metallb.
I have a few questions:
- have you seen problem in CI or just your machine?
- when you saw the problem can you share the log? I would like to understand how ARP possibly play a role here. It a layer 2 protocol, you usually don't want to mess with it unless it's strictly necessary.
- there are plenty of private address spaces you can pick and just assume they are not taking by other things installed on your/CI machine. And if we'll ever find a collision we will just pick a new one

ale8k: I can't, and you can see why, the latest test actually did fail because one of the IPs was taken. (https://github.com/canonical/jimm/actions/runs/17678580243/job/50246897337?pr=1702) 

Secondly metallb works at level 2? I'm not really following you here?

Yeah definitely. But they need to be in the same subnet cidr range? So may as well just ping a couple too then right?...

.github/actions/test-server/action.yaml
kian99: Looks like this change was done in the wrong place.
If the jimm-version is `dev` we call the make target but if we're starting a non-dev version of JIMM it was just calling the manual steps.

ale8k: Whoops lol changed

.github/workflows/integration-test.yaml
```diff
@@ -15,6 +16,16 @@ jobs:
...
         uses: docker/setup-compose-action@v1
 
+      - name: Set up Microk8s
+        run: |
+          sudo snap install microk8s --classic
+          sudo usermod -a -G microk8s "$USER"
+          newgrp microk8s
+
+          sudo microk8s enable storage
+          ./scripts/enable_metal_lb_microk8s.sh
```
SimoneDutto: can we inline the script here?

tests/bootstrap/bootstrap-k8s.sh
```diff
@@ -0,0 +1,30 @@
...
+set -euo pipefail
+source "local/jimm/detect-jaas.sh"
+
+JIMM_CONTROLLER_NAME="${JIMM_CONTROLLER_NAME:-jimm-dev}"
+
+sudo microk8s config | juju add-k8s testk8s --cluster-name=microk8s-cluster --client
+
+$JAAS bootstrap testk8s test-controller 3.6.8 --controller-service-type=loadbalancer
+
+# Would be nice to check "jaas controller" output?
```
SimoneDutto: I would say: either do it or remove the comment

.github/actions/test-server/action.yaml
```diff
@@ -82,6 +91,7 @@ runs:
       working-directory: ${{ github.action_path }}/../../..
       env:
         JIMM_VERSION: ${{ inputs.jimm-version }}
+        JIMM_BOOTSTRAP_LOGIN_TOKEN_REFRESH_URL: "http://${{ env.MACHINE_IP }}:17070/.well-known/jwks.json"
```
SimoneDutto, line 94: Here i'm ignorant. Doesn't juju require an https endpoint for the refresh url?

SimoneDutto (commented):
very minor things and LGTM

.github/actions/test-server/action.yaml
SimoneDutto: can you also remind me why we need this one?

ale8k: No no, the lib we used doesn't enforce it (thankfully).

SimoneDutto (approved):
thanks!

.github/actions/test-server/action.yaml
ale8k: This is to bootstrap the controller with the correct JWKS url, so that when we send JWTs to the controller they can verify the token.

kian99 (approved):

PR #1685, opened 2025-08-25
ci: add integration tests with the dashboard

kian99:
## Description
This PR builds on #1684 and adds a new workflow to perform integration tests with the Juju dashboard.

The workflow is almost identical to our current integration test setup but instead of running our integration tests, we run an action from the Juju dashboard repo that tests the dashboard against JIMM.

The workflow is currently set to run on `workflow_dispatch` and `pull_request` events. The `pull_request` event is only to see it work in this PR and then a follow-up will change it to only run when we release a new version of JIMM.

Fixes [JUJU-7937](https://warthogs.atlassian.net/browse/JUJU-7937)

## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests
- [x] Covered by integration tests

[JUJU-7937]: https://warthogs.atlassian.net/browse/JUJU-7937?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1703, opened 2025-09-10
ci: enhance release workflow

kian99:
## Description
This PR follows up on https://github.com/canonical/jimm/pull/1685 to only run the dashboard tests periodically and before releases.

In order to make it easier to run the dashboard tests before releases I've needed to slightly adjust our release workflows. We currently have 3 - release rocks, release snaps, run go-releaser (for release notes). Instead of updating each, I've created a "Release pre-checks" workflow that runs any pre-checks then triggers the individual release jobs. You can see an example of this working successfully on my fork here https://github.com/kian99/jimm/actions/runs/17610849517/job/50033609571, everything succeeded to the point where it attempted to release and failed due to lack of secrets.

I've also made a few changes to the dashboard-tests workflow.
1. It runs on a Github runner instead of a self-hosted. I had switched it to a self-hosted runner previously as I thought it was needed but running it on my fork on Github runners showed it was slightly faster there. Also, if jobs don't need the beefier machines of a self-hosted runner the suggestion is to use Github runners.
2. The dashboard tests can be triggered by adding a label to the PR called "test-dashboard". I think this will be useful when we want to update the dashboard image to a new revision and verify the change in the PR before landing it. I tested this on my fork here https://github.com/kian99/jimm/pull/2 and when the label is added the dashboard tests run. To retrigger the tests, the label must be removed and added again.

I've also tried to leave some comments for the non-intuitive bits of Github Actions.

### DIFF
--- a/.github/workflows/dashboard-tests.yaml
+++ b/.github/workflows/dashboard-tests.yaml
@@ -1,13 +1,30 @@
 name: Juju dashboard integration test
 
 on:
+  workflow_call:
   workflow_dispatch:
   pull_request:
+    types: [labeled]
+  schedule:
+    # Run twice a week: Mondays and Wednesdays at 03:00 UTC
+    - cron: '0 3 * * 1,3'
 
 jobs:
-  startjimm:
+  test-dashboard:
     name: Test JIMM with the Juju dashboard
-    runs-on: [self-hosted-linux-amd64-jammy-large]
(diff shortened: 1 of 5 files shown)


luci1900 (approved):
Looks reasonable. If it runs, should be good.

SimoneDutto (approved):

PR #1704, opened 2025-09-10
feat(bootstrap): adds progress bar to cli file download

alesstimec:
## Description

```
 ./jaas bootstrap testk8scloud test-controller 3.6.8 --controller-service-type=loadbalancer                                                                                                 v3

Starting bootstrap job.

Should you cancel this process, you can track the progress via bootstrap-status with the job ID:
        juju [jaas] bootstrap-status 2d9d165d-58f7-4f3c-a959-0f01b9e927d3

        Downloading the Juju CLI, version 3.6.8 for bootstrap. This may take a few minutes
[##......................................]   6%
[#####...................................]  13%
[########................................]  21%
[###########.............................]  28%
[##############..........................]  35%
[################........................]  42%
[###################.....................]  49%
[######################..................]  56%
[#########################...............]  64%
[############################............]  71%
[###############################.........]  78%
[##################################......]  85%
[#####################################...]  92%
[#######################################.]  99%
Creating Juju controller "test-controller" on testk8scloud/localhost
Bootstrap to Kubernetes cluster identified as microk8s/localhost
Creating k8s resources for controller "controller-test-controller"
Starting controller pod
Bootstrap agent now started
Contacting Juju controller at 192.168.1.0 to verify accessibility...
{}
```
## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests


internal/jujuclistore/store.go
```diff
@@ -123,7 +125,7 @@ func (b *Binary) Done() {
 // It retries the download on server errors or rate limiting.
 // The retry logic uses exponential backoff.
 // The context can be used to cancel the operation.
-func (p *jujuCLIStore) Get(ctx context.Context, spec JujuBinarySpec) (*Binary, error) {
+func (p *jujuCLIStore) Get(ctx context.Context, spec JujuBinarySpec, logFunction func(string)) (*Binary, error) {
```
kian99, line 138: I would prefer to see an interface here instead of a function.

internal/jujuclistore/store.go
```diff
@@ -312,3 +321,53 @@ func (p *jujuCLIStore) downloadFile(ctx context.Context, downloadUrl string) (*o
...
+
+	logFunction func(string)
+}
+
+// Read implements the io.Reader interface.
+func (pr *progressReader) Read(p []byte) (n int, err error) {
+	n, err = pr.Reader.Read(p)
+	pr.totalRead += n
+	if n > 0 && time.Since(pr.startTime) > pr.logInterval {
+		pr.startTime = time.Now()
```
kian99: Seems like startTime should be named `lastLogTime`

kian99 (approved):

internal/jujuclistore/store.go
luci1900: FWIW I think either is fine. Callbacks aren't entirely uncommon.

internal/jujuclistore/store.go
```diff
@@ -312,3 +321,53 @@ func (p *jujuCLIStore) downloadFile(ctx context.Context, downloadUrl string) (*o
 
 	return binaryFile, nil
 }
+
+type progressReader struct {
```
luci1900, line 335: There seems to be some prior art on this, any reason we wrote our own?

alesstimec: because it is easy enough to write one instead of searching for something and importing it

internal/jimm/cloudcred/attr.go
SimoneDutto, line 1: i've never seen this file before. What is it?

internal/jujuclistore/store.go
luci1900: Fair enough.

luci1900 (approved):
I approve of progress indicators, their lack can lead to users doing the wrong thing.

internal/jujuclistore/store.go
```diff
@@ -312,3 +321,53 @@ func (p *jujuCLIStore) downloadFile(ctx context.Context, downloadUrl string) (*o
...
+		int(percent),
+	)
+}
+
+func newProgressReader(r io.Reader, totalBytes int64, width int, logInterval time.Duration, logFunction func(string)) *progressReader {
+	if logInterval <= 0 {
+		logInterval = time.Second
+	}
+	if width <= 0 {
+		width = 40
```
SimoneDutto: can we put this 40 into a const to explain it's the expected width of the terminal. I think every terminal should be at least 40 chars wide.

internal/jujuclistore/store.go
```diff
@@ -312,3 +321,53 @@ func (p *jujuCLIStore) downloadFile(ctx context.Context, downloadUrl string) (*o
...
+	if n > 0 && time.Since(pr.startTime) > pr.logInterval {
+		pr.startTime = time.Now()
+		pr.logFunction(pr.logLine(pr.totalRead))
+	}
+	return n, err
+}
+
+func (pr *progressReader) logLine(progress int) string {
+	numStar := math.Floor(float64(progress) / float64(pr.totalBytes) * float64(pr.width))
+	percent := math.Floor(float64(progress) / float64(pr.totalBytes) * float64(100))
```
SimoneDutto, line 363: if totalBytes is 0 this will panic.
https://go.dev/play/p/--pxWO0HCv5

I can see some proxies or random http stuff setting content length to 0. Maybe validate isn't 0 in `new`

internal/jujuclistore/store.go
```diff
@@ -251,15 +253,22 @@ func (p *jujuCLIStore) downloadFile(ctx context.Context, downloadUrl string) (*o
 		return nil, jujuerrors.Errorf("request failed with status %d", resp.StatusCode)
 	}
 
-	xzReader, err := xz.NewReader(resp.Body)
+	var reader io.Reader
+	reader = resp.Body
+	if logFunction != nil {
+		reader = newProgressReader(reader, resp.ContentLength, 40, time.Second, logFunction)
```
SimoneDutto: are we sure ContentLength is always honored? With proxies and stuff?

ale8k:
Why can't we take cursor to start of line so it prints more nicely?

alesstimec:
@ale8k incremental improvements..

ale8k:
> @ale8k incremental improvements..

Why not now? We are adding the progress bar in this PR. 

Just try:
```go
	return fmt.Sprintf("\r[%s%s] %3d%%",
		strings.Repeat("#", int(numStar)),
		strings.Repeat(".", pr.width-int(numStar)),
		int(percent),
	)
```

I think \r should work.

internal/jujuclistore/store.go
luci1900: It can be missing https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Length

I don't think that'd make it `0`, though.

SimoneDutto:
> > @ale8k incremental improvements..
> 
> Why not now? We are adding the progress bar in this PR.
> 
> Just try:
> 
> ```go
> 	return fmt.Sprintf("\r[%s%s] %3d%%",
> 		strings.Repeat("#", int(numStar)),
> 		strings.Repeat(".", pr.width-int(numStar)),
> 		int(percent),
> 	)
> ```
> 
> I think \r should work.

I don't like the cursor thing, because if you try to c/p something (for example the command to stop the job) while it is writing with the cursor is a pain. I think it's fine to log each line.

internal/jimm/cloudcred/attr.go
ale8k: It's used to redact certain credential attributes when users call get cloud cred.

internal/jujuclistore/store.go
ale8k: Agreed, interface or function both provide a means to drop a mock in.

internal/jujuclistore/store.go
ale8k: It would because it isn't a pointer. I think we should probably have a check like "if it is 0, just print a simple Downloading.... message".

ale8k (approved):
lgtm, but think @SimoneDutto is right about the content length. We should probably guard against it.

internal/jujuclistore/store.go
alesstimec: keeping it as a function for now..

internal/jujuclistore/store.go
alesstimec: changed

internal/jujuclistore/store.go
alesstimec: well, ideally you'd get the terminal width and display accordingly, but there's no terminal involved here. there is on the client side, but the client does not send us that information.

alesstimec: done as a const

internal/jujuclistore/store.go
alesstimec: added a check

PR #1705, opened 2025-09-17
fix: use correct context for bootstrap unlock

kian99:
## Description
This PR fixes an issue where the bootstrap lock was not released when the user cancelled bootstrap. The issue was that we were using a cancelled context.

I've added a test for this, which tests that contexts used after the line `jobCtx = context.WithoutCancel(jobCtx)` correctly aren't cancelled when the bootstrap job is cancelled. It also tests that the bootstrap lock context is not cancelled.
I've verified that the test fails in scenarios where the corresponding logic is removed.

To speed up the test a bit I've lowered the minimum polling interval on the job-tracker to 1s from 5s otherwise the test may take up to 5s polling and waiting for the job to be cancelled. There may be other ways to handle this but this felt the most straightforward.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

### DIFF
--- a/internal/jujuclient/dial.go
+++ b/internal/jujuclient/dial.go
@@ -348,6 +348,11 @@ func (c *Connection) ConnectStream(path string, attrs url.Values) (base.Stream, 
 	if err != nil {
 		return nil, errors.E(op, err)
 	}
+	ok = names.IsValidUser(user)
+	if !ok {
+		zapctx.Error(c.ctx, "invalid controller credentials", zap.String("user", user))
+		return nil, errors.E(op, "invalid/missing controller credentials")
+	}
 	requestHeader := jujuhttp.BasicAuthHeader(names.NewUserTag(user).String(), pass)
 
 	conn, err := rpc.Dial(c.ctx, c.ctl, modelTag, path, requestHeader)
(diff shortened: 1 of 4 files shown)


internal/jujuclient/dial.go
```diff
@@ -348,6 +348,11 @@ func (c *Connection) ConnectStream(path string, attrs url.Values) (base.Stream,
 	if err != nil {
 		return nil, errors.E(op, err)
 	}
+	ok = names.IsValidUser(user)
+	if !ok {
+		zapctx.Error(c.ctx, "invalid controller credentials", zap.String("user", user))
+		return nil, errors.E(op, "invalid/missing controller credentials")
+	}
```
kian99, line 355: This is an unrelated change I had sitting locally. You can technically hit a panic without this change if the controller's credentials are not present in Vault due to the `names.NewUserTag()` function. So I've just added some validation.

luci1900 (approved):
Typically you'd want to be able to interrupt the cancel/cleanup path too, but in this case I can't imagine what that would be.

Seems fine.

SimoneDutto (approved):

PR #1708, opened 2025-09-19
refactor(jujuclient): updates facade version to 3.6.9 versions

alesstimec:
## Description

Updates facade versions.

As we prepare for the juju 4.0 release we want to make sure JIMM uses the latest 3.6 facade versions when talking to Juju controllers - this means it will be able to talk to both 3.6.9 and 4.0 controllers as 4.0 will support facade versions of 3.6.9.


internal/jujuclient/cloud.go
```diff
@@ -103,26 +96,16 @@ func (c Connection) RevokeCredential(ctx context.Context, cred names.CloudCreden
...
-		if err := c.Call(ctx, "Cloud", 1, "", "RevokeCredentials", &in, &out); err != nil {
-			return errors.E(op, jujuerrors.Cause(err))
-		}
+	in := jujuparams.RevokeCredentialArgs{
+		Credentials: []jujuparams.RevokeCredentialArg{{
+			Tag:   cred.String(),
+			Force: true,
+		}},
 	}
+	if err := c.CallHighestFacadeVersion(ctx, "Cloud", []int{7}, "", "RevokeCredentialsCheckModels", &in, &out); err != nil {
```
kian99, line 105: I assume this can be simplified because JIMM only supports recent versions of Juju controller so they will definitely have this method?

internal/jimm/juju/interface.go
```diff
@@ -171,10 +171,6 @@ type API interface {
 	// RevokeModelAccess revokes model access from a user.
 	RevokeModelAccess(context.Context, names.ModelTag, names.UserTag, jujuparams.UserAccessPermission) error
 
-	// SupportsCheckCredentialModels returns true if the
-	// CheckCredentialModels method can be used.
-	SupportsCheckCredentialModels() bool
```
kian99, line 176: Why was this removed?

kian99 (approved):

SimoneDutto (approved):
Can you add to the description the reason behind the update?

internal/jimm/juju/interface.go
alesstimec: it was used to decide which version of the facade to call.. since i'm removing older facade versions, this is no longer necessary..

PR #1709, opened 2025-09-22
refactor: use jujuparams over jujucloud for bootstrap credential

kian99:
## Description
This PR replaces our usage of `jujucloud.CloudCredential` with `jujuparams.CloudCredential` when sending the data over the wire. The former is used by the Juju CLI and holds a map of credentials for the cloud while the latter holds a single credential. Now, during bootstrap, if your client has multiple credentials for the cloud you are attempting to bootstrap, you must specify the credential name, unless a default is set. If your client only holds a single credential for that cloud, that credential will be used implicitly.

Fixes [JUJU-8531](https://warthogs.atlassian.net/browse/JUJU-8531)

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

[JUJU-8531]: https://warthogs.atlassian.net/browse/JUJU-8531?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cmd/jaas/cmd/bootstrap.go
+++ b/cmd/jaas/cmd/bootstrap.go
@@ -83,6 +83,7 @@ type bootstrapCommand struct {
 
 	// Flags
 
+	credentialName         string
 	timeout                int
 	publicDNSAddress       string
 	controllerServiceType  string
@@ -127,6 +128,7 @@ func (c *bootstrapCommand) SetFlags(f *gnuflag.FlagSet) {
 		"yaml": cmd.FormatYaml,
 		"json": cmd.FormatJson,
 	})
+	f.StringVar(&c.credentialName, "credential", "", "The name of the cloud credential to use for bootstrapping. Only required if more than one credential is available for the cloud.")
 	f.IntVar(&c.timeout, "timeout", 0, "The timeout in seconds for the bootstrap operation.")
 	f.StringVar(&c.publicDNSAddress, "public-dns-address", "", "Public DNS address (with port) of the controller")
 	f.StringVar(
@@ -162,9 +164,29 @@ func (c *bootstrapCommand) Run(ctxt *cmd.Context) error {
 		return fmt.Errorf("failed to get cloud %q: %w", c.cloud, err)
 	}
 
-	bootstrapCredential, err := c.store.CredentialForCloud(c.cloud)
+	cloudCreds, err := c.store.CredentialForCloud(c.cloud)
 	if err != nil {
 		return fmt.Errorf("failed to get credential for cloud %q: %w", c.cloud, err)
+	}
+
+	var bootstrapCred jujucloud.Credential
+	switch {
+	case len(cloudCreds.AuthCredentials) == 1 && c.credentialName == "":
+		// If there's only one credential and the user didn't specify a credential name, use it.
+		for _, cred := range cloudCreds.AuthCredentials {
+			bootstrapCred = cred
+			break
+		}
+	case c.credentialName != "":
+		// If a credential name is provided, use it.
+		var ok bool
+		bootstrapCred, ok = cloudCreds.AuthCredentials[c.credentialName]
+		if !ok {
+			return fmt.Errorf("no credential found with name %q", c.credentialName)
+		}
+	default:
+		// If there are multiple credentials and no name is provided, return an error.
+		return fmt.Errorf("multiple credentials found for cloud %q, please specify one using --credential", c.cloud)
 	}
 
 	req := apiparams.BootstrapStartParams{
@@ -173,7 +195,10 @@ func (c *bootstrapCommand) Run(ctxt *cmd.Context) error {
 		ControllerName:    c.controllerName,
 		ControllerVersion: c.controllerVersion,
 		Cloud:             cloudToParams(*bootstrapCloud),
-		Credential:        *bootstrapCredential,
+		Credential: jujuparams.CloudCredential{
+			Attributes: bootstrapCred.Attributes(),
+			AuthType:   string(bootstrapCred.AuthType()),
+		},
 
 		Flags: apiparams.BootstrapFlags{
 			Timeout:                c.timeout,
--- a/cmd/jaas/cmd/bootstrap.go
+++ b/cmd/jaas/cmd/bootstrap.go
@@ -184,9 +184,16 @@ func (c *bootstrapCommand) Run(ctxt *cmd.Context) error {
 		if !ok {
 			return fmt.Errorf("no credential found with name %q", c.credentialName)
 		}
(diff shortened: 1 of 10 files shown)


cmd/jaas/cmd/bootstrap.go
```diff
@@ -127,6 +128,7 @@ func (c *bootstrapCommand) SetFlags(f *gnuflag.FlagSet) {
 		"yaml": cmd.FormatYaml,
 		"json": cmd.FormatJson,
 	})
+	f.StringVar(&c.credentialName, "credential", "", "The name of the cloud credential to use for bootstrapping. Only required if more than one credential is available for the cloud.")
```
ale8k, line 131: Shouldn't this be "default" for the default credential? (I'm not certain if all defaults are called default or the cloud name so double check this)

cmd/jaas/cmd/bootstrap.go
```diff
@@ -162,18 +164,41 @@ func (c *bootstrapCommand) Run(ctxt *cmd.Context) error {
 		return fmt.Errorf("failed to get cloud %q: %w", c.cloud, err)
 	}
 
-	bootstrapCredential, err := c.store.CredentialForCloud(c.cloud)
+	cloudCreds, err := c.store.CredentialForCloud(c.cloud)
 	if err != nil {
 		return fmt.Errorf("failed to get credential for cloud %q: %w", c.cloud, err)
 	}
 
+	var bootstrapCred jujucloud.Credential
```
ale8k, line 172: Instead of checking for a singular credential and defaulting to doing nothing when no cred defined, I think we should be checking if a default cred name is defined and then looking that up in the map. I'm not 100% what default cred names are though. When you create one via the constructors it is just "default", and when you look at the clouds available, they always reference a credential name (which I think is the default credential).

ale8k: So maybe something like:
```go
	bootstrapCredential, err := c.store.CredentialForCloud(c.cloud)
	...
	credential := bootstrapCredential.AuthCredentials[bootstrapCredential.DefaultCredential]
```

internal/jujucommands/bootstrap.go
```diff
@@ -176,7 +176,15 @@ func (c *bootstrapCmd) Run(ctx context.Context, p BootstrapCmdParams) (<-chan Ou
 		}
 	}
 
-	if err := store.UpdateCredential(cloudName, p.CloudCred); err != nil {
+	// Create a cloudCredential at the last possible moment from the provided credential.
+	// A cloudCredential holds a map of credentials for the cloud with optional defaults.
+	// We only accept a single credential for bootstrapping.
+	cloudCred := jujucloud.CloudCredential{
```
ale8k, line 182: I don't know if this will work as it isn't setting a default - how would Juju know what to use? Or does it just take the first?

ale8k (changes requested):
I'm not sure we're doing this right, I think we should be looking up the default credential.

Also was this QA'd to work? If it does then maybe we can drop my comments get it in.

cmd/jaas/cmd/bootstrap.go
kian99: Seems a credential's name can be anything since user chooses the name and then default will then be the name of whichever credential the user defined as the default.
I've tested this now adding 2 AWS credentials to my client, this is what is inside `credentials.yaml`,
````
credentials:
  aws:
    aws-1:
      auth-type: access-key
      access-key: abc
      secret-key: abc
    aws-2:
      auth-type: access-key
      access-key: def
      secret-key: def
```
Then after running `juju default-credential aws aws-2` it becomes,
```
credentials:
  aws:
    default-credential: aws-2
    aws-1:
      auth-type: access-key
      access-key: abc
      secret-key: abc
    aws-2:
      auth-type: access-key
      access-key: def
      secret-key: def
```

cmd/jaas/cmd/bootstrap.go
kian99: Just a question, should we use the default credential in the bootstrap scenario? It seems fine but just want to confirm.

In any case, I'll add to the logic so that even if there are multiple credentials available, if a default is specified, that will be used. If there are multiple credentials available and no default, then we'll return the error asking the user to specify one.

internal/jujucommands/bootstrap.go
kian99: It seems to be working since the integration test is passing, but we can be even more explicit by setting `--credential=<credential-name>` on the bootstrap command. But since we currently only have 1 credential, this seems okay too.

kian99 (commented):
> Also was this QA'd to work?

@ale8k I didn't qa locally, but the integration test passes and bootstraps.

luci1900 (approved):
Seems like a straightforward improvement, as long as it still works.

cmd/jaas/cmd/bootstrap.go
ale8k: I'll resolve this and we can discuss in the other comment.

cmd/jaas/cmd/bootstrap.go
ale8k: I'm not sure either really... But yeah that sounds sensible. I do think default should always be specified though, perhaps if you could test this with a singular credential and then two credentials and see if a default is always set? Then we can simplify the logic?

internal/jujucommands/bootstrap.go
ale8k: Ah nice! Will resolve and leave other the comment open with the suggestion.

ale8k:
> > Also was this QA'd to work?
> 
> @ale8k I didn't qa locally, but the integration test passes and bootstraps.

That's fine then imo.

cmd/jaas/cmd/bootstrap.go
kian99: I'm not following here sorry. When I tested it, if you add credentials for a cloud, there is no default set until you run `juju default-credential <cloud> <credential-name>`.

ale8k (approved):
Approving but please check if default is always set and if it is, we can remove the picking of the first credential and simplify the logic? And if possible please ping me and let me know aha, interested to know if it is always set.

cmd/jaas/cmd/bootstrap.go
ale8k: Ah sorry I think I misunderstood, ok that's interesting, so there's no default until explicitly set? So it must just be iterating and taking the first then right?

kian99: Yeah that seems to be what it must be doing

PR #1710, opened 2025-09-23
feat: bootstrap config options

kian99:
## Description
This PR allows arbitrary config values to be passed to the JAAS plugin when bootstrapping a controller. These config values are then used when running the bootstrap command. We validate that the map of config values does not contain values like the login-token-refresh-url which JIMM sets.

There is a security implication worth considering here, if the Juju CLI has a vulnerability that allows it to do _something_ malicious when a particular combination of config values are set, we would be enabling that behavior. I'm also wary about any kind of injection attacks here where we are trusting arbitrary user input and passing it to the Juju CLI, but I can't come up with any concrete issues or mitigation.

Fixes [JUJU-8536](https://warthogs.atlassian.net/browse/JUJU-8536)

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

## Test instructions
I've attempted to QA this locally, but I'm getting the error `ERROR config without operator-storage value not valid.` which comes from Juju code and might have to do with my Microk8s setup.

[JUJU-8536]: https://warthogs.atlassian.net/browse/JUJU-8536?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


SimoneDutto (approved):

cmd/jaas/cmd/bootstrap.go
```diff
@@ -36,35 +37,38 @@ printing only the job ID, without waiting for the job to complete.
...
+These config options must match the config options supported by the Juju CLI
+for the version of Juju being bootstrapped. See the Juju documentation for
+the version specified for the full list of supported bootstrap config
+options.
+
+Note that some config options may not be specified as they will automatically
+be set.
+These are:
+
+- login-token-refresh-url
```
luci1900, line 56: This would generally suggest it should be its own separate top-level thing, but I'm not convinced that's an option here.

luci1900 (approved):
Seems like an improvement overall.

PR #1711, opened 2025-09-23
feat: make jobs more generic

luci1900:
## Description

Rename the various bootstrap job APIs to be generic and applicable to other cases of running a `juju` cli command in a JIMM facade API triggered from the `jaas` CLI.

Starting a job is not generic, so there's `GetJobInfo` and `StopJob`, but `StartBootstrapJob`. I will be adding `StartDestroyControllerJob` in a future PR.

The manager is also not generic, under the assumption that "job" is generally bootstrap-related or similar.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

### DIFF
--- a/cmd/jaas/cmd/bootstrap_unit_test.go
+++ b/cmd/jaas/cmd/bootstrap_unit_test.go
@@ -99,8 +99,8 @@ func (s *bootstrapCmdSuite) TestBootstrapRunDetached(c *gc.C) {
 			"cred-1": jujucloud.NewCredential(jujucloud.UserPassAuthType, map[string]string{}),
 		},
 	}, nil)
-	s.client.EXPECT().Bootstrap(gomock.Any()).DoAndReturn(func(bsp *params.BootstrapStartParams) (*params.BootstrapStartResponse, error) {
-		expected := &params.BootstrapStartParams{
+	s.client.EXPECT().StartBootstrapJob(gomock.Any()).DoAndReturn(func(bsp *params.BootstrapParams) (*params.StartJobResponse, error) {
+		expected := &params.BootstrapParams{
 			ControllerName: "controller-name",
 			CloudName:      cloudName,
 			RegionName:     "region",
@@ -125,7 +125,7 @@ func (s *bootstrapCmdSuite) TestBootstrapRunDetached(c *gc.C) {
 		c.Assert(bsp.Config, gc.DeepEquals, expected.Config)
 		c.Assert(bsp.ControllerVersion, gc.Equals, expected.ControllerVersion)
 
-		return &params.BootstrapStartResponse{
+		return &params.StartJobResponse{
 			JobID: "test-job-id",
 		}, nil
 	})
@@ -172,12 +172,12 @@ func (s *bootstrapCmdSuite) TestBootstrapWatchLogs(c *gc.C) {
 			"cred-1": jujucloud.NewCredential(jujucloud.UserPassAuthType, map[string]string{}),
 		},
 	}, nil)
-	s.client.EXPECT().Bootstrap(gomock.Any()).Return(&params.BootstrapStartResponse{
+	s.client.EXPECT().StartBootstrapJob(gomock.Any()).Return(&params.StartJobResponse{
 		JobID: "test-job-id",
 	}, nil)
 	s.client.EXPECT().Close().Return(nil)
 
-	s.client.EXPECT().BootstrapStatus(gomock.Any()).Return(params.BootstrapStatusResponse{
+	s.client.EXPECT().GetJobInfo(gomock.Any()).Return(params.GetJobInfoResponse{
 		Status:    params.StatusSuccessful,
 		Logs:      []string{"log-line", "log-line"},
 		Watermark: 2,
@@ -189,7 +189,7 @@ func (s *bootstrapCmdSuite) TestBootstrapWatchLogs(c *gc.C) {
 	}).Times(2)
 
 	s.writer.EXPECT().Write(gomock.Any()).DoAndReturn(func(b []byte) (int, error) {
-		c.Check(string(b), gc.Equals, "Bootstrap job completed successfully.\n")
+		c.Check(string(b), gc.Equals, "Job completed successfully.\n")
 		return len(b), nil
 	})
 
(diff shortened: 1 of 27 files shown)


cmd/jaas/cmd/bootstrap_unit_test.go
```diff
@@ -295,7 +295,7 @@ func (s *bootstrapCmdSuite) TestBootstrapMultipleCredentials(c *gc.C) {
 	}).Times(2)
 
 	s.writer.EXPECT().Write(gomock.Any()).DoAndReturn(func(b []byte) (int, error) {
-		c.Check(string(b), gc.Equals, "Bootstrap job completed successfully.\n")
+		c.Check(string(b), gc.Equals, "Job completed successfully.\n")
```
kian99, line 298: A nit but I think the ideal output here is `Bootstrap completed successfully` even though internally it's a job.

kian99: Ah reading lower dwn, I see since we're using the generic watchJobLogs logic we're limited to this, happy to leave in that case.

internal/db/job_logs.go
```diff
@@ -14,27 +14,27 @@ import (
...
 		return errors.E(op, err)
 	}
 
 	return d.Transaction(func(d *Database) error {
 		// Blocks all other operations, including reads, writes, and other locks.
-		if err := d.DB.Exec(bootstrapLoglockQuery).Error; err != nil {
-			return errors.E(op, "failed to lock bootstrap_logs table", err)
+		if err := d.DB.Exec(jobLoglockQuery).Error; err != nil {
+			return errors.E(op, "failed to lock job_logs table", err)
 		}
```
kian99, line 32: Huh, why do we have a lock around writing job logs.

pkg/api/params/params.go
```diff
@@ -586,21 +578,56 @@ const (
...
+// StartJobResponse holds the response for starting
+// a controller job.
+type StartJobResponse struct {
+	// JobID is the ID of the job that was started.
+	JobID string `json:"job-id"`
+}
+
+// BootstrapFlags holds the flags that can be used
+// when bootstrapping a new controller.
+type BootstrapFlags struct {
```
kian99: This struct is gone now since https://github.com/canonical/jimm/pull/1710, so we can remove this.

internal/jimm/bootstrap/bootstrap.go
```diff
@@ -73,7 +73,7 @@ type BinaryStore interface {
 	Get(ctx context.Context, spec jujuclistore.JujuBinarySpec, logFunction func(string)) (*jujuclistore.Binary, error)
 }
 
-// JujuCommands defines the Juju CLI methods that the bootstrap job requires.
+// JujuCommands defines the Juju CLI methods that the jobs require.
```
SimoneDutto: i think here is fine to leave the original comment.

kian99 (approved):
Looks good, I think it needs a rebase on `v3` to get the latest changes for the config options map.

internal/jimm/bootstrap/bootstrap.go
```diff
@@ -125,32 +125,32 @@ func NewBootstrapManager(
 	}, nil
 }
 
-// GetBootstrapStatusAndLogs retrieves the status and logs of a bootstrap job.
+// GetJobInfo retrieves the status and logs of a bootstrap job.
 // It requires the user to be an admin and returns the status, error message, logs,
 // and a watermark for pagination.
-func (b *bootstrapManager) GetBootstrapStatusAndLogs(ctx context.Context, _ *openfga.User, jobId uuid.UUID, offset int) (params.BootstrapStatusResponse, error) {
-	const op = errors.Op("jimm.GetBootstrapStatusAndLogs")
+func (b *bootstrapManager) GetJobInfo(ctx context.Context, _ *openfga.User, jobId uuid.UUID, offset int) (params.GetJobInfoResponse, error) {
```
SimoneDutto, line 131: mmm, looking at this it seems it should be called `JobManager` now.
But then we would need to find a place for the StartBootstrap

SimoneDutto: Here we have a mix of generic facade implementation mixed with very specific stuff. Like `tryToCleanupController`. I don't like it

cmd/jaas/cmd/bootstrap_unit_test.go
luci1900: We could get it to write the job type, but it'd require more rearranging. I was trying to keep the changes minimal.

internal/jimm/bootstrap/bootstrap.go
SimoneDutto: what do you think about having a bootstrap manager having a job manager as a dependency? So you can put all of your methods for start, stop, log in a single manager but the specific one in its own bootstrap manager.
And we can reuse the jobmanager elsewhere in the future.

internal/db/job_logs.go
luci1900: I'm also not sure why that happens. We don't automatically restart jobs on a different jimm instance, for instance.

SimoneDutto (commented):
Nice stuff, i can totally see the amount of back and forth refactoring all this.

One thing we can discuss though is whether we like this hybrid implementation.
Having half generic methods with half facade name inside the bootstrap manager is something i don't necessarly like.

Maybe we can extract and using the jobmanager as a dependency of the bootstrap manager.
This would allow us to use to JobManager from other managers or god forbid swapping the JobManager implementation without touching the other bits.

What do you think?

If we want to proceed with this implementation instead i'm happy to +1

internal/jimm/bootstrap/bootstrap.go
luci1900: I had started on just that https://github.com/canonical/jimm/commit/9235a532c68250fcfd32d46cb33fa26b4e6a1835, I was talking to Alex about it yesterday.

The initial implementation was already mixing generic and specific, so I wasn't sure how far it's worth going. If we did want the ability to call juju commands from jimm APIs, we'd probably want to make it generic also in how they're started and how they're invoked from the jaas cli, so users could call arbitrary juju commands on any of their jimm-registered controllers.

Do we expect this dodgy bootstrap approach to remain long-term? If not, I wouldn't try to make it much nicer.

kian99: I agree on not spending the time right now to make a jobManager while we think it's only going to be "short-lived" in the bootstrap logic.

SimoneDutto: why do we say "short-lived"? Are we planning on changing this implementation?

luci1900: I guess it's a hope that the bootstrap logic will get redone after juju 4?

I struggle with what the overall least bad choice would be.

SimoneDutto: mmm, the bootstrap story in 4 is still the same though, i don't think we will change this implementation soon tbh but i don't want to block this pr

internal/db/job_logs.go
luci1900: @ale8k Can you remember why?

SimoneDutto (approved):

PR #1715, opened 2025-09-26
chore(dependencies): updates dependencies

alesstimec:
## Description

Updates dependencies to address high vulnerabilities. 

**NOTE**: removes generate of credenital attributes in internal/jimm/cloudcred -> we need to address this in a followup.


kian99:
@alesstimec Looks like this needs a `go mod tidy` and the linter is complaining that 2 methods `CommandBase` and `NewAPIRootWithDialOpts` are missing.

Issue #1717, opened 2025-10-09
SSH Keys not being added to machines


jsimpso:
When adding SSH keys to a model managed by JAAS, the keys don't seem to be propagating to the machines as expected.

```
jsimpso@is-bastion-ps7:~$ juju ssh-keys
No keys to display.
jsimpso@is-bastion-ps7:~$ juju add-ssh-key "$(cat ~/.ssh/id_ed25519.pub)"
jsimpso@is-bastion-ps7:~$ juju ssh-keys
Keys used in model: d7629a7e-3e09-4b0e-bef5-3bae563632dc@serviceaccount/test-machine-model-1-ps7
92:f3:c9:ee:64:18:c6:4d:82:78:61:28:75:dd:91:3b (jsimpso@is-bastion-ps7)
jsimpso@is-bastion-ps7:~$ juju status
Model                     Controller  Cloud/Region           Version  SLA          Timestamp
test-machine-model-1-ps7  jaas-ps7    prodstack7/prodstack7  3.6.8    unsupported  05:15:47Z

App     Version  Status  Scale  Charm   Channel        Rev  Exposed  Message
ubuntu  24.04    active      1  ubuntu  latest/stable   26  no

Unit       Workload  Agent  Machine  Public address  Ports  Message
ubuntu/0*  active    idle   0        10.151.199.13

Machine  State    Address        Inst id                               Base          AZ   Message
0        started  10.151.199.13  25d3f81d-a60f-42ee-936a-cf2bf7234382  ubuntu@24.04  AZ2  ACTIVE
jsimpso@is-bastion-ps7:~$ juju ssh 0
ubuntu@10.151.199.13: Permission denied (publickey).
jsimpso@is-bastion-ps7:~$ ssh ubuntu@10.151.199.13 -i .ssh/id_ed25519
ubuntu@10.151.199.13: Permission denied (publickey).
jsimpso@is-bastion-ps7:~$ juju exec --unit ubuntu/0 -- cat /home/ubuntu/.ssh/authorized_keys
ssh-rsa <truncated> Juju:juju-system-key
```

There's no mention of the key operation in `/var/log/juju/machine-0.log`. They *do* exist in the JIMM database.

If I swap to talk directly with one of the backing controller's `controller` model, the operation works as expected:
```
jsimpso@is-bastion-ps7:~$ juju switch jaas-controller-36-testing-ps7
[...]

PR #1718, opened 2025-10-13
fix: don't intercept ssh key methods

kian99:
## Description
This PR addresses https://github.com/canonical/jimm/issues/1717 preventing users from adding SSH keys to models through JIMM.

The model-proxy was intercepting ssh key requests to work correctly with the ssh-proxy feature that was previously being developed. 
Without that feature complete in Juju we were unintentionally preventing the addition of SSH keys to models. The fix is to remove this logic and allow the commands to go through the model as normal.

In Juju 4 SSH key methods are not model-scoped so they can be implemented without intercepting them with the model proxy.

Fixes [JUJU-8626](https://warthogs.atlassian.net/browse/JUJU-8626)

## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests

## Test instructions
Tested by doing the following:
1. `make qa-lxd`
2. `juju add-model foo`
3. `juju add-machine`
4. `juju ssh 0` <- verify SSH to machine works
5. `ssh-keygen` create the key in a file like `test-key`
6. `juju add-ssh-key "$(cat test-key.pub)"`
7. `juju list-ssh-keys` Verify the key exists in the model
8. `juju ssh 0`
9. `cat ~/.ssh/authorized_keys` There should be 3 keys present. One Juju system key, one which was your original host key, and a final one which should match the key you just created.

[JUJU-8626]: https://warthogs.atlassian.net/browse/JUJU-8626?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


SimoneDutto (approved):
our code, our beautiful code is on fire 🔥

luci1900 (approved):

PR #1720, opened 2025-10-17
ci: add permissions for releasing server rock

kian99:
## Description
The release server rock workflow is failing due to the error `denied: installation not allowed to Write organization package`. It seems the workflow is missing the necessary permissions. Because of the changes in https://github.com/canonical/jimm/pull/1703 I think I misunderstood the permissions of nested workflows. The pre-release workflow has permissions `content: write` but the workflow it invokes does not have the necessary permissions. This PR aims to address that.

I tried to test the changes on my fork, but there I get a different error that I'm unsure is related.

### DIFF
--- a/.github/workflows/release-server-rock.yaml
+++ b/.github/workflows/release-server-rock.yaml
@@ -4,6 +4,12 @@ on:
   workflow_dispatch:
   workflow_call:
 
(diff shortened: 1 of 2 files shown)


SimoneDutto (approved):

luci1900 (approved):
Let's try it, I don't think there's much potential downside.

PR #1722, opened 2025-10-29
fix(add-model): dials juju as external user on add-model

alesstimec:
## Description

Changes the way JIMM dials the controller when adding a model. This is to achieve Juju 4 compatibility when adding a model - juju 4 has strong referential integrity, so atm JIMM fails to add cloud credentials for an external user, because that external user does not exist in controller's database. After [this](https://github.com/juju/juju/pull/21027) change, this PR should at least enable us to add credentials and a model.

## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests


internal/openfga/user.go
```diff
@@ -301,7 +301,7 @@ func (u *User) CheckPermission(ctx context.Context, resourceTag string, permissi
 		relation = ofganames.ReaderRelation
 	case string(params.OfferConsumeAccess):
 		relation = ofganames.ConsumerRelation
-	case string("add-cloud"):
+	case string("add-model"):
```
luci1900: Was this a bug or are we changing semantics?

alesstimec: uncaught bug, i think

luci1900 (approved):
Seems cleaner than granting permissions separately, but obviously incomplete.

kian99 (commented):
I assume there's no test changes because this was already tested and we're just refactoring?

internal/jimm/juju/modelbuilder.go
```diff
@@ -420,15 +430,20 @@ func (b *modelBuilder) CreateControllerModel() *modelBuilder {
 		return b
 	}
 
+	permissions := []permission{}
+	if err := b.creator.CheckPermission(b.ctx, b.cloud.ResourceTag().String(), "add-model"); err != nil {
```
SimoneDutto: don't we have a const for this permission?

internal/openfga/user.go
SimoneDutto: let's create a const

internal/jimm/juju/modelbuilder.go
```diff
@@ -474,17 +489,6 @@ func (b *modelBuilder) CreateControllerModel() *modelBuilder {
...
-	// the local database entry will be deleted but the model
-	// will remain on the controller and will trigger the "already exists
-	// in the backend controller" message above when the user
-	// attempts to create a model with the same name again.
-	if err := api.GrantJIMMModelAdmin(b.ctx, names.NewModelTag(info.UUID)); err != nil {
-		zapctx.Error(b.ctx, "leaked model", zap.String("model", info.UUID), zaputil.Error(err))
-		b.err = errors.E(err)
-		return b
-	}
-
```
SimoneDutto, line 487: why this deletion?

SimoneDutto:
out of scope for this pr: why don't we add a `non-required` workflow to run the integration test with juju 4 controller?

internal/jimm/juju/modelbuilder.go
kian99: This is a leftover from long ago when JIMM had to grant itself access to things, now it's all in the JWT but there are sometimes these leftovers we didn't delete.

kian99 (approved):

internal/jimm/juju/modelbuilder.go
alesstimec: there is no need to grant jimm permissions

.vscode/launch.json
```diff
@@ -23,7 +23,7 @@
                     "to": "/jimm/",
                 },
             ],
-            "port": 2345,
```
SimoneDutto: what's up with these changes?

SimoneDutto (changes requested):

.vscode/launch.json
alesstimec: shutup and take them :D:D

SimoneDutto (approved):

alesstimec:
Closing this old PR in favor of Luci's new work

Issue #1727, opened 2025-11-04
Error parsing login access token


jsimpso:
When trying to create a model under JAAS (channel 3/edge revision 90), using Terraform, I see this error:

```
module.test_jaas_machine_model_ps7.juju_model.service_model: Creating...
╷
│ Error: Client Error
│ 
│   with module.test_jaas_machine_model_ps7.juju_model.service_model,
│   on .terraform/modules/test_jaas_machine_model_ps7/services/compute/machine_model_jaas/main.tf line 67, in resource "juju_model" "service_model":
│   67: resource "juju_model" "service_model" {
│ 
│ Unable to create model "\"test-jaas-machine-model-ps7\"", got error: parsing login access token: key provider 0 failed: failed to find key with key ID
│ "3cbdfb30-89b8-4c90-9b74-c4a3e374e899" in key set
```

The model has actually been created:
```
jsimpso@nautilus:~$ juju models
Controller: jaas-ps7

Model                                                                             Cloud/Region           Type       Status     Machines  Cores  Units  Access  Last connection
d69974a4-a959-4528-ba99-daafa088c978@serviceaccount/test-jaas-machine-model-ps7*  prodstack7/prodstack7  openstack  available         1      2  1      admin   never connected
```

I see the same error when trying to import the model:
```
$ terraform import module.test_jaas_machine_model_ps7.juju_model.service_model test-jaas-machine-model-ps7

module.test_jaas_machine_model_ps7.juju_model.service_model: Import prepared!
  Prepared juju_model for import
╷
│ Error: Client Error
│ 
│ parsing login access token: key provider 0 failed: failed to find key with key ID "3cbdfb30-89b8-4c90-9b74-c4a3e374e899" in key set
╵
```

I also see this error when trying to `juju destroy-model` directly via the CLI, where I'm authenticated as my own OIDC user rather than the service account used by Terraform:
```
simpso@nautilus:~$ juju destroy-model d69974a4-a959-4528-ba99-daafa088c978@serviceaccount/test-jaas-machine-model-ps7
WARNING This command will destroy the "d69974a4-a959-4528-ba99-daafa088c978@serviceaccount/test-jaas-machine-model-ps7" model and affect the following resources. It cannot be stopped.

 - 0 application will be removed
 - 0 filesystem and 0 volume will be destroyed

To continue, enter the name of the model to be unregistered: d69974a4-a959-4528-ba99-daafa088c978@serviceaccount/test-jaas-machine-model-ps7
[...]

PR #1728, opened 2025-11-05
refactor: use lower jwks cache age

kian99:
## Description
This PR lowers the cache header timeout in our JWKS handler. This will tell clients to refresh their JWKS public key more frequently which should help in scenarios where users are recovering a JIMM instance and will allow them to avoid restarting Juju controllers to refresh the JIMM public key. We don't expect many clients to be fetching the JWKS, one per Juju controller.

The new max age is 10 minutes unless the JWKS expiry is <10 min.

This may also help scenarios described in #1727. To further alleviate issues with public key expiry, we also need to generate a new JWKS before the old one expires and store both.

## Engineering checklist
- [ ] Documentation updated
- [x] Covered by unit tests

### DIFF
--- a/internal/jimmhttp/wellknown_handler.go
+++ b/internal/jimmhttp/wellknown_handler.go
@@ -96,12 +96,17 @@ func (wkh *WellKnownHandler) JWKS(w http.ResponseWriter, r *http.Request) {
 	actualMaxAge := int64(maxAgeSeconds)
 	expiresTime := now.Add(maxAgeSeconds * time.Second)
 
+	// If the expiry is sooner than now + maxAgeSeconds, use the expiry instead.
+	// If the expiry is in the past, set a low cache value of 30s.
 	if expiry.After(now) {
 		remaining := int64(expiry.Sub(now).Seconds())
 		if remaining < maxAgeSeconds {
 			actualMaxAge = remaining
 			expiresTime = expiry
 		}
+	} else {
+		actualMaxAge = 10
+		expiresTime = now.Add(10 * time.Second)
 	}
 	w.Header().Set("Cache-Control", fmt.Sprintf("must-revalidate, max-age=%d, immutable", actualMaxAge))
 	w.Header().Set("Expires", expiresTime.Format(time.RFC1123))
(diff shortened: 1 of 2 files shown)


ale8k (approved):
LGTM.

internal/jimmhttp/wellknown_handler.go
```diff
@@ -89,19 +88,22 @@ func (wkh *WellKnownHandler) JWKS(w http.ResponseWriter, r *http.Request) {
...
-	// 	- immutable (so the client of this JWKS knows it will not change until the rotation date)
-
-	w.Header().Add("Cache-Control", fmt.Sprintf("must-revalidate, max-age=%d, immutable", int64(math.Floor(maxAge.Seconds()))))
-	// We also use expires as I've noticed some JWK cache clients in golang specifically
-	// look at the expires header over max-age directive... No idea why.
-	w.Header().Add("Expires", expires)
+	// Set cache headers to a maximum age of 10 minutes (600 seconds)
+	// to ensure clients do not cache for too long.
+	// If expiry is less than 10 minutes, use the actual expiry
+	const maxAgeSeconds = 600
```
alesstimec, line 94: why don't we make this configurable and save ourselves future PRs tweaking this value?

alesstimec (approved):

internal/jimmhttp/wellknown_handler.go
kian99: I would prefer to see this in a follow-up if it becomes necessary.

internal/jimmhttp/wellknown_handler.go
```diff
@@ -89,19 +88,22 @@ func (wkh *WellKnownHandler) JWKS(w http.ResponseWriter, r *http.Request) {
...
+	now := time.Now().UTC()
+	actualMaxAge := int64(maxAgeSeconds)
+	expiresTime := now.Add(maxAgeSeconds * time.Second)
+
+	if expiry.After(now) {
```
luci1900, line 101: Explanation to negative value case would be nice.

luci1900 (approved):

internal/jimmhttp/wellknown_handler.go
luci1900: So I don't forget, it may be better to only set the header to cache if it's not already expired

PR #1732, opened 2025-11-18
fix(batch-openfga-req): batch openfga requests

### LINKED ISSUE #1729: OpenFGA operation limit trying to add >100 users to a group
I'm trying to create a large JAAS group via Terraform, and I'm seeing an error returned from OpenFGA:

```
juju_jaas_access_group.iam_group: Creating...
╷
│ Error: Client Error
│
│   with juju_jaas_access_group.iam_group,
│   on main.tf line 57, in resource "juju_jaas_access_group" "iam_group":
│   57: resource "juju_jaas_access_group" "iam_group" {
│
│ Unable to create access relationships for group-58ce1c7f-c625-463a-8a6e-fe2c8d014b1e, got error: cannot add or remove relations: Write validation error for POST Write with body {"code":"exceeded_entity_limit","message":"The number of write
│ operations exceeds the allowed limit of 100"}
│  with error code exceeded_entity_limit error message: The number of write operations exceeds the allowed limit of 100 (failed request to OpenFGA)
```

It looks like the default for `maxTuplesPerWrite` is `100` per [OpenFGA documentation](https://openfga.dev/docs/getting-started/setup-openfga/configuration), which tracks with the error we're seeing here.

Could we please consider implementing some kind of transaction batching in JAAS when writing to OpenFGA, splitting up larger requests into chunks? 

The OpenFGA charm doesn't currently expose `maxTuplesPerWrite` as a config option, but I think we should consider providing the `maxTuplesPerWrite` value via the `openfga` interface regardless. Then we'd be able to chunk batches against a real number rather than a magic number, in case that option becomes exposed in the future, is changed out of band, or the default changes at some point.

Let me know if that sounds reasonable and if you'd like me to raise a feature request for that somewhere on the OpenFGA side :)

SimoneDutto:
## Description

openfga has a maxTuplesPerWrite config value (default to 100). We want to support JIMM.AddRelation with more than the openfga config value, so we are batching requests to OpenFga. This has some challenges:
 - we need the openfga charm to share this config value with us.
 - we need to handle partial failures. 

Both are working in progress because we are waiting on other projects' work, but this is a good intermediate step to solve the issue to avoid the error. 

The full solution to this issue is to make this operation idempotempt (which currently is not) because it would leverage a new feature in the openfga 1.10 to avoid `confict on duplicated write` (Read more: https://github.com/openfga/go-sdk/blob/v0.7.3/README.md#conflict-options-for-write-operations).
But it has a broader scope crossing some teams:
- update openfga charm
- update openfga canonical wrapper client
- (potentially) update terraform

partially fix #1729

### DIFF
--- a/internal/jimm/permissions/relations.go
+++ b/internal/jimm/permissions/relations.go
@@ -18,6 +18,13 @@ import (
 	apiparams "github.com/canonical/jimm/v3/pkg/api/params"
 )
 
+// BATCH_SIZE_OPENFGA defines the maximum number of tuples to process in a single batch operation.
+// This is the default for maxTuplesPerWrite
+// See: https://openfga.dev/docs/getting-started/setup-openfga/configuration
+// TODO: this value should be received from the OpenFGA charm's relation, so we make sure that we batch
+// requests according to the deployed OpenFGA instance configuration.
+const BATCH_SIZE_OPENFGA = 100
+
 // AddRelation checks user permission and add given relations tuples.
 // At the moment user is required be admin.
 func (j *permissionManager) AddRelation(ctx context.Context, user *openfga.User, tuples []apiparams.RelationshipTuple) error {
@@ -29,11 +36,19 @@ func (j *permissionManager) AddRelation(ctx context.Context, user *openfga.User,
 	if err != nil {
 		return errors.E(err)
 	}
-	err = j.authSvc.AddRelation(ctx, parsedTuples...)
-	if err != nil {
-		return errors.E(op, errors.CodeOpenFGARequestFailed, err)
+	for i := 0; i < len(parsedTuples); i += BATCH_SIZE_OPENFGA {
+		end := i + BATCH_SIZE_OPENFGA
+		if end > len(parsedTuples) {
+			end = len(parsedTuples)
+		}
+		batch := parsedTuples[i:end]
+
+		err = j.authSvc.AddRelation(ctx, batch...)
+		if err != nil {
+			return errors.E(op, errors.CodeOpenFGARequestFailed, err)
+		}
+		j.logUserUpdates(ctx, user, batch, true)
 	}
-	j.logUserUpdates(ctx, user, parsedTuples, true)
 	return nil
 }
 
@@ -48,11 +63,19 @@ func (j *permissionManager) RemoveRelation(ctx context.Context, user *openfga.Us
 	if err != nil {
 		return errors.E(op, err)
 	}
-	err = j.authSvc.RemoveRelation(ctx, parsedTuples...)
-	if err != nil {
-		return errors.E(op, errors.CodeOpenFGARequestFailed, err)
+	for i := 0; i < len(parsedTuples); i += BATCH_SIZE_OPENFGA {
+		end := i + BATCH_SIZE_OPENFGA
+		if end > len(parsedTuples) {
+			end = len(parsedTuples)
+		}
+		batch := parsedTuples[i:end]
+
+		err = j.authSvc.RemoveRelation(ctx, batch...)
+		if err != nil {
+			return errors.E(op, errors.CodeOpenFGARequestFailed, err)
+		}
+		j.logUserUpdates(ctx, user, batch, true)
 	}
-	j.logUserUpdates(ctx, user, parsedTuples, false)
 	return nil
 }
 
--- a/internal/jimm/permissions/relations_test.go
+++ b/internal/jimm/permissions/relations_test.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"slices"
 
+	petname "github.com/dustinkirkland/golang-petname"
 	qt "github.com/frankban/quicktest"
 	"github.com/juju/zaputil/zapctx"
 	"go.uber.org/zap"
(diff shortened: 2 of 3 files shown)


internal/jimm/permissions/relations.go
```diff
@@ -29,11 +36,19 @@ func (j *permissionManager) AddRelation(ctx context.Context, user *openfga.User,
...
-	if err != nil {
-		return errors.E(op, errors.CodeOpenFGARequestFailed, err)
+	for i := 0; i < len(parsedTuples); i += BATCH_SIZE_OPENFGA {
+		end := i + BATCH_SIZE_OPENFGA
+		if end > len(parsedTuples) {
+			end = len(parsedTuples)
+		}
+		batch := parsedTuples[i:end]
+
+		err = j.authSvc.AddRelation(ctx, batch...)
```
luci1900, line 46: Do we care about partial failure here and eventual retry?

I haven't checked if `AddRelation` is idempotent.

internal/jimm/permissions/relations.go
```diff
@@ -29,11 +36,19 @@ func (j *permissionManager) AddRelation(ctx context.Context, user *openfga.User,
...
+		if err != nil {
+			return errors.E(op, errors.CodeOpenFGARequestFailed, err)
+		}
```
kian99, line 49: Will we add in a follow-up, the number of tuples written if we fail?

internal/jimm/permissions/relations_test.go
```diff
@@ -464,3 +466,72 @@ func (s *permissionManagerSuite) TestRelationshipLogUserUpdated(c *qt.C) {
...
+		Relation:     names.ReaderRelation.String(),
+		TargetObject: model.ResourceTag().String(),
+	}, 100, "")
+	c.Assert(err, qt.IsNil)
+	c.Assert(tuplesListed, qt.HasLen, len(tuples))
+}
+
+func (s *permissionManagerSuite) TestRemoveRelationBatch(c *qt.C) {
+	c.Parallel()
+	ctx := context.Background()
```
kian99: I think use `c.Context()` if it has it.

internal/jimm/permissions/relations.go
SimoneDutto: We care because if you try to add 1000 things from terraform and you fail midway when you retry you will receive an error because there is a conflict on write.

I'll expand the PR description a bit to give more context

kian99 (approved):

internal/jimm/permissions/relations.go
SimoneDutto: I was re-thinking about this, the best option is to make this `AddRelation` idempotent with the new openfga feature. I've messaged the Identity team to have an ETA for the update.

Returning the number and persisting it in the state would feel just an hack, and i'm ok to do it if we know the openfga charm will not be updated.

internal/jimm/permissions/relations.go
luci1900: In that case, we may need to swallow the error from `AddRelation`/`RemoveRelation` in the case that the state is already what we're trying to make it be.

I suppose we already had that problem in case one of the sent relations was already added/removed.

Should be fine as a potential follow-up.

luci1900 (approved):

internal/jimm/permissions/relations.go
luci1900: It doesn't seem like the number actually helps on its own. You still wouldn't be able to do something about it without knowing specifically which caused the call to fail.

internal/jimm/permissions/relations.go
SimoneDutto: if i've understood correctly you are saying to check the err type and not return it in case it's `writeDuplicate`.
However i believe openfga will return this error if one single tuple in the list is duplicated, so it might be that the state is not what you want but just one single element is duplicated.

internal/jimm/permissions/relations.go
kian99: The thinking was that if the client has the number of tuples written, at least in the case of Terraform, it can record that in its state but still mark the resource creation as failed. That will create a tainted resource that the user can either `untaint` making their next apply an `update` or allow Terraform to recreate the tainted resource, deleting the semi-added tuples then re-add everything.

There is more to this story that was discussed but isn't written here. I would like to see the number of tuples returned to give clients the chance to be more "correct" if possible.

internal/jimm/permissions/relations.go
luci1900: Exactly, the write of X tuples can fail because of a single duplicate, whether we're trying to send 100 or 10.

The solution would be for the operation to be idempotent. Either openfga does that or we can do that by retrying the call without the offending duplicate.

Either way, even without the batching this was already a potential problem that the caller would have to resolve by manually removing the duplicate anyway. The batching merely makes it more likely to cause duplicates by only executing some of the batches until an error.

internal/jimm/permissions/relations.go
SimoneDutto: But this is not needed in case this operation is idempotent. So i'm waiting to understand how much will it take to release the 1.10 openfga

SimoneDutto: or even as @luci1900 suggested in the other comment, making this operation idempotent regardless of openfga

PR #1733, opened 2025-11-18
fix: tweak cache-control headers

kian99:
## Description
This PR tweaks the cache-control headers we use when sending our JWKS to clients.
Namely,
1. Don't set the `Expires` header, based on [MSDN docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Caching#expires_or_max-age) it is out-dated and we can almost certainly safely stop using it.
> In HTTP/1.0, freshness used to be specified by the Expires header.
>
> The time format is difficult to parse, many implementation bugs were found, and it is possible to induce problems by intentionally shifting the system clock; therefore, max-age — for specifying an elapsed time — was adopted for Cache-Control in HTTP/1.1.
> If both Expires and Cache-Control: max-age are available, max-age is defined to be preferred. So it is not necessary to provide Expires now that HTTP/1.1 is widely used.
2. Remove the `immutable` keyword. I don't have a very strong argument on this one. When asking AI to evaluate the current cache-control headers it said `immutable` was not valid for a JWKS endpoint because 
> JWKS documents contain mutable signing keys used for JWT validation.
Keys can rotate:
> automatically
> due to compromise
> due to policy
> during deployment
> unexpectedly

JWKS endpoints from [Auth0](https://samples.auth0.com/.well-known/jwks.json) and [Google](https://www.googleapis.com/oauth2/v3/certs) also don't include this keyword. Digging further, the [MSDN docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Cache-Control) don't say too much on this header except that it,
> The immutable response directive indicates that the response will not be updated while it's [fresh](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Caching#fresh_and_stale_based_on_age).

Partially addresses [JUJU-8776](https://warthogs.atlassian.net/browse/JUJU-8776)

[JUJU-8776]: https://warthogs.atlassian.net/browse/JUJU-8776?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/jimmhttp/wellknown_handler.go
+++ b/internal/jimmhttp/wellknown_handler.go
@@ -94,7 +94,6 @@ func (wkh *WellKnownHandler) JWKS(w http.ResponseWriter, r *http.Request) {
 	const maxAgeSeconds = 600
 	now := time.Now().UTC()
 	actualMaxAge := int64(maxAgeSeconds)
-	expiresTime := now.Add(maxAgeSeconds * time.Second)
 
 	// If the expiry is sooner than now + maxAgeSeconds, use the expiry instead.
 	// If the expiry is in the past, set a low cache value of 30s.
@@ -102,13 +101,10 @@ func (wkh *WellKnownHandler) JWKS(w http.ResponseWriter, r *http.Request) {
 		remaining := int64(expiry.Sub(now).Seconds())
 		if remaining < maxAgeSeconds {
 			actualMaxAge = remaining
-			expiresTime = expiry
 		}
 	} else {
 		actualMaxAge = 10
-		expiresTime = now.Add(10 * time.Second)
 	}
-	w.Header().Set("Cache-Control", fmt.Sprintf("must-revalidate, max-age=%d, immutable", actualMaxAge))
-	w.Header().Set("Expires", expiresTime.Format(time.RFC1123))
+	w.Header().Set("Cache-Control", fmt.Sprintf("must-revalidate, max-age=%d", actualMaxAge))
 	render.JSON(w, r, ks)
 }
--- a/internal/jimmhttp/wellknown_handler_test.go
+++ b/internal/jimmhttp/wellknown_handler_test.go
@@ -143,8 +143,7 @@ func TestWellknownAPIJWKSJSONHandles200(t *testing.T) {
 	c.Assert(err, qt.IsNil)
 	c.Assert(b, qt.JSONEquals, jwks)
 	c.Assert(code, qt.Equals, http.StatusOK)
-	c.Assert(resp.Header.Get("Cache-Control"), qt.Equals, "must-revalidate, max-age=600, immutable")
-	c.Assert(resp.Header.Get("Expires"), qt.Matches, now.Add(10*time.Minute).Format(time.RFC1123))
+	c.Assert(resp.Header.Get("Cache-Control"), qt.Equals, "must-revalidate, max-age=600")
 
 	// Test with expiry < 5min (should use actual expiry)
 	expiryShort := now.Add(3 * time.Minute)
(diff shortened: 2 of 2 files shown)


luci1900 (approved):
I hadn't noticed this uses `immutable`. That's for the case where you're returning something like a content-addressed file. Typically only CDNs use it.

SimoneDutto (approved):
How are we testing the cache-control header is doing what we think it does?

Also, why don't we just remove this? :D I think it's giving us more issues than the actual performance gain we get from it

kian99:
> How are we testing the cache-control header is doing what we think it does?

I don't think we really are or really can, it's just an HTTP standard so we set them and expect clients to follow the standard.

> Also, why don't we just remove this? :D I think it's giving us more issues than the actual performance gain we get from it

I believe it's a necessary optimization, we need to tell the Juju controller that it can use this public key to verify tokens for N minutes/hours before refreshing it, otherwise requesting the public key everytime we connect to a Juju controller could become a deluge of requests to JIMM to fetch the public key.

internal/jimmhttp/wellknown_handler.go
```diff
@@ -94,21 +94,17 @@ func (wkh *WellKnownHandler) JWKS(w http.ResponseWriter, r *http.Request) {
...
 			actualMaxAge = remaining
-			expiresTime = expiry
 		}
 	} else {
 		actualMaxAge = 10
-		expiresTime = now.Add(10 * time.Second)
 	}
-	w.Header().Set("Cache-Control", fmt.Sprintf("must-revalidate, max-age=%d, immutable", actualMaxAge))
-	w.Header().Set("Expires", expiresTime.Format(time.RFC1123))
+	w.Header().Set("Cache-Control", fmt.Sprintf("must-revalidate, max-age=%d", actualMaxAge))
```
ale8k, line 108: Shouldn't we check max age != 0 and if it is 0, set no-cache?
```
But now that HTTP/1.1-conformant servers are widely deployed, there's no reason to ever use that max-age=0 and must-revalidate combination — you should instead just use no-cache.
```
From MSDN doc.

ale8k: Does the lib we're using honour must-revalidate? Presumably the lib will just try the call and if it fails update it's own internal JWKS cache?

ale8k (approved):
LGTM

On the immutable,

I also read `The immutable response directive indicates that the response will not be updated while it's [fresh](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Caching#fresh_and_stale_based_on_age).`

Perhaps I misunderstood it, but I understood it as there are other factors that may make it refresh (like a hard browser reload or whatever [obviously not the case here]) and I was worried there may be things I was unaware of that may cause it to refresh despite being fresh, that's the only reason I added immutable if that helps understand why it was there.

Two minor comments, up to you if you want to do them, approving regardless!

internal/jimmhttp/wellknown_handler.go
kian99: > Shouldn't we check max age != 0 and if it is 0, set no-cache?

If the JWKS has already expired but we haven't refreshed it yet for whatever reason (e.g. a bug), we're setting the age to 10s, that way the client will refresh soon but shouldn't spam us.

> Does the lib we're using honour must-revalidate? Presumably the lib will just try the call and if it fails update it's own internal JWKS cache?

I don't know the answer to this one.

ale8k: I just meant to be defensive in the code we should set no-cache instead (like the doc suggests)

And I only ask because if it isn't necessary, perhaps we can just drop it. I wonder what it is "revalidating", especially in the context of a JWKS.

kian99: > there's no reason to ever use that max-age=0 and must-revalidate combination — you should instead just use no-cache.

Ah gotcha so if max-age=0 change it to no-cache. I wouldn't change the code if it doesn't actually do anything as I think it reads simpler as is.

PR #1735, opened 2025-11-19
ci: free up space in release-binaries job

kian99:
## Description
The release-binaries job is failing [consistently](https://github.com/canonical/jimm/actions/runs/19494919491/job/55796610484) with an issue that seems to be related to disk space because logs abruptly stop, which I've normally seen due to low disk space. It's also happening during compilation/download for a `go build` another indicator this is likely due to disk space.

We can either, 
- Free up space on the Github hosted runner.
- Switch to a self-hosted runner with more disk.
- Avoid building for multiple architectures.

I've opted to go with the 1st option but would also be happy with the second.

### DIFF
--- a/.github/workflows/release-binaries.yaml
+++ b/.github/workflows/release-binaries.yaml
@@ -11,6 +11,23 @@ jobs:
   goreleaser:
     runs-on: ubuntu-22.04
     steps:
+      - name: Free Disk Space (Ubuntu)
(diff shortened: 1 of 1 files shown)


SimoneDutto (approved):

luci1900 (approved):
I don't think any of the options are ideal, they can't guarantee it won't happen again. This should do for now.

PR #1738, opened 2025-11-24
feat: add can_addmodel permission to controllers

kian99:
## Description
This PR adds a new permission to our authorization model. Specifically, `can_addmodel` on the controller type, this fields dictates whether a user has been granted access to create models on a specific controller, much like the `can_addmodel` permission on clouds. 

This is the first step in a series of upcoming changes to allow clients to select the controller where their model will be created in JAAS.

Resolves [JUJU-8813](https://warthogs.atlassian.net/browse/JUJU-8813)

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

## Test instructions
I've updated the openfga auth model tests.

[JUJU-8813]: https://warthogs.atlassian.net/browse/JUJU-8813?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/openfga/user.go
+++ b/internal/openfga/user.go
@@ -34,9 +34,19 @@ type User struct {
 	JimmAdmin bool
 }
 
-// IsAllowedAddModed returns true if the user is allowed to add a model on the
+// IsAllowedAddModelToController returns true if the user is allowed to add a model on the
+// specified controller.
+func (u *User) IsAllowedAddModelToController(ctx context.Context, resource names.ControllerTag) (bool, error) {
+	allowed, err := checkRelation(ctx, u, resource, ofganames.CanAddModelRelation)
+	if err != nil {
+		return false, errors.E(err)
+	}
+	return allowed, nil
+}
+
+// IsAllowedAddModelToCloud returns true if the user is allowed to add a model on the
 // specified cloud.
-func (u *User) IsAllowedAddModel(ctx context.Context, resource names.CloudTag) (bool, error) {
+func (u *User) IsAllowedAddModelToCloud(ctx context.Context, resource names.CloudTag) (bool, error) {
 	allowed, err := checkRelation(ctx, u, resource, ofganames.CanAddModelRelation)
 	if err != nil {
 		return false, errors.E(err)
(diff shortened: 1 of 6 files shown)


luci1900 (approved):

internal/openfga/user_test.go
```diff
@@ -347,6 +347,45 @@ func (s *userTestSuite) TestControllerAccess(c *gc.C) {
...
+	c.Assert(err, gc.IsNil)
+	aliceIdentity, err := dbmodel.NewIdentity(alice.Id())
+	c.Assert(err, gc.IsNil)
+
+	eveUser := openfga.NewUser(eveIdentity, s.ofgaClient)
+	aliceUser := openfga.NewUser(aliceIdentity, s.ofgaClient)
+
+	canAddModel, err := eveUser.IsAllowedAddModelToController(ctx, controller)
+	c.Assert(err, gc.IsNil)
+	c.Assert(canAddModel, gc.Equals, true)
```
alesstimec, line 396: this check asserts that controller administrator has permission to add a model to the controller. can we also please have check to assert that direct relations grant you the same? and via a group, too

alesstimec (approved):

internal/openfga/user.go
```diff
@@ -34,9 +34,19 @@ type User struct {
 	JimmAdmin bool
 }
 
-// IsAllowedAddModed returns true if the user is allowed to add a model on the
+// IsAllowedAddModelToController returns true if the user is allowed to add a model on the
```
ale8k, line 37: Can we have the setter on the user please that creates the tuple? We have this for other things.

ale8k (approved):
LGTM

Would like the setter too though please. Like SetUserCanAddModelToController or w/e you call it lol.

internal/openfga/user.go
kian99: Went to add it, seems there is already `SetControllerAccess` where you can pass in the desired relation. If we need a more specific one I'll add it in a future PR.

PR #1739, opened 2025-11-24
chore: remove opcode from errors

kian99:
## Description
This PR removes opcodes from the errors package. The opcodes were meant to supplement error info to indicate which function/method generated the error. This can be done better with tracing and stacktraces and since the opcode was not currently being used, I have dropped it.

In places where we were using opcodes as a monitoring label, I have converted the constants to strings.

I didn't use AI for this btw, just good old find and replace.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests


luci1900 (approved):
Should just need the database bits replaced with strings or similar.

alesstimec (approved):

PR #1740, opened 2025-11-24
feat: PrepadeUpgradeTo call

ale8k:
## Description
This PR contains the PrepareUpgradeTo call. I've decided to split up the work of Prepare, MigrateAndUpgrade and the plugging of the facade in to make it more easily reviewable. (So please ignore the branch name aha.)

The next PRs will be the migrate and upgrade and finally the facade.

You'll notice there's a TODO in this PR, around the call of ControllerModelSummary - we don't actually need to do this call. We can instead import controller models. I'll do this change last after everything else is done and remove this.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests


internal/jimm/upgrade/mocks/api.go
kian99, line 1: Just a minor one before I do a full review, but what do you think about defining a smaller API interface that only contains the methods this manager needs?

ale8k: For the return type of Dial?

internal/jimm/upgrade/types.go
```diff
@@ -19,3 +26,25 @@ type UpgradeParams struct {
...
+// UpgradeManagerAPI defines the subset of the Juju API used by the UpgradeManager.
+type UpgradeManagerAPI interface {
+	// API implements the base.APICallCloser so that we can
+	// use the juju api clients to interact with juju controllers.
+	base.APICallCloser
+
+	// ControllerModelSummary fetches the model summary of the model on the
+	// controller that hosts the controller machines.
+	ControllerModelSummary(context.Context, *jujuparams.ModelSummary) error
+}
```
SimoneDutto: i think we tend to use `types.go` for actual struct we need to export to other pkgs, interface definitions would be in `upgrade.go`.

It's not a big deal, but i think it's better to have the interface close to the struct

internal/jimm/upgrade/upgrade.go
```diff
@@ -23,26 +29,126 @@ type BootstrapManager interface {
...
+	dialer Dialer,
 ) (*upgradeManager, error) {
 	if bootstrapManager == nil {
 		return nil, errors.E("bootstrap manager cannot be nil")
 	}
 	return &upgradeManager{
 		bootstrapManager: bootstrapManager,
+		jujuManager:      jujumanager,
+		store:            store,
+		dialer:           dialer,
```
SimoneDutto, line 66: when we'll plug in the NewUpgraderManager in the internal/jimm/jimm.go where do we get the `Dialer` from? Do we have an implementation somewhere in a pkg?

internal/jimm/upgrade/upgrade.go
```diff
@@ -23,26 +29,126 @@ type BootstrapManager interface {
...
+		return bootstrapCloud, bootstrapCredential, errors.E(errors.CodeBadRequest, "target version must be greater than current version")
+	}
+
+	api, err := j.dialer.Dial(ctx, &m.Controller, names.ModelTag{}, nil, nil)
+	if err != nil {
+		return bootstrapCloud, bootstrapCredential, errors.E(fmt.Errorf("failed to dial the controller: %w", err))
+	}
+
+	// TODO: When adding controller, import controller models, then we can find the controller model
+	// based on it's UUID via ModelInfo and not iterate through model summaries checking IsController.
```
SimoneDutto: is this a real TODO we'd do at some point or a wish for the future?

And, this todo is about the implementation so should be where the implementation is

internal/jimm/upgrade/upgrade.go
```diff
@@ -23,26 +29,126 @@ type BootstrapManager interface {
...
+	var ctrlModelSummary jujuparams.ModelSummary
+	if err := api.ControllerModelSummary(ctx, &ctrlModelSummary); err != nil {
+		return bootstrapCloud, bootstrapCredential, errors.E(fmt.Errorf("failed to get controller model summary: %w", err))
+	}
+
+	cloudClient := cloud.NewClient(api)
```
SimoneDutto: is this normal? We usually deal with raw facade methods, we don't use the client in the Juju pkg

SimoneDutto (commented):
some comments

internal/jimm/upgrade/types.go
kian99: If you want to match the Juju manager, I think that puts this stuff inside of an `interface.go`.

internal/jimm/upgrade/upgrade.go
```diff
@@ -23,26 +29,126 @@ type BootstrapManager interface {
 	StartBootstrapJob(ctx context.Context, user *openfga.User, params bootstrap.BootstrapParams) (string, error)
 }
 
+// JujuManager defines the juju manager methods required by the upgrade manager.
+type JujuManager interface {
```
kian99, line 33: @SimoneDutto I think you were suggesting that we don't put this upgrade manager inside the Juju manager. But if we're using this logic from the Juju manager, I'm inclined to say it should just go in there. 
Wdyt? I'm not sure what we gain from making a separate manager here except for more layers of abstraction.

internal/jimm/upgrade/upgrade.go
kian99: We have a top-level dialer in `jujuclient` I think or somewhere else.

internal/jimm/upgrade/upgrade.go
kian99: Is the comment up to date? I see we're calling `ControllerModelSummary`, no loops? Unless it's looping inside the method?

internal/jimm/upgrade/upgrade.go
kian99: Yeah this is also surprising to me, shouldn't our client already have the methods you need? I.e. just add `CredentialContents` to the `UpgradeManagerAPI` interface.

internal/jimm/upgrade/types.go
ale8k: Moved them over to upgrade.go, if you guys want an interface.go lmk and I'll move again

internal/jimm/upgrade/upgrade.go
ale8k: Yeah it's top level and we'll cast it into this minimal interface like so:

```go
upgradeAPI := j.Dialer.(upgrade.Dialer)
upgrade.NewUpgradeManager2(nil, nil, nil, upgradeAPI)
```

internal/jimm/upgrade/upgrade.go
ale8k: No no comment is slightly out of date. I'll leave the comment in though. Yeah @SimoneDutto I plan on doing it, there's a few other times in the past (I can't remember from the top of my head) where we needed controller model details but didn't have them, so just to slim off a call and get some basic details I think we should just track controller models.

ale8k: Actually I'll remove the TODO and put it in my notes.

internal/jimm/upgrade/upgrade.go
ale8k: No it's not normal, I was being a little lazy here and thinking "It's only Phase 1", but fair points guys I'll add CredentialContents and call our API.

internal/jimm/upgrade/upgrade.go
kian99: I'm so confused, why would we need to do `upgradeAPI := j.Dialer.(upgrade.Dialer)`? Why wouldn't `j.Dialer` work?

ale8k: Ignore me I'm being a dope it'll implicitly type cast it lol.

internal/jimm/upgrade/upgrade.go
```diff
@@ -23,26 +29,151 @@ type BootstrapManager interface {
...
+	// Cloud procedure on the Cloud facade.
+	Cloud(tag names.CloudTag, cloud *jujucloud.Cloud) error
+}
+
 // upgradeManager provides a means to manage controller upgrades within JIMM.
 type upgradeManager struct {
 	bootstrapManager BootstrapManager
+	jujuManager      JujuManager
+	store            Store
+	dialer           Dialer
```
SimoneDutto: i'm struggling to see the reasons behind using half thing from Juju manager and half from the Dial + Juju client.

I would put methods in the new JujuManager or using the dialer for everything. Half here and half there isn't more confusing to read?

SimoneDutto: discussing it offline, it's a thing we need to discuss and for now it's fine!

SimoneDutto (approved):

internal/jimm/upgrade/upgrade.go
ale8k: We discussed it, all good.

kian99 (approved):
Approved, we'll switch back to juju.API for the dialer and in a follow-up, I'll make some adjustments

PR #1741, opened 2025-11-24
fix(get-identity): concurrency safe implementation of GetIdentity

alesstimec:
## Description

Fixes the GetIdentity db method.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

### DIFF
--- a/internal/db/identity.go
+++ b/internal/db/identity.go
@@ -5,6 +5,8 @@ package db
 import (
 	"context"
 
+	"gorm.io/gorm/clause"
+
 	"github.com/canonical/jimm/v3/internal/dbmodel"
 	"github.com/canonical/jimm/v3/internal/errors"
 	"github.com/canonical/jimm/v3/internal/logger"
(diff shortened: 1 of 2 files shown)


luci1900 (approved):
This is doable in a single query, but I think it's a little tricky with gorm and it shouldn't matter either way.

Interesting test.

ale8k (approved):
lgtm

internal/db/identity.go
```diff
@@ -36,14 +38,24 @@ func (d *Database) GetIdentity(ctx context.Context, u *dbmodel.Identity) (err er
...
 	}
-	if result.RowsAffected == 1 {
+
+	// Check if a new identity was created.
+	if result.RowsAffected > 0 {
 		logger.LogUserCreated(ctx, u.Name)
+		return nil
 	}
+
+	// If we didn't create it, it must exist (or we raced and lost). Fetch it.
```
SimoneDutto, line 54: Why do we need this?
Create doesn't fill the fields when the `ON CONFLICT` triggers?

SimoneDutto (commented):
1 comment

internal/db/identity.go
luci1900: It only triggers when the row had already been inserted, so whoever won the race has already filled the values.

SimoneDutto: and the other is empty so you need to re-fetch it?

ale8k: Yup

SimoneDutto (approved):

PR #1742, opened 2025-11-25
feat(upgrade-manager): create UpgradeManager

SimoneDutto:
# Description

(for now) The UpgradeManager is in charge of starting and waiting for a bootstrap job to complete.

I've taken 3 decisions worth of noticing:
- put the `WaitForJobCompletion` method in BootstapManager, because i would like to keep the logic to poll job status, etc inside the BootstrapManager.
- `WaitForJobCompletion` is not collecting logs for now.
- create the UpgradeParams in `internal/jimm/upgrade` which is just a c/p from bootstrap params right now, i guess it will expand in the future and i didn't want just to embed the type from another pkg.


internal/jimm/upgrade/upgrade.go
```diff
@@ -0,0 +1,76 @@
...
+		return nil, errors.E("bootstrap manager cannot be nil")
+	}
+	return &upgradeManager{
+		bootstrapManager: bootstrapManager,
+	}, nil
+}
+
+// UpgradeTo upgrades a controller by fetching its configuration and initiating
+// a bootstrap job with that configuration, then waits for the bootstrap to complete.
+func (u *upgradeManager) UpgradeTo(ctx context.Context, user *openfga.User, params UpgradeParams) error {
```
ale8k, line 45: Purrrrfecto, I'll plug this https://github.com/canonical/jimm/pull/1740 in once this merges here.

luci1900 (approved):

kian99 (approved):

internal/jimm/bootstrap/bootstrap.go
```diff
@@ -186,6 +187,36 @@ func (b *bootstrapManager) StopJob(ctx context.Context, user *openfga.User, jobI
 	return nil
 }
 
+// WaitForJobCompletion waits for a bootstrap job to complete by polling its status.
+// It returns an error if the user is nil, if the job ID is nil, or if the job fails.
+// It returns nil on successful completion.
+func (b *bootstrapManager) WaitForJobCompletion(ctx context.Context, user *openfga.User, jobId uuid.UUID) error {
```
ale8k: User isn't used?

internal/jimm/bootstrap/bootstrap.go
```diff
@@ -186,6 +187,36 @@ func (b *bootstrapManager) StopJob(ctx context.Context, user *openfga.User, jobI
...
+	const op = errors.Op("jimm.WaitForJobCompletion")
+	if user == nil {
+		return errors.E(op, "user cannot be nil")
+	}
+	if jobId == uuid.Nil {
+		return errors.E(op, "job ID cannot be nil")
+	}
+
+	for {
```
ale8k, line 210: Can we use a ticker instead please and not time.Sleep

internal/jimm/bootstrap/bootstrap.go
```diff
@@ -41,6 +41,7 @@ const (
 	bootstrapJobType         = "bootstrap"
 	destroyControllerJobType = "destroy-controller"
 	maxJobDuration           = 60 * time.Minute
+	defaultPollingInterval   = 5 * time.Second
```
ale8k, line 44: Let the caller pass this in and construct a ticker? Then the caller can just cancel context to end it, this will speed up tests too.

internal/jimm/upgrade/upgrade.go
```diff
@@ -0,0 +1,76 @@
+// Copyright 2025 Canonical.
+
+// upgrade package provides functionality to manage the upgrade process
+// for controllers in JIMM.
+package upgrade
```
ale8k, line 5: automatedupgrade would be a better name, just in case we have any "upgrade" only logic later on and renaming is annoying.

internal/jimm/upgrade/types.go
```diff
@@ -0,0 +1,21 @@
+// Copyright 2025 Canonical.
+
+package upgrade
+
+import (
+	jujucloud "github.com/juju/juju/cloud"
+)
+
+// UpgradeParams defines the parameters required for bootstrapping a Juju controller.
```
ale8k, line 9: Defines params for upgrading a controller, by means of bootstrapping a new controller and migrating the models.

ale8k (changes requested):
Some minor comments

internal/jimm/bootstrap/bootstrap.go
SimoneDutto: done

internal/jimm/bootstrap/bootstrap.go
SimoneDutto: done

internal/jimm/upgrade/types.go
SimoneDutto: i don't understand this comment

internal/jimm/upgrade/upgrade.go
SimoneDutto: i think this is fine, in case we'll change it

docker-compose.common.yaml
```diff
@@ -89,6 +89,7 @@ services:
         -----END OPENSSH PRIVATE KEY-----
       # To Q/A this, use any reachable address for this compose JIMM. 
       JIMM_BOOTSTRAP_LOGIN_TOKEN_REFRESH_URL:  "${JIMM_BOOTSTRAP_LOGIN_TOKEN_REFRESH_URL:-https://jimm.localhost/.well-known/jwks.json}"
+      GOTOOLCHAIN: auto
```
SimoneDutto, line 92: this is to avoid clash between go version and air version, it's fly-by and i can remove it

internal/jimm/upgrade/types.go
ale8k: Just to reword it but it's ok it'll probably change anyway as it grows.

ale8k (approved):

PR #1743, opened 2025-11-25
chore: upgrade Juju dependency to 3.6.12

kian99:
## Description
This PR updates our dependency on Juju to the latest release 3.6.12.

There were 2 breaking changes that required fixing,
- The interface definition for `base.MacaroonDischarger` changed. This is used in our implementation of a Juju client and was changed by myself in a PR fixing a bug with `juju login`. It required a small wrapper to expose some extra functionality.
- The providers in `github.com/juju/juju/provider/` were moved into `internal/` we relied on these providers to generate a mapping of attribute names that are sensitive which is then used in the implemention of GetCloudCredentials to decide which attributes of a cloud-credential should be hidden by default. The existing mapping is kept but no new mappings can be generated with the current approach, so I've removed it. I've created [JUJU-8837](https://warthogs.atlassian.net/browse/JUJU-8837) to address this separately.

Additionally, Juju 3.6.12 is using Go 1.25 so we have upgraded as well. A side-effect of that is that golangci-lint needs an update to [at least](https://github.com/golangci/golangci-lint/issues/5873) v2.4.0 which is an upgrade from the v1 release we were using. I've used their CLI tool to migrate our config file to their new version and fixed the new linter issues.

I suggest reviewing this PR commit-by-commit to see the changes more easily.

[JUJU-8837]: https://warthogs.atlassian.net/browse/JUJU-8837?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cmd/jaas/cmd/group.go
```diff
@@ -268,7 +268,7 @@ func (c *removeGroupCommand) Run(ctxt *cmd.Context) error {
 			return errors.E(err, "Failed to read from input.")
 		}
 		text = strings.ReplaceAll(text, "\n", "")
-		if !(text == "y" || text == "Y") {
+		if text != "y" && text != "Y" {
```
luci1900, line 271: yay

.golangci.yaml
```diff
@@ -1,90 +1,78 @@
...
-  sort-results: true
-  formats:
-    - format: github-actions
-
 linters:
-  disable-all: true
+  default: none
   enable:
-    # The following linters are enabled by default
+    - bodyclose
```
luci1900, line 10: Most of this seems just re-ordered?

kian99: Yeah there are some other changes like [this one](https://golangci-lint.run/docs/product/migration-guide/#lintersenablestylecheckgosimplestaticcheck)
> The linters stylecheck, gosimple, and staticcheck has been merged inside the staticcheck.

internal/jimmhttp/rebac_admin/identities.go
```diff
@@ -171,9 +171,10 @@ func (s *identitiesService) PatchIdentityRoles(ctx context.Context, identityId s
 			Relation:     ofganames.AssigneeRelation.String(),
 			TargetObject: jimmnames.NewRoleTag(p.Role).String(),
 		}
-		if p.Op == resources.IdentityRolesPatchItemOpAdd {
+		switch p.Op {
+		case resources.IdentityRolesPatchItemOpAdd:
 			additions = append(additions, t)
-		} else if p.Op == "remove" {
+		case "remove":
```
luci1900, line 177: Huh. Odd that one is a name and the other a literal.

luci1900 (approved):
Looks fine.

internal/jimmhttp/rebac_admin/identities.go
kian99: Yeah that is weird

SimoneDutto (approved):
lgtm

PR #1744, opened 2025-11-25
feat(upgrade-to-cmd): add `juju upgrade-to <model-uuid> <version>`

SimoneDutto:
## Description

For now it calls the jimm facade that returns some simple `Logs` to print to stdout or an error.

Some changes from the spec which i think make more sense:
- no <model-uuid | model name>, i don't like the idea of having commands with mixed input and do custom logic to resolve one into the other. Users can just put the model uuid, also because model-name is not enough, expecially in JAAS, you need `owner/model-name`
<s> - version is not another arg but it's an optional flag. This is to conform with other upgrade command 
Ex. juju upgrade-controller  
```
You can upgrade the controller to a new patch version by specifying
the `--agent-version` flag. If not specified, the upgrade candidate
will default to the most recent patch version matching the current
major and minor version. Upgrading to a new major or minor version is
not supported.

Examples:

    juju upgrade-controller --dry-run
    juju upgrade-controller --agent-version 2.0.1
```
</s>

### DIFF
--- a/cmd/jaas/cmd/upgradeto.go
+++ b/cmd/jaas/cmd/upgradeto.go
@@ -1,0 +1,117 @@
+// Copyright 2025 Canonical.
+
+package cmd
+
+import (
+	"github.com/juju/cmd/v3"
+	"github.com/juju/gnuflag"
+	jujuapi "github.com/juju/juju/api"
+	jujucmd "github.com/juju/juju/cmd"
+	"github.com/juju/juju/cmd/modelcmd"
+	"github.com/juju/juju/jujuclient"
+
+	"github.com/canonical/jimm/v3/internal/errors"
+	"github.com/canonical/jimm/v3/pkg/api"
+	apiparams "github.com/canonical/jimm/v3/pkg/api/params"
+)
+
+const (
+	upgradeToDoc = `
+Upgrades a controller to a specified version.
+`
+	upgradeToExample = `
+    juju upgrade-to 2cb433a6-04eb-4ec4-9567-90426d20a004 3.6.11
+`
+)
+
+// NewUpgradeToCommand returns a command to upgrade a controller to a specified version.
+func NewUpgradeToCommand() cmd.Command {
+	cmd := &upgradeToCommand{
+		store: jujuclient.NewFileClientStore(),
+	}
+	cmd.jimmAPIFunc = cmd.newClient
+
+	return modelcmd.WrapBase(cmd)
+}
+
+// upgradeToCommand upgrades a controller to a specified version.
+type upgradeToCommand struct {
+	modelcmd.ControllerCommandBase
+	out cmd.Output
+
+	store       jujuclient.ClientStore
+	dialOpts    *jujuapi.DialOpts
+	version     string
+	modelUUID   string
+	jimmAPIFunc func() (JIMMAPI, error)
+}
+
+func (c *upgradeToCommand) Info() *cmd.Info {
+	return jujucmd.Info(&cmd.Info{
+		Name:     "upgrade-to",
+		Args:     "<model-uuid> <version>",
+		Purpose:  "Upgrades a controller to a specified version",
+		Doc:      upgradeToDoc,
+		Examples: upgradeToExample,
+	})
+}
+
+// SetFlags implements Command.SetFlags.
+func (c *upgradeToCommand) SetFlags(f *gnuflag.FlagSet) {
+	c.CommandBase.SetFlags(f)
+	c.out.AddFlags(f, "yaml", map[string]cmd.Formatter{
+		"yaml": cmd.FormatYaml,
+		"json": cmd.FormatJson,
+	})
+}
+
+// Init implements the cmd.Command interface.
+func (c *upgradeToCommand) Init(args []string) error {
+	if len(args) < 2 {
+		return errors.E("missing required arguments: model UUID and version")
+	}
+	if len(args) > 2 {
+		return errors.E("too many arguments")
+	}
+	c.modelUUID = args[0]
+	c.version = args[1]
+	return nil
+}
+
+// Run implements Command.Run.
+func (c *upgradeToCommand) Run(ctxt *cmd.Context) error {
+	client, err := c.jimmAPIFunc()
+	if err != nil {
+		return errors.E(err, "failed to create JIMM client")
+	}
+	defer client.Close()
+
+	resp, err := client.UpgradeTo(&apiparams.UpgradeToRequest{
+		TargetVersion: c.version,
+		ModelUUID:     c.modelUUID,
+	})
+	if err != nil {
+		return errors.E(err)
+	}
+
+	err = c.out.Write(ctxt, resp)
+	if err != nil {
+		return errors.E(err)
+	}
+	return nil
+}
+
+// newClient creates a new JIMM API client.
+func (c *upgradeToCommand) newClient() (JIMMAPI, error) {
+	currentController, err := c.store.CurrentController()
+	if err != nil {
+		return nil, errors.E(err, "could not determine controller")
+	}
+
+	apiCaller, err := c.NewAPIRootWithDialOpts(c.store, currentController, "", c.dialOpts)
+	if err != nil {
+		return nil, err
+	}
+
+	return api.NewClient(apiCaller), nil
+}
--- a/cmd/jaas/cmd/upgradeto.go
+++ b/cmd/jaas/cmd/upgradeto.go
@@ -9,6 +9,8 @@ import (
 	jujucmd "github.com/juju/juju/cmd"
 	"github.com/juju/juju/cmd/modelcmd"
 	"github.com/juju/juju/jujuclient"
+	"github.com/juju/names/v5"
+	jujuversion "github.com/juju/version/v2"
 
 	"github.com/canonical/jimm/v3/internal/errors"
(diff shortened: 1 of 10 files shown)


cmd/jaas/cmd/upgradeto.go
```diff
@@ -0,0 +1,124 @@
...
+	"github.com/canonical/jimm/v3/pkg/api"
+	apiparams "github.com/canonical/jimm/v3/pkg/api/params"
+)
+
+const (
+	upgradeToDoc = `
+Upgrades a controller to a specified version.
+`
+	upgradeToExample = `
+    juju upgrade-to 2cb433a6-04eb-4ec4-9567-90426d20a004 --version 2.0.1
```
kian99: I think it would be fine to accept a modelOwner/modelName as well, we already do it for some other commands and the server can determine if it's a owner+name or UUID. But happy to just go with whatever we all think is easiest/best.

cmd/jaas/cmd/upgradeto.go
```diff
@@ -0,0 +1,124 @@
...
+    juju upgrade-to mymodel --version 3.5.0
```
kian99: Is this valid? Just the `myModel` part I mean. I think it needs an `owner/` portion. But then the PR description says we won't do names and UUIDs.

cmd/jaas/cmd/upgradeto.go
```diff
@@ -0,0 +1,124 @@
...
+	"github.com/juju/juju/jujuclient"
+
+	"github.com/canonical/jimm/v3/internal/errors"
+	"github.com/canonical/jimm/v3/pkg/api"
+	apiparams "github.com/canonical/jimm/v3/pkg/api/params"
+)
+
+const (
+	upgradeToDoc = `
+Upgrades a controller to a specified version.
```
kian99, line 22: We say "specified version" but the version is not required.
The PR description says this is to conform with `upgrade-controller` but we can't do that here. `upgrade-controller` upgrades to the latest patch version e.g. `3.6.6` -> `3.6.7`.

With this feature you actually have no idea where the user wants to go, is it the next major, the next minor, etc. So the version is required.

kian99 (commented):
Needs some changes

cmd/jaas/cmd/upgradeto.go
SimoneDutto: yeah, should be removed

cmd/jaas/cmd/upgradeto.go
SimoneDutto: yeah, that's what i wanted to discuss. Should we default to latest major always and we let the user specify it everytime?

kian99: I think just letting the user specify it is the most straightforward thing for this feature, at least for now.

luci1900: I agree on requiring the user to always specify.

I don't think that stops it from being a flag necessarily, though. The ergonomics are separate, I think.

cmd/jaas/cmd/upgradeto.go
SimoneDutto: i prefer having just model-uuid

cmd/jaas/cmd/upgradeto.go
SimoneDutto: i've put both required and positional args.

We had a discussion and most of the team think that a flag should be considered optional.

cmd/jaas/cmd/upgradeto.go
```diff
@@ -0,0 +1,117 @@
...
+		"json": cmd.FormatJson,
+	})
+}
+
+// Init implements the cmd.Command interface.
+func (c *upgradeToCommand) Init(args []string) error {
+	if len(args) < 2 {
+		return errors.E("missing required arguments: model UUID and version")
+	}
+	if len(args) > 2 {
```
kian99, line 75: Let's do basic validation of the args here. I'd say the version string at least can be validated.

pkg/api/params/params.go
```diff
@@ -236,6 +236,20 @@ type SetControllerDeprecatedRequest struct {
...
+	// TargetVersion is the target controller version to upgrade to.
+	TargetVersion string `json:"target-version"`
+
+	// ModelUUID is the UUID of the model to upgrade.
+	ModelUUID string `json:"model-uuid"`
+}
+
+// UpgradeToResponse is the response that is sent in an UpgradeTo method.
+type UpgradeToResponse struct {
+	Logs string `json:"logs"`
```
kian99: Why just a string called `Logs`? I'd imagine just an error would be more appropriate.

cmd/jaas/cmd/upgradeto.go
```diff
@@ -0,0 +1,117 @@
...
+	"github.com/canonical/jimm/v3/pkg/api"
+	apiparams "github.com/canonical/jimm/v3/pkg/api/params"
+)
+
+const (
+	upgradeToDoc = `
+Upgrades a controller to a specified version.
+`
+	upgradeToExample = `
+    juju upgrade-to 2cb433a6-04eb-4ec4-9567-90426d20a004 3.6.11
```
ale8k: Version must go first, if you wanted N models (which we will eventually) it's a bit funky to take the version as the last argument.

cmd/jaas/cmd/upgradeto.go
```diff
@@ -0,0 +1,117 @@
...
+	out cmd.Output
+
+	store       jujuclient.ClientStore
+	dialOpts    *jujuapi.DialOpts
+	version     string
+	modelUUID   string
+	jimmAPIFunc func() (JIMMAPI, error)
+}
+
+func (c *upgradeToCommand) Info() *cmd.Info {
```
ale8k, line 51: godoc

internal/errors/errors.go
```diff
@@ -63,8 +63,6 @@ func (e *Error) ErrorInfo() map[string]any {
 // is constructed by processing the given arguments. The meaning of the
 // arguments is as follows:
 //
-//	errors.Op   - string representation of the operation being
```
ale8k, line 66: Lol I removed this in mine too.

pkg/api/params/params.go
```diff
@@ -236,6 +236,20 @@ type SetControllerDeprecatedRequest struct {
 	Deprecated bool `json:"deprecated"`
 }
 
+// UpgradeToRequest is the request that is sent in an UpgradeTo method.
+type UpgradeToRequest struct {
+	// TargetVersion is the target controller version to upgrade to.
+	TargetVersion string `json:"target-version"`
+
+	// ModelUUID is the UUID of the model to upgrade.
+	ModelUUID string `json:"model-uuid"`
```
ale8k: I think these should look like this:
```go
// UpgradeToRequest holds the parameters for phase 1 for automated upgrades.
type UpgradeToRequest struct {
    ModelTag                string `json:"model-tag"`
    TargetControllerVersion string `json:"target-controller-version"`
}

    Success bool   `json:"success"`
    Error   string `json:"error,omitempty"`
}
```

We wouldn't return logs in phase 1?

ale8k (changes requested):
Params need changing then lgtm.

cmd/jaas/cmd/upgradeto.go
ale8k: Idm uuid for phase 1. Can make this better later.

cmd/jaas/cmd/upgradeto.go
SimoneDutto: when we change it to multiple models we would need to change the params and the parsing, and we can even change this one.

pkg/api/params/params.go
SimoneDutto: let's add the logs to make possible to display something in case of success or error. Worst case we don't use it

ale8k: I really don't think we should, like we said just the very simplest thing. JIMM can log for our debugging purposes.

ale8k: It's gonna change anyways, you won't know what exit code to use without knowing if there's an error anyway.

cmd/jaas/cmd/upgradeto.go
ale8k: I'd really appreciate it if we did just put the version first.

kian99: I agree with Alex on this one, seems like an easy one to setup for the future.

kian99 (commented):
Still need to come back and review the rest, just leaving the comments I had for now

pkg/api/params/params.go
SimoneDutto: done

pkg/api/params/params.go
SimoneDutto: changed to alex's suggestion

cmd/jaas/cmd/upgradeto.go
SimoneDutto: done

cmd/jaas/cmd/upgradeto.go
SimoneDutto: done

ale8k (approved):
lgtm, nice!

luci1900 (approved):
Looks fine.

I wish the commands were structured differently, but that's in general not for this one specifically.

PR #1745, opened 2025-11-26
feat: change add model logic to include target-controller 

kian99:
## Description
This PR updates JIMM's add-model logic to incorporate a new parameteter - `TargetController`. Setting this field allows a user to specify which controller will host a model. We also incorporate the new `can-addmodel` permission introduced on controllers.

I recommend reviewing commit-by-commit. 
- The first commit removes some unused functions/methods.
- The second refactors the model builder to be easier to follow, incorporating the new logic mentioned above and ensures the existing add-model tests pass.
- The third adds additional test cases for the new logic.
- The fourth fixes tests where the user didn't have add-model permissions.

The changes to the model builder take the following high-level approach:
1. If the user specified `TargetController`, check they have access to it and use that as the only candidate controller.
2. If the user didn't specify anything, find all controllers they have access to and store these as potential candidates.
3. Filter the candidate controllers by those that support the desired cloud.
4. Filter the candidate controllers by those that support the cloud-region.
5. Shuffle the candidate controllers, incorporating controller<->cloud-region priority (as before) and select the first.

This PR doesn't assume where the target controller field comes from. In https://github.com/juju/juju/pull/21306 it is proposed to place it inside the `ModelCreateArgs` struct, but discussions may lead it to come from the URL. Either way is okay.

Fixes [JUJU-8815](https://warthogs.atlassian.net/browse/JUJU-8815)

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

[JUJU-8815]: https://warthogs.atlassian.net/browse/JUJU-8815?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cmd/jaas/cmd/updatemigratedmodel_test.go
```diff
@@ -25,9 +25,9 @@ var _ = gc.Suite(&updateMigratedModelSuite{})
 func (s *updateMigratedModelSuite) TestUpdateMigratedModelSuperuser(c *gc.C) {
 	s.AddController(c, "controller-1", s.APIInfo(c))
 
-	cct := names.NewCloudCredentialTag(jimmtest.TestCloudName + "/charlie@canonical.com/cred")
+	cct := names.NewCloudCredentialTag(jimmtest.TestCloudName + "/alice@canonical.com/cred")
```
alesstimec, line 28: because alice is jimm admin?

internal/jimm/juju/modelbuilder.go
```diff
@@ -93,6 +114,16 @@ func (b *modelBuilder) WithOwner(owner *dbmodel.Identity) *modelBuilder {
 	return b
 }
 
+// WithAuthorizer returns a builder with the specified OpenFGA user
+// that will be used to authorize the model creation.
+func (b *modelBuilder) WithAuthorizer(ofgaUser *openfga.User) *modelBuilder {
```
alesstimec, line 121: hmm.. not sure about this.. shouldn't this user be the model owner (see WithOwner)?

cmd/jaas/cmd/updatemigratedmodel_test.go
kian99: Yip

internal/jimm/juju/modelbuilder.go
kian99: I think it should be the person making the call, I.e. if Alice is trying to create a model on behalf of someone else, we should check Alice's permissions. It seems like previously we were checking the permissions of the eventual owner.

internal/jimm/juju/modelbuilder.go
```diff
@@ -113,11 +144,76 @@ func (b *modelBuilder) WithConfig(cfg map[string]interface{}) *modelBuilder {
...
+			return nil
+		}
+		ok, err := b.ofgaUser.IsAllowedAddModelToController(b.ctx, c.ResourceTag())
+		if err != nil {
+			return fmt.Errorf("failed to verify permissions for adding model to controller: %v", err)
+		}
+		if ok {
+			candidateControllers = append(candidateControllers, candidateController{
+				controller: *c,
+				priority:   0, // priority is unknown until we select the clou-region
```
luci1900: As in we use the priority the cloud region already has?

kian99: Not sure I understood the question correctly but more like, we can't assign priorities to controllers until we know the cloud + region where the model will be hosted. Because the priority is based on picking a controller that is in the same cloud + region.

internal/jimm/juju/modelbuilder.go
```diff
@@ -175,50 +292,53 @@ func (b *modelBuilder) WithCloudRegion(region string) *modelBuilder {
...
-
-		// and select the first controller in the slice
-		b.cloudRegion = region
-		b.cloudRegionID = r.ID
-		b.cloudRegionVirtual = r.Virtual
-		b.controller = &regionControllers[0].Controller
+	}
 
-		break
+	// If there is no such region we will get a zero valued object below.
```
luci1900, line 336: That's a bit naff. Not for this PR, but would be nice to change the callee, possibly to return an `ok` boolean.

internal/jimm/juju/modelbuilder.go
```diff
@@ -175,50 +292,53 @@ func (b *modelBuilder) WithCloudRegion(region string) *modelBuilder {
...
+
+	// Filter the candidate controllers based on the specified region.
+	// If the region is not specified, we pick the first cloud region
+	// with any associated candidate controller
 	if region == "" {
-		for _, r := range b.cloud.Regions {
-			regionControllers := r.Controllers
-			if len(regionControllers) == 0 {
-				continue
+		// Make a map of supported regions in the cloud from among the candidate controllers.
```
luci1900, line 317: Semantically a set, I guess.

Absolutely a nitpick, but would we consider a set library like https://github.com/deckarep/golang-set?

internal/jimm/juju/modelbuilder.go
```diff
@@ -133,8 +229,29 @@ func (b *modelBuilder) WithCloud(user *openfga.User, cloud names.CloudTag) *mode
...
+	matchCloud := func(r dbmodel.CloudRegionControllerPriority) bool {
+		return r.CloudRegion.CloudName == c.Name
+	}
+	var candidateControllers []candidateController
+	for _, controller := range b.candidates {
+		if slices.ContainsFunc(controller.controller.CloudRegions, matchCloud) {
+			candidateControllers = append(candidateControllers, controller)
+		}
+	}
+	b.candidates = candidateControllers
```
luci1900, line 270: Does this mean the builder methods must be called in a specific order?

internal/jimm/juju/modelbuilder.go
kian99: Yeah agreed.

internal/jimm/juju/modelbuilder.go
kian99: Yip, it was like that previously too.

internal/jimm/juju/modelbuilder.go
```diff
@@ -18,6 +20,23 @@ import (
...
+	// priority decides an ordering for controller selection
+	// based on the cloud-region.
+	priority uint
+}
+
+func shuffleCandidateControllers(controllers []candidateController) {
+	shuffle(len(controllers), func(i, j int) {
+		controllers[i], controllers[j] = controllers[j], controllers[i]
+	})
+	sort.SliceStable(controllers, func(i, j int) bool {
```
luci1900, line 37: It took me a second to figure out this does a stable sort to preserve the previous shuffling. If it's just me, it probably doesn't need explaining.

.github/workflows/integration-test.yaml
```diff
@@ -89,3 +89,7 @@ jobs:
             echo "Graceful shutdown message not found!"
             exit 1
           fi
+
+      - name: Print JIMM logs on failure
+        if: failure()
+        run: docker logs jimm
```
luci1900, line 95: Good idea!

internal/jimm/juju/modelbuilder.go
luci1900: That's unfortunate. I guess potentially for fixing another time.

internal/jimm/juju/modelbuilder.go
luci1900: It wasn't clear to me whether it comes from `CloudRegionControllerPriority` in the database. I've since checked out your branch and that was made clear.

luci1900 (approved):
Looks reasonable.

The split by commit is nice, but I don't know if it's a standard I'm happy to be held to. The bribe of deleted code helps, though.

internal/jimm/juju/modelbuilder.go
kian99: I think Juju actually has one as well https://github.com/juju/collections

luci1900: Ah, but without generics.

internal/jimm/juju/model.go
```diff
@@ -42,20 +42,24 @@ func shuffleRegionControllers(controllers []dbmodel.CloudRegionControllerPriorit
 
 // ModelCreateArgs contains parameters used to add a new model.
 type ModelCreateArgs struct {
```
ale8k, line 44: Can we upgrade the comment to explain target controller is optional and takes a controller name. Maybe even call it TargetControllerName to be super explicit.

internal/jimm/juju/model.go
```diff
@@ -42,20 +42,24 @@ func shuffleRegionControllers(controllers []dbmodel.CloudRegionControllerPriorit
...
+	Config           map[string]interface{}
+	Cloud            names.CloudTag
+	CloudRegion      string
+	CloudCredential  names.CloudCredentialTag
+	TargetController string
 }
 
 // FromJujuModelCreateArgs converts jujuparams.ModelCreateArgs into AddModelArgs.
-func (a *ModelCreateArgs) FromJujuModelCreateArgs(args *jujuparams.ModelCreateArgs) error {
+// TODO(Kian): refactor to move targetController into jujuparams.ModelCreateArgs
```
ale8k: What does this TODO mean? Just move target controller into ModelCreateArgs?

internal/testutils/jimmtest/env.go
```diff
@@ -183,9 +183,26 @@ func (m Model) addModelRelations(c *qt.C, db *db.Database, client *openfga.OFGAC
...
+		var relation openfga.Relation
+		switch u.Access {
+		case "admin":
+			relation = ofganames.AdministratorRelation
+		case "add-model":
+			relation = ofganames.CanAddModelRelation
+		default:
+			c.Fatalf("unknown controller access level: %s", u.Access)
+		}
+		if client != nil {
```
ale8k: We just used the client on line 187, also as it's just for tests it's ok to let it panic.

ale8k: Actually wouldn't a simple lookup map look a bit better here

```go
relation, ok := map[string]openfga.Relation{
    "admin":     ofganames.AdministratorRelation,
    "add-model": ofganames.CanAddModelRelation,
}[u.Access]
if !ok {
    c.Fatalf("unknown controller access level: %s", u.Access)
}
dbUser := ctl.env.User(u.User).DBObject(c, db)
// ... so forth
```

Could also just remove the ok check and let the SetControllerAccess fail with "" or whatever it is.

internal/jujuapi/modelmanager.go
```diff
@@ -186,7 +186,8 @@ func (r *controllerRoot) CreateModel(ctx context.Context, args jujuparams.ModelC
 	defer cancel()
 
 	var mca juju.ModelCreateArgs
-	if err := mca.FromJujuModelCreateArgs(&args); err != nil {
+	var targetController string // This will be populated at a later stage based on user input.
+	if err := mca.FromJujuModelCreateArgs(&args, targetController); err != nil {
```
ale8k: I'm not following this here, why the var? I think "" with the doc is fine?

internal/jimm/juju/modelbuilder.go
```diff
@@ -264,17 +384,10 @@ func (b *modelBuilder) CreateDatabaseModel() *modelBuilder {
...
-	// regions/controllers for the specified cloud
-	if b.cloudRegionID == 0 {
-		// if selectCloudRegion returns an error that means we have
-		// no regions/controllers for the specified cloud - we
-		// error and abort
-		if err := b.selectCloudRegion(); err != nil {
-			b.err = errors.E(err)
-			return b
-		}
+	// make a final controller selection
```
ale8k: Don't need this doc, just godoc selectController

internal/jimm/juju/model.go
```diff
@@ -108,11 +112,17 @@ func (j *JujuManager) AddModel(ctx context.Context, user *openfga.User, args *Mo
 
 	builder := newModelBuilder(ctx, j)
 	builder = builder.WithOwner(owner)
+	builder = builder.WithAuthorizer(user)
 	builder = builder.WithName(args.Name)
 	if err := builder.Error(); err != nil {
 		return nil, errors.E(err)
 	}
 
+	builder = builder.WithController(args.TargetController)
```
ale8k: This reads kinda funny.

We're doing the if controller set inside, but I would've expected something like uhm:
```go
if args.TargetController != "" {
    builder.WithTargetController(args.TargetController)
} else {
    builder.WithHighedPriorityRandomController() 
}
```
Then we can godoc each path super clearly and it reads nicely.

Could we do something like this?

internal/jimm/juju/modelbuilder.go
```diff
@@ -113,11 +144,81 @@ func (b *modelBuilder) WithConfig(cfg map[string]interface{}) *modelBuilder {
...
+// An empty controllerName means no specific controller is targeted.
+func (b *modelBuilder) WithController(controllerName string) *modelBuilder {
+	if b.err != nil {
+		return b
+	}
+	if b.ofgaUser == nil {
+		b.err = errors.E("authorizer not specified")
+		return b
+	}
+	if controllerName != "" {
```
ale8k: Just to fight for my other comment (I know it's a bit of a hassle aha), you can see there's two distinct paths here that could be split and nicely documented.

internal/jimm/juju/modelbuilder.go
```diff
@@ -113,11 +144,81 @@ func (b *modelBuilder) WithConfig(cfg map[string]interface{}) *modelBuilder {
...
+			return nil
+		}
+		ok, err := b.ofgaUser.IsAllowedAddModelToController(b.ctx, c.ResourceTag())
+		if err != nil {
+			return fmt.Errorf("failed to verify permissions for adding model to controller: %v", err)
+		}
+		if ok {
+			candidateControllers = append(candidateControllers, candidateController{
+				controller: *c,
+				priority:   0, // priority is unknown until we select the clou-region
```
ale8k: clou-region ;)

internal/jimm/juju/modelbuilder.go
```diff
@@ -133,8 +234,29 @@ func (b *modelBuilder) WithCloud(user *openfga.User, cloud names.CloudTag) *mode
 		b.err = err
 		return b
 	}
+	ok, err := b.ofgaUser.IsAllowedAddModelToCloud(b.ctx, c.ResourceTag())
```
ale8k, line 249: I'm really starting to dislike the "builder", it isn't really a builder but most just a sequence of steps and having it able to fail on an authcheck in a func called "WithCloud" isn't clear when reading the builder. 

It's also just not ideal - we've run the WithController logic, but then realised after getting a set of candidate controllers they aren't authorised to the cloud?

I think we need a card to come back to this later and just rewrite it a little so it's more semantic. And probably remove the "builder".

internal/jimm/juju/modelbuilder.go
```diff
@@ -113,11 +144,81 @@ func (b *modelBuilder) WithConfig(cfg map[string]interface{}) *modelBuilder {
...
+	b.candidates = candidateControllers
+	// Error early with a helpful message if no
+	// candidate controllers are available.
+	if len(b.candidates) == 0 {
+		b.err = errors.E("no available controllers - check permissions to controllers and list of available controllers")
+	}
+	return b
+}
+
 // WithCloud returns a builder with the specified cloud.
```
ale8k, line 223: For now can we update the godoc to explain it's really "FilterControllersByCloud"

Because it should read that we're essentially slimming down a set of candidates, and from the final selection randomly picking and right now it doesn't read that way tbh.

internal/jimm/juju/modelbuilder.go
```diff
@@ -175,50 +297,53 @@ func (b *modelBuilder) WithCloudRegion(region string) *modelBuilder {
 		b.err = errors.E("cloud not specified")
 		return b
 	}
-	// if the region is not specified, we pick the first cloud region
-	// with any associated controllers
+
+	// Filter the candidate controllers based on the specified region.
+	// If the region is not specified, we pick the first cloud region
+	// with any associated candidate controller
```
ale8k: This doc can live on the godoc,

internal/jimm/juju/modelbuilder.go
```diff
@@ -359,29 +476,20 @@ func (b *modelBuilder) UpdateDatabaseModel() *modelBuilder {
...
+		return fmt.Errorf("unsupported cloud region %s/%s - confirm access to the cloud/controller", b.cloud.Name, b.cloudRegion)
 	}
 
 	// shuffle controllers according to their priority
-	shuffleRegionControllers(regionControllers)
+	shuffleCandidateControllers(b.candidates)
 
-	b.cloudRegionID = regionControllers[0].CloudRegionID
-	b.controller = &regionControllers[0].Controller
+	b.controller = &b.candidates[0].controller
```
ale8k, line 504: This doesn't need to be b.controller, can just return, it seems b.controller is only used in CreateDbModel?

internal/jimm/juju/modelbuilder.go
```diff
@@ -18,6 +20,23 @@ import (
...
 
+type candidateController struct {
+	// controller is a candidate controller for the new model.
+	controller dbmodel.Controller
+	// priority decides an ordering for controller selection
+	// based on the cloud-region.
+	priority uint
+}
+
+func shuffleCandidateControllers(controllers []candidateController) {
```
ale8k, line 33: Lil godoc would be nice here. It doesn't really just shuffle, it shuffles then sorts the priority.

And side question, did we c/p from the original because you wanted to leave shuffleRegionControllers just for the K8S stuff?

internal/jimm/juju/modelbuilder.go
ale8k, line 37: I noticed this isn't tested, could we add a test where it sets rand.Seed? It is our most important intelligent function of course aha 😝

internal/jimm/juju/model_test.go
```diff
@@ -1376,7 +1534,7 @@ func TestAddModel(t *testing.T) {
 
 			test.args.CloudCredentialTag = test.cloudCredTag.String()
 			args := juju.ModelCreateArgs{}
-			err = args.FromJujuModelCreateArgs(&test.args)
+			err = args.FromJujuModelCreateArgs(&test.args, test.TargetController)
```
ale8k: We're missing a test / could update this test where two controllers have the same priority and rand.Seed the test to make it go on 0a then 0b.

ale8k (changes requested):
Nice! I like how we filter the candidates down. I do think semantically it could be re-ordered so we don't do some unnecessary stuff - mainly around the auth checks and region priorities, we could look up the region before getting the controllers, then we know the priority and don't need a new slice (right?)

But I also think that's probably for another PR where we clean this up and just delete the builder structure to be honest. 

I've left some kind of "cleanup" comments, but once those are done I think it LGTM and we follow up some cleanup later?

internal/jimm/juju/model.go
kian99: Yip, I'll update our Juju dep and move it to the right place since the PR in Juju landed today.

internal/jimm/juju/model.go
kian99: Can do

internal/jimm/juju/modelbuilder.go
kian99: > It doesn't really just shuffle, it shuffles then sorts the priority.

So it does shuffle >.>

I c/p it instead of removing the original because the original is also used elsewhere when adding a hosted cloud.

Added a godoc.

internal/jimm/juju/modelbuilder.go
kian99: It's also used in `CreateControllerModel` where we dial the controller so I'll leave it like this for now.

internal/testutils/jimmtest/env.go
kian99: This was just c/p, removed the `if client != nil` check but left the rest the same.

internal/jujuapi/modelmanager.go
kian99: Changed now anyway

internal/jimm/juju/modelbuilder.go
kian99: Added some tests with AI and tweaked a bit. It doesn't use rand.Seed and just mocks out `shuffle` which I think is easier.

internal/jimm/juju/model_test.go
kian99: Since we're now testing the random shuffle logic I'm going to leave this out as I think we've got sufficient coverage.

internal/jimm/juju/model.go
ale8k: Ah sure, ty for confirming!

internal/jimm/juju/modelbuilder_internal_test.go
```diff
@@ -0,0 +1,137 @@
...
+import (
+	"math/rand/v2"
+	"testing"
+
+	qt "github.com/frankban/quicktest"
+
+	"github.com/canonical/jimm/v3/internal/dbmodel"
+)
+
+func TestShuffleCandidateControllers(t *testing.T) {
```
ale8k, line 14: Ty! Surprised this wasn't here tbh lol. Imagine if we had a bug in this 😆

internal/jimm/juju/modelbuilder.go
ale8k: Ty ty and what I meant is it shuffles and then sorts. This is a sorted sliced by priority, but not other fields.

internal/jimm/juju/model.go
```diff
@@ -118,7 +118,11 @@ func (j *JujuManager) AddModel(ctx context.Context, user *openfga.User, args *Mo
 		return nil, errors.E(err)
 	}
 
-	builder = builder.WithController(args.TargetController)
+	if args.TargetController != "" {
```
ale8k, line 81: Ty this reads much nicer now & muuuch appreciated. Sorry for being fastidious here...

internal/testutils/jimmtest/env.go
ale8k: Ah sure, ty for removing nil check.

internal/jimm/juju/modelbuilder.go
ale8k: Yeah same thing, ty for that.

ale8k (approved):

internal/jimm/juju/modelbuilder.go
alesstimec: I still think we should be checking the permission of the eventual owner.. just as we are verifying that cloud credentials belong to the eventual owner..

kian99: But then there is no point to creating models on behalf of someone else right? You'd need to grant them permissions, then their account could do it.

alesstimec: sgtm

PR #1747, opened 2025-11-27
Juju 8806/migrateandupgrade method

ale8k:
## Description
This PR builds upon #1740 and introduces the migrate and upgrade method. So, given we can prepare, bootstrap, we can now migrate and upgrade. 

There is a few hiccups in this PR, namely:
- I've left a TODO for when the dry-run of migrate is merged and where to add it.
- A todo, which is more of a question about user removal in ModelInfo, happy to remove this TODO once answered in this PR
- A TODO of where to backup, but that also has issues in that only IAAS can be backed up. 
- A TODO questioning what IgnoreAgentVersions means, I'm guessing it is the option to filter out versions?
- A comment below the "upgrade in progress", this is a situation we cannot handle right now. But I believe once we have a workflow engine, we can surface a user manual intervention required error, and then they could retry this upgrade. My final intent is that these will be done in separate retry-able jobs to handle this scenario.
- The controller actually "chooses" the version based on if it is possible to upgrade to it. I haven't dove into how it chooses, but I guess that's something to be aware of and needs a little bit of investigation.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

### DIFF
--- a/internal/jimm/upgrade/upgrade.go
+++ b/internal/jimm/upgrade/upgrade.go
@@ -27,6 +27,10 @@ import (
 	"github.com/canonical/jimm/v3/internal/openfga"
 )
 
+var (
+	AlreadyUpgradedError = jujuerrors.New("model has already been upgraded")
+)
+
 // BootstrapManager defines the bootstrap manager methods required by the upgrade manager.
 type BootstrapManager interface {
 	WaitForJobCompletion(ctx context.Context, jobId uuid.UUID, config bootstrap.WaitConfig) error
@@ -213,11 +217,22 @@ func (u *upgradeManager) MigrateAndUpgradeModel(ctx context.Context, user *openf
 			Attempts: 10,
 			Delay:    5 * time.Second,
 			Func: func() error {
-				// TODO: We don't care about the user here. We just wanna know if internal migration completed.
-				// Our ModelInfo handles the redirect, it'll error until the migration is complete (traversing the redirect err to
-				// the new controller). Can we remove the user?
 				mi, err = u.jujuManager.ModelInfo(ctx, user, mt)
-				return err
+				if err != nil {
+					return err
+				}
+
+				m, err := u.jujuManager.GetModel(ctx, mi.UUID)
+				if err != nil {
+					return err
+				}
+
+				// It hasn't migrated yet, so error out.
+				if m.Controller.Name != targetControllerName {
+					return errors.E("model has not yet migrated to target controller")
+				}
+
+				return nil
 			},
 			Clock: clock.WallClock,
 		},
@@ -245,7 +260,7 @@ func (u *upgradeManager) MigrateAndUpgradeModel(ctx context.Context, user *openf
 		// that manual intervention is required.
 	}
 	if jujuerrors.Is(upgradeErr, jujuerrors.AlreadyExists) {
-		upgradeErr = jujuerrors.New("model has already been upgraded")
+		upgradeErr = AlreadyUpgradedError
 	}
 
 	if upgradeErr != nil {
--- a/internal/jimm/upgrade/upgrade.go
+++ b/internal/jimm/upgrade/upgrade.go
@@ -7,11 +7,15 @@ package upgrade
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/google/uuid"
+	"github.com/juju/clock"
+	jujuerrors "github.com/juju/errors"
 	jujucloud "github.com/juju/juju/cloud"
 	jujuparams "github.com/juju/juju/rpc/params"
 	"github.com/juju/names/v5"
+	"github.com/juju/retry"
 	"github.com/juju/version/v2"
 	"github.com/juju/zaputil/zapctx"
 	"go.uber.org/zap"
@@ -59,6 +63,15 @@ func NewUpgradeManager(
 	if bootstrapManager == nil {
 		return nil, errors.E("bootstrap manager cannot be nil")
 	}
+	if jujumanager == nil {
+		return nil, errors.E("juju manager cannot be nil")
+	}
+	if store == nil {
+		return nil, errors.E("store cannot be nil")
+	}
+	if dialer == nil {
+		return nil, errors.E("dialer cannot be nil")
+	}
 	return &upgradeManager{
 		bootstrapManager: bootstrapManager,
 		jujuManager:      jujumanager,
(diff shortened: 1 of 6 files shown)


internal/jimm/upgrade/upgrade.go
```diff
@@ -173,3 +190,73 @@ func (u *upgradeManager) CloneController(ctx context.Context, user *openfga.User
...
+	if err := retry.Call(
+		retry.CallArgs{
+			Attempts: 10,
+			Delay:    5 * time.Second,
+			Func: func() error {
+				// TODO: We don't care about the user here. We just wanna know if internal migration completed.
+				// Our ModelInfo handles the redirect, it'll error until the migration is complete (traversing the redirect err to
+				// the new controller). Can we remove the user?
+				mi, err = u.jujuManager.ModelInfo(ctx, user, mt)
+				return err
```
kian99: This check seems insufficient. I.e. when you reach this step and do the `ModelInfo` call, the model may have migrated and everything is fine (true for a small/empty model) but if it's a large model, migration will take time and this call will still succeed but without telling you whether the model was moved yet.

You should do `ModelInfo` followed by a call to `Database.GetModel()` and check the model's controller. If it's not the new target controller, then the move hasn't completed yet.

internal/jimm/upgrade/upgrade.go
```diff
@@ -173,3 +190,73 @@ func (u *upgradeManager) CloneController(ctx context.Context, user *openfga.User
...
+	}
+
+	api, err := u.dialer.Dial(ctx, dbCtrl, names.ModelTag{}, nil, nil)
+	if err != nil {
+		return controllerChosenVersion, errors.E(fmt.Errorf("failed to dial target controller: %w", err))
+	}
+
+	var upgradeErr error
+	controllerChosenVersion, upgradeErr = api.UpgradeModel(mi.UUID, targetVersion, "", false, true)
+	//nolint:staticcheck // Has a description to highlight this possibility.
```
kian99, line 255: What's the nolint escape for?

internal/jimm/upgrade/upgrade.go
```diff
@@ -173,3 +190,73 @@ func (u *upgradeManager) CloneController(ctx context.Context, user *openfga.User
...
+	if jujuparams.IsCodeUpgradeInProgress(upgradeErr) {
```
kian99, line 256: We're not doing anything inside this `if`?

kian99 (commented):
I like how simple it is, thanks for this.

internal/jimm/upgrade/upgrade.go
```diff
@@ -173,3 +190,73 @@ func (u *upgradeManager) CloneController(ctx context.Context, user *openfga.User
...
+	var mi *jujuparams.ModelInfo
+	if err := retry.Call(
+		retry.CallArgs{
+			Attempts: 10,
+			Delay:    5 * time.Second,
+			Func: func() error {
+				// TODO: We don't care about the user here. We just wanna know if internal migration completed.
+				// Our ModelInfo handles the redirect, it'll error until the migration is complete (traversing the redirect err to
+				// the new controller). Can we remove the user?
+				mi, err = u.jujuManager.ModelInfo(ctx, user, mt)
```
SimoneDutto, line 220: i would try to use `IsFatalError` to model fatal error from migration in progress.
For not it can be really simple and just checking the error type, but i guess it's easy it will expand as we have more knowledge on how to check the model has been migrated.

internal/jimm/upgrade/upgrade.go
```diff
@@ -173,3 +190,73 @@ func (u *upgradeManager) CloneController(ctx context.Context, user *openfga.User
...
+	controllerChosenVersion, upgradeErr = api.UpgradeModel(mi.UUID, targetVersion, "", false, true)
+	//nolint:staticcheck // Has a description to highlight this possibility.
+	if jujuparams.IsCodeUpgradeInProgress(upgradeErr) {
+		// Apparently upgrades can have issues, that can be manually resolved, then you can run
+		// the upgrade-model command with the --reset-previous-upgrade option.
+		// We can't do that here, so we just return an error. We could possible indicate to the user here
+		// that manual intervention is required.
+	}
+	if jujuerrors.Is(upgradeErr, jujuerrors.AlreadyExists) {
+		upgradeErr = jujuerrors.New("model has already been upgraded")
```
SimoneDutto: define a const error for this, it will be useful

internal/jimm/upgrade/upgrade.go
```diff
@@ -173,3 +190,73 @@ func (u *upgradeManager) CloneController(ctx context.Context, user *openfga.User
...
+		return controllerChosenVersion, errors.E(fmt.Errorf("failed to upgrade model after migration: %w", upgradeErr))
+	}
+
+	zapctx.Info(ctx, "model migrate and upgrade complete",
+		zap.String("model-uuid", modelUUID),
+		zap.String("target-controller", targetControllerName),
+		zap.String("target-version", targetVersion.String()),
+		zap.String("controller-chosen-version", controllerChosenVersion.String()),
+	)
+	return controllerChosenVersion, nil
```
SimoneDutto, line 276: why do we return the controller version? Can it be something different from the version we asked for?

internal/jujuclient/modelupgrader.go
```diff
@@ -0,0 +1,22 @@
...
+// best version is selected by the controller and returned as ChosenVersion
+// in the result.
+func (c Connection) UpgradeModel(
+	modelUUID string,
+	targetVersion version.Number,
+	stream string,
+	ignoreAgentVersions bool,
+	dryRun bool,
+) (version.Number, error) {
+	return modelupgrader.NewClient(&c).UpgradeModel(modelUUID, targetVersion, stream, ignoreAgentVersions, dryRun)
```
SimoneDutto, line 21: why are we steering away from the raw facade pattern? This mix and match is going to bite us when we need to support both 3 and 4, or we do a refactor

kian99: I can answer this one. I started the move from raw calls to Juju's API since that reduces our burden of testing. But depending on the outcome of JIMM's compatability matrix with Juju, we may need to go back to raw facade calls. For now I don't mind either way (and this approach is simpler).

internal/jimm/upgrade/upgrade.go
ale8k: Answering the no-lint here too.

I wanted to keep this block so we remember this is what Juju does, an UpgradeInProgress is all they tell us, it could be taking an indefinite amount of time and I guess we'd never know the progress. 

I think in a later phase we'll come back and need to remember this check, as in perhaps some kind of upgrade health check should be run that can eventually prompt for user intervention OR we give them a way to reset the previous upgrade and let the workflow eventually finish OR we ignore it, fire it off, and call it job done.

No matter what, it seems we require the ability to allow users to reset upgrades to controllers behind JIMM and I didn't want to forget it.

internal/jimm/upgrade/upgrade.go
ale8k: Yep it appears it can, I need to dive in and see why the controller may pick a version for us. I think for our case, where we're specifying the same version as the controller's agent it *should* be OK but I need to confirm.

internal/jujuclient/modelupgrader.go
ale8k: Yeah... A while back I wanted to do this for newer client facade methods too. We can switch it later if need be.

internal/jimm/upgrade/upgrade.go
ale8k: There's are not fatal errors right now / yet. So is this OK?

SimoneDutto (approved):

kian99 (approved):

PR #1748, opened 2025-11-28
Juju 8806/upgrade to method

ale8k:
## Description
Adds the UpgradeTo method to JIMM. It's super primitive, and has some TODOs scattered to remind us of things to consider when moving up in phases. Namely:
  - Currently this only works with personal clouds.
    We need to extend it to work with public clouds
    (i.e., not send the cloud and just region and credential)
  - K8S clouds are determined by the HostCloudRegion of the controller model summary.
    We need to handle that case here. It may look something like "if caas" again.
  - No original configuration from the source controller is carried through, this will need adding.
  - We need to name the controller, at the moment it's just a timestamp.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests


pkg/api/params/params.go
```diff
@@ -653,15 +653,3 @@ type DestroyControllerRequest struct {
 	// ControllerName of the controller to destroy
 	ControllerName string `json:"controller-name"`
 }
-
```
ale8k: Simone added in his PR

internal/jujuapi/jimm.go
```diff
@@ -705,30 +702,3 @@ func (r *controllerRoot) StartDestroyControllerJob(ctx context.Context, req apip
 		JobID: jobID,
 	}, nil
 }
-
-func (r *controllerRoot) UpgradeTo(ctx context.Context, req apiparams.UpgradeToRequest) (apiparams.UpgradeToResponse, error) {
```
ale8k: Doing in a separate PR

internal/jimm/upgrade/upgrade.go
```diff
@@ -275,3 +280,58 @@ func (u *upgradeManager) MigrateAndUpgradeModel(ctx context.Context, user *openf
...
+// automated upgrade process.
+//
+// TODO:
+//   - Currently this only works with personal clouds.
+//     We need to extend it to work with public clouds
+//     (i.e., not send the cloud and just region and credential)
+//   - K8S clouds are determined by the HostCloudRegion of the controller model summary.
+//     We need to handle that case here. It may look something like "if caas" again.
+//   - No original configuration from the source controller is carried through, this will need adding.
+//   - We need to name the controller, at the moment it's just a timestamp.
```
kian99: Can we put this in a Jira card with clearer wording and put a link to the card here?

internal/jimm/upgrade/upgrade.go
```diff
@@ -275,3 +280,58 @@ func (u *upgradeManager) MigrateAndUpgradeModel(ctx context.Context, user *openf
...
+func (u *upgradeManager) UpgradeTo(ctx context.Context, user *openfga.User, modelUUID string, targetVersion version.Number) (version.Number, error) {
+	var chosenVersion version.Number
+	var newControllerName string = fmt.Sprintf("controller-%d", time.Now().Unix())
```
kian99, line 293: What about adding this as an input param for now? I.e. it comes from the user.

kian99 (commented):
Looks good, just some minors comments.

SimoneDutto (approved):
lgtm

internal/jimm/upgrade/upgrade.go
ale8k: Yeah I'll make a card and place it on phase 2.

internal/jimm/upgrade/upgrade.go
ale8k: I was thinking about this, and as it's phase 1 and the later phases ideally won't take them as input from the user I think this is fine. I've added to the card anyways. I do think we need some kind of logical linking between controller upgrades for decomissioning purposes, so I imagine we'll tackle it then. Right now It doesn't matter so much I don't think and it has a UUID.

internal/jimm/upgrade/upgrade.go
```diff
@@ -148,18 +153,18 @@ func (u *upgradeManager) PrepareUpgradeTo(ctx context.Context, modelUUID string,
...
+		return bootstrapCloud, bootstrapCloudRegion, bootstrapCredential, errors.E("controller cloud is not marked as a controller cloud")
 	}
 
 	if !bootstrapCloud.IsControllerCloud {
-		return bootstrapCloud, bootstrapCredential, errors.E("controller cloud is not marked as a controller cloud")
+		return bootstrapCloud, bootstrapCloudRegion, bootstrapCredential, errors.E("controller cloud is not marked as a controller cloud")
 	}
 
-	return bootstrapCloud, bootstrapCredential, nil
+	return bootstrapCloud, bootstrapCloudRegion, bootstrapCredential, nil
```
luci1900, line 167: The more I look at this the more I wonder if it should return a struct. Not for this PR, though.

internal/jimm/upgrade/upgrade.go
```diff
@@ -275,3 +280,52 @@ func (u *upgradeManager) MigrateAndUpgradeModel(ctx context.Context, user *openf
 	)
 	return controllerChosenVersion, nil
 }
+
+// UpgradeTo upgrades a model.
+// That is, it bootstraps a new controller with the target version
+// and migrates the model to that controller. This is "Phase 1" of the
+// automated upgrade process.
+//
+// This currently only works with personal, none-kubernetes clouds.
```
luci1900: non-kubernetes

luci1900 (approved):

internal/jimm/upgrade/upgrade.go
ale8k: Fair point... I completely missed the region earlier but yeah it should probably be a struct. In the later phases this is going to be deleted and more of a "hey here's how to do it" reference.

PR #1749, opened 2025-11-28
ci: configure dockerhub mirror

kian99:
## Description
Thanks to some [help](https://chat.canonical.com/canonical/pl/3czj48u17fdzietpci6ebtyjjy) from platform engineering, it was found that Microk8s was failing to complete its setup because it was hitting Dockerhub rate-limiting errors when our integration test job runs on PS7. This PR ensures Microk8s uses our Dockerhub registry mirror.

Used https://github.com/canonical/github-runner-operator/pull/662 and https://canonical-self-hosted-github-runner-docs.readthedocs-hosted.com/en/latest/usage/faq/how-to-avoid-dockerhub-rate-limits/ as references.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -5,6 +5,10 @@ on:
   workflow_dispatch:
   merge_group:
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   build_test:
     name: Build and Test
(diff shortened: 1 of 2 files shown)


ale8k (approved):

SimoneDutto (approved):
why don't we have the same error in tf?

kian99:
> why don't we have the same error in tf?

Using the https://github.com/charmed-kubernetes/actions-operator action, the mirror is setup automatically. We don't want to use that action here because it also bootstraps a Juju controller (we optionally use it in the action that sets up JIMM, Juju, etc).

PR #1751, opened 2025-12-01
Getting rid of the conn suite 

SimoneDutto:
# Description

Approach to get rid of the conn suite.

Rationale: keep most of the test suites, and setup as we currently have but gently switch from the connsuite to an externally bootstrapped controller.

Outline:
- the juju controller is bootstrapped via the Juju cli and configured to use the same jwks is used by the test's jimm
- jimm's tests will still create a running jimm per test via the test harness as usual and they'll attach the externally bootstrapped controller.
- jimm will run command on the backing juju controller

# JWKS
I've added an alpine nginx to serve a fixed jwks.json file that the backing juju controller will be bootstrapped with.
`JWKS_DNS=jwks.localhost CONTROLLER_NAME=test-e2e ./local/jimm/setup-controller.sh`

And reuse the same jwks setting up JIMM inside the `JIMMSuite`

# Adding the controller
Right now in the E2ESuite we parse the yaml config file defined at JIMM_BACKING_CONTROLLER_CONFIG path.
And we add the controller.
For now it's just a single controller but the code is ready to add multiple controllers via the same controllers.yaml file.

```
s.AddController(c, "controller-1", &api.Info{
		ControllerUUID: "b2746812-29be-47a6-8cae-40717c390a19",
		Addrs:          []string{"10.165.178.29:17070"},
		Tag:            names.NewUserTag("admin"),
		Password:       "",
		CACert:         ""
	})
```
This file can be generated via `make generate-test-env <controller>`

# Credentials
Here the approach is similar to the tf provider, we list the credentials on the client and we add to JIMM from the filestore.

# Change to the tests
- using the E2ESuite instead of the BootstrapSuite
- remove all ConnSuite specific calls. Ex.
` f := factory.NewFactory(stateMachine3.State, s.StatePool) `
and calling the JIMM's client directly, that will target the backing controller.
- do proper cleanup. Ex. adding a model to the backing controller, now requires cleanup because there isn't a controller per test.

# Downsides

- Bootstrapping multiple controllers will require a new suite called "E2EMultipleControllersSuite" and a different way to pass the connection details (maybe a yaml file).
- Bootstrapping outside the test harness, so there is no control over isolation for tests requiring a full controller.


docker-compose.yaml
```diff
@@ -152,6 +152,18 @@ services:
       traefik.http.routers.juju-dashboard.tls: true
       traefik.http.services.juju-dashboard.loadbalancer.server.port: 8080
 
+  jwks-server:
```
kian99, line 154: A comment above this line about what the purpose of this container would be helpful, or maybe better would be an update to `local/README.md`

testing/modelmanager_test.go
```diff
@@ -0,0 +1,68 @@
...
+	"github.com/canonical/jimm/v3/internal/testutils/jimmtest"
+)
+
+type modelE2eManagerSuite struct {
+	websocketE2ESuite
+}
+
+var _ = gc.Suite(&modelE2eManagerSuite{})
+
+func (s *modelE2eManagerSuite) TestListModelSummaries(c *gc.C) {
```
kian99, line 25: What do you think about moving this test outside of `internal/jujuapi`. The spec states we'll write these new tests in `testing/e2e/`, a nice step towards integration tests that don't need to reach for internals.

.vscode/settings.json
SimoneDutto, line 1: the reason why we can't track .vscode anymore it's described in `testing/README.md` at the bottom.
Basically, babak's extension doesn't support ${workspaceFolder}` expansion and `go.testEnvFile`, so we can't really track this file no more. Because everybody here will have a different absolute path for when he wants to run the E2E_test

luci1900: It's fairly common to commit a `settings.example.json` or similar. That way everyone can have their own, but it's easier for new joiners to set up.

SimoneDutto: yeah, added

luci1900 (approved):
I like this approach.

testing/modelmanager_test.go
kian99, line 1: Name of the file needs fixing

testing/package_test.go
```diff
@@ -0,0 +1,21 @@
...
+)
+
+// Registers Go Check tests into the Go test runner.
+func TestPackage(t *testing.T) {
+	gc.TestingT(t)
+}
+
+func onlyRunE2ETests(c *gc.C) {
+	if _, ok := os.LookupEnv("RUN_E2E_TESTS"); !ok {
+		c.Skip("Skipping e2e tests. Set RUN_E2E_TESTS=true to run them.")
```
kian99: Can we move this check into `TestPackage`?

.vscode/.settings.json.example
kian99, line 1: Should we add the .settings file to .gitignore now since everyone will need a different config potentially?

testing/README.md
```diff
@@ -0,0 +1,48 @@
...
+This guide outlines the steps to set up and run the e2e test suite.
+
+## Prerequisites
+
+1. **Start the test environment**
+
+   Start the dev environment to run these tests:
+
+   ```bash
+   make test-env
```
kian99, line 12: Shouldn't this be dev-env?

testing/README.md
kian99, line 1: Seems like we could create a make target like `e2e_test` that does these 3 steps as a convenience.

kian99 (commented):
Thanks for this

.vscode/.settings.json.example
SimoneDutto: this is just the .example, it's not used by vscode

testing/package_test.go
SimoneDutto: i've tried it but it wasn't working. I'll retry

.vscode/.settings.json.example
kian99: What I mean is, the other file should not be commited anymore since it will need to contain user specific paths. We could add it to gitignore, if it's not already there.

SimoneDutto: in fact i've deleted the `settings.json` and added to .gitignore

testing/README.md
SimoneDutto: i think this is the right one, because we don't need jimm

.vscode/.settings.json.example
kian99: Ah I see that now thanks!

kian99 (approved):

testing/modelmanager_test.go
SimoneDutto: fixed

testing/package_test.go
SimoneDutto: it works

testing/README.md
SimoneDutto: it's just that you don't want to do all 3 at all time, and i prefer them to be separated for now. We can create it later

PR #1753, opened 2025-12-02
feat: plugin-upgradeto-facade

ale8k:
## Description
Plugs in the upgrade to facade into JIMM. This PR is built on top of #1748 so please review that first. If you want to review this, just review commit 3ba9172.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

## Test instructions
This description will be updated when it has been QA'd and steps placed.


internal/jujuapi/jimm.go
```diff
@@ -702,3 +707,33 @@ func (r *controllerRoot) StartDestroyControllerJob(ctx context.Context, req apip
...
+// UpgradeTo upgrades the controller hosting the given model by cloning a new controller
+// at the requested version and migrating the model to it (phase 1 automated upgrade).
+func (r *controllerRoot) UpgradeTo(ctx context.Context, req apiparams.UpgradeToRequest) (apiparams.UpgradeToResponse, error) {
+	if !r.user.JimmAdmin {
+		return apiparams.UpgradeToResponse{}, errors.E(errors.CodeUnauthorized, "unauthorized")
+	}
+
+	mt, err := names.ParseModelTag(req.ModelTag)
+	if err != nil {
+		return apiparams.UpgradeToResponse{}, errors.E(errors.CodeBadRequest, fmt.Errorf("invalid model tag %q: %w", req.ModelTag, err))
```
kian99, line 720: Shouldn't these all be 
```go
return apiparams.UpgradeToResponse{
			Success: false,
			Error:   err.Error(),
		}
```
Since the error goes in the response object?

internal/jimm/jimm.go
```diff
@@ -316,6 +317,14 @@ type BootstrapManager interface {
 	StartDestroyControllerJob(ctx context.Context, user *openfga.User, params bootstrap.DestroyControllerParams) (string, error)
 }
 
+// UpgradeManager provides methods to manage controller cloning and model automated upgrades.
+type UpgradeManager interface {
+	CloneController(ctx context.Context, user *openfga.User, params upgrade.CloneControllerParams) error
+	MigrateAndUpgradeModel(ctx context.Context, user *openfga.User, modelUUID string, targetControllerName string, targetVersion version.Number) (version.Number, error)
+	PrepareUpgradeTo(ctx context.Context, modelUUID string, targetVersion version.Number) (jujucloud.Cloud, string, jujucloud.Credential, error)
```
kian99: Do these 3 methods need to be exposed?

ale8k: Nope good spot, what c/p.

internal/jujuapi/jimm.go
ale8k: As discussed it probably shouldn't have an Error field as they're for iterative responses. The CLI expects this error to be filled though on a failed success (hence why I added it), but I think later on we should just come and remove the error from the response / phase 2 will rewrite this anyway.

ale8k: Scratch that, I've got a none-nil err in the linter due to trying to return nil after checking error is not nil. I'll remove the err field entirely.

pkg/api/params/params.go
```diff
@@ -246,8 +246,7 @@ type UpgradeToRequest struct {
 
 // UpgradeToResponse holds the response for phase 1 of an automated upgrade.
 type UpgradeToResponse struct {
-	Success bool   `json:"success"`
-	Error   string `json:"error,omitempty"`
+	Success bool `json:"success"`
```
SimoneDutto, line 249: Basically we could just a facade with no return type. This is useless right?

ale8k: Tbh yeah lol

ale8k: I mean it doesn't matter so much as this is phase 1 and it's gonna be replaced anyway.

SimoneDutto (approved):
approving, if you can remove the return type is better.

kian99 (approved):

PR #1754, opened 2025-12-02
Juju 8806/qa fixes automated upgrades

ale8k:
## Description
Contains some small fixes for automated upgrades.

I think the code is fairly self explanatory, so I'll save the shpeel. 

## Test instructions
The QA steps to run an automated upgrade are as so:
Turn off controller masking manually, set this to false:
https://github.com/ale8k/jimm/blob/4c572adc7c3056ec3e92623c13b150be98123a53/internal/jujuapi/controllerroot.go#L57

```bash
# Export this variable export JIMM_BOOTSTRAP_LOGIN_TOKEN_REFRESH_URL="http://<an IP/DNS that can reach JIMM from your machines local LXD>:17070/.well-known/jwks.json"
sudo snap install juju --channel=3/stable --revision=32912 # 32912 = 3.6.11 (arm), 32911 = 3.6.11 (amd64)
go build ./cmd/jaas
./local/jimm/qa-lxd.sh
juju status # Verify the model is 3.6.11
./jaas upgrade-to 3.6.12 $(juju show-model test-lxd | yq '.test-lxd.model-uuid')
juju status # Verify the model version is 3.6.12 
juju show-model test-lxd # Verify controller UUID is the same as the newly bootstrapped controller (find UUID in ./jaas controllers)
juju switch qa-lxd
juju models # Verify model is gone
```


internal/jimm/bootstrap/bootstrap.go
```diff
@@ -574,9 +574,11 @@ func (b *bootstrapManager) tryCleanupController(ctx context.Context, jujuCmd Juj
 func (b *bootstrapManager) consumeCommandOutput(ctx context.Context, outputCh <-chan jujucommands.OutputLine, jobId uuid.UUID) error {
 	for output := range outputCh {
 		if output.Err != nil {
+			zapctx.Error(ctx, "command output error", zap.Error(output.Err))
 			b.writeJobLog(ctx, jobId, output.Err.Error())
 			return errors.E(fmt.Errorf("command failed: %w", output.Err))
```
kian99: This is a good example of double logging. We're returning the error so you shouldn't also log it here.

internal/jimm/upgrade/upgrade.go
```diff
@@ -105,8 +105,8 @@ func (u *upgradeManager) PrepareUpgradeTo(ctx context.Context, modelUUID string,
 		return bootstrapCloud, bootstrapCloudRegion, bootstrapCredential, errors.E(err)
 	}
 
-	if currentVersion.Compare(targetVersion) >= 0 {
-		return bootstrapCloud, bootstrapCloudRegion, bootstrapCredential, errors.E(errors.CodeBadRequest, "target version must be greater than current version")
+	if currentVersion.Compare(targetVersion) == 1 {
+		return bootstrapCloud, bootstrapCloudRegion, bootstrapCredential, errors.E(errors.CodeBadRequest, "target version must be greater than current version or equal to current version")
```
kian99:
```suggestion
		return bootstrapCloud, bootstrapCloudRegion, bootstrapCredential, errors.E(errors.CodeBadRequest, "target version must be greater than or equal to current version")
```

internal/jujucommands/bootstrap.go
```diff
@@ -90,6 +90,9 @@ func (b BootstrapCmdParams) BuildBootstrapCmdArgs() []string {
 	// Always add controller name & cloud at the end
 	args = append(args, b.CloudNameAndRegion, b.ControllerName)
 
+	// why
```
kian99: dd

internal/jimm/upgrade/upgrade.go
```diff
@@ -206,9 +206,29 @@ func (u *upgradeManager) MigrateAndUpgradeModel(ctx context.Context, user *openf
...
+			Delay:    5 * time.Second,
+			// We could consider all errors fatal, bar: target prechecks failed: machine 0 not running (pending)
+			// IsFatalError:
+			Func: func() error {
+				zapctx.Debug(ctx, "Attempting to initiate internal migration")
+				r, err := u.jujuManager.InitiateInternalMigration(ctx, user, modelUUID, targetControllerName)
+				if err != nil {
+					zapctx.Error(ctx, "Failed to initiate internal migration", zap.Error(err))
+					return err
+				}
```
kian99, line 225: Is this idempotent?

ale8k: I haven't checked our side of the code, but IIRC when it was implemented you said it was? I'd be surprised if it wasn't either honestly as Juju's side of things is.

internal/jimm/bootstrap/bootstrap.go
```diff
@@ -577,6 +577,7 @@ func (b *bootstrapManager) consumeCommandOutput(ctx context.Context, outputCh <-
...
 		}
+		zapctx.Debug(ctx, "command output", zap.String("line", output.Line))
```
ale8k: I'd like to keep this as when in debug it's helpful to also see the logs on JIMM's side and not go to DB to see if it is running.

internal/jimm/upgrade/upgrade.go
```diff
@@ -105,8 +105,8 @@ func (u *upgradeManager) PrepareUpgradeTo(ctx context.Context, modelUUID string,
 		return bootstrapCloud, bootstrapCloudRegion, bootstrapCredential, errors.E(err)
 	}
 
-	if currentVersion.Compare(targetVersion) >= 0 {
-		return bootstrapCloud, bootstrapCloudRegion, bootstrapCredential, errors.E(errors.CodeBadRequest, "target version must be greater than current version")
+	if currentVersion.Compare(targetVersion) == 1 {
+		return bootstrapCloud, bootstrapCloudRegion, bootstrapCredential, errors.E(errors.CodeBadRequest, "target version must be greater than or equal to current version")
```
SimoneDutto, line 109: so we don't support migrating to same version anymore?

internal/jimm/upgrade/upgrade.go
```diff
@@ -206,9 +206,29 @@ func (u *upgradeManager) MigrateAndUpgradeModel(ctx context.Context, user *openf
...
+	// As the controller has just been bootstrapped, the controller machine can still be in a pending state.
+	// Unfortunately Juju doesn't return a typed error for us to examine, and the best we've got is the error message string.
+	// So we retry a few times here to allow the controller machine to come up.
+	var iimResult jujuparams.InitiateMigrationResult
+	if err := retry.Call(
+		retry.CallArgs{
+			Attempts: 10,
+			Delay:    5 * time.Second,
+			// We could consider all errors fatal, bar: target prechecks failed: machine 0 not running (pending)
+			// IsFatalError:
```
SimoneDutto, line 218: leftover?

internal/jujucommands/bootstrap.go
```diff
@@ -90,6 +91,10 @@ func (b BootstrapCmdParams) BuildBootstrapCmdArgs() []string {
 	// Always add controller name & cloud at the end
 	args = append(args, b.CloudNameAndRegion, b.ControllerName)
 
+	// Add architecture constraint to ensure juju bootstraps with the correct arch.
+	// Without this, the controller application that is deployed can be deployed with the wrong architecture.
+	args = append(args, fmt.Sprintf("--bootstrap-constraints=%s", fmt.Sprintf("arch=%s", runtime.GOARCH)))
```
SimoneDutto, line 96: mmm, this is ok because we always bootstrap controller from the same jimm's arch. Right?

SimoneDutto (approved):
lgtm, bar some comments

internal/jimm/upgrade/upgrade.go
ale8k: This would allow that, this is just checking if the current version > target version.

internal/jimm/upgrade/upgrade.go
ale8k: This is more of a reminder for when we come around to phase 2. Tbh I don't really like what I've done here and think it deserves a bit more due diligence, but I think we can do that in phase 2.

internal/jujucommands/bootstrap.go
ale8k: Well... I think it's probably too opinionated and should probably be an option that is passed in & sourced from the OG controller. This is more of a "let's make it work for now". But to pass it in requires a bit more effort and I'm trying to keep things lean.

internal/jimm/upgrade/upgrade.go
SimoneDutto: i'm not sure, what if it's 2 or 0?

ale8k: It can only be -1, 0 and 1. Less, equal, greater.

SimoneDutto: aaa, i thought it was a number. Don't they have enum?

I don't quite understand, so here we are returning an error in case target is less than current?

ale8k: Aha no no, they just return a number lol. If it's greater, current cannot be greater than target. But we allow it to be less or equal.

SimoneDutto: ok, thanks for clarification

internal/jimm/bootstrap/bootstrap.go
kian99: What I mean is, you can keep the debug logging but it should be done here. It should be done higher up. Otherwise you fall victim to logging at different layers and logging the same thing multiple times. So at the point where we no longer return the error, I.e. The place where we are saving it somewhere, we can also log it.

kian99: If that's feasible ofc. I assume it is

ale8k: Ahh you mean at the call site. Not sure how I'd do it higher up because we're reading from the chan here.

internal/jimm/bootstrap/bootstrap.go
kian99: @ale8k Can you add this patch to your PR, I was taking a look and if a job fails we don't actually log it, we just write the error to DB.
```
@@ -115,6 +115,7 @@ func (j *Tracker) monitorJob(id uuid.UUID, jobErrCh chan error, cancelJob contex
                select {
                case err := <-jobErrCh:
                        if err != nil {
+                               zapctx.Error(ctx, "job failed", zap.String("id", id.String()), zap.Error(err))
                                if err := j.store.SetJobFailed(ctx, id, err); err != nil {
                                        zapctx.Error(ctx, "error marking the job as failed", zap.Error(err), zap.String("id", id.String()))
                                }
```

internal/jimm/bootstrap/bootstrap.go
kian99: Seems like I put my comments in the wrong place. This one is just logging the command output so feel free leave it in. Although be aware if we have multiple jobs running at the same time, the log lines will be interleaved and with a job id in the log you can't correlate them.

internal/jimm/upgrade/upgrade.go
alesstimec: i'd still suggest a more helpful error message.. something along the lines "migrating models to an older version of Juju is not supported".. this "target version" terminology might be unfamiliar to everybody outside of this team :)

internal/jimm/upgrade/upgrade.go
```diff
@@ -132,7 +132,7 @@ func (u *upgradeManager) PrepareUpgradeTo(ctx context.Context, modelUUID string,
 		return bootstrapCloud, bootstrapCloudRegion, bootstrapCredential, errors.E(fmt.Errorf("failed to parse cloud credential tag from controller model summary: %w", err))
 	}
 
-	credentialContents, err := api.CredentialContents(ctrlCloud.Id(), ctrlCloudCred.Id(), true)
+	credentialContents, err := api.CredentialContents(ctrlCloud.Id(), ctrlCloudCred.Name(), true)
```
alesstimec, line 135: will this work.. afaics juju compiles the credential ID by using the cloud name, the name of the logged in user and the credential name.. since we dial as "admin" on L112, wouldn't this mean that it will try to get credential contents for admin's credentials?  

see https://github.com/juju/juju/blob/3.6/apiserver/facades/client/cloud/cloud.go#L883

internal/jimm/upgrade/upgrade.go
```diff
@@ -206,9 +206,29 @@ func (u *upgradeManager) MigrateAndUpgradeModel(ctx context.Context, user *openf
 
 	// TODO: Once precheck PR merged, run precheck.
 
-	iimResult, err := u.jujuManager.InitiateInternalMigration(ctx, user, modelUUID, targetControllerName)
-	if err != nil {
-		return controllerChosenVersion, errors.E(fmt.Errorf("failed to initiate internal migration for upgrade: %w", err))
+	// As the controller has just been bootstrapped, the controller machine can still be in a pending state.
```
alesstimec, line 209: is there another call that would tell us whether the newly bootstrapped controller is ready to be used? i'd be more comfortable retrying a more harmless call (e.g. `juju models`) to see if the controller up&running

internal/jujucommands/bootstrap.go
alesstimec: i agree, we should copy the architecture of the source controller here.. we shouldn't assume the cloud is running machines of the same arch as JIMM.. although this is complicated..

alesstimec (commented):
a few comments

internal/jimm/upgrade/upgrade.go
ale8k: It's changed because I caught it during QA. It prefixed cloud and owner to it already somehow.

internal/jujucommands/bootstrap.go
ale8k: Created a card for phase 2 to make sure this gets done.

internal/jimm/upgrade/upgrade.go
ale8k: As juju is checking the state of the machines in the precheck, I'd argue this is the correct call to make. It is verifying everything that should be in X state is in X state. If we switch to another call we may miss this. 

Note, I am implementing a dry run of this for phase 2. In which case we'll be dry running this against many models.

internal/jimm/bootstrap/bootstrap.go
ale8k: We did originally log it... Wonder why it was removed? But yeah will do!

internal/jimm/bootstrap/bootstrap.go
ale8k: Fair point and we'll eventually test this in parallel, I've added the job id to the log.

internal/jimm/upgrade/upgrade.go
alesstimec: that's not what i'm asking.. i understand the need for the change.. but look at the juju code and what it does - it uses the authenticated user (admin in this case) to construct the cloud credential ID

ale8k: As discussed, resolved.

kian99 (approved):

PR #1755, opened 2025-12-03
test(modelmanagersuite): move/refactor modelManagerSuite to use the new websocketE2ESuite

SimoneDutto:
# Description

Tests are moved mostly as is. I've just avoided to create machines where wasn't necessary for the reason of the test. See TestModelInfo for example.

> Fun fact. The new suite with a live Juju controller takes 86 seconds, and running the same suite with the connsuite takes 90 seconds.


internal/utils/utils.go
```diff
@@ -21,3 +21,7 @@ func NewConversationID() string {
 	}
 	return hex.EncodeToString(buf)
 }
+
+func Ptr[T any](v T) *T {
+	return &v
+}
```
kian99, line 27: I remember reading Go added a language feature so that you don't have to do this anymore. But it's only coming in Go 1.26 - https://github.com/golang/go/issues/45624

kian99 (approved):

kian99:
Very nice btw!

testing/modelmanager_test.go
```diff
@@ -50,12 +66,772 @@ func (s *modelE2eManagerSuite) TestListModelSummaries(c *gc.C) {
...
+	c.Assert(err, gc.IsNil)
+	bob := openfga.NewUser(bobIdentity, s.OFGAClient)
+	err = bob.SetModelAccess(context.Background(), mt4, ofganames.WriterRelation)
+	c.Assert(err, gc.Equals, nil)
+
+	model5Name := petname.Generate(2, "-")
+	mt5 := s.AddModel(c, names.NewUserTag("charlie@canonical.com"), model5Name, names.NewCloudTag(jimmtest.TestE2ECloudName), jimmtest.TestE2ECloudRegionName, s.Model2.CloudCredential.ResourceTag())
+	// TODO (alesstimec) change once granting has been re-implemented
+	// err = client.GrantModel("bob@canonical.com", "admin", mt5.Id())
+	// c.Assert(err, gc.Equals, nil)
```
kian99: You can delete this btw

testing/modelmanager_test.go
```diff
@@ -50,12 +66,772 @@ func (s *modelE2eManagerSuite) TestListModelSummaries(c *gc.C) {
...
+	err = bob.SetModelAccess(context.Background(), mt5, ofganames.AdministratorRelation)
+	c.Assert(err, gc.Equals, nil)
+
```
kian99, line 187: Just to confirm, does adding a machine give us nothing with regards to this test? Seems like we could add a single machine.

testing/modelmanager_test.go
```diff
@@ -50,12 +66,772 @@ func (s *modelE2eManagerSuite) TestListModelSummaries(c *gc.C) {
...
+	c.Assert(models[0].Result.CloudCredentialTag, gc.Equals, s.Model.CloudCredential.Tag().String())
+	c.Assert(models[0].Result.OwnerTag, gc.Equals, names.NewUserTag("bob@canonical.com").String())
+	// As admin, bob can see all users
+	c.Assert(findUserAccess(models[0].Result.Users, "bob@canonical.com"), gc.Equals, "admin")
+	c.Assert(findUserAccess(models[0].Result.Users, "alice@canonical.com"), gc.Equals, "admin")
+
+	// Model 2 (s.Model2) - bob has no access
+	c.Assert(models[1].Result, gc.IsNil)
+	c.Assert(models[1].Error, gc.Not(gc.IsNil))
+	c.Assert(models[1].Error.Code, gc.Equals, jujuparams.CodeUnauthorized)
```
kian99, line 229: It's a bit strange we return models where the user doesn't have access.
Ah but that's because we specifically asked for it, okay that's fine.

testing/modelmanager_test.go
```diff
@@ -50,12 +66,772 @@ func (s *modelE2eManagerSuite) TestListModelSummaries(c *gc.C) {
...
+	mt4 := s.AddModel(c, names.NewUserTag("charlie@canonical.com"), model4Name, names.NewCloudTag(jimmtest.TestE2ECloudName), jimmtest.TestE2ECloudRegionName, s.Model2.CloudCredential.ResourceTag())
+
+	bobIdentity, err := dbmodel.NewIdentity("bob@canonical.com")
+	c.Assert(err, gc.IsNil)
+	bob := openfga.NewUser(bobIdentity, s.OFGAClient)
+	err = bob.SetModelAccess(context.Background(), mt4, ofganames.WriterRelation)
+	c.Assert(err, gc.Equals, nil)
+
+	model5Name := petname.Generate(2, "-")
+	mt5 := s.AddModel(c, names.NewUserTag("charlie@canonical.com"), model5Name, names.NewCloudTag(jimmtest.TestE2ECloudName), jimmtest.TestE2ECloudRegionName, s.Model2.CloudCredential.ResourceTag())
```
kian99, line 184: Don't we need to clean this model up?

internal/testutils/jimmtest/e2e_suite.go
```diff
@@ -142,9 +147,65 @@ func (s *WebsocketE2ESuite) SetUpTest(c *gc.C) {
...
+	cct := names.NewCloudCredentialTag(TestE2ECloudName + "/charlie@canonical.com/cred")
+	cloudCredentials := s.GetExistingClientCredentialsForCloud(c, TestE2ECloudName)
+	s.UpdateCloudCredential(c, cct, cloudCredentials)
+	s.Credential2 = new(dbmodel.CloudCredential)
+	s.Credential2.SetTag(cct)
+	err := s.JIMM.Database.GetCloudCredential(ctx, s.Credential2)
+	c.Assert(err, gc.Equals, nil)
+	mt := s.AddModel(c, names.NewUserTag("charlie@canonical.com"), petname.Generate(2, "-"), names.NewCloudTag(TestE2ECloudName), TestE2ECloudRegionName, cct)
+	s.Model2 = new(dbmodel.Model)
+	s.Model2.SetTag(mt)
```
kian99, line 162: add these for cleanup here

internal/testutils/jimmtest/e2e_suite.go
```diff
@@ -142,9 +147,65 @@ func (s *WebsocketE2ESuite) SetUpTest(c *gc.C) {
...
+
 }
 
 func (s *WebsocketE2ESuite) TearDownTest(c *gc.C) {
+	if s.Model2 != nil {
+		s.DestroyModelAndDeleteFromDatabase(c, s.Model2.ResourceTag())
+	}
+	if s.Model3 != nil {
+		s.DestroyModelAndDeleteFromDatabase(c, s.Model3.ResourceTag())
+	}
```
kian99: Avoid these and just run cleanup based on what was set during setup and test creation.

internal/testutils/jimmtest/e2e_suite.go
```diff
@@ -142,9 +147,65 @@ func (s *WebsocketE2ESuite) SetUpTest(c *gc.C) {
...
+	s.Credential2.SetTag(cct)
+	err := s.JIMM.Database.GetCloudCredential(ctx, s.Credential2)
+	c.Assert(err, gc.Equals, nil)
+	mt := s.AddModel(c, names.NewUserTag("charlie@canonical.com"), petname.Generate(2, "-"), names.NewCloudTag(TestE2ECloudName), TestE2ECloudRegionName, cct)
+	s.Model2 = new(dbmodel.Model)
+	s.Model2.SetTag(mt)
+	err = s.JIMM.Database.GetModel(ctx, s.Model2)
+	c.Assert(err, gc.Equals, nil)
+	mt = s.AddModel(c, names.NewUserTag("charlie@canonical.com"), petname.Generate(2, "-"), names.NewCloudTag(TestE2ECloudName), TestE2ECloudRegionName, cct)
+	s.Model3 = new(dbmodel.Model)
```
kian99, line 166: Why do we need to add so many arbitrary models? Matching the old suite?

kian99 (commented):
I was a bit quick with the previous review. Still lgtm but some suggestions.

testing/modelmanager_test.go
SimoneDutto: I think regarding the thing we want to test it gives us nothing.
Because the test is basically testing that the merge of info between what we have in JIMM and the ModelInfo call to Juju controllers work. And machines have nothing to do with it.
Since adding a machine is an expensive op, i wouldn't add it because i don't think it's useful for this, but lmk

internal/testutils/jimmtest/e2e_suite.go
SimoneDutto: you made me realize i could do something better:

```
func (s *E2ESuite) AddModel(c *gc.C, owner names.UserTag, name string, cloud names.CloudTag, region string, cred names.CloudCredentialTag) names.ModelTag {
	mt := s.JIMMSuite.AddModel(c, owner, name, cloud, region, cred)
	s.ModelsToCleanup = append(s.ModelsToCleanup, mt)
	return mt
}
```

and 
```
func (s *E2ESuite) TearDownTest(c *gc.C) {
	for _, mt := range s.ModelsToCleanup {
		s.DestroyModelAndDeleteFromDatabase(c, mt)
	}
	s.JIMMSuite.TearDownTest(c)
}
```

internal/testutils/jimmtest/e2e_suite.go
SimoneDutto: yep

testing/modelmanager_test.go
SimoneDutto: yep, now it's taken care of

internal/testutils/jimmtest/e2e_suite.go
alesstimec: maybe we could remove the creation of all these arbitrary models and move their creation to the specific test that needs them?

internal/testutils/jimmtest/e2e_suite.go
kian99: This can be even more generic. Add a method `Cleanup(f func())` which adds the function to a slice and iterate over the funcs in the cleanup The benefit of this is that we can more easily drop-in replace gocheck with a regular `*testing.T` which already has the same method with the same signature `func (c *testing.common) Cleanup(f func())`

luci1900 (approved):
Looks good.

internal/testutils/jimmtest/e2e_suite.go
```diff
@@ -230,39 +296,47 @@ func (s *E2ESuite) SetUpTest(c *gc.C) {
...
+	defer s.mutexModelsToCleanup.Unlock()
+	mt := s.JIMMSuite.AddModel(c, owner, name, cloud, region, cred)
+	s.ModelsToCleanup = append(s.ModelsToCleanup, mt)
+	return mt
+}
+
 func (s *E2ESuite) TearDownTest(c *gc.C) {
-	if s.Model != nil {
-		s.DestroyModelAndDeleteFromDatabase(c, s.Model.ResourceTag())
+	for _, mt := range s.ModelsToCleanup {
```
alesstimec: what happens if there are multiple tests running at the same time, adding models.. then one tests finishes and tries to delete all models, even those being used in other tests?

alesstimec (changes requested):
a bit worried about the model cleanup..

internal/testutils/jimmtest/e2e_suite.go
alesstimec: if anywhere, this should live in the TearDownSuite

kian99: But these models are per-test so that's fine, unless you are thinking of tests like `ListModels()`. I think for those you can just avoid marking them as `t.Parallel()` then they won't run in parallel with other tests.

SimoneDutto: yeah, the models are created/cleaned up per-test. So no concurrency issues.

SimoneDutto: from what i got from looking at the doc, gocheck doesn't support running tests in parallel

internal/testutils/jimmtest/e2e_suite.go
SimoneDutto: done!

alesstimec (approved):

PR #1757, opened 2025-12-05
feat: add import model guardrails

kian99:
## Description
This PR addresses an issue with model imports into JIMM.

JIMM handles model imports by interpreting the serialised model description and then using the v9 description package to deserialise it, modify it and re-serialise it before sending the bytes to the target controller. When Juju migrates to a new description package e.g. v10, JIMM will fail to deserialise this new format. In this PR we create our own `description` package that imports all the major versions of `github.com/juju/description` that we care about and create wrappers that expose only the parts that we need.

Inside this new package is also a mapping of Juju controller version to description package version. I.e. in 3.6.9 - 3.6.12 were using v9 and 3.6.13 will be using v10. Other versions of Juju are not accommodated for. They can be added in a follow-up or discarded.

As an alternative approach JIMM could only support the latest major version of the description package, the only downside here is that the destination controller must be at a specific version range. In keeping with Juju's paradigm of target-controller-version >= source-controller-version, we add a bit more complexity to JIMM to allow you to migrate to more versions of Juju.

During the import we use the target controller's version to determine the correct version of the `description` package to use. For this approach to be valid we need to ensure we keep track of the controller version in case it undergoes an upgrade. We already had some logic for this but I've updated it to make sure the version is updated and added a test for this.

As a final note, regardless of the solution taken, when a new major version of the description package is released, JIMM still needs to be updated to import it. This normally happens when we import a new version of Juju.

Fixes [JUJU-8914](https://warthogs.atlassian.net/browse/JUJU-8914)

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

[JUJU-8914]: https://warthogs.atlassian.net/browse/JUJU-8914?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/jimm/juju/migrationtarget.go
```diff
@@ -10,11 +10,12 @@ import (
 	"strings"
 	"time"
 
-	"github.com/juju/description/v9"
+	"github.com/canonical/jimm/v3/internal/description"
 	jujucrossmodel "github.com/juju/juju/core/crossmodel"
 	coremigration "github.com/juju/juju/core/migration"
 	"github.com/juju/juju/environs/config"
 	"github.com/juju/juju/rpc/params"
```
luci1900: Huh, it's complaining about duplicate imports.

kian99: Fixed

luci1900 (approved):
Ugly workaround, but I suppose so is this whole thing.

Should be fine.

SimoneDutto (approved):
this pr seems a lot of fiddling around. Good job.

Is it possible to add a test to migrate a model from 3.6.9 to 3.6.12 in the ci? Just to make sure all of this work

kian99:
> this pr seems a lot of fiddling around. Good job.
> 
> Is it possible to add a test to migrate a model from 3.6.9 to 3.6.12 in the ci? Just to make sure all of this work

We test migration from 3.6.12 -> 3.6.12 already when we test importing a model, since the controller attached to JIMM is bootstrapped with the same version as the external one. So I believe that covers the same thing.

PR #1758, opened 2025-12-05
feat: add internal description package

kian99:
## Description
This PR adds a new package at `internal/description` that wraps Juju `github.com/juju/description` package across major versions. This will allow model imports to work across a wider range of controller versions. This PR splits off the work from https://github.com/canonical/jimm/pull/1757 into a smaller change-set.

Partially fixes [JUJU-8914](https://warthogs.atlassian.net/browse/JUJU-8914)

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

[JUJU-8914]: https://warthogs.atlassian.net/browse/JUJU-8914?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/description/description.go
+++ b/internal/description/description.go
@@ -1,0 +1,128 @@
+// Copyright 2025 Canonical.
+
+// Package description provides version-agnostic wrappers around major versions
+// of `github.com/juju/description` which hold Juju model descriptions, to
+// facilitate migrations between different Juju versions.
+package description
+
+import (
+	"fmt"
+
+	"github.com/canonical/jimm/v3/internal/errors"
+	descriptionv10 "github.com/juju/description/v10"
+	descriptionv9 "github.com/juju/description/v9"
+	"github.com/juju/juju/environs/config"
+	"github.com/juju/names/v5"
+	"github.com/juju/version/v2"
+)
+
+const latestDescriptionVersion = 10
+
+// CloudCredentialArgs holds the arguments needed to set a cloud credential.
+type CloudCredentialArgs struct {
+	Owner      names.UserTag
+	Cloud      names.CloudTag
+	Name       string
+	AuthType   string
+	Attributes map[string]string
+}
+
+// Model provides a version-agnostic interface for accessing
+// model description data during migrations.
+type Model interface {
+	Owner() names.UserTag
+	Users() []User
+	CloudRegion() string
+	Applications() []Application
+	SetOwner(owner names.UserTag)
+	ClearUsers()
+	CloudCredential() CloudCredential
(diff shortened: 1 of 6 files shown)


luci1900 (approved):
Unfortunate, but necessary. Looks fine, good test.

ale8k (approved):
LGTM, but I don't think it should be in internal/description. It's only here because we want to switch versions of what is a publicly exported package. So shouldn't it be in /pkg really? Or something like /mypackagecopieslol

kian99:
> LGTM, but I don't think it should be in internal/description. It's only here because we want to switch versions of what is a publicly exported package. So shouldn't it be in /pkg really? Or something like /mypackagecopieslol

I'm not sure I follow. I don't believe we're looking to expose this package to others. Even the stuff in `pkg` gets mirrored out to https://github.com/canonical/jimm-go-sdk for clients like Terraform, etc. They won't need this either.

ale8k:
> > LGTM, but I don't think it should be in internal/description. It's only here because we want to switch versions of what is a publicly exported package. So shouldn't it be in /pkg really? Or something like /mypackagecopieslol
> 
> I'm not sure I follow. I don't believe we're looking to expose this package to others. Even the stuff in `pkg` gets mirrored out to https://github.com/canonical/jimm-go-sdk for clients like Terraform, etc. They won't need this either.

I more meant we only bring it in due to multiple import issues. It isn't some internal logic to JIMM?

kian99:
> It isn't some internal logic to JIMM?

Isn't it? In any case, I'd be more inclined to have it in `internal` and only moving to public consumption when there's a need, that's easier than doing it the other way around.

internal/description/description.go
```diff
@@ -0,0 +1,128 @@
...
+
+// Package description provides version-agnostic wrappers around major versions
+// of `github.com/juju/description` which hold Juju model descriptions, to
+// facilitate migrations between different Juju versions.
+package description
+
+import (
+	"fmt"
+
+	"github.com/canonical/jimm/v3/internal/errors"
```
alesstimec, line 11: sort the imports, please

alesstimec (approved):

internal/description/description.go
kian99: I'll do that in https://github.com/canonical/jimm/pull/1759 to avoid re-running the CI.

luci1900:
FWIW, I think everything should be in `internal` until proven otherwise.

PR #1759, opened 2025-12-05
fix: find offers not-found error

### LINKED ISSUE #990: Client error creating offers under JAAS
### Description

When the provider is configured to talk to JAAS, I'm seeing client errors when creating offers:

```
╷
│ Error: Client Error
│
│   with juju_offer.postgresql_juju_offer,
│   on main.tf line 1, in resource "juju_offer" "postgresql_juju_offer":
│    1: resource "juju_offer" "postgresql_juju_offer" {
│
│ Unable to create offer, got error: model "795798e4-922f-49c7-9169-004ffc17df90@serviceaccount/dbaas-is-jaas-testing-ps7" not found (not found) (not found)
╵
```

This does actually create the offer successfully, and I'm able to import it afterwards

```
jsimpso@desktop:~/tmp/terraform $ juju offers
Offer       User  Relation ID  Status  Endpoint  Interface  Role  Ingress subnets
postgresql  -
```

```
juju_offer.postgresql_juju_offer: Importing from ID "795798e4-922f-49c7-9169-004ffc17df90@serviceaccount/dbaas-is-jaas-testing-ps7.postgresql"...
juju_offer.postgresql_juju_offer: Import prepared!
  Prepared juju_offer for import
juju_offer.postgresql_juju_offer: Refreshing state... [id=795798e4-922f-49c7-9169-004ffc17df90@serviceaccount/dbaas-is-jaas-testing-ps7.postgresql]
ephemeral.vault_kv_secret_v2.ps7_jaas_credentials: Closing...
ephemeral.vault_kv_secret_v2.ps7_jaas_credentials: Closing complete after 0s

Import successful!

The resources that were imported are shown above. These resources are now in
your Terraform state and will henceforth be managed by Terraform.
```

### Urgency

Blocker for our release

### Terraform Juju Provider version

0.23.1,1.0.0

### Terraform version

v1.12.2

### Juju version

3.6.12

### Terraform Configuration(s)

```terraform
resource "juju_offer" "postgresql_juju_offer" {
  model            = "set this when testing <1.0"
  model_uuid       = "set this when testing >=1.0"
  application_name = "postgresql"
  endpoints        = ["database"]
}
```

### Reproduce / Test

```shell
terraform init
terraform apply
juju offers
terraform apply
terraform import juju_offer.postgresql_juju_offer {import value per docs}
terraform plan
```

### Debug/Panic Output

```shell

```

### Notes & References

_No response_

kian99:
## Description
This PR fixes https://github.com/juju/terraform-provider-juju/issues/990.

The issue raised is that the Terraform client is calling `FindApplicationOffer` and passing the model name + model owner + offer name. In JIMM's implementation of `FindApplicationOffer` we dial each controller and call `FindApplicationOffer` on each, passing those original query parameters, acting as an aggregator. 

The problem comes in because when Juju receives the filter parameters and the model owner and model name fields are set, a `NotFound` error is returned if the model doesn't exist. This causes the issue because aggregating across multiple controllers, the model can only exist on one of them.

The solution is either to avoid failing the aggregation process when 1 controller fails, for any reasons, or catch the `NotFound` error. The former solution seems attractive at first since it seems like a bad idea to return an error if we fail to dial only 1 controller but if the offer was hosted on that controller and we skip it because it is down, we might give the client the impression that the offer doesn't exist which can trigger reconciliation loops in tools like ArgoCD/Flux. So I've gone with the latter approach to capture the `NotFound` error. I've added a test and tested manually.

Fixes [JUJU-8913](https://warthogs.atlassian.net/browse/JUJU-8913)

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

[JUJU-8913]: https://warthogs.atlassian.net/browse/JUJU-8913?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/jimm/juju/applicationoffer.go
+++ b/internal/jimm/juju/applicationoffer.go
@@ -516,6 +516,9 @@ func (j *JujuManager) queryControllersForOffers(ctx context.Context, user *openf
 
 	for _, ctl := range controllers {
 		eg.Go(func() error {
+			// Return early if a single controller has an error
+			// to avoid misleading clients about what exists which
+			// could cause unneeded reconciliation.
 			api, err := j.dial(ctx, ctl, names.ModelTag{}, nil)
 			if err != nil {
 				return errors.E(err)
@@ -523,6 +526,9 @@ func (j *JujuManager) queryControllersForOffers(ctx context.Context, user *openf
 			defer api.Close()
 			controllerOffers, err := query(api)
 			if err != nil {
+				if errors.ErrorCode(err) == errors.CodeNotFound {
+					return nil
+				}
 				return errors.E(err)
 			}
 			for _, offer := range controllerOffers {
--- a/internal/jimm/juju/applicationoffer_test.go
+++ b/internal/jimm/juju/applicationoffer_test.go
@@ -2151,3 +2151,101 @@ func TestRevokeOfferAccessOnController(t *testing.T) {
 		}
 	}
 }
+
+const offerNotFoundTestEnv = `clouds:
+- name: test-cloud
+  type: test-provider
+  regions:
+  - name: test-cloud-region
+cloud-credentials:
+- owner: alice@canonical.com
+  name: cred-1
+  cloud: test-cloud
+controllers:
+- name: controller-1
+  uuid: 00000001-0000-0000-0000-000000000001
+  cloud: test-cloud
+  region: test-cloud-region
+- name: controller-2
+  uuid: 00000001-0000-0000-0000-000000000002
+  cloud: test-cloud
+  region: test-cloud-region
+models:
+- name: model-1
+  uuid: 00000002-0000-0000-0000-000000000001
+  controller: controller-2
+  cloud: test-cloud
+  region: test-cloud-region
+  cloud-credential: cred-1
+  owner: bob@canonical.com
+  life: alive
+  users:
+  - user: bob@canonical.com
+    access: admin
+application-offers:
+- name: offer-1
+  url: test-offer-url
+  uuid: 00000012-0000-0000-0000-000000000001
+  model-name: model-1
+  model-owner: bob@canonical.com
+  application-name: application-1
+  application-description: app description 1
+`
+
+func TestFindApplicationOffers_MultipleControllers(t *testing.T) {
+	c := qt.New(t)
+	ctx := context.Background()
+
+	// Define expected offer from the "good" controller
+	// Details copy the offer defined in the test environment
+	expectedOffer := jujuparams.ApplicationOfferAdminDetailsV5{
+		ApplicationOfferDetailsV5: jujuparams.ApplicationOfferDetailsV5{
+			OfferUUID: "00000012-0000-0000-0000-000000000001",
+			OfferURL:  "test-offer-url",
+			OfferName: "offer-1",
+		},
+	}
+
+	controller1Dialed := false
+	// Setup dialers for two controllers
+	dialers := jimmtest.DialerMap{
+		"controller-1": &jimmtest.Dialer{
+			API: &jimmtest.API{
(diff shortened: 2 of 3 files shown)


luci1900 (approved):
Looks good.

As follow-up, we should consider peeking into the filter and potentially replacing the errgroup with something that returns an array of results.

internal/jimm/juju/applicationoffer.go
```diff
@@ -516,13 +516,19 @@ func (j *JujuManager) queryControllersForOffers(ctx context.Context, user *openf
...
+			// to avoid misleading clients about what exists which
+			// could cause unneeded reconciliation.
 			api, err := j.dial(ctx, ctl, names.ModelTag{}, nil)
 			if err != nil {
 				return errors.E(err)
 			}
 			defer api.Close()
 			controllerOffers, err := query(api)
 			if err != nil {
+				if errors.ErrorCode(err) == errors.CodeNotFound {
```
alesstimec, line 529: a more "correct" or "flexible" way would be to give the called the ability to specify which errors are "ok"

internal/jimm/juju/applicationoffer_test.go
```diff
@@ -2151,3 +2151,101 @@ func TestRevokeOfferAccessOnController(t *testing.T) {
...
+- name: offer-1
+  url: test-offer-url
+  uuid: 00000012-0000-0000-0000-000000000001
+  model-name: model-1
+  model-owner: bob@canonical.com
+  application-name: application-1
+  application-description: app description 1
+`
+
+func TestFindApplicationOffers_MultipleControllers(t *testing.T) {
```
alesstimec, line 2195: a simpler test would be if we tried querying the existing controller for offers in a model that does not exist..

alesstimec (approved):
LGTM with a few comments

internal/jimm/juju/applicationoffer.go
kian99: I don't _think_ it needs to be so generic, but I see your point.

internal/jimm/juju/applicationoffer_test.go
kian99: Ah that's true. I guess I'll leave it for now since it's already like this.

SimoneDutto:
Commenting on this super late, maybe you've already discuss it in a standup.

I think the error group is fundamentally wrong here, what do we gain from returning the first error when querying multiple controllers for an offer? I see zero usecases where this is a feature, all use cases where this is a genuine issue (ex. one of the controller is not available, it is bugged, doesn't have the offer, etc)

kian99:
> Commenting on this super late, maybe you've already discuss it in a standup.
> 
> I think the error group is fundamentally wrong here, what do we gain from returning the first error when querying multiple controllers for an offer? I see zero usecases where this is a feature, all use cases where this is a genuine issue (ex. one of the controller is not available, it is bugged, doesn't have the offer, etc)

I left a comment in the change. If you skip controllers that return an error (maybe the controller is down), you risk returning a NotFound error. That can cause tools downstream to think the infra was modified and try to recreate it.

SimoneDutto:
> > Commenting on this super late, maybe you've already discuss it in a standup.
> > I think the error group is fundamentally wrong here, what do we gain from returning the first error when querying multiple controllers for an offer? I see zero usecases where this is a feature, all use cases where this is a genuine issue (ex. one of the controller is not available, it is bugged, doesn't have the offer, etc)
> 
> I left a comment in the change. If you skip controllers that return an error (maybe the controller is down), you risk returning a NotFound error. That can cause tools downstream to think the infra was modified and try to recreate it.

I think the correct approach would be:
- all controller NotFound -> NotFound
- one controller found the offer -> return the offer
- all controllers other error -> something like 503 service not available

kian99:
Yeah it could be more optimistic and keep trying to find it.

PR #1760, opened 2025-12-08
fix: remove dry-run from model upgrade

ale8k:
## Description
Increases the retry to a few minutes, sometimes it can run on that long waiting for the machine to become ready leading to inconsistent results.

It was also only dry running the upgrades. 

https://asciinema.org/a/b1tcSrSM4ZhyD4qNBBb7cQhgw 

Video of it is linked above.

## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests

### DIFF
--- a/internal/jimm/upgrade/upgrade.go
+++ b/internal/jimm/upgrade/upgrade.go
@@ -277,12 +277,7 @@ func (u *upgradeManager) MigrateAndUpgradeModel(ctx context.Context, user *openf
 		return controllerChosenVersion, errors.E(fmt.Errorf("failed to dial target controller: %w", err))
 	}
 
-	zapctx.Debug(ctx, "upgrading model agent after migration")
-
-	controllerChosenVersion, err = u.upgradeModel(ctx, api, modelUUID, targetVersion, true)
-	if err != nil {
-		return controllerChosenVersion, errors.E(fmt.Errorf("failed to dry-run upgrade model after migration: %w", err))
-	}
+	zapctx.Debug(ctx, "upgrading model agent after migration", zap.String("model-uuid", mi.Name))
 
 	controllerChosenVersion, err = u.upgradeModel(ctx, api, modelUUID, targetVersion, false)
 	if err != nil {
--- a/internal/jimm/upgrade/upgrade.go
+++ b/internal/jimm/upgrade/upgrade.go
@@ -212,8 +212,8 @@ func (u *upgradeManager) MigrateAndUpgradeModel(ctx context.Context, user *openf
 	var iimResult jujuparams.InitiateMigrationResult
 	if err := retry.Call(
 		retry.CallArgs{
-			Attempts: 10,
-			Delay:    5 * time.Second,
+			Attempts: 30,
+			Delay:    10 * time.Second,
 			// We could consider all errors fatal, bar: target prechecks failed: machine 0 not running (pending)
 			// IsFatalError:
 			Func: func() error {
@@ -241,8 +241,8 @@ func (u *upgradeManager) MigrateAndUpgradeModel(ctx context.Context, user *openf
 	var mi *jujuparams.ModelInfo
 	if err := retry.Call(
 		retry.CallArgs{
-			Attempts: 10,
-			Delay:    5 * time.Second,
+			Attempts: 30,
+			Delay:    10 * time.Second,
 			Func: func() error {
 				mi, err = u.jujuManager.ModelInfo(ctx, user, mt)
 				if err != nil {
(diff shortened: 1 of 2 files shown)


internal/jimm/upgrade/upgrade.go
```diff
@@ -277,21 +277,16 @@ func (u *upgradeManager) MigrateAndUpgradeModel(ctx context.Context, user *openf
...
-		// the upgrade-model command with the --reset-previous-upgrade option.
-		// We can't do that here, so we just return an error. We could possible indicate to the user here
-		// that manual intervention is required.
-	}
-	if jujuerrors.Is(upgradeErr, jujuerrors.AlreadyExists) {
-		upgradeErr = AlreadyUpgradedError
+	zapctx.Debug(ctx, "upgrading model agent after migration")
+
+	controllerChosenVersion, err = u.upgradeModel(ctx, api, modelUUID, targetVersion, true)
+	if err != nil {
```
kian99: What's the point of doing this? If it were to fail during the dry-run steps it'll fail similarly when we run it without the dry-run.

ale8k: If we fail on the upgrade, the only way to revert appears to be that flag I mentioned in the comment in the last PR. And currently, there's no manual intervention a user can do - so it's just to be on the safe side really. It also returns the chosen controller version - there's no validation right now as to whether that's the same as target, but we could also error if it doesn't match.

kian99: What I mean is, since we're not able to take advantage of the dry-run for the upgrade. Why not just remove this until we can take advantage of it.

internal/jimm/upgrade/upgrade.go
```diff
@@ -277,21 +277,16 @@ func (u *upgradeManager) MigrateAndUpgradeModel(ctx context.Context, user *openf
...
-	//nolint:staticcheck // Has a description to highlight this possibility.
-	if jujuparams.IsCodeUpgradeInProgress(upgradeErr) {
-		// Apparently upgrades can have issues, that can be manually resolved, then you can run
-		// the upgrade-model command with the --reset-previous-upgrade option.
-		// We can't do that here, so we just return an error. We could possible indicate to the user here
-		// that manual intervention is required.
-	}
-	if jujuerrors.Is(upgradeErr, jujuerrors.AlreadyExists) {
-		upgradeErr = AlreadyUpgradedError
+	zapctx.Debug(ctx, "upgrading model agent after migration")
```
kian99: Add the model name here

internal/jimm/upgrade/upgrade.go
```diff
@@ -351,3 +346,28 @@ func (u *upgradeManager) UpgradeTo(ctx context.Context, user *openfga.User, mode
 
 	return chosenVersion, nil
 }
+
+func (u *upgradeManager) upgradeModel(ctx context.Context, api juju.API, modelUUID string, targetVersion version.Number, dryRun bool) (version.Number, error) {
+	var upgradeErr error
+	var controllerChosenVersion version.Number
+
+	controllerChosenVersion, upgradeErr = api.UpgradeModel(modelUUID, targetVersion, "", false, dryRun)
```
kian99, line 349:
```suggestion
	controllerChosenVersion, upgradeErr := api.UpgradeModel(modelUUID, targetVersion, "", false, dryRun)
```

internal/jimm/upgrade/upgrade.go
ale8k: I get you, but I was more seeing phase 1 as a skeleton of sorts. So that when we come to do phase 2+ - we have some reminders here and there (same as the TODOs I left into the UpgradeTo method).

kian99 (approved):
I think we should delete the step where we set dry-run to `true` since it's not useful in this context.

internal/jimm/upgrade/upgrade.go
```diff
@@ -212,8 +212,8 @@ func (u *upgradeManager) MigrateAndUpgradeModel(ctx context.Context, user *openf
 	var iimResult jujuparams.InitiateMigrationResult
 	if err := retry.Call(
 		retry.CallArgs{
-			Attempts: 10,
-			Delay:    5 * time.Second,
+			Attempts: 30,
```
luci1900, line 215: That's an awful lot of attempts. Is it necessary even with the increased timeout?

ale8k: It might not be... But I wanted to just be safe really. I didn't add backoff as I just wanted it to simply retry a bunch every 10 seconds for 5 minutes.

kian99:
@ale8k Please update the PR title btw

internal/jimm/upgrade/upgrade.go
luci1900: I'd prefer we have a comment here to revisit this decision, even if we go with this for now for the sake of the demo.

luci1900 (approved):
Fine except for the minor comment.

PR #1761, opened 2025-12-08
feat: implement bare-minimum model get

ale8k:
## Description
Juju's show-model sources the agent-version for the controller from ModelGet. When not implemented, it causes the Juju CLI to error out detailing ModelGet wasn't implemented like so:
```
juju client not compatible with server: no such request - method ModelConfig.ModelGet is not implemented (not implemented)
re-install your juju client to match the version running on the controller
```

This now populates the version set in show-controller with JIMM's VersionInfo.Version.
And displays it like so:
```
jimm-dev:
  details:
    controller-uuid: 3217dbc9-8ea9-4381-9e97-01eab0b3f6bb
    api-endpoints: ['jimm.localhost:443']
    cloud: ""
    agent-version: 3.6.12
    agent-git-commit: unknown git commit
    controller-model-version: unknown version <- this was the offending line.
```

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

## Test instructions
Simply run show-controller.

### DIFF
--- a/internal/jujuapi/modelconfig.go
+++ b/internal/jujuapi/modelconfig.go
@@ -1,0 +1,31 @@
+// Copyright 2025 Canonical.
+
+package jujuapi
+
+import (
+	"context"
+
+	"github.com/canonical/jimm/v3/internal/jujuapi/rpc"
+	jimmversion "github.com/canonical/jimm/v3/version"
+	"github.com/juju/juju/rpc/params"
+)
+
+func init() {
+	facadeInit["ModelConfig"] = func(r *controllerRoot) []int {
+		modelGetMethod := rpc.Method(r.ModelGet)
+
+		r.AddMethod("ModelConfig", 3, "ModelGet", modelGetMethod)
+		return []int{3}
+	}
+}
+
+func (r *controllerRoot) ModelGet(ctx context.Context) (params.ModelConfigResults, error) {
+	return params.ModelConfigResults{
+		Config: map[string]params.ConfigValue{
+			"agent-version": {
+				Value:  jimmversion.VersionInfo.Version,
+				Source: "jimm",
+			},
+		},
+	}, nil
+}
--- a/internal/jujuapi/modelconfig.go
+++ b/internal/jujuapi/modelconfig.go
@@ -24,7 +24,7 @@ func (r *controllerRoot) ModelGet(ctx context.Context) (params.ModelConfigResult
 	return params.ModelConfigResults{
 		Config: map[string]params.ConfigValue{
 			"agent-version": {
-				Value:  jimmversion.JIMM_CONTROLLER_VERSION,
+				Value:  jimmversion.ControllerVersion,
 				Source: "jimm",
 			},
 		},
--- a/internal/jujuapi/modelconfig.go
+++ b/internal/jujuapi/modelconfig.go
@@ -20,7 +20,10 @@ func init() {
 	}
 }
 
-// ModelGet returns the model configuration.
+// ModelGet returns the model configuration in JIMM's case, this is used to return the "controller" model config.
+// As JIMM doesn't have a controller model, or even an agent for that matter, we simulate this.
+//
+// It is required because the CLI reports on the agent-version during a show-controller call.
 func (r *controllerRoot) ModelGet(ctx context.Context) (params.ModelConfigResults, error) {
 	return params.ModelConfigResults{
 		Config: map[string]params.ConfigValue{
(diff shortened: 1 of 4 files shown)


internal/jujuapi/modelconfig.go
```diff
@@ -0,0 +1,31 @@
...
+		r.AddMethod("ModelConfig", 3, "ModelGet", modelGetMethod)
+		return []int{3}
+	}
+}
+
+func (r *controllerRoot) ModelGet(ctx context.Context) (params.ModelConfigResults, error) {
+	return params.ModelConfigResults{
+		Config: map[string]params.ConfigValue{
+			"agent-version": {
+				Value:  jimmversion.VersionInfo.Version,
```
kian99: This will display jimm's version rather than the version of the Juju controller we claim to be.
I.e. if the previous output was,
```
    agent-version: 3.6.12
    agent-git-commit: unknown git commit
    controller-model-version: unknown version
```
When I run `juju show-controller` against a regular Juju controller I get,
```
    agent-version: 3.6.12
    agent-git-commit: d8a727fd219fea140f915a24ac267cc7b64e751a
    controller-model-version: 3.6.12
```
Where the agent-version and controller-model-version match.
With this change, the versions will be something like,
```
```
    agent-version: 3.6.12
    agent-git-commit: unknown git commit
    controller-model-version: 3.3.5 <- JIMM's version
```
I don't know what the right answer is, but we could just return the Juju ("3.6.12") version instead.

**Review by ale8k — COMMENTED** [review]
> This will display jimm's version rather than the version of the Juju controller we claim to be.
> I.e. if the previous output was,
> ```
>     agent-version: 3.6.12
>     agent-git-commit: unknown git commit
>     controller-model-version: unknown version
> ```
> When I run `juju show-controller` against a regular Juju controller I get,
> ```
>     agent-version: 3.6.12
>     agent-git-commit: d8a727fd219fea140f915a24ac267cc7b64e751a
>     controller-model-version: 3.6.12
> ```
> Where the agent-version and controller-model-version match.
> With this change, the versions will be something like,
> ```
> ```
>     agent-version: 3.6.12
>     agent-git-commit: unknown git commit
>     controller-model-version: 3.3.5 <- JIMM's version
> ```
> I don't know what the right answer is, but we could just return the Juju ("3.6.12") version instead. 

I don't really know either. I just made a presumption here that we'd use JIMM's version. They'd need to have a --target-controller to pick a controller behind JIMM. So I think for now, this is probably ok?

File: internal/jujuapi/modelconfig.go (hunk 1 above)
Comment (luci1900) [reply internal/jujuapi/modelconfig.go]: I was actually going to ask what version we should be returning. I don't think that's entirely clear to me, the argument could be made that it should be the protocol version of the proxy regardless of which controller it's proxying to.

Comment (ale8k) [reply internal/jujuapi/modelconfig.go]: I've responded below explaining why trying to pick the target controller's version *might* be problematic.

Comment (kian99) [reply internal/jujuapi/modelconfig.go]: In another method, where we need to return the controller version, we return the lowest version of controller connected to JIMM. I.e. 2 controllers connected to JIMM v3.6.9 and v3.6.12, we return v3.6.9

File: version/version.go [hunk 2]
```diff
@@ -4,6 +4,11 @@ package version
 
 import "strings"
 
+const (
+	// JIMM claims to be a 3.6.12 controller and client.
+	JIMM_CONTROLLER_VERSION = "3.6.12"
```
kian99: Could we skip the all-caps and also change this to  something like `ReportedJujuVersion` or `ControllerVersion`.
AI suggested `ReportedAgentVersion` based on the prompt,
>  What would be a good name for a variable that indicates the version of Juju that JIMM pretends to be?

kian99 (approved):

luci1900 (approved):
Looks good.

Since we control it explicitly, it's our promise that we act like that particular version. At least in theory.

internal/jujuapi/modelconfig.go
```diff
@@ -0,0 +1,32 @@
...
+func init() {
+	facadeInit["ModelConfig"] = func(r *controllerRoot) []int {
+		modelGetMethod := rpc.Method(r.ModelGet)
+
+		r.AddMethod("ModelConfig", 3, "ModelGet", modelGetMethod)
+		return []int{3}
+	}
+}
+
+func (r *controllerRoot) ModelGet(ctx context.Context) (params.ModelConfigResults, error) {
```
alesstimec, line 27: godoc, please

version/version.go
alesstimec: +inf on not using all caps!

alesstimec (approved):
LGTM modulo two comments

PR #1762, opened 2025-12-09
ci: use hostname -I first entry for login-token-refresh-url qa-microk8s script

ale8k:
## Description
Use hostname -I for the login-token-refresh URL in qa-microk8s. This should work for mosto f our developers, enabling them to run qa-microk8s consistently - given the developer is connected to a home network / has some kind of DHCP as their primary network interface (eth0 / whatever).


kian99 (commented):
If you're having trouble because address `10.0.1.1` is not available, you're probably just missing the host-access addon, see https://canonical.com/microk8s/docs/addon-host-access, try running `sudo microk8s enable host-access`

ale8k:
If an issue arises again this is an alternative and it's in our PR history. Closing.

PR #1764, opened 2025-12-09
feat: modify jimm ListControllers to list controllers users have can_addmodel permission on

ale8k:
## Description
This PR modifies the ListControllers method of JIMM to list only controllers the user has can_addmodel on.

For admin users, it returns all controllers.

I modified it further in that we no longer return the JIMM controller, wasn't really sure there's a point of returning JIMM in the set and it was only returned when the user wasn't an admin anyway.

I added an additional test to Juju API, which might appear kind of pointless, but it at least tests that when calling JIMM's API it does behave as expected - even if we are returning dummy database entries.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

### DIFF
--- a/cmd/jaas/cmd/listcontrollers_test.go
+++ b/cmd/jaas/cmd/listcontrollers_test.go
@@ -47,21 +47,6 @@ var (
     data: {}
     since: null
 `
-
-	expectedOutput = `- name: jaas
-  uuid: 6acf4fd8-32d6-49ea-b4eb-dcb9d1590c11
-  publicaddress: ""
-  apiaddresses: \[\]
-  cacertificate: ""
-  cloudtag: ""
-  cloudregion: ""
-  agentversion: .*
-  status:
-    status: available
-    info: ""
-    data: {}
-    since: null
-`
 )
 
 type listControllersSuite struct {
@@ -87,5 +72,5 @@ func (s *listControllersSuite) TestListControllers(c *gc.C) {
 	bClient := s.SetupCLIAccess(c, "bob")
 	context, err := cmdtesting.RunCommand(c, cmd.NewListControllersCommandForTesting(s.ClientStore(), bClient))
 	c.Assert(err, gc.IsNil)
-	c.Assert(cmdtesting.Stdout(context), gc.Matches, expectedOutput)
+	c.Assert(cmdtesting.Stdout(context), gc.Matches, `\[\]\n?`)
 }
--- a/internal/jimm/juju/jimm.go
+++ b/internal/jimm/juju/jimm.go
@@ -76,7 +76,7 @@ func (j *JujuManager) ListControllers(ctx context.Context, user *openfga.User) (
 			zapctx.Error(ctx, "error checking user permissions for controller", zap.String("controller", c.Name), zap.Error(err))
 			return nil
 		}
-		if user.JimmAdmin || canAddModel {
+		if canAddModel {
 			controllers = append(controllers, *c)
 		}
 		return nil
--- a/internal/jimm/juju/jimm.go
+++ b/internal/jimm/juju/jimm.go
@@ -12,6 +12,8 @@ import (
 	jujuparams "github.com/juju/juju/rpc/params"
 	"github.com/juju/names/v5"
 	"github.com/juju/version/v2"
+	"github.com/juju/zaputil/zapctx"
+	"go.uber.org/zap"
 	"golang.org/x/sync/errgroup"
 
 	"github.com/canonical/jimm/v3/internal/db"
@@ -64,15 +66,19 @@ func (j *JujuManager) ControllerInfo(ctx context.Context, name string) (*dbmodel
 	return &ctl, nil
 }
 
-// ListControllers returns a list of controllers the user has access to.
+// ListControllers returns a list of controllers the user has can_addmodel to.
+// JIMM admins get all controllers.
 func (j *JujuManager) ListControllers(ctx context.Context, user *openfga.User) ([]dbmodel.Controller, error) {
-	if !user.JimmAdmin {
-		return nil, errors.E(errors.CodeUnauthorized, "unauthorized")
-	}
-
 	var controllers []dbmodel.Controller
 	err := j.Database.ForEachController(ctx, func(c *dbmodel.Controller) error {
-		controllers = append(controllers, *c)
+		canAddModel, err := user.IsAllowedAddModelToController(ctx, c.ResourceTag())
+		if err != nil {
+			zapctx.Error(ctx, "error checking user permissions for controller", zap.String("controller", c.Name), zap.Error(err))
+			return nil
+		}
+		if user.JimmAdmin || canAddModel {
+			controllers = append(controllers, *c)
+		}
 		return nil
 	})
 	if err != nil {
--- a/internal/jimm/juju/jimm_test.go
+++ b/internal/jimm/juju/jimm_test.go
@@ -39,24 +39,27 @@ controllers:
   cloud: test
   region: test-region-1
   agent-version: 3.2.1
+  users:
+    - username: bob@canonical.com
+      access: add-model
+    - username: eve@canonical.com
+      access: add-model
 - name: test2
   uuid: 00000001-0000-0000-0000-000000000002
   cloud: test
   region: test-region-2
   agent-version: 3.2.0
+  users:
+    - username: bob@canonical.com
+      access: add-model
 - name: test3
   uuid: 00000001-0000-0000-0000-000000000003
   cloud: test
   region: test-region-3
   agent-version: 2.1.0
-users:
-- username: alice@canonical.com
-  controller-access: superuser
-- username: bob@canonical.com
-  can-add-model: [00000001-0000-0000-0000-000000000001, 00000001-0000-0000-0000-000000000002]
-- username: eve@canonical.com
-  can-add-model: [00000001-0000-0000-0000-000000000001, 00000001-0000-0000-0000-000000000003]
-- username: notallowedanycontrollers@canonical.com
+  users:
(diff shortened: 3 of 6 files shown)


internal/jimm/juju/jimm.go
```diff
@@ -64,15 +66,19 @@ func (j *JujuManager) ControllerInfo(ctx context.Context, name string) (*dbmodel
...
-
 	var controllers []dbmodel.Controller
 	err := j.Database.ForEachController(ctx, func(c *dbmodel.Controller) error {
-		controllers = append(controllers, *c)
+		canAddModel, err := user.IsAllowedAddModelToController(ctx, c.ResourceTag())
+		if err != nil {
+			zapctx.Error(ctx, "error checking user permissions for controller", zap.String("controller", c.Name), zap.Error(err))
+			return nil
+		}
+		if user.JimmAdmin || canAddModel {
```
kian99: You can remove the `user.JimmAdmin` that will get checked in the `IsAllowedAddModel` call

internal/testutils/jimmtest/env.go
```diff
@@ -127,10 +127,27 @@ func (e *Environment) User(name string) *User {
...
+			}
+
+			err := openfgaUser.SetControllerAccess(
+				context.Background(),
+				names.NewControllerTag(controllerUUID),
+				ofganames.CanAddModelRelation,
+			)
+			c.Assert(err, qt.IsNil)
+		}
+	}
```
kian99: You can delete this as the permissions for the resource are normally defined on the resource itself rather than on the user.

kian99: See in this PR (not merged yet) https://github.com/canonical/jimm/pull/1745/files#diff-4cc4a751474a3a0cd8bbdd149ded07c916c37c380880bfdc2487f6a827688cb7
Then you can do the following in the yaml for the env,
```
  users:
  - user: alice@canonical.com
    access: add-model
```
This just matches the pattern elsewhere.

internal/jimm/juju/jimm.go
ale8k: Ahh yeah, good shout!

internal/testutils/jimmtest/env.go
ale8k: Yeah sure. Although perhaps we should be explicit and call it controller-access: can_addmodel and be really explicit?

ale8k: Actually nvm think it's fine.

Would've preferred it on the user really but it is ok.

kian99: I've finally merged https://github.com/canonical/jimm/pull/1745 so you should have the new stuff without any issue

cmd/jaas/cmd/listcontrollers_test.go
```diff
@@ -87,5 +72,5 @@ func (s *listControllersSuite) TestListControllers(c *gc.C) {
 	bClient := s.SetupCLIAccess(c, "bob")
 	context, err := cmdtesting.RunCommand(c, cmd.NewListControllersCommandForTesting(s.ClientStore(), bClient))
 	c.Assert(err, gc.IsNil)
-	c.Assert(cmdtesting.Stdout(context), gc.Matches, expectedOutput)
+	c.Assert(cmdtesting.Stdout(context), gc.Matches, `\[\]\n?`)
```
kian99, line 75: This is a good example of a cli test we can delete (feel free to delete now but in the future is fine too).

internal/jimm/juju/jimm.go
```diff
@@ -64,15 +66,19 @@ func (j *JujuManager) ControllerInfo(ctx context.Context, name string) (*dbmodel
 	return &ctl, nil
 }
 
-// ListControllers returns a list of controllers the user has access to.
+// ListControllers returns a list of controllers the user has can_addmodel to.
+// JIMM admins get all controllers.
```
kian99, line 70: > JIMM admins get all controllers.

This bit is probably not necessary since the code isn't actually doing anything special for admins.

kian99 (approved):
Lgtm, ty!

internal/jimm/juju/jimm.go
```diff
@@ -64,15 +66,19 @@ func (j *JujuManager) ControllerInfo(ctx context.Context, name string) (*dbmodel
...
 func (j *JujuManager) ListControllers(ctx context.Context, user *openfga.User) ([]dbmodel.Controller, error) {
-	if !user.JimmAdmin {
-		return nil, errors.E(errors.CodeUnauthorized, "unauthorized")
-	}
-
 	var controllers []dbmodel.Controller
 	err := j.Database.ForEachController(ctx, func(c *dbmodel.Controller) error {
```
SimoneDutto, line 73: in JIMM we have a `func (j *JujuManager) ForEachUserModel` which is basically iterating plus checking permissions, do you think it's worth to recreate the same pattern for controllers as well? Otherwise, things like `JujuManager.FindApplicationOffers` will always iterate over all controllers skipping this permission check?

cmd/jaas/cmd/listcontrollers_test.go
ale8k: Yeah I thought so too. I'll just leave it as is for now though.

internal/jimm/juju/jimm.go
ale8k: I just like it being clear is all really.

internal/jimm/juju/jimm.go
ale8k: Yeah I like that idea. We can have a ForEachUserController. I'll add it now!

kian99: Fwiw I don't think ForEachUserX is a great pattern, it doesn't tell you what permission is needed. For `ForEachUserModel` it returns all models the user has read/write/admin access on but you can't limit to models where the user has write access. Controllers face a slightly similar problem since we have `admin`, `can_addmodel`, and `audit_log_viewer` so depending on what you're after, the list of controllers will change.

SimoneDutto (approved):
we can go with whatever, this is fine as well

internal/jimm/juju/jimm.go
SimoneDutto: the only issue i see, if we forgetting that we are adding a new permission and we return the list of all controllers inadvertentently

ale8k: As discussed, for another time. Kinda hard to do without generics and our J struct wouldn't work for this, and a huge func signature isn't ideal either. So we'd end up with very specific functions.

alesstimec: If we are going to continue the ForEachX pattern, then we should, perhaps, reconsider the implementation.. currently it fetches all X from the database then consults OpenFGA to see if users have access.. i propose we do it the other way around - first check in OpenFGA all the Xs user has access to, then fetch them from the database.. wdyt?

internal/jimm/juju/jimm.go
```diff
@@ -64,15 +66,19 @@ func (j *JujuManager) ControllerInfo(ctx context.Context, name string) (*dbmodel
...
-		controllers = append(controllers, *c)
+		canAddModel, err := user.IsAllowedAddModelToController(ctx, c.ResourceTag())
```
alesstimec, line 74: i will assume this returns true if everybody has this permission on the controller

internal/jimm/juju/jimm_test.go
```diff
@@ -92,30 +96,49 @@ func TestListControllers(t *testing.T) {
...
 		expectedError       string
-	}{{
-		about:     "superuser can list controllers",
-		user:      env.User("alice@canonical.com").DBObject(c, j.Database),
-		jimmAdmin: true,
-		expectedControllers: []dbmodel.Controller{
-			env.Controller("test1").DBObject(c, j.Database),
-			env.Controller("test2").DBObject(c, j.Database),
-			env.Controller("test3").DBObject(c, j.Database),
+	}{
```
alesstimec, line 102: can we add a test to check the user has this permission if `user:*` (everybody) has this permission

alesstimec (changes requested):
changes look good, but i'm still not 100% OK with this change.. it is a bit conceptual change - JIMM no longer presents itself as one controller, but instead exposes the fact that there are multiple controllers behind it..

has this change been QAd with Juju CLI?

internal/jimm/juju/jimm.go
ale8k: Yep:
`define can_addmodel: [user, user:*, group#member, role#assignee] or administrator`

internal/jimm/juju/jimm.go
ale8k: I think it makes more sense what you're suggesting, but isn't it just putting the load on OpenFGA vs putting the load on our DB? And probably a bit more load on OpenFGA, which in turn is putting load on our DB, very likely with a more complex query?

ale8k:
> changes look good, but i'm still not 100% OK with this change.. it is a bit conceptual change - JIMM no longer presents itself as one controller, but instead exposes the fact that there are multiple controllers behind it..
> 
> has this change been QAd with Juju CLI?

I haven't QA'd with a controller no... I will.

internal/jimm/juju/jimm.go
kian99: > first check in OpenFGA all the Xs user has access to, then fetch them from the database

+1 on Alex's point, I believe listing of all resources a user has access to with OpenFGA is a lot more complex than with the DB. OpenFGA is then better suited to access checks. If we wanted to optimise this (since it's currently N calls to OpenFGA where N=number of resources), we could fetch all from the DB and do a bulk check to OpenFGA. But definitely a future optimisation imo.

alesstimec (approved):

PR #1766, opened 2025-12-10
doc: update add-permission command doc

kian99:
## Description
This PR updates some outdated help text on the `add-permission` command on the jaas cli.

The easiest way to cross-reference these changes is to look at our tag resolver code at `internal/jimm/permissions/tagresolver.go` and at our OpenFGA auth model at `openfga/authorisation_model.fga`.

### DIFF
--- a/cmd/jaas/cmd/permission.go
+++ b/cmd/jaas/cmd/permission.go
@@ -41,6 +41,10 @@ These are used to define access control between resources.
 The object and target object must be of the form <tag>-<objectname> or <tag>-<object-uuid>
 E.g. "user-Alice" or "controller-MyController"
 
+Certain reserved tags exist to denote specific resource types:
+- The user-everyone@external tag represents all users.
+- The controller-jimm tag represents the JIMM controller itself.
+
 -f    Read from a file where filename is the location of a JSON encoded file of the form:
     [
         {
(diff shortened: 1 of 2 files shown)


cmd/jaas/cmd/permission.go
```diff
@@ -56,21 +56,28 @@ E.g. "user-Alice" or "controller-MyController"
...
     member
 
+If target_object is a role, the relation can only be:
+
+	assignee
+
 If target_object is a controller, the relation can be one of:
 
-    loginer
+    audit_log_viewer
```
alesstimec: we should mention this relation only makes sense with the JIMM controller, but will have no effects for others

cmd/jaas/cmd/permission.go
```diff
@@ -79,17 +86,26 @@ If target_object is a model, the relation can be one of:
...
 
-Additionally, if the object is a group, a userset can be applied by adding #member as follows.
+If the object is a group, a userset must be applied by adding #member as follows.
 This will grant/revoke access to all users within TeamA:
 
     group-TeamA#member administrator controller-MyController
+
+Similarly if the object is a role, a userset must be applied by adding #member as follows.
+
+	role-Auditor#member audit_log_viewer controller-MyController
```
alesstimec: i don't think this is correct.. it should be
```
role-Auditor#assignee audit_log_viewer controller-MyController
```

alesstimec (changes requested):

ale8k (approved):
Really? We were missing this? LGTM

cmd/jaas/cmd/permission.go
kian99: Ah yes, I thought the same initially and then read the wrong bit in the OpenFGA model. Fixing it now.

alesstimec (approved):

PR #1767, opened 2025-12-11
test(migrate-e2e): migrate jimmSuite to websocketE2Esuite

SimoneDutto:
# Description

In this pr i migrate the jimmSuite to the new websocketE2ESuite.

To do so we need to do some adjustments:
- modelName and cloudName are not hardcoded but generated with `petname`, to make sure we can then parallelize these tests in the future
- change the `controllers.yaml` file to contain all ip addresses for a controller
- change the assert func to check for controller info to check all the expected addrs are present in the response. We need a custom assert function because JIMM is setting controller addresses from the response from the Juju controller, which contains 127.0.0.1 and :: we want to exclude from the comparison.

### DIFF
--- a/internal/jujuapi/jimm_unit_test.go
+++ b/internal/jujuapi/jimm_unit_test.go
@@ -4,16 +4,21 @@ package jujuapi_test
 
 import (
 	"context"
+	"testing"
+	"time"
 
+	qt "github.com/frankban/quicktest"
 	"github.com/google/uuid"
 	jujuparams "github.com/juju/juju/rpc/params"
 	"github.com/juju/names/v5"
 	gc "gopkg.in/check.v1"
 
+	"github.com/canonical/jimm/v3/internal/db"
 	"github.com/canonical/jimm/v3/internal/dbmodel"
 	"github.com/canonical/jimm/v3/internal/errors"
 	"github.com/canonical/jimm/v3/internal/jimm"
 	"github.com/canonical/jimm/v3/internal/jimm/bootstrap"
+	"github.com/canonical/jimm/v3/internal/jujuapi"
 	"github.com/canonical/jimm/v3/internal/openfga"
 	"github.com/canonical/jimm/v3/internal/testutils/jimmtest"
 	"github.com/canonical/jimm/v3/internal/testutils/jimmtest/mocks"
@@ -322,4 +327,66 @@ func (s *jimmUnitTestSuite) TestStartDestroyControllerJob(c *gc.C) {
 	ctrlInfo.Models = append(ctrlInfo.Models, dbmodel.Model{})
 	_, err = root.StartDestroyControllerJob(ctx, req)
 	c.Assert(err, gc.ErrorMatches, "cannot destroy controller with models")
+}
+
+// TestAuditLogAPIParamsConversion tests the conversion of API params to a AuditLogFilter struct.
+// Note that this test doesn't require a running Juju/JIMM controller so it doesn't use gc + the jimmSuite.
+func TestAuditLogAPIParamsConversion(t *testing.T) {
+	c := qt.New(t)
+	testCases := []struct {
+		about   string
+		request apiparams.FindAuditEventsRequest
+		result  db.AuditLogFilter
+		err     error
+	}{
+		{
+			about: "Test basic conversion",
+			request: apiparams.FindAuditEventsRequest{
+				After:    "2023-08-14T00:00:00Z",
+				Before:   "2023-08-14T00:00:00Z",
+				UserTag:  "user-alice",
+				Model:    "123",
+				Method:   "Deploy",
+				Offset:   10,
+				Limit:    10,
+				SortTime: false,
+			},
+			result: db.AuditLogFilter{
+				Start:       time.Date(2023, 8, 14, 0, 0, 0, 0, time.UTC),
+				End:         time.Date(2023, 8, 14, 0, 0, 0, 0, time.UTC),
+				IdentityTag: "user-alice",
+				Model:       "123",
+				Method:      "Deploy",
+				Offset:      10,
+				Limit:       10,
+				SortTime:    false,
+			},
+		}, {
+			about: "Test limit lower bound",
+			request: apiparams.FindAuditEventsRequest{
+				Limit: 0,
+			},
+			result: db.AuditLogFilter{
+				Limit: jujuapi.AuditLogDefaultLimit,
+			},
+		}, {
+			about: "Test limit upper bound",
+			request: apiparams.FindAuditEventsRequest{
+				Limit: jujuapi.AuditLogUpperLimit + 1,
+			},
+			result: db.AuditLogFilter{
+				Limit: jujuapi.AuditLogUpperLimit,
+			},
+		},
+	}
+	for _, test := range testCases {
+		c.Log(test.about)
+		res, err := jujuapi.AuditParamsToFilter(test.request)
+		if test.err == nil {
+			c.Assert(err, qt.IsNil)
+			c.Assert(res, qt.DeepEquals, test.result)
+		} else {
+			c.Assert(err, qt.ErrorMatches, test.err)
+		}
+	}
 }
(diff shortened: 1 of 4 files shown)


testing/jimm_test.go
```diff
@@ -449,81 +458,26 @@ func (s *jimmSuite) TestAuditLog(c *gc.C) {
...
 	defer conn.Close()
 	client := api.NewClient(conn)
 	evs, err := client.FindAuditEvents(&apiparams.FindAuditEventsRequest{Method: "Deploy"})
 	c.Assert(err, gc.Equals, nil)
 	c.Assert(len(evs.Events), gc.Equals, 0)
 }
 
-// TestAuditLogAPIParamsConversion tests the conversion of API params to a AuditLogFilter struct.
-// Note that this test doesn't require a running Juju/JIMM controller so it doesn't use gc + the jimmSuite.
-func TestAuditLogAPIParamsConversion(t *testing.T) {
```
SimoneDutto, line 527: moved to `jimm_unit_test.go`

internal/jujuapi/jimm_unit_test.go
```diff
@@ -4,16 +4,21 @@ package jujuapi_test
 
 import (
 	"context"
+	"testing"
+	"time"
 
+	qt "github.com/frankban/quicktest"
```
luci1900, line 10: Took me a minute to figure out this has nothing to do with https://hackage.haskell.org/package/QuickCheck

testing/jimm_test.go
```diff
@@ -140,41 +143,43 @@ func (s *jimmSuite) TestAddControllerPublicAddressWithoutPort(c *gc.C) {
...
 	c.Assert(err, gc.Equals, nil)
-	c.Assert(ci, jc.DeepEquals, apiparams.ControllerInfo{
-		Name:          "controller-2",
-		UUID:          info.ControllerUUID,
-		APIAddresses:  info.Addrs,
-		CACertificate: info.CACert,
-		CloudTag:      names.NewCloudTag(jimmtest.TestCloudName).String(),
-		CloudRegion:   jimmtest.TestCloudRegionName,
-		AgentVersion:  s.Model.Controller.AgentVersion,
+	ciExpected := apiparams.ControllerInfo{
```
luci1900, line 247: This is a slight readability improvement, imo.

internal/testutils/jimmtest/e2e_suite.go
```diff
@@ -80,10 +80,22 @@ func GetControllersConfig(c *gc.C) *ControllersConfig {
 	return &config
 }
 
+// GetOneControllerConfig retrieves one controller configuration
+// from the controllers config file. It can be used in tests when
+// a valid controller config is needed.
+func (s *E2ESuite) GetOneControllerConfig(c *gc.C) (string, ControllerInfo) {
```
luci1900, line 86: I'd say `GetFirstControllerConfig`.

luci1900 (approved):
Looks good, random names are nice.

Just one nitpick on method name.

internal/testutils/jimmtest/e2e_suite.go
kian99: First could also be tricky since we're iterating over a map (I think?) so the order would be random.

testing/jimm_test.go
```diff
@@ -102,32 +103,34 @@ func (s *jimmSuite) TestListControllersUnauthorized(c *gc.C) {
...
+
 	tests := []struct {
 		req           apiparams.AddControllerRequest
 		expectedError string
 	}{{
 		req: apiparams.AddControllerRequest{
 			Name:          "controller-2",
 			PublicAddress: "controller.test.com",
-			CACertificate: s.APIInfo(c).CACert,
+			CACertificate: conf.ToAPIInfo().CACert,
```
kian99, line 202: Can't we just keep the method signature the same and put it on `jimmSuite` then there will be less work to change future suites.

kian99 (commented):
Looks good but I think we can make some tweaks to have less LOC going forward. Lmk

internal/testutils/jimmtest/e2e_suite.go
SimoneDutto: yeah, it's a random one because ranging on a map is random. And it is for the better because when we add multiple controllers these tests should pass regardless of the one it's picked

testing/jimm_test.go
SimoneDutto: the only issue i see with this is that there is no such a thing as `s.APIINfo(c)` because we need to pick a random controller first to extract the APIInfo out of.

So it would be something like `s.GetOneApiInfo` which is a change nonetheless

kian99: But you can just rename it afterwards. Keep it called `ApiInfo`, do the same thing as `GetOneApiInfo()` and leave a note in the docstring to rename it later.

SimoneDutto: thinking about it, you can't.

Because there are other examples where you do:
`conf.ToAPIInfo().CACert` and afterwards `conf.ToAPIInfo().Addrs`

if we create a s.APIInfo(c) which randomly pick one it wouldn't work.

So in anycase this needs changing

internal/testutils/jimmtest/e2e_suite.go
luci1900: I was trying to figure out if it deliberately randomises on iteration or if it's just bucket/insertion order. If it's not stably ordered, it'd be nice to make it deliberately random ourselves.

luci1900: Merely "not guaranteed to be ordered" could mean "often the same order" https://go.dev/blog/maps#iteration-order

SimoneDutto: yeah, but i wouldn't bother too much. It doesn't matter, it can pick always the same or always different ones, the test should never fail because of it

testing/jimm_test.go
kian99: I don't follow, isn't `GetOneControllerConfig` randomly picking one? Surely we can do the same thing as before, i.e. the functions assume only 1 controller exists attached to JIMM then we can add a new function like `s.ToSpecificAPiInfo()` later when we need it?

SimoneDutto: i'll try to explain the best as i can:
there is a test that was previously like this:
```
c.Check(ci, jc.DeepEquals, apiparams.ControllerInfo{
		Name:          "controller-1",
		UUID:          s.Model.Controller.UUID,
		APIAddresses:  s.APIInfo(c).Addrs,
		CACertificate: s.APIInfo(c).CACert,
		CloudTag:      names.NewCloudTag(jimmtest.TestCloudName).String(),
		CloudRegion:   jimmtest.TestCloudRegionName,
		AgentVersion:  s.Model.Controller.AgentVersion,
		Status: jujuparams.EntityStatus{
			Status: "available",
		},
	})
```

if we create:
```
func (suite) APIInfo {
   // randomly pick one from the conf and returns the APIInfo for it 
}
```

then this test would be wrong, because the `APIAddresses` and `CaCertificate` can potentially belong to different confs.

`s.APIInfo` before was working because it was assuming always a single controller was associated to the `ConnSuite` but this assumption is wrong now.
And i think it's more correct to get the conf from a random controller and then convert it to `APIInfo` after, even if it is 3 LOC changes.

Also, there are only 2 other occurencies of s.APIINfo in the `jujuapi` pkg we are migrating.
image

kian99: I think I understand a bit better, so `s.APIInfo()` would return a different controller each time, which means you might get different controllers when it calls,
```
APIAddresses:  s.APIInfo(c).Addrs,
CACertificate: s.APIInfo(c).CACert,
```
But you can fix this behind the scenes by picking one controller at the start of the test/suite and always returning that from s.APIInfo. Wdyt?

kian99: I'd just like to see as few LOC changed while we migrate

SimoneDutto: yeah, i guess i could keep the "one controller" on the struct. However:
- i don't see the point, these changes are already in and only other 2 are expected
- if i pick it per test once we parallelize the suite we need to handle this per test randomization
- if i pick it per suite we run all the tests against the same controller, which is not bad, but it's nice to have the randomization

kian99: Resolving, I didn't there were not much more uses of `s.APIInfo(c)`

kian99 (approved):

ale8k (approved):
lgtm after reviewing the existing comments.

PR #1768, opened 2025-12-11
test(migrate): migrate apiSuite and userManagerSuite

SimoneDutto:


testing/api_test.go
```diff
@@ -0,0 +1,42 @@
...
+	c.Assert(err, gc.Equals, nil)
+	u := url.URL{
+		Scheme: "wss",
+		Host:   serverURL.Host,
+		Path:   fmt.Sprintf("/models/%s/commands", s.Model.UUID.String),
+	}
+	dial := websocket.DefaultDialer
+	dial.TLSClientConfig = &tls.Config{
+		InsecureSkipVerify: true,
+	}
```
SimoneDutto, line 35: we are just testing that that endpoint is not found, so i think it's fair to skip tls

luci1900: Sure, but I think you need the lint silencing comment.

luci1900 (approved):

testing/usermanager_test.go
```diff
@@ -1,24 +1,25 @@
...
 type userManagerSuite struct {
-	websocketSuite
+	jimmtest.WebsocketE2ESuite
 }
 
 var _ = gc.Suite(&userManagerSuite{})
 
 func (s *userManagerSuite) TestAddUser(c *gc.C) {
-	conn := s.open(c, nil, "alice")
+	conn := s.Open(c, nil, "alice", nil)
```
kian99, line 22: Kinda minor but I would love it if we had a suite inside of this `testing package` which wraps the `jimmtest.WebsocketE2ESuite` then `s.Open()` could stay as `s.open()` and the signature would stay the same. Then copying over everything becomes even easier and at the end we can do the refactorings with a single rename.

kian99 (approved):

testing/usermanager_test.go
SimoneDutto: i'll merge this and then create another pr so we can discuss it there

PR #1769, opened 2025-12-12
feat: new AddModelToController method

kian99:
## Description
This PR adds a new method called `AddModelToController` on the JIMM facade. It follows the same semantics as the `AddCloudToController` method.

I've added a unit and integration test. For the integration test there is only ever 1 controller attached to JIMM so it's not very effective but eventually when we have multiple controllers attached to jIMM during a test this will become more valuable.

Fixes [JUJU-8945](https://warthogs.atlassian.net/browse/JUJU-8945)

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

[JUJU-8945]: https://warthogs.atlassian.net/browse/JUJU-8945?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/jujuapi/jimm.go
+++ b/internal/jujuapi/jimm.go
@@ -182,6 +182,7 @@ func (r *controllerRoot) AddCloudToController(ctx context.Context, req apiparams
 	return nil
 }
 
+// AddModelToController adds a new model to a specific controller.
 func (r *controllerRoot) AddModelToController(ctx context.Context, req apiparams.AddModelToControllerRequest) (jujuparams.ModelInfo, error) {
 	mca, err := toAddModelArgs(req.ModelCreateArgs)
 	if err != nil {
--- a/internal/jujuapi/jimm.go
+++ b/internal/jujuapi/jimm.go
@@ -188,7 +188,7 @@ func (r *controllerRoot) AddModelToController(ctx context.Context, req apiparams
 		return jujuparams.ModelInfo{}, errors.E(err)
 	}
 	// Add JIMM specific field.
-	mca.TargetController = req.ControllerName
+	mca.ControllerName = req.ControllerName
 
 	info, err := r.jimm.JujuManager().AddModel(ctx, r.user, mca)
 	if err != nil {
--- a/internal/jujuapi/jimm.go
+++ b/internal/jujuapi/jimm.go
@@ -24,6 +24,7 @@ import (
 	"github.com/canonical/jimm/v3/internal/jimm/juju"
 	"github.com/canonical/jimm/v3/internal/jujuapi/rpc"
 	ofganames "github.com/canonical/jimm/v3/internal/openfga/names"
+	"github.com/canonical/jimm/v3/internal/servermon"
 	apiparams "github.com/canonical/jimm/v3/pkg/api/params"
 	"github.com/canonical/jimm/v3/version"
 )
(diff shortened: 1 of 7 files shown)


internal/jujuapi/jimm.go
```diff
@@ -179,6 +182,27 @@ func (r *controllerRoot) AddCloudToController(ctx context.Context, req apiparams
...
+	// Add JIMM specific field.
+	mca.TargetController = req.ControllerName
+
+	info, err := r.jimm.JujuManager().AddModel(ctx, r.user, mca)
+	if err != nil {
+		servermon.ModelsCreatedFailCount.Inc()
+		return jujuparams.ModelInfo{}, errors.E(err)
+	}
+
+	servermon.ModelsCreatedCount.Inc()
```
kian99, line 200: I was thinking to move these monitoring bits into the business logic layer since they are duplicated but I'm not sure it's particularly necessary so I've just left them duplicated here for now.

internal/jujuapi/jimm.go
```diff
@@ -179,6 +182,27 @@ func (r *controllerRoot) AddCloudToController(ctx context.Context, req apiparams
 	return nil
 }
 
+func (r *controllerRoot) AddModelToController(ctx context.Context, req apiparams.AddModelToControllerRequest) (jujuparams.ModelInfo, error) {
```
alesstimec, line 186: suggestion: let's just call it `CreateModel` to shadow the same method on the modelmanager facade, or `AddModel`..  and a godoc, please

internal/jujuapi/jimm.go
```diff
@@ -179,6 +182,27 @@ func (r *controllerRoot) AddCloudToController(ctx context.Context, req apiparams
...
+	mca, err := toAddModelArgs(req.ModelCreateArgs)
+	if err != nil {
+		return jujuparams.ModelInfo{}, errors.E(err)
+	}
+	// Add JIMM specific field.
+	mca.TargetController = req.ControllerName
```
alesstimec: it's a bit confusing that in wireformat type we call it `ControllerName`, but in internal type we call it `TargetController`

testing/modelmanager_test.go
```diff
@@ -428,6 +430,59 @@ func (s *modelE2EManagerSuite) TestCreateModel(c *gc.C) {
 	}
 }
 
+func (s *modelE2EManagerSuite) TestCreateModelOnTargetController(c *gc.C) {
```
alesstimec: this is not really a modelmanager facade test..  a specific jimm facade suite would be a better place for it

internal/jujuapi/jimm.go
luci1900: Won't let me add a shrug emoji.

internal/jujuapi/jimm.go
kian99: You wouldn't prefer it to be consistent with `AddCloudToController`? I don't mind either way but it seems there is some existing precedence.

internal/jujuapi/jimm.go
kian99: I'll change the internal type

luci1900 (approved):

internal/jujuapi/jimm.go
alesstimec: actually i'd prefer both to be consistent.. AddCloud and AddModel.. but we can't all get what we want and naming is hard.. so, yeah, fine with AddModelToController and AddCloudToController for now..

testing/jimm_test.go
```diff
@@ -1124,3 +1125,56 @@ func (s *jimmSuite) TestUpgradeTo_TargetVersionLower(c *gc.C) {
 	_, err := client.UpgradeTo(&req)
 	c.Assert(err, gc.ErrorMatches, `failed to run upgrade to: failed to prepare for upgrade: target version must be greater than or equal to current version \(bad request\)`)
 }
+
+func (s *modelE2EManagerSuite) TestCreateModelOnTargetController(c *gc.C) {
```
alesstimec: jimmSuite?

internal/jujuapi/jimm.go
alesstimec, line 186: godoc, please

alesstimec (approved):

PR #1771, opened 2025-12-15
test(migrate-suite): removed modelManagerStorageSuite and migrated caasModelManagerSuite

SimoneDutto:
## Description

Removed modelManagerStorageSuite because it wasn't testing any particular JIMM feature, and it would have required to deploy an application with storage (ex. postgres) with all the hassle of waiting for it to be ready, creating the storage etc

caasModelManagerSuite migrated.


internal/testutils/kubetest/kubetest.go
```diff
@@ -48,5 +50,24 @@ func NewFakeKubernetes(c *gc.C) *httptest.Server {
 			w.WriteHeader(http.StatusNotFound)
 		}
 	}))
+	l, err := net.Listen("tcp", fmt.Sprintf("%s:0", getOutboundIP()))
```
SimoneDutto: before it was enough to just listen on "127.0.0.1" because the juju in the connsuite was able to reach it without any problem.
But now that we have an actual controller in lxd, it wasn't working anymore

internal/testutils/kubetest/kubetest.go
```diff
@@ -48,5 +50,24 @@ func NewFakeKubernetes(c *gc.C) *httptest.Server {
...
+	srv.Start()
+
 	return srv
 }
+
+// getOutboundIP gets the preferred outbound IP of this machine.
+// This is needed because httptest.NewServer by default listens on localhost only,
+// which is not reachable from the Juju controller used by the testing harness.
+func getOutboundIP() string {
+	conn, err := net.Dial("udp", "8.8.8.8:80")
```
luci1900: Maybe we should use a Canonical property instead of Google.

testing/caasmodelmanager_test.go
```diff
@@ -0,0 +1,196 @@
...
+	s.WebsocketE2ESuite.SetUpTest(c)
+
+	s.ksrv = kubetest.NewFakeKubernetes(c)
+
+	conn := s.Open(c, nil, "bob", nil)
+	defer conn.Close()
+
+	cloudclient := cloudapi.NewClient(conn)
+	s.cloudName = petname.Generate(2, "-")
+	err := cloudclient.AddCloud(cloud.Cloud{
```
kian99: We should remove the cloud at the end of the test

internal/testutils/kubetest/kubetest.go
```diff
@@ -48,5 +50,24 @@ func NewFakeKubernetes(c *gc.C) *httptest.Server {
...
+	c.Assert(err, gc.IsNil)
+	srv.Listener = l
+	srv.Start()
+
 	return srv
 }
+
+// getOutboundIP gets the preferred outbound IP of this machine.
+// This is needed because httptest.NewServer by default listens on localhost only,
+// which is not reachable from the Juju controller used by the testing harness.
```
kian99: Why can't the test server listen on `0.0.0.0`?

SimoneDutto: because then the fakek8s reports "0.0.0.0" as the address the juju controller is trying to reach the k8s cluster API

internal/testutils/kubetest/kubetest.go
SimoneDutto: good idea

internal/testutils/kubetest/kubetest.go
kian99: Ah I think I misunderstood, we're just trying to find the host's address to tell the Juju controller to contact this fake K8s server. In which case this seems like an interesting way to go about it but if I'm understanding correctly, there might be an easier way. 

Since the Juju controller is in LXC, it can't reach services running on the host directly, this k8s server can then listen on broadcast `0.0.0.0` (I think it might even work listening on `localhost` but I'm not 100% sure) and then we just grab the IP from `ip addr show lxdbr0` and pass that as the cloud's endpoint below when we do `Endpoint:        s.ksrv.URL,`. 

What might be even easier, according to AI in some cases there is a special `host.lxd` address,
> If you’re using LXD (snap), you can often reach the host via:
>     host.lxd
> Example:
>     curl http://host.lxd:8080
> This only works if:
> LXD is managing the container
> the bridge DNS is enabled
> Nice when available, but not guaranteed.

SimoneDutto: i'll try if it works with the default config for lxd. Otherwise i'll keep my solution if it's ok

kian99: This should work https://serverfault.com/questions/1078399/how-do-i-refer-to-the-lxd-host-from-an-lxd-container tldr use `_gateway.lxd:<port>` I tested this locally and running a little hello-world server on `0.0.0.0:8085` locally I could `curl _gateway.lxd:8085` successfully from inside the LXC container.

SimoneDutto: it works!

internal/testutils/kubetest/kubetest.go
SimoneDutto: removed completely

testing/caasmodelmanager_test.go
SimoneDutto: this is an issue i still need to work on, because right now a model created in fake k8s cloud can't be deleted, because i think the fake k8s server doesn't really comply with all the things needed and the model gets stuck, so the cloud can't be deleted.

SimoneDutto: the model is stuck there, and can't even be deleted with `force`.

luci1900 (dismissed):
Looks good, barring the linter errors left.

.github/workflows/e2e-test.yaml
```diff
@@ -54,7 +66,10 @@ jobs:
 
       - name: Generate controller config
         run: CONTROLLER=test-e2e make generate-test-env
-
+      
+      # - name: Setup tmate session
+      #   uses: mxschmitt/action-tmate@v3
```
kian99: dd

kian99 (dismissed):

SimoneDutto:
converting it to a draft because it has some challenge i prefer to discuss once we've moved all the others

testing/caasmodelmanager_test.go
SimoneDutto, line 1: this should be a git mv from modelmanager_test.go plus some changes, but git is stupid when you change too many things

.github/workflows/e2e-test.yaml
```diff
@@ -43,7 +43,20 @@ jobs:
           sudo usermod -a -G lxd $USER && \
           sudo iptables -F FORWARD && \
           sudo iptables -P FORWARD ACCEPT
-
+
```
luci1900, line 46: Unintended whitespace?

testing/README.md
```diff
@@ -28,6 +28,30 @@ This guide outlines the steps to set up and run the e2e test suite.
...
    ```
 
+
+## Setup microk8s cloud
+
+Our tests run using an lxd backing controller against the lxd cloud, but there are suites requiring a microk8s cloud to test specific
+JAAS functionalities.
+
+After setting up microk8s we add the k8s cloud to the client:
+`juju add-k8s microk8s-cp --client`
```
luci1900, line 38: Why `-cp`? Maybe `-e2e` would be more obvious?

luci1900 (approved):
Looks good, I have mere nits to pick.

Tasty deleted code.

kian99 (approved):
Great stuff!

PR #1770, opened 2025-12-15
ci: grant all users with access to backing controller

kian99:
## Description
This PR fixes our failing dashboard integration tests by granting all users with add-model access to the backing controller. This preserves existing behavior for users of our setup-jimm action.

## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests
- [x] Covered by integration tests

PR #1773, opened 2025-12-15
chore: use the detect-jaas script to reliably execute JAAS commands

kian99:
## Description
This PR tweaks the changes in https://github.com/canonical/jimm/pull/1770 which modified our `test-server` action to grant all users with access to a backing JAAS controller by running `juju jaas ...`. As it turns out, this is not the right way to do it because depending on the environment in which the action runs, the Juju CLI may be available as a Snap or built locally (in cases where this action is running in the Juju repo and the Juju CLI is built from source).

The right thing to do is to utilise the `detect-jaas.sh` script which handles these intricacies.

### DIFF
--- a/.github/actions/test-server/action.yaml
+++ b/.github/actions/test-server/action.yaml
@@ -176,8 +176,11 @@ runs:
         CONTROLLER_NAME: ${{ steps.lxd-controller.outputs.name }}
 
     - name: Grant all users with add-model access to backing controller
-      run: juju jaas add-permission user-everyone@external can_addmodel controller-${{ steps.lxd-controller.outputs.name }}
+      run: ./local/jimm/grant-controller-access.sh
+      working-directory: ${{ github.action_path }}/../../..
       shell: bash
+      env:
+        CONTROLLER_NAME: ${{ steps.lxd-controller.outputs.name }}
 
     - name: Provide service account with cloud-credentials
       run: ./local/jimm/setup-service-account.sh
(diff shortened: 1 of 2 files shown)


local/jimm/grant-controller-access.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+# This script is used to grant all users with access to a controller
+# controller in JAAS, this is useful for CI where we want to allow
```
ale8k: controller controller

.github/actions/test-server/action.yaml
```diff
@@ -176,8 +176,11 @@ runs:
         CONTROLLER_NAME: ${{ steps.lxd-controller.outputs.name }}
 
     - name: Grant all users with add-model access to backing controller
-      run: juju jaas add-permission user-everyone@external can_addmodel controller-${{ steps.lxd-controller.outputs.name }}
+      run: ./local/jimm/grant-controller-access.sh
+      working-directory: ${{ github.action_path }}/../../..
```
ale8k, line 180: I think there's something like GitHub.root or whatever it is called we could use instead here right?

ale8k (approved):
Ty

luci1900 (approved):
Sure.

I think I dislike the CLI overlay thing even more.

.github/actions/test-server/action.yaml
kian99: Aha I'm not sure, I'll look into that. I know this works since we use it lower down.

alesstimec (approved):

PR #1775, opened 2025-12-17
test(migrate): migrate controllerSuite and watcherSuite

SimoneDutto:


luci1900 (approved):

alesstimec (commented):
can we please do `git mv` to move the file first, then modify it? would make it much easier to review just the changes needed for the new integration testing framework.

SimoneDutto:
> then modify it?

i usually do it but here i had to keep a test in the original file, and git didn't understand the file was just moved cause of it. I think

ale8k (approved):
lgtm, but with the removal of all watcher soon, could we remove our watcher code too? I.e., this test and the other bits?

SimoneDutto:
> lgtm, but with the removal of all watcher soon, could we remove our watcher code too? I.e., this test and the other bits?

soon :skull:

PR #1776, opened 2025-12-17
fix(import-model): fix for import-model where juju does not return models' cloud region

alesstimec:
## Description
 
In case of a k8s cluster Juju sometimes does not return a cloud region for a model (as there is only one default region) and we need to set a cloud region on a model when inserting it. To that end if and only if there is one region on the cloud (default) we select that one.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

### DIFF
--- a/internal/jimm/juju/controller.go
+++ b/internal/jimm/juju/controller.go
@@ -512,13 +512,23 @@ func (m *modelImporter) setModelCloud(ctx context.Context) error {
 		return err
 	}
 
-	cr := cloud.Region(m.modelInfo.CloudRegion)
-	if cr.Name != m.modelInfo.CloudRegion {
-		return errors.E("cloud region not found")
-	}
+	if m.modelInfo.CloudRegion != "" {
+		cr := cloud.Region(m.modelInfo.CloudRegion)
+		if cr.Name != m.modelInfo.CloudRegion {
+			return errors.E("cloud region not found")
+		}
 
-	m.model.CloudRegionID = cr.ID
-	m.model.CloudRegion = cr
+		m.model.CloudRegionID = cr.ID
+		m.model.CloudRegion = cr
+	} else {
+		if len(cloud.Regions) != 1 {
+			return errors.E("expected one cloud region, found %d", len(cloud.Regions))
+		}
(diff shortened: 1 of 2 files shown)


internal/jimm/juju/controller.go
```diff
@@ -512,13 +512,21 @@ func (m *modelImporter) setModelCloud(ctx context.Context) error {
...
 
-	m.model.CloudRegionID = cr.ID
-	m.model.CloudRegion = cr
+		m.model.CloudRegionID = cr.ID
+		m.model.CloudRegion = cr
+	} else {
+		if len(cloud.Regions) != 1 {
+			return errors.E("multiple regions found for cloud but model has no region specified")
+		}
+		m.model.CloudRegionID = cloud.Regions[0].ID
```
ale8k, line 529: I think this is worth a go doc, explaining why using 0 is ok (because there's literally only ever 1 and perhaps why there's only 1)

ale8k (approved):

internal/jimm/juju/controller.go
alesstimec: i commented on why we choose it, i cannot sanely assume to know why there is only one and why juju does not report it in the model info call

internal/jimm/juju/controller.go
```diff
@@ -512,13 +512,23 @@ func (m *modelImporter) setModelCloud(ctx context.Context) error {
...
+		if cr.Name != m.modelInfo.CloudRegion {
+			return errors.E("cloud region not found")
+		}
 
-	m.model.CloudRegionID = cr.ID
-	m.model.CloudRegion = cr
+		m.model.CloudRegionID = cr.ID
+		m.model.CloudRegion = cr
+	} else {
+		if len(cloud.Regions) != 1 {
```
SimoneDutto, line 524: if len(cloud.Regions) == 0 this code would panic

SimoneDutto: maybe this if should check also for 0, just to be defensive

SimoneDutto (approved):
looks good

PR #1777, opened 2025-12-17
feat(ofga client): use idempotent add/remove relations

SimoneDutto:
## Description

This new feature permits adding/removing tuples idempotently. We can safely use it for all the operations that we do in JAAS.

It is safe to call this method even if JIMM is connected to a <=1.10 openfga server because the flag to ignore duplicates/missing tuples would simply be ignored by the openfga server. So this update won't affect people using an old openfga charm, and will benefit people using the new charm.

The idempotency is needed, for example, in the case Terraform fails to add tuples mid-through, because now it can just re-add everything all together, before it would have error out on the duplicates.


ale8k (approved):

luci1900 (approved):

alesstimec (approved):

local/openfga/entrypoint.sh
```diff
@@ -19,7 +19,25 @@ psql -Atx "$OPENFGA_DATASTORE_URI" -c "DELETE FROM authorization_model;"
...
+
+# Openfga generates a random authorization_model_id when we add the model.
+# However, JIMM expects a specific hardcoded authorization_model_id to function correctly.
+# So, we replace the generated ID with our hardcoded ID in both the column and the serialized protobuf
+# within the database. 
+psql -Atx "$OPENFGA_DATASTORE_URI" <<-SQL
+  UPDATE authorization_model 
+  SET 
+    authorization_model_id = '01GP1EC038KHGB6JJ2XXXXCXKB',
+    serialized_protobuf = decode(
```
SimoneDutto, line 31: before openfga 1.10 the `serialized_protobuf` column didn't exist, and it was enough to just update the `authorization_model_id`.

Sadly the model-id is in the column and in the protobuf column as well, so we need to amend both

PR #1779, opened 2025-12-17
test(test): test

SimoneDutto:
### DRAFT

Just trying to understand where the regression is.


SimoneDutto:
Ok, so it's not the update to V1.10 per SE but the new update in ofga

PR #1780, opened 2025-12-17
test(migrate): migrate application offers suite

luci1900:
## Description
Migrate application offers tests to E2E suite. Change`DeployApplication` in the suite to take struct as parameters.

Switch action to run on hosted runner, since it was running out of disk space on the GitHub one.

## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests

### DIFF
--- a/.github/workflows/e2e-test.yaml
+++ b/.github/workflows/e2e-test.yaml
@@ -13,10 +13,13 @@ jobs:
   e2e-test:
     name: E2E Tests with External Controller
     timeout-minutes: 60
-    runs-on: ubuntu-22.04
+    runs-on: [self-hosted-linux-amd64-jammy-large]
     steps:
       - name: Checkout JIMM repo
         uses: actions/checkout@v4
+
+      - name: Set up Docker Compose
+        uses: docker/setup-compose-action@v1
 
       - name: Setup Go
         uses: actions/setup-go@v4
@@ -40,13 +43,13 @@ jobs:
           sudo usermod -a -G lxd $USER && \
           sudo iptables -F FORWARD && \
           sudo iptables -P FORWARD ACCEPT
-      
+
       - name: Generate certs
         working-directory: local/traefik/certs
         run: ./certs.sh
       
       - name: Start test environment
-        run: make test-env
+        run: docker compose up -d --wait
 
       - name: Bootstrap controller with JWKS endpoint
         run: |
--- a/internal/testutils/jimmtest/e2e_suite.go
+++ b/internal/testutils/jimmtest/e2e_suite.go
@@ -12,6 +12,7 @@ import (
 	"net/url"
 	"os"
 	"runtime"
+	"strings"
 	"sync"
 
 	petname "github.com/dustinkirkland/golang-petname"
@@ -78,6 +79,14 @@ func (s *E2ESuite) GetControllersConfig(c *gc.C) *ControllersConfig {
 	c.Assert(err, gc.IsNil, gc.Commentf("failed to parse controller config file"))
 
 	c.Assert(len(config.Controllers), gc.Not(gc.Equals), 0, gc.Commentf("no controllers found in config"))
+
+	// If cert ending newline was lost to YAML parsing, add it back
+	for name, info := range config.Controllers {
+		if !strings.HasPrefix(info.CACert, "\n") {
+			info.CACert += "\n"
+			config.Controllers[name] = info
+		}
+	}
 
 	return &config
 }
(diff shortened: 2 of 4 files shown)


testing/applicationoffers_test.go
```diff
@@ -10,127 +10,105 @@ import (
...
 
 func (s *applicationOffersSuite) TearDownTest(c *gc.C) {
-	s.endpoint = state.Endpoint{}
-	s.factory = nil
-	if s.state != nil {
-		s.state.Release()
-		s.state = nil
-	}
-	s.websocketSuite.TearDownTest(c)
+	s.endpointName = ""
```
SimoneDutto: do we need that?

testing/applicationoffers_test.go
```diff
@@ -10,127 +10,105 @@ import (
...
-	})
-	s.factory.MakeUnit(c, &factory.UnitParams{
-		Application: app,
-	})
-	s.endpoint, err = app.Endpoint("url")
-	c.Assert(err, gc.Equals, nil)
+	s.WebsocketE2ESuite.SetUpTest(c)
+
+	s.DeployApplication(c, s.AdminUser, s.Model.Tag(), "test-app", "juju-qa-dummy-sink")
+	s.endpointName = "source"
```
SimoneDutto: do we need this?

internal/testutils/jimmtest/e2e_suite.go
```diff
@@ -194,7 +194,7 @@ func (s *WebsocketE2ESuite) SetUpTest(c *gc.C) {
 	c.Assert(err, gc.Equals, nil)
 }
 
-func (s *WebsocketE2ESuite) DeployApplication(c *gc.C, user *openfga.User, modelTag names.Tag, appName string) {
+func (s *WebsocketE2ESuite) DeployApplication(c *gc.C, user *openfga.User, modelTag names.Tag, appName string, charmName string) {
```
SimoneDutto: i would go for a struct, so we can support multiple different options without changing signature in the future

.github/workflows/e2e-test.yaml
```diff
@@ -40,13 +43,13 @@ jobs:
...
           sudo iptables -P FORWARD ACCEPT
-      
+
       - name: Generate certs
         working-directory: local/traefik/certs
         run: ./certs.sh
       
       - name: Start test environment
-        run: make test-env
+        run: docker compose up -d --wait
```
SimoneDutto, line 52: why?

testing/applicationoffers_test.go
```diff
@@ -204,36 +182,35 @@ func (s *applicationOffersSuite) TestGetConsumeDetails(c *gc.C) {
 		ControllerInfo: &jujuparams.ExternalControllerInfo{
 			ControllerTag: s.Model.Controller.Tag().String(),
 			Addrs:         info.Addrs,
-			Alias:         "controller-1",
-			CACert:        caCert,
+			Alias:         s.Model.Controller.Name,
+			CACert:        info.CACert + "\n",
```
SimoneDutto: we need this?

testing/applicationoffers_test.go
luci1900: Not entirely, it could just be a constant. I'll do that.

.github/workflows/e2e-test.yaml
luci1900: `make test-env` runs `sys-deps`, which fails because of missing microk8s and juju-db.

testing/applicationoffers_test.go
luci1900: I don't know why, but the other side doesn't strip whitespace from around the cert. That may be a better solution instead, but I wasn't sure where to do that.

internal/testutils/jimmtest/e2e_suite.go
luci1900: I wasn't sure what should actually be in that struct. The one we pass onwards is a bit unwieldy.

I was going to just change it again when I look at https://github.com/canonical/jimm/blob/v3/internal/jujuapi/localcharm_test.go#L57

.github/workflows/e2e-test.yaml
SimoneDutto: mmm, yeah. That's annoying

testing/applicationoffers_test.go
SimoneDutto: strip them where we parse the yaml

internal/testutils/jimmtest/e2e_suite.go
SimoneDutto: pass the conf we suppor now, name, charmname, etc

luci1900: What about units? If I put it as a field on a struct, it'll default to `0` if not set. That's why the big struct we pass on has a pointer.

I could choose to set it to `1` if it's `0`. After all, it's just these tests and if you mean something else you might not use the method.

testing/applicationoffers_test.go
luci1900: Potentially, but that's inside `client.GetConsumeDetails`. I don't think I want to change that.

luci1900: Ok, I've tracked it down to here https://github.com/canonical/jimm/blob/v3/local/backing_test_controller/generate.sh#L50

The output of `juju show-controller` in YAML is ambiguous, where the newline inside a string field can be confused with the newline to signal a new field. The output in JSON however does contain a newline quite clearly.

I guess I can change the script to parse JSON instead of yaml.

SimoneDutto: as long as it works i'm ok with it

luci1900: It turns out the certs *should* end with a newline. It is however the YAML parsing that ends up eating them, since there is no practical way to express a trailing newline.

I have added a workaround where we parse the YAML for the test suite to add the trailing newline if it isn't there.

SimoneDutto (approved):

ale8k (approved):
lgtm!

alesstimec (approved):

PR #1788, opened 2026-01-05
chore(migrate): migrate cloudSuite

SimoneDutto:
## Description
migrate cloudSuite.

A few challenges:
- i needed to fetch the current cloud credential instead of creating a fake `user-pass` cred, that's because that cred is then used to create a model, and since this is an actual test, we need an actual credential, so you'll see something like:
```go
credentialTag := names.NewCloudCredentialTag(fmt.Sprintf("%s/test@canonical.com/%s", jimmtest.TestE2ECloudName, credentialName))
existingCloudCred := s.GetExistingClientCredentialsForCloud(c, jimmtest.TestE2ECloudName)
_, err := client.UpdateCredentialsCheckModels(credentialTag,
	cloud.NewCredential("certificate", existingCloudCred.Attributes),
)
```
- petname.Generate everywhere i saw it might be needed once we try to make these tests parallel in the future
- add a suite method to addCloud and cleanup


luci1900 (approved):
Looks reasonable.

kian99 (approved):

PR #1789, opened 2026-01-06
test(migrate): migrate local charm suite

luci1900:
## Description
Migrate local charm tests to E2E suite.

I had to redesign the second test a little, since I couldn't find a way to reach into resource-related internals otherwise.

## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests

### DIFF
--- a/testing/localcharm_test.go
+++ b/testing/localcharm_test.go
@@ -1,34 +1,34 @@
 // Copyright 2024 Canonical.
 
-package jujuapi_test
+package testing
 
 import (
 	"fmt"
 	"strings"
 
+	"github.com/canonical/jimm/v3/internal/testutils/jimmtest"
 	"github.com/juju/charm/v12"
 	"github.com/juju/charm/v12/resource"
 	"github.com/juju/juju/api"
 	"github.com/juju/juju/api/client/charms"
 	"github.com/juju/juju/api/client/resources"
 	"github.com/juju/juju/testcharms"
-	"github.com/juju/juju/testing/factory"
 	"github.com/juju/version/v2"
 	gc "gopkg.in/check.v1"
 )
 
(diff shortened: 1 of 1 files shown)


testing/localcharm_test.go
```diff
@@ -43,35 +43,42 @@ func (s *localCharmSuite) TestLocalCharmDeploy(c *gc.C) {
...
+	c.Assert(err, gc.IsNil)
+
+	appName := "test-app"
+
 	uploadClient, err := resources.NewClient(conn)
 	c.Assert(err, gc.IsNil)
 
-	// test
-	err = uploadClient.Upload(app.Name(), "test", "file", pendingId, strings.NewReader("<data>"))
+	pendingIDs, err := uploadClient.AddPendingResources(resources.AddPendingResourcesArgs{
```
ale8k, line 67: Not you, but it'd be good to see the ids aren't pending anymore. Or do we realise that from the no error returned in  uploadClient.Upload()?

ale8k (approved):
LGTM, I don't really know enough in this area to be 100% confident.  But generally this looks like what I'd expect if we're just testing the resource endpoint(s now right?).

luci1900:
> LGTM, I don't really know enough in this area to be 100% confident. But generally this looks like what I'd expect if we're just testing the resource endpoint(s now right?).

From what @SimoneDutto was saying, the test is primarily for ensuring HTTP proxying works too, besides the WebSocket proxying we generally do.

testing/localcharm_test.go
luci1900: Not sure how that would be done, I don't see an API for listing pending resources.

Ultimately, yes, the success of the overall operation should be sufficient for proving the black box is behaving as expected.

kian99 (approved):

PR #1790, opened 2026-01-07
test: tweak upgrade-to test

kian99:
## Description
This PR modifies the upgrade test. It avoids installing a fixed version of the Juju snap to bootstrap an old agent version and instead relies on the bootstrap command's `--agent-version` flag.

The test bootstraps a controller with a fixed agent version of 3.6.11, creates a model on that controller, and then attempts to upgrade it. It verifies the upgrade is complete by polling so that when the API becomes async minimal changes to the test will be required.

In order to create the model on the correct controller we use a JAAS command that is currently WIP i.e. `juju jaas add-model --target-controller`. Once this functionality lands this test should be ready.

One hurdle faced here was how the bootstrapped controller (done during `upgrade-to` execution) would connect to `jimm.localhost`. In the integration test workflow we the login-token-refresh-url to `"http://${{ env.MACHINE_IP }}:17070/.well-known/jwks.json"` where `MACHINE_IP` is the address of the default network interface on the device. This routes traffic from containers to the host to the JIMM container exposed alongside Traefik without TLS and makes the testing much easier. We may want to, in the future, consider cloning the controller with all its config (controller and model config).

On a final note, this test is quite slow because it is bootstrapping 2 controllers. It is worth a dicussion on how to improve this. Some ideas include,
- Speed up bootstrap
- Run the integration test workflow periodically (maybe daily) and allow a label on PRs to trigger it.

Fixes [JUJU-8938](https://warthogs.atlassian.net/browse/JUJU-8938)
 
## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests

[JUJU-8938]: https://warthogs.atlassian.net/browse/JUJU-8938?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/local/jimm/setup-controller.sh
+++ b/local/jimm/setup-controller.sh
@@ -33,6 +33,10 @@ else
   echo "Skipping connecting the controller to JIMM"
 fi
 
+if [[ -n "${AGENT_VERSION:-}" ]]; then
+  BOOTSTRAP_ARGS+=(--agent-version "${AGENT_VERSION}")
+fi
+
 echo "Bootstrapping controller"
 JUJU_DEV_FEATURE_FLAGS=ssh-jump juju bootstrap "${BOOTSTRAP_ARGS[@]}"
 rm "$CLOUDINIT_FILE"
--- a/tests/upgrades/upgrade.sh
+++ b/tests/upgrades/upgrade.sh
@@ -1,54 +1,76 @@
 #!/bin/bash
 
-# This test upgrades a single model juju controller from one version to another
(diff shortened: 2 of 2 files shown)


luci1900 (approved):

alesstimec (approved):
LGTM one the `juju jaas add-model` command lands

PR #1793, opened 2026-01-07
feat(cmd): add-model command to add a model to a specific controller

alesstimec:
## Description

Introduces the add-model jaas command enabling users to add new models to specific JAAS controllers.

## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests


cmd/jaas/cmd/addmodel.go
```diff
@@ -0,0 +1,617 @@
...
+	"github.com/juju/names/v5"
+
+	jimmapi "github.com/canonical/jimm/v3/pkg/api"
+	jimmapiparams "github.com/canonical/jimm/v3/pkg/api/params"
+)
+
+// NewAddModelCommand returns a command to add a model.
+func NewAddModelCommand() cmd.Command {
+	command := &addModelCommand{
+		providerRegistry: environs.GlobalProviderRegistry(),
```
kian99: What does this do?

cmd/jaas/cmd/addmodel.go
```diff
@@ -0,0 +1,617 @@
...
+	messageArgs := []interface{}{c.Name}
+
+	details := jujuclient.ModelDetails{
+		ModelUUID: model.UUID,
+		ModelType: coremodel.ModelType(model.Type),
+	}
+	if featureflag.Enabled(feature.Branches) || featureflag.Enabled(feature.Generations) {
+		// Default target is the master branch.
+		details.ActiveBranch = coremodel.GenerationMaster
+	}
```
kian99: Not sure what this feature flag is for, maybe we should drop this for our CLI?

cmd/jaas/cmd/addmodel.go
```diff
@@ -0,0 +1,617 @@
...
+this controller manages more than one cloud.
+Please specify which cloud/region to use:
+
+    juju add-model [options] <model-name> cloud[/region]
+`[1:]
+	}
+	return errors.Errorf("%s\nThe clouds/regions supported by this controller are:\n\n%s", prefix, buf.String())
+}
+
+func maybeGetControllerCloud(cloudClient *cloudapi.Client) (names.CloudTag, jujucloud.Cloud, error) {
```
kian99: This logic is probably unnecessary since JIMM's cloud has no bearing on where to deploy models.

cmd/jaas/cmd/addmodel_test.go
```diff
@@ -0,0 +1,164 @@
...
+	gc "gopkg.in/check.v1"
+
+	"github.com/canonical/jimm/v3/cmd/jaas/cmd"
+	"github.com/canonical/jimm/v3/internal/dbmodel"
+	"github.com/canonical/jimm/v3/internal/testutils/cmdtest"
+	"github.com/canonical/jimm/v3/internal/testutils/jimmtest"
+)
+
+type addModelSuite struct {
+	cmdtest.JimmCmdSuite
```
kian99: Since we're already going to make all the jaas cli tests unit based, can you change this so that these tests don't rely on `cmdtest.JimmCmdSuite` since that relies on the connSuite. Have a look at the bootstrap command to see its tests.

cmd/jaas/cmd/addmodel.go
kian99: One thing it makes me think about - we could do Jimm's `ListControllers` to show what controllers are available if something goes wrong but I also think that's overloading what this command does so I probably wouldn't add it, at least not now.

kian99 (commented):
Mostly lgtm, there's a lot of the Juju code that is hard to say is all needed but seems mostly fine.
Not approving yet because of the tests and whether we can refactor those to be unit tests.

cmd/jaas/cmd/addmodel.go
SimoneDutto, line 1: how do we track changes in this file in juju we need to bring to JAAS? Do we have something in place?

cmd/jaas/cmd/addmodel.go
```diff
@@ -0,0 +1,617 @@
...
+	f.StringVar(&c.CredentialName, "credential", "", "Specify the credential to be used by the model")
+	f.Var(&c.Config, "config", "Specify the path to a YAML model configuration file or individual configuration options (`--config config.yaml [--config key=value ...]`)")
+	f.BoolVar(&c.noSwitch, "no-switch", false, "Choose not to switch to the newly created model")
+	f.StringVar(&c.targetController, "target-controller", "", "Target controller for the model")
+}
+
+// Init initializes the command with the given arguments.
+func (c *addModelCommand) Init(args []string) error {
+	if len(args) == 0 {
+		return common.MissingModelNameError("add-model")
```
SimoneDutto, line 115: is it the first time we don't use errors.E in jaas command?

cmd/jaas/cmd/addmodel.go
```diff
@@ -0,0 +1,617 @@
...
+		return errors.Errorf("%q is not a valid name: model names may only contain lowercase letters, digits and hyphens", c.Name)
+	}
+
+	if c.Owner != "" && !names.IsValidUser(c.Owner) {
+		return errors.Errorf("%q is not a valid user", c.Owner)
+	}
+
+	if c.targetController == "" {
+		return errors.Errorf("target controller not specified")
+	}
```
SimoneDutto, line 133: since in Terraform we still leave the user the ability of leaving this empty and relying on the random choice by jaas i feel like we should remove this check.

Otherwise we could find people using both:
`juju add-model`
`juju jaas add-model` 
in the same setup script, which seems weird.
What do you think?

cmd/jaas/cmd/addmodel.go
```diff
@@ -0,0 +1,617 @@
...
+
+	return cmd.CheckEmpty(args)
+}
+
+// Run executes the add-model command.
+//
+//nolint:gocognit
```
SimoneDutto, line 147: 😆

cmd/jaas/cmd/addmodel.go
```diff
@@ -0,0 +1,617 @@
...
+		modelOwner:  modelOwner,
+	})
+	if err != nil {
+		logger.Errorf("%v", err)
+		ctx.Infof("Use \n* 'juju add-credential -c' to upload a credential to a controller or\n" +
+			"* 'juju autoload-credentials' to add credentials from local files or\n" +
+			"* 'juju add-model --credential' to use a local credential.\n" +
+			"Use 'juju credentials' to list all available credentials.\n")
+		return cmd.ErrSilent
+	}
```
SimoneDutto: this made me realize if we do this `juju jaas add-model` it would be difficult to move jaas away from the juju cli. Because we would need to bring these commands as well.

cmd/jaas/cmd/addmodel.go
SimoneDutto, line 1: After the break somebody can remind me why we can't just add a flag to `juju add-model --target-controller`? I thought we've merged the pr, but i remember there was another issue.

cmd/jaas/cmd/addmodel_test.go
ale8k: +1

cmd/jaas/cmd/addmodel.go
ale8k: Gets the registry where all providers are registered (you know the init funcs? they go into a single global var and this returns it IIRC)

ale8k: Not sure we need providers though right?

cmd/jaas/cmd/addmodel.go
alesstimec: if you look at the implementation, you'll see that the Clouds method on the cloud facade returns all clouds the logged in user has access to.. which is wrong in the case where we're creating a model on behalf of another user

cmd/jaas/cmd/addmodel.go
alesstimec: if the `--target-controller` flag is not specified control will never be handed over to the jaas plugin's `add-model` command..

SimoneDutto: ah ok!

cmd/jaas/cmd/addmodel.go
kian99: We did add that flag, but we don't import the JAAS SDK into the Juju client package (at least not yet), so the Juju client can't call JIMM's `add-model` method.

kian99 (approved):

ale8k (commented):
I think we should embed their command publicly into a wrapping struct. Make the AddModel call a callback function in juju and override it with our target controller. That way we stay in sync, it is easier to maintain and clearer what we're trying to do here. I don't think c/ping the command is the right approach to this problem.

cmd/jaas/cmd/addmodel.go
```diff
@@ -0,0 +1,646 @@
...
+			OwnerTag:           names.NewUserTag(c.modelOwner).String(),
+			CloudTag:           cloudTag.String(),
+			CloudRegion:        cloudRegion,
+			CloudCredentialTag: credentialTag.String(),
+			Config:             modelConfig,
+		},
+		ControllerName: c.targetController,
+	}
+
+	model, err := c.jimmClient.AddModelToController(req)
```
ale8k, line 243: this could be c.addModelCallback within Juju

where addModel is a func passed to the client. It returns the same in Juju
```go
	addModelClient := c.newAddModelAPI(root)
	model, err := addModelClient.CreateModel(c.Name, modelOwner, cloudTag.Id(), cloudRegion, credentialTag, attrs)
	if err != nil {
```

? Then my wrapping idea makes this much simpler?

alesstimec: there are also differences in how we deduce the cloud..

ale8k (approved):
Righty ho ho ho, perhaps later we can improve it and export some bits and bobs from juju to remove the c/p. Approved! Good work!

PR #1794, opened 2026-01-09
feat(implement fatal login err code): err code

SimoneDutto:
## Description
Implement the new error code introduced here.

https://github.com/juju/juju/pull/21541

# QA

QA against juju 3.12 without the new logic to handle the new code implemented, and against tip of 3 with the changes for the fatal login error and it works fine.

### DIFF
--- a/internal/jujuapi/admin.go
+++ b/internal/jujuapi/admin.go
@@ -49,7 +49,7 @@ func (r *controllerRoot) LoginDevice(ctx context.Context) (params.LoginDeviceRes
 
 	deviceResponse, err := r.jimm.LoginManager().LoginDevice(ctx)
 	if err != nil {
-		return response, errors.E(err, errors.CodeUnauthorized)
+		return response, err
 	}
 	// NOTE: As this is on the controller root struct, and a new controller root
 	// is created per WS, it is EXPECTED that the subsequent call to GetDeviceSessionToken
--- a/internal/auth/auth.go
+++ b/internal/auth/auth.go
@@ -388,10 +388,6 @@ func (as *AuthenticationService) NewMigrationToken(ctx context.Context, username
 // The error code returned here is used by the Juju CLI to know when to start a
 // device login flow, prompting the user to login again.
 func (as *AuthenticationService) VerifySessionToken(token string) (_ jwt.Token, err error) {
-
-	errorFn := func(message string) error {
-		return errors.E(message, errors.CodeSessionTokenInvalid)
-	}
 	defer func() {
 		if err != nil {
 			servermon.AuthenticationFailCount.WithLabelValues("VerifySessionToken").Inc()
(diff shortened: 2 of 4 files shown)


internal/jujuapi/admin.go
```diff
@@ -49,7 +49,7 @@ func (r *controllerRoot) LoginDevice(ctx context.Context) (params.LoginDeviceRes
 
 	deviceResponse, err := r.jimm.LoginManager().LoginDevice(ctx)
 	if err != nil {
-		return response, errors.E(err, errors.CodeUnauthorized)
+		return response, err
```
SimoneDutto, line 52: i've checked in the JUJU Cli and we don't use `CodeUnauthorized` for any business logic.

But please pay extra attention to this change

kian99 (approved):
Reading this PR makes me feel like this code wasn't the best approach in the end because it's so specific, like the error code is just for that tryInOrder logic. I still think it's a fine step forward though.

But I had one more thought - if we put a code on the unsupported login then if the client hits unsupported login, it doesn't keep track of that error.

SimoneDutto:
> Reading this PR makes me feel like this code wasn't the best approach in the end because it's so specific, like the error code is just for that tryInOrder logic. I still think it's a fine step forward though.
> 
> But I had one more thought - if we put a code on the unsupported login then if the client hits unsupported login, it doesn't keep track of that error.

i didn't full get the last sentence

SimoneDutto:
> Reading this PR makes me feel like this code wasn't the best approach in the end because it's so specific, like the error code is just for that tryInOrder logic. I still think it's a fine step forward though.
> 
> But I had one more thought - if we put a code on the unsupported login then if the client hits unsupported login, it doesn't keep track of that error.

i still think this is the best approach, but i would have preferred to have different type (not codes) of errors coming from lower layers and doing the conversion to `FatalLogin` or `SessionInvalid` codes in an higher level (jujuapi pkg for example).

So we direct whether or not we want to trigger the device flow or sending a fatal depending an application error type.
But yeah, we don't really have that in JIMM

luci1900 (approved):

kian99:
> > Reading this PR makes me feel like this code wasn't the best approach in the end because it's so specific, like the error code is just for that tryInOrder logic. I still think it's a fine step forward though.
> > But I had one more thought - if we put a code on the unsupported login then if the client hits unsupported login, it doesn't keep track of that error.
> 
> i didn't full get the last sentence

Just flipping the logic essentially. Instead of stopping the `tryInOrderProvider` early when we get a fatal login error, if the UsernamePassword login returns a `NotSupportedCode` back, the client can do something like,
```go
func (p *tryInOrderLoginProviders) Login(ctx context.Context, caller base.APICaller) (*api.LoginResultParams, error) {
	var lastError error
	for i, provider := range p.providers {
		result, err := provider.Login(ctx, caller)
		if err != nil {
			p.logger.Debugf("login error using provider %d - %s", i, err.Error())
			// Check the code instead of using errors.Is, but you get the idea.
			if errors.Code(err, api.ErrNotSupported) {
				continue
			}
			lastError = err
		} else {
			p.logger.Debugf("successful login using provider %d", i)
			p.authHeader = func() (http.Header, error) { return provider.AuthHeader() }
			return result, nil
		}
		lastError = err
	}
	if lastError == nil {
		lastError = "login failed, no supported provider defined"
	}
	return nil, errors.Trace(lastError)
}
```
Lmk what you think.

SimoneDutto:
> > > Reading this PR makes me feel like this code wasn't the best approach in the end because it's so specific, like the error code is just for that tryInOrder logic. I still think it's a fine step forward though.
> > > But I had one more thought - if we put a code on the unsupported login then if the client hits unsupported login, it doesn't keep track of that error.
> > 
> > 
> > i didn't full get the last sentence
> 
> Just flipping the logic essentially. Instead of stopping the `tryInOrderProvider` early when we get a fatal login error, if the UsernamePassword login returns a `NotSupportedCode` back, the client can do something like,
> 
> ```go
> func (p *tryInOrderLoginProviders) Login(ctx context.Context, caller base.APICaller) (*api.LoginResultParams, error) {
> 	var lastError error
> 	for i, provider := range p.providers {
> 		result, err := provider.Login(ctx, caller)
> 		if err != nil {
> 			p.logger.Debugf("login error using provider %d - %s", i, err.Error())
> 			// Check the code instead of using errors.Is, but you get the idea.
> 			if errors.Code(err, api.ErrNotSupported) {
> 				continue
> 			}
> 			lastError = err
> 		} else {
> 			p.logger.Debugf("successful login using provider %d", i)
> 			p.authHeader = func() (http.Header, error) { return provider.AuthHeader() }
> 			return result, nil
> 		}
> 		lastError = err
> 	}
> 	if lastError == nil {
> 		lastError = "login failed, no supported provider defined"
> 	}
> 	return nil, errors.Trace(lastError)
> }
> ```
> 
> Lmk what you think.

ah yeah, we've discussed this in the other one. But i'm more in favor of having a new specific code for fatal, then using an existing code because it's more worrying and the other pr is already merged in.

PR #1798, opened 2026-01-12
chore(migrate jujuclient): migrate juju client from connsuite to websockete2esuite

SimoneDutto:


.github/workflows/e2e-test.yaml
```diff
@@ -80,9 +80,18 @@ jobs:
             connect=tcp:$host_ip_addr:16443 \
             bind=container
       
-      - name: Run E2E tests
+      - name: Run E2E tests ./internal/jujuclient
```
kian99, line 83: Maybe we should rename this to "Run controller integration tests" or something like that.

.github/workflows/e2e-test.yaml
```diff
@@ -80,9 +80,18 @@ jobs:
...
           RUN_E2E_TESTS: "1"
           JIMM_MICROK8S_TEST_CLOUD_NAME: "microk8s-cp"
           JIMM_BACKING_CONTROLLER_CONFIG: ${{ github.workspace }}/controllers.yaml
+
+      - name: Run E2E tests testing
+        run: go test -v ./testing/... -timeout 30m
+        env:
+          RUN_E2E_TESTS: "1"
+          JIMM_MICROK8S_TEST_CLOUD_NAME: "microk8s-cp"
+          JIMM_BACKING_CONTROLLER_CONFIG: ${{ github.workspace }}/controllers.yaml
```
kian99, line 95: It would be better to run them at the same time

kian99 (commented):
Before I approve, wouldn't it be better to just switch to Juju's API client? Then we can delete all the tests?

SimoneDutto:
closing this because we are thinking about not using jujuclient all together

PR #1800, opened 2026-01-13
fix(debug-log args): fix debug logs args ignored

### LINKED ISSUE #1799: juju debug-log on a JAAS managed model does not respect CLI arguments
Hello,

I am working with a model managed by JAAS and have encountered a bug with `juju debug-log`. 

## Bug

When a user provides arguments to debug-log, e.g. `--include <unit>/11` or `--limit 50`, these items are disregarded and the full debug-log output is displayed. For long-standing models, this output can be quite large. Additionally the `--no-tail` argument is not respected.

For example, I captured the output of a sample command and found the output was > 10k lines:

```bash
<alexlukens@laptop>~$ juju debug-log --include jenkins-agent/11 --limit 100 > output.txt
^C
<alexlukens@laptop>~$ cat output.txt | wc -l
11166
```

## Expected Outcome

The command line arguments provided to `juju debug-log` are respected as would be on a non-JAAS managed model.

SimoneDutto:
# Description

our https proxy was dropping url query values, now it doesn't.

juju debug-logs args are passed as url args, and our proxy was dropping those, which is wrong.

I've taken the chance to pass url query params also for the controlller http proxy.

fix #1799

# QA

`juju debug-log --limit 1 --no-tail` now works and returns a single log.

### DIFF
--- a/internal/rpc/dial.go
+++ b/internal/rpc/dial.go
@@ -90,14 +90,14 @@ func GetAddressesAndTLSConfig(ctx context.Context, ctl *dbmodel.Controller) ([]s
 // Dial connects to the controller/model and returns a raw websocket
 // that can be used as is.
 // It accepts the endpoints to dial, normally /api or /commands.
-func Dial(ctx context.Context, ctl *dbmodel.Controller, modelTag names.ModelTag, finalPath string, headers http.Header) (*websocket.Conn, error) {
+func Dial(ctx context.Context, ctl *dbmodel.Controller, modelTag names.ModelTag, finalPath string, headers http.Header, attrs url.Values) (*websocket.Conn, error) {
 	addrs, tlsConfig := GetAddressesAndTLSConfig(ctx, ctl)
 	dialer := Dialer{
 		TLSConfig: tlsConfig,
 	}
 	var websocketUrls []string
 	for _, addr := range addrs {
-		websocketUrls = append(websocketUrls, websocketURL(addr, modelTag, finalPath))
+		websocketUrls = append(websocketUrls, websocketURL(addr, modelTag, finalPath, attrs))
 	}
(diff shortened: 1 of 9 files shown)


testing/streammodelproxy_test.go
```diff
@@ -30,6 +31,30 @@ func (s *streamProxySuite) TestDebugLogs(c *gc.C) {
...
+		NoTail: true,
+		Limit:  1,
+	})
+	c.Assert(err, gc.IsNil)
+	messages := 0
+	for {
+		select {
+		case _, ok := <-logChan:
+			if !ok {
+				c.Assert(messages, gc.Equals, 1)
```
ale8k, line 48: Is it possible to assert content at all?

ale8k (approved):
lgtm 1 minor question

testing/streammodelproxy_test.go
SimoneDutto: i can tell you the message is:
`"provisioning in zones: [duttospc]"`

but i wouldn't test on it because i'm expecting it to be really flaky

internal/rpc/dial.go
```diff
@@ -90,14 +90,14 @@ func GetAddressesAndTLSConfig(ctx context.Context, ctl *dbmodel.Controller) ([]s
 // Dial connects to the controller/model and returns a raw websocket
 // that can be used as is.
 // It accepts the endpoints to dial, normally /api or /commands.
-func Dial(ctx context.Context, ctl *dbmodel.Controller, modelTag names.ModelTag, finalPath string, headers http.Header) (*websocket.Conn, error) {
+func Dial(ctx context.Context, ctl *dbmodel.Controller, modelTag names.ModelTag, finalPath string, headers http.Header, attrs url.Values) (*websocket.Conn, error) {
```
kian99, line 93: At some point we'll probably want to look at using an `http.Request` object.

kian99 (approved):

PR #1801, opened 2026-01-13
fix: fixes add-model response when no controllers are registered

ale8k:
## Description
Fixes https://warthogs.atlassian.net/browse/JUJU-9051, a bug that @alesstimec found.

When running add-model against JIMM, and no controllers were registered, the code path for [this](https://github.com/ale8k/juju/blob/2c70d10b5744dcd076d237a7a64147cac6116347/cmd/juju/controller/addmodel.go#L430-L434) was running. This is because Juju was attempting to get the cloud [here](https://github.com/ale8k/juju/blob/2c70d10b5744dcd076d237a7a64147cac6116347/cmd/juju/controller/addmodel.go#L454-L457) in order to retrieve the credential to add the model.

I've solved this by adding a count controllers DB call, and checking in ForEachUserCloud that controllers are registered. This error is propogated through the facade and now returned from the Clouds() call. Surfaced like so:
```
➜  jimm git:(juju-9051/fix-add-model-no-controllers) ✗ juju add-model a
ERROR no controllers registered
``` 
I did think of putting it in the facade call but any call to ForEachUserCloud is moot without having controllers, hence why I added the check here. We should probably add this to all Cloud retrieval calls, but I didn't want to scope creep this PR.

## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests

## Test instructions
Spin up jimm without controllers, login, and run add-model.

### DIFF
--- a/internal/db/controller.go
+++ b/internal/db/controller.go
@@ -180,7 +180,7 @@ func (d *Database) CountControllers(ctx context.Context) (count int, err error) 
 	const op = "db.CountControllers"
 
 	if err := d.ready(); err != nil {
-		return 0, errors.E(err)
+		return -1, errors.E(err)
 	}
 
 	durationObserver := servermon.DurationObserver(servermon.DBQueryDurationHistogram, op)
--- a/internal/db/controller.go
+++ b/internal/db/controller.go
@@ -176,7 +176,7 @@ func (d *Database) ForEachControllerModel(ctx context.Context, ctl *dbmodel.Cont
 }
 
 // CountControllers returns the number of controllers stored in the database.
-func (d *Database) CountControllers(ctx context.Context) (count int64, err error) {
+func (d *Database) CountControllers(ctx context.Context) (count int, err error) {
 	const op = "db.CountControllers"
 
 	if err := d.ready(); err != nil {
@@ -188,8 +188,9 @@ func (d *Database) CountControllers(ctx context.Context) (count int64, err error
 	defer servermon.ErrorCounter(servermon.DBQueryErrorCount, &err, op)
 
 	db := d.DB.WithContext(ctx)
-	if err := db.Model(&dbmodel.Controller{}).Count(&count).Error; err != nil {
-		return 0, errors.E(dbError(err))
+	var count64 int64
+	if err := db.Model(&dbmodel.Controller{}).Count(&count64).Error; err != nil {
+		return -1, errors.E(dbError(err))
 	}
-	return count, nil
+	return int(count64), nil
 }
--- a/internal/db/controller.go
+++ b/internal/db/controller.go
@@ -174,3 +174,22 @@ func (d *Database) ForEachControllerModel(ctx context.Context, ctl *dbmodel.Cont
 	}
 	return nil
 }
+
+// CountControllers returns the number of controllers stored in the database.
+func (d *Database) CountControllers(ctx context.Context) (count int64, err error) {
+	const op = "db.CountControllers"
+
+	if err := d.ready(); err != nil {
+		return 0, errors.E(err)
+	}
+
+	durationObserver := servermon.DurationObserver(servermon.DBQueryDurationHistogram, op)
+	defer durationObserver()
+	defer servermon.ErrorCounter(servermon.DBQueryErrorCount, &err, op)
+
+	db := d.DB.WithContext(ctx)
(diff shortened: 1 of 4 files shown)


internal/db/controller.go
```diff
@@ -174,3 +174,22 @@ func (d *Database) ForEachControllerModel(ctx context.Context, ctl *dbmodel.Cont
...
+		return 0, errors.E(err)
+	}
+
+	durationObserver := servermon.DurationObserver(servermon.DBQueryDurationHistogram, op)
+	defer durationObserver()
+	defer servermon.ErrorCounter(servermon.DBQueryErrorCount, &err, op)
+
+	db := d.DB.WithContext(ctx)
+	if err := db.Model(&dbmodel.Controller{}).Count(&count).Error; err != nil {
+		return 0, errors.E(dbError(err))
```
alesstimec: `return -1, errors.E(dbError(err)`

to make a clear distinction between "no controller here" and "omg. there was an error getting the controller count"

internal/db/controller.go
```diff
@@ -174,3 +174,22 @@ func (d *Database) ForEachControllerModel(ctx context.Context, ctl *dbmodel.Cont
 	}
 	return nil
 }
+
+// CountControllers returns the number of controllers stored in the database.
+func (d *Database) CountControllers(ctx context.Context) (count int64, err error) {
```
alesstimec: let's return `(int, error)`

i don't think any JIMM instance will ever have number of controller that would exceed golang's int capabilities..

internal/jimm/juju/cloud.go
```diff
@@ -59,8 +59,20 @@ func (j *JujuManager) GetCloud(ctx context.Context, user *openfga.User, tag name
 // true then f will be called with all clouds known to JIMM. If f returns
 // an error then iteration will stop immediately and the error will be
 // returned unchanged. The given function should not update the database.
+//
+// Additionally, if there are no controllers registered with JIMM then an
+// error of "no controllers registered" will be returned.
 func (j *JujuManager) ForEachUserCloud(ctx context.Context, user *openfga.User, f func(*dbmodel.Cloud) error) error {
 
+	ctrlCount, err := j.Database.CountControllers(ctx)
```
kian99, line 67: Instead of putting this here, let's put it at the top of `AddModel()` (in the domain layer). As we're trying to define some business logic,
- When the user runs `add-model`
- And JIMM has no controllers
- We should return an error "no controllers registered"

It seems clearer there then nested inside here everytime we count a user's clouds.

ale8k: It doesn't reach the AddModel code path, this is a check previous to AddModel within the CLI. They retrieve the clouds.

alesstimec: I agree with @kian99

alesstimec: oh, actually, @ale8k is right.. CLI does check for clouds user has access to.. perhaps we could solve this in another way: in the ForEachUserCloud check if the cloud/region has any registered controller - if not, skip the cloud?

ale8k: That's overcomplicating this I think, a simple "do we have controllers" suffices I think.

kian99: Since I misunderstood and the undesirable error message comes from the CLI when there are 0 clouds, and in JAAS we seem to want to say "consider it an error to query for things when there are no controllers attached" then the change is okay.

kian99 (approved):
Loosk good, truncating the int64 to int is also fine and I'd make that change.

internal/db/controller.go
```diff
@@ -174,3 +174,22 @@ func (d *Database) ForEachControllerModel(ctx context.Context, ctl *dbmodel.Cont
...
+	const op = "db.CountControllers"
+
+	if err := d.ready(); err != nil {
+		return 0, errors.E(err)
```
alesstimec: -1 here, too

internal/jimm/juju/cloud.go
```diff
@@ -59,8 +59,20 @@ func (j *JujuManager) GetCloud(ctx context.Context, user *openfga.User, tag name
...
+	if err != nil {
+		return errors.E(fmt.Errorf("cannot count controllers: %w", err))
```
alesstimec: suggestion: `cannot get controller count`

alesstimec (approved):
LGTM with final two comments

PR #1805, opened 2026-01-14
docs: add doc anchor tags

kian99:
## Description
This PR updates the JAAS CLI reference doc to add anchor tags. These can be used in the MyST markdown format for cross-referencing between pages - see https://myst-parser.readthedocs.io/en/latest/syntax/cross-referencing.html#cross-references.

To update the auto-generated file I've updated our script with a step that runs `sed` on matching lines. I used an LLM to generate the `sed` and the commits in this PR show the iterations on how it evolved as I tried to simplify (with some hand tweaks too).

### DIFF
--- a/docs/reference/list-of-jaas-plugin-commands/index.md
+++ b/docs/reference/list-of-jaas-plugin-commands/index.md
@@ -57,6 +57,7 @@ Add group to jimm.
 Adds a group.
 
 
+(command-jaas-add-model)=
 # jaas add-model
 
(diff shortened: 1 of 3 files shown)


tmihoc (approved):
This looks very nice, thanks!

PS We still need to update the links to use the new anchors. I'll do this in a separate PR tomorrow. (There are a few more little things I want to fix.)

ale8k (approved):
lgtm.

PR #1816, opened 2026-01-19
feat(add-river-queue): add river to JIMM, register workers, and add wrapper client

SimoneDutto:
## Description

two pkgs:
- `riverworker` contains the worker definition, and the register and start
- `riverclient` sets up the client and insert jobs

I imagine the river client being passed around like the openfga client, and from the `jujuapi` pkg or another pkg we call.

`c.riverClient.InsertUpgradeTo(args ...)`

The worker receives just the args and uses the manager which it is configured with to call the business logic, so the business logic doesn't leave `internal/jimm`.

The testing harness is very handy for both riverclient and riverworker. And it plays nice with the managers and mocks.

> Using `jimmsvc.JIMM().Database` to setup workers and client is nice and tidy to see but it comes with some limitations. In fact, using sql pkg doesn't support `NOTIFY/LISTEN`, and it does a simple polling to get events. If we want to use these functionalities we need to pass the database url and create a different postgresql driver just for river. Read more here: https://riverqueue.com/docs/gorm


internal/dbmodel/sql/postgres/034_river_bootstrap.up.sql
kian99, line 1: Hmmm what's this? We don't need to include River's DB migrations in our DB migrations. Unless things have changed since the last time I'd looked at their stuff.

internal/riverclient/riverclient.go
kian99, line 1: This package is quite small, what about merging it with `riverworker`? I wouldn't try create too much separation until we know its needed.

internal/river/migrationworker.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+
+	return &upgradeToWorker{
+		migrationManager: migrationManager,
+	}, nil
+}
+
+// UpgradeToArgs defines the arguments for the upgradeToWorker job.
+type UpgradeToArgs struct {
+	UUID string `json:"uuid"`
```
kian99, line 29: Is this the model's UUID?

internal/dbmodel/sql/postgres/034_river_bootstrap.up.sql
SimoneDutto: i think it's needed, otherwise how would river be setup in postgresql?
https://riverqueue.com/docs#running-migrations

internal/riverclient/riverclient.go
SimoneDutto: i like the separation, because one thing is running the worker and one is submitting jobs.

It makes sense there is a client and worker pkg to me

internal/river/migrationworker.go
SimoneDutto: just an example yeah

internal/dbmodel/sql/postgres/034_river_bootstrap.up.sql
kian99: Take a look at the section https://riverqueue.com/docs/migrations#go-migration-api

SimoneDutto: i'll check

internal/river/migrationworker.go
kian99, line 1: Nit: I think `riverqueue` would help first time readers.

kian99 (approved):

luci1900 (approved):
Noice

PR #1817, opened 2026-01-19
test: addcloudtocontroller unit test

ale8k:
## Description
Refactors add cloud to controller into a unit test. I've also added a single place to create JIMM API clients as:

1. We're not dialing any real controller anymore, so we don't care about the store
2. It removes the need to do "newClient" in every command and use the store. The store seems to only be used to dial for our cmd.
## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

### DIFF
--- a/cmd/jaas/cmd/addcloudtocontroller.go
+++ b/cmd/jaas/cmd/addcloudtocontroller.go
@@ -8,19 +8,16 @@ import (
 
 	"github.com/juju/cmd/v3"
 	"github.com/juju/gnuflag"
-	jujuapi "github.com/juju/juju/api"
 	"github.com/juju/juju/cloud"
 	jujucmd "github.com/juju/juju/cmd"
 	jujucmdcloud "github.com/juju/juju/cmd/juju/cloud"
 	jujucmdcommon "github.com/juju/juju/cmd/juju/common"
 	"github.com/juju/juju/cmd/modelcmd"
-	"github.com/juju/juju/jujuclient"
 	"github.com/juju/juju/rpc/params"
 	"github.com/juju/names/v5"
 
 	"github.com/canonical/jimm/v3/internal/errors"
 	jimmjujuapi "github.com/canonical/jimm/v3/internal/jujuapi"
-	"github.com/canonical/jimm/v3/pkg/api"
 	apiparams "github.com/canonical/jimm/v3/pkg/api/params"
 )
 
@@ -42,7 +39,6 @@ is already known and will error otherwise.
 // controller in JIMM.
 func NewAddCloudToControllerCommand() cmd.Command {
 	cmd := &addCloudToControllerCommand{
-		store:           jujuclient.NewFileClientStore(), // Can be removed alongside all "newClient" calls if we like the new newClient approach.
 		cloudByNameFunc: jujucmdcommon.CloudByName,
 		jimmAPIFunc:     NewClient,
 	}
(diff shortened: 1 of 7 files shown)


internal/testutils/cmdtest/jimmsuite.go
kian99, line 1: We might need this around until ALL the tests are refactored no? Or was the only usage now removed?

cmd/jaas/cmd/apiroot.go
kian99, line 1: This looks good. 

On a separate not, it makes me think, we could inject the API client into the commands as a dependency rather than creating the client in the command. If we were using a more modern CLI library, from what I've seen you can run a "pre-command" function where we could instantiate the client.

With the Juju library, we don't seem to have that but we could consider adding the logic to `main()` in `cmd/jaas/main.go` and passing the client through the context, then the tests can create a mock client and the cmd files become a bit simpler. The tricky bit is not all commands probably want to create an API client. Lmk what you think.

cmd/jaas/cmd/addcloudtocontroller_test.go
```diff
@@ -1,211 +1,76 @@
...
-	})
-	c.Assert(err, gc.IsNil)
-
-	err = s.JIMM.OpenFGAClient.AddController(context.Background(), s.JIMM.ResourceTag(), names.NewControllerTag(info.ControllerUUID))
-	c.Assert(err, gc.IsNil)
+	return jimmAPI
 }
 
-func (s *addCloudToControllerSuite) TestAddCloudToController(c *gc.C) {
-	c.Skip("This test is failing since 3.6.12 since the providers moved to internal.")
```
kian99, line 83: We have a card on the backlog to fix this, since we are removing this please check that we have an equivalent test for this in our e2e testing suite.

cmd/jaas/cmd/addcloudtocontroller.go
```diff
@@ -139,10 +141,29 @@ func (c *addCloudToControllerCommand) Run(ctxt *cmd.Context) error {
...
+	}); err != nil {
+		return errors.E(err)
+	}
+	ctxt.Infof("Cloud %q added to controller %q.", c.cloudName, c.dstControllerName)
+
+	// Old:
+	// err = c.addCloudToController(ctxt, newCloud)
+	// if err != nil {
+	// 	return errors.E(err, fmt.Sprintf("error adding cloud to controller: %v", err))
+	// }
```
kian99: dd

kian99 (approved):

cmd/jaas/cmd/apiroot.go
ale8k: I thought about this too, but we're defining the interface in cmd. We'd have to move the testing interface into a separate 3rd package tp prevent circular dependency. 

I think the pre-command / pass via context would be better 100%. 

I was under the impression all of our commands would create an API client? But if not, we could go this approach for now, and then improve on it later and pass it in via context. It should be a smaller change I think. Wdyt?

internal/testutils/cmdtest/jimmsuite.go
ale8k: This was the only usage, I plan on just removing them as we go if we remove the last ref.

cmd/jaas/cmd/apiroot.go
kian99: Yeah can do later.

cmd/jaas/cmd/addcloudtocontroller_test.go
kian99: Taking another look at this, I recall that it was the CLI logic that was failing to parse a cloud due to the lack of providers. I believe we still need some kind of test that will be failing due to this issue.

kian99 (changes requested):
Just requesting changes to override my previous approval. We need a test like the previous `TestAddCloudToController` that is verifying the parsing of clouds logic.

kian99 (approved):

luci1900 (approved):

PR #1819, opened 2026-01-20
feat: add migration worker and update manage to have a singular migra…

ale8k:
## Description
…te method. This PR:

1. Adds the worker
2. Updates river to take an UpgradeManager interface
3. Adds a singular method for migrations on the UpgradeManager.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests


internal/jimm/upgrade/upgrade.go
```diff
@@ -366,3 +366,72 @@ func (u *upgradeManager) upgradeModel(ctx context.Context, api juju.API, modelUU
...
+// Phase 2 of automated upgrades is from here onwards.
+
+// MigrateModel migrates a model to a new controller without upgrading the model's agent.
+// This is Phase 2 of automated upgrades and to be called from a river worker.
+func (u *upgradeManager) MigrateModel(ctx context.Context, user *openfga.User, modelUUID string, targetControllerName string) error {
+	// As the controller has just been bootstrapped, the controller machine can still be in a pending state.
+	// Unfortunately Juju doesn't return a typed error for us to examine, and the best we've got is the error message string.
+	// So we retry a few times here to allow the controller machine to come up.
+	var iimResult jujuparams.InitiateMigrationResult
+	if err := retry.Call(
```
kian99: Because the steps in the `UpgradeTo` workflow are being orchestrated, I think we should work on the assumption that the previous step has completed its work. I.e. below it seems like it would be sensible to assume the new controller is up and ready.

With the above in mind, it seems to me like we shouldn't need this retry loop around `InitiateInternalMigration`. If it fails, that's unexpected and we should let the orchestrator decide to retry. The second retry loop around checking the migration completion is good though because we're waiting time for that to complete.

And, depending on the behaviour of `InitiateInternalMigration`, we need something at the start of the `MigrateModel()` method to check if the model is already on the new controller otherwise the function is not idempotent. I.e. what does `InitiateInternalMigration` do if its called again and the migration completed / is in-progress.

internal/river/migrationworker.go
```diff
@@ -5,46 +5,49 @@ import (
...
 )
 
-// UpgradeToManager defines the interface for managing upgrade operations.
-type UpgradeToManager interface {
-	UpgradeTo(ctx context.Context, user *openfga.User, modelUUID string, targetVersion version.Number) (version.Number, error)
+// UpgradeManagerMigrationWorker defines the methods required from this manager for the migration worker.
+type UpgradeManagerMigrationWorker interface {
+	// MigrateModel migrates a model to a new controller without upgrading the model's agent.
+	MigrateModel(ctx context.Context, user *openfga.User, modelUUID string, targetControllerName string) error
 }
```
kian99: Do we really need another interface when we've got `UpgradeManager` in `internal/river/river.go`? Seems like we're just jumping through extra hoops for nothing.

internal/jimm/upgrade/upgrade.go
ale8k: The check for machine-0 is more of a timing issue, unless we poll the machines state prior we won't know if the migrate is OK. But I think you're right in that we let the supervising job insert this with N retries, and that should cover it. I'm not sure if you can retry if X error occurs only. But we can check that out later I guess. I'll remove now under the presumption we'll insert with X retries.

On the last point, if it's already in progress it fails. But that would be an expected result.

However if it's completed... I don't know. I'll test it manually using raw juju (everything looks OK until the actual migration).

internal/river/migrationworker.go
ale8k: It limits the scope of what it is using. Not harm done by having it really and makes it super clear what methods are going to be consumed within the worker.

internal/jimm/upgrade/upgrade.go
ale8k: To answer the question what happens if you migrate a model that's been migrated:
```bash
➜  jimm git:(juju-9053/migrate-model-job) jj migrate migrateme mb
Migration started with ID "97cfbc86-1b0b-4013-826f-e47145133eaa:0"
➜  jimm git:(juju-9053/migrate-model-job) jj models
Controller: ma

Model       Cloud/Region         Type  Status     Machines  Units  Access  Last connection
controller  localhost/localhost  lxd   available         1      1  admin   just now
➜  jimm git:(juju-9053/migrate-model-job) jj migrate migrateme mb
ERROR model ma:admin/migrateme not found
```
But I think this is expected too.. The model isn't there anymore. I do wonder if we need idempotency though, if the model doesn't exist, it doesn't exist, therefore we cannot migrate it. And I think this makes the most sense as it will fail fast anyway.

We could mark it as migrated (maybe we do but I don't think we do?) and if it is, return no error perhaps.

internal/river/migrationworker.go
kian99: I think it isn't worth it to jump through the extra indirections when navigating the code. It's not like it needs to be scoped down from 50 methods to 1, it's 2 methods -> 1.

internal/jimm/upgrade/upgrade.go
kian99: The idempotency is for handling scenarios where JIMM crashes or the DB is down and River can't save the outcome of the job. Then after a delay when it retries and the DB is back up, things should proceed if the overall process was successful.

ale8k: Yeah but it doesn't need to be idempotent to achieve that (I'm also not even sure if it is possible given the outcome). I think we're better off checking if it's on the target controller after a crash and ignoring model not found errors...

internal/river/migrationworker.go
ale8k: Yeah sure, it's not many, was just keeping to good practices really. If you feel strongly about it, I'll remove it?

internal/jimm/upgrade/upgrade.go
kian99: > I think we're better off checking if it's on the target controller after a crash and ignoring model not found errors

I believe that's what I suggested when I said,

> we need something at the start of the MigrateModel() method to check if the model is already on the new controller

ale8k: Ah sorry my bad, I miscommunicated what I meant. I was suggesting we do it in the supervising job before calling migrate model? Otherwise we'd be naming it something funky like "MaybeMigrateModel"

kian99: I think you should keep that business logic in the the UpgradeManager method and let the supervising job just handle River logic.

ale8k: Yeah I agree, just means we gotta name it funky or I guess it's fine in the godoc?

kian99: Fine by me to just go in the godoc

internal/jimm/upgrade/upgrade.go
```diff
@@ -366,3 +366,72 @@ func (u *upgradeManager) upgradeModel(ctx context.Context, api juju.API, modelUU
...
+		retry.CallArgs{
+			Attempts: 30,
+			Delay:    10 * time.Second,
+			// We could consider all errors fatal, bar: target prechecks failed: machine 0 not running (pending)
+			// IsFatalError:
```
SimoneDutto: the comment and the commented IsFatal don't match

internal/jimm/upgrade/upgrade.go
```diff
@@ -366,3 +366,72 @@ func (u *upgradeManager) upgradeModel(ctx context.Context, api juju.API, modelUU
...
+		return errors.E(fmt.Errorf("failed to initiate internal migration: %w", err))
+	}
+
+	mt, err := names.ParseModelTag(iimResult.ModelTag)
+	if err != nil {
+		return errors.E(fmt.Errorf("failed to parse model tag from initiate internal migration result: %w", err))
+	}
+
+	var mi *jujuparams.ModelInfo
+	if err := retry.Call(
```
SimoneDutto, line 403: i would implement a `IsFatal` here.
For example if the database is down, or the juju controller is not available i wouldn't want to wait for 30 retries to give some feedback to the user

internal/river/migrationworker.go
```diff
@@ -5,46 +5,49 @@ import (
...
+		upgradeManager: upgradeManager,
 	}, nil
 }
 
-// UpgradeToArgs defines the arguments for the upgradeToWorker job.
-type UpgradeToArgs struct {
-	UUID string `json:"uuid"`
+// UpgradeMigrationWorker defines the arguments for the upgradeMigrationWorker job.
+type UpgradeMigrationWorker struct {
+	User                 *openfga.User `json:"user"` // TODO: Will this work without JSON tags Alex? How will it serialise the user between instances?
```
SimoneDutto: i don't think this would work, you need to refactor the other method to extract a `openfga.User` from the `username`.

At the end of the day i don't thing `opefga.User` is serializable, and even if it is. It's not worth to do it

SimoneDutto (commented):
looks good, a few comments here and there

internal/jimm/upgrade/upgrade.go
```diff
@@ -366,3 +368,68 @@ func (u *upgradeManager) upgradeModel(ctx context.Context, api juju.API, modelUU
...
+		zapctx.Error(ctx, "Failed to initiate internal migration", zap.Error(err))
+		return errors.E(fmt.Errorf("failed to initiate internal migration: %w", err))
+	}
+
+	mt, err := names.ParseModelTag(iimResult.ModelTag)
+	if err != nil {
+		return errors.E(fmt.Errorf("failed to parse model tag from initiate internal migration result: %w", err))
+	}
+
+	modelNotMigratedErr := errors.E("model has not yet migrated to target controller")
```
SimoneDutto, line 400: does errors.E work with `stderrors.Is`?

internal/jimm/upgrade/upgrade_test.go
```diff
@@ -429,6 +429,64 @@ func (s *upgradeManagerSuite) TestUpgradeTo_Success(c *qt.C) {
 	c.Assert(time.Since(time.Unix(ts, 0)) < 2*time.Second, qt.IsTrue)
 }
 
+func (s *upgradeManagerSuite) TestMigrateModel_Success(c *qt.C) {
```
SimoneDutto, line 433: can you add a test to check the retry works as expected

SimoneDutto (approved):
one more test please and it's approved

internal/jimm/upgrade/upgrade.go
ale8k: Will test.

internal/jimm/upgrade/upgrade_test.go
kian99: Can you also please also add a test to check that the migration is skipped if the model is already on the target controller.

For the test Simone is suggesting to exercise the retry, you can take a look at my PR where I used `synctest` it was super simple and made the test fast without requiring you to pass in a fake clock.

ale8k: Yep will do, good shout!

I already seen the synctest, will add!

internal/river/migrationworker.go
```diff
@@ -3,48 +3,61 @@ package river
...
-// Work performs the upgrade operation receiving the job with UpgradeToArgs.
-func (w *upgradeToWorker) Work(ctx context.Context, job *river.Job[UpgradeToArgs]) error {
-	_, err := w.migrationManager.UpgradeTo(ctx, nil, job.Args.UUID, version.Number{})
-	if err != nil {
+// Work implements the [river.Worker] interface.
+func (w *upgradeMigrationWorker) Work(ctx context.Context, job *river.Job[UpgradeMigrationWorker]) error {
+	u := &dbmodel.Identity{Name: job.Args.Username}
+	if err := w.store.FetchIdentity(ctx, u); err != nil {
 		return err
 	}
```
kian99, line 56: This looks good.

It would be awesome, if you feel like it, to refactor the permission checks so that they're in `jujuapi` and make it so that the manager doesn't need the OpenFGA user at all.

internal/river/package_test.go
```diff
@@ -1,5 +1,5 @@
 package river
 
-var NewUpgradeToWorker = newUpgradeToWorker
+var NewUpgradeMigrationWorker = newUpgradeMigrationWorker
```
kian99: Should we just make these test whitebox to avoid this? I've done it in my PR.

internal/river/migrationworker.go
ale8k: How could we do it in `jujuapi`? Any worker could pick this up, and the time between their session could have been invalidated. Unless I'm misunderstanding?

internal/river/package_test.go
ale8k: Yeah sure.

internal/jimm/upgrade/upgrade_test.go
```diff
@@ -429,6 +430,145 @@ func (s *upgradeManagerSuite) TestUpgradeTo_Success(c *qt.C) {
...
+	s.jujuManager.EXPECT().
+		GetModel(gomock.Any(), targetMt.Id()).
+		DoAndReturn(func(ctx context.Context, uuid string) (dbmodel.Model, error) {
+			getModelCalls++
+			if getModelCalls < 3 {
+				return dbmodel.Model{Controller: dbmodel.Controller{Name: "still-source"}}, nil
+			}
+			return dbmodel.Model{Controller: dbmodel.Controller{Name: targetController}}, nil
+		}).
+		Times(3)
```
SimoneDutto, line 535: can't you just simplify this by doing:
 ```
s.jujuManager.EXPECT().
		GetModel(gomock.Any(), targetMt.Id()).
		DoAndReturn(func(ctx context.Context, uuid string) (dbmodel.Model, error) {
				return dbmodel.Model{Controller: dbmodel.Controller{Name: "still-source"}}, nil
		}).Times(3)
s.jujuManager.EXPECT().
		GetModel(gomock.Any(), targetMt.Id()).
		DoAndReturn(func(ctx context.Context, uuid string) (dbmodel.Model, error) {
				return dbmodel.Model{Controller: dbmodel.Controller{Name: targetController}}, nil
		}).

ale8k: Yeah but then our test for the amount is conteigent on our mocking framework, whereas having an explicit var (getModelCalls) is explicit. So I'll leave as is.

internal/river/migrationworker.go
kian99: Can discuss separately another time.

kian99 (approved):

internal/river/migrationworker.go
ale8k: Yeah it's a bigger issue I think but I'm not really too sure on how we solve it... I think each call should re-validate what it intends to do in a smart way.

PR #1820, opened 2026-01-20
feat: introduce model-upgrade river worker

kian99:
## Description
This PR introduces two new pieces for the model upgrade story:
1. A new method in the upgrade manager that only upgrades a model. It can be called repeatedly after the model has completed upgrading and returns early without error.
2. A river worker that handles the model upgrade.

Two extra bits:
1. One of the tests in `internal/jimm/upgrade/upgrade_test.go` uses `synctest` to avoid waiting 10s for each iteration of the retry loop, but I couldn't use it when the test was inside of `qtsuite` so I've removed the use of `qtsuite` for these tests. The use of synctest was surprisingly simple and worked well here.
2. I've also removed the example `UpgradeTo()` river worker that was initially present as an example and reorganized the files a bit. I've moved the tests out of the `river_test` package and into `river` when I noticed that the `new<worker>` functions were exported but will never be used outside of tests.

The commits split the work in two, the first commit is just the upgrade manager changes and the second shows the river changes.

Fixes [JUJU-9054](https://warthogs.atlassian.net/browse/JUJU-9054)
## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

[JUJU-9054]: https://warthogs.atlassian.net/browse/JUJU-9054?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ale8k:
Please undelte the migrationworker file, I've added my worker into that one. It'll make merge conflicts easier for us.

internal/jimm/upgrade/upgrade_test.go
```diff
@@ -429,11 +432,173 @@ func (s *upgradeManagerSuite) TestUpgradeTo_Success(c *qt.C) {
...
+		}
+		v := targetVersion
+		mi.AgentVersion = &v
+		return nil
+	})
+
+	s.api.EXPECT().UpgradeModel(modelUUID, targetVersion, "", false, false).Return(targetVersion, nil)
+
+	// Wrap the test in a synctest so that time advances instantly.
+	synctest.Test(t, (func(t *testing.T) {
```
ale8k, line 690: Can use:
```go
	tt, ok := (c.TB).(*testing.T)
	c.Assert(ok, qt.Equals, true)

	synctest.Test(tt, func(t *testing.T) {
		err = upgradeMgr.MigrateModel(ctx, &openfga.User{}, targetMt.Id(), targetController)
		c.Assert(err, qt.IsNil)
	})
```

And then you could use a suite.

kian99: Ah fair enough

internal/jimm/upgrade/upgrade.go
```diff
@@ -293,6 +295,68 @@ func (u *upgradeManager) MigrateAndUpgradeModel(ctx context.Context, user *openf
...
+				}
+				if mi.AgentVersion == nil {
+					return fmt.Errorf("model agent version is nil before upgrade")
+				}
+				if mi.AgentVersion.Compare(targetVersion) >= 0 {
+					// Already at target version or newer.
+					return nil
+				}
+
+				// UpgradeModel is safe to call multiple times.
```
ale8k, line 337: Should we run the precheck here? I forget which argument is which

internal/jimm/upgrade/upgrade.go
```diff
@@ -293,6 +295,68 @@ func (u *upgradeManager) MigrateAndUpgradeModel(ctx context.Context, user *openf
...
+				mi := jujuparams.ModelInfo{}
+				err := api.ModelInfo(ctx, &mi)
+				if err != nil {
+					return fmt.Errorf("failed to get model info before upgrade: %w", err)
+				}
+				if mi.AgentVersion == nil {
+					return fmt.Errorf("model agent version is nil before upgrade")
+				}
+				if mi.AgentVersion.Compare(targetVersion) >= 0 {
+					// Already at target version or newer.
```
ale8k, line 332: Zapctx log would help

internal/jimm/upgrade/upgrade.go
```diff
@@ -293,6 +295,68 @@ func (u *upgradeManager) MigrateAndUpgradeModel(ctx context.Context, user *openf
...
+					return nil
+				}
+				if err != nil {
+					return errors.E(fmt.Errorf("failed to upgrade model: %w", err))
+				}
+
+				return fmt.Errorf("model upgrade started/in-progress")
+			},
+			NotifyFunc: func(lastError error, attempt int) {
+				zapctx.Info(ctx, "model upgrade attempt", zap.Error(lastError), zap.Int("attempt", attempt))
```
ale8k: Maybe make it debug, info is probably a bit much

internal/jimm/upgrade/upgrade.go
```diff
@@ -293,6 +295,68 @@ func (u *upgradeManager) MigrateAndUpgradeModel(ctx context.Context, user *openf
...
+				}
+				if jujuerrors.Is(err, jujuerrors.AlreadyExists) {
+					// Model is already upgraded
+					return nil
+				}
+				if err != nil {
+					return errors.E(fmt.Errorf("failed to upgrade model: %w", err))
+				}
+
+				return fmt.Errorf("model upgrade started/in-progress")
```
ale8k: Shouldn't we errors.E this for consistency? And the two at the top too.

internal/jimm/upgrade/upgrade_test.go
ale8k: Yeah prefer to just leave it as a suite for less changes. Can remove suite in a subsequent PR based on your feedback from Francesco?

internal/jimm/upgrade/upgrade_test.go
```diff
@@ -429,11 +432,173 @@ func (s *upgradeManagerSuite) TestUpgradeTo_Success(c *qt.C) {
...
+		err := upgradeMgr.UpgradeModel(ctx, modelUUID, targetVersion)
+		c.Assert(err, qt.IsNil)
+	}))
+}
+
+func TestUpgradeModel_AlreadyExistsErrFromUpgradeIsOK(t *testing.T) {
```
ale8k: TestUpgradeModel_HasAlreadyBeenUpgradedReturnsNoError or something. Calling the test  AlreadyExistsErr... means I gotta read code to know what you're testing here.

internal/jimm/upgrade/upgrade_test.go
```diff
@@ -429,11 +432,173 @@ func (s *upgradeManagerSuite) TestUpgradeTo_Success(c *qt.C) {
...
+		v := targetVersion
+		mi.AgentVersion = &v
+		return nil
+	})
+
+	err = upgradeMgr.UpgradeModel(ctx, modelUUID, targetVersion)
+	c.Assert(err, qt.IsNil)
+}
+
+func TestUpgradeModel_RetriesUntilModelReportsTargetVersion(t *testing.T) {
```
ale8k, line 646: Just RetriesUntilUpgraded is OK aha.

internal/river/migrationworker_test.go
ale8k, line 1: Yeah please undelete so our history is easier.  Or perhaps we merge mine first, then you can rebase mine in? Sorry I didn't create a new file so I guess it is kind of my fault.

internal/river/upgradeworker.go
```diff
@@ -0,0 +1,45 @@
...
+
+	return &upgradeWorker{
+		migrationManager: migrationManager,
+	}, nil
+}
+
+// UpgradeArgs defines the arguments for the upgradeWorker job.
+type UpgradeArgs struct {
+	ModelUUID     string         `json:"model-uuid"`
+	TargetVersion version.Number `json:"target-version"`
```
ale8k, line 24: Does this have JSON tags? Surprised it does lol.

internal/river/upgradeworker_test.go
```diff
@@ -0,0 +1,43 @@
...
+	qt "github.com/frankban/quicktest"
+	"github.com/juju/version/v2"
+	"github.com/riverqueue/river"
+	"github.com/riverqueue/river/riverdriver/riverdatabasesql"
+	"github.com/riverqueue/river/rivertest"
+	"github.com/riverqueue/river/rivertype"
+	gomock "go.uber.org/mock/gomock"
+)
+
+func TestUpgradeWorker(t *testing.T) {
```
ale8k, line 16: Error test is missing.

internal/river/utils_test.go
```diff
@@ -0,0 +1,18 @@
...
+
+	"github.com/canonical/jimm/v3/internal/db"
+	"github.com/canonical/jimm/v3/internal/testutils/jimmtest"
+	qt "github.com/frankban/quicktest"
+)
+
+func setupTestDB(c *qt.C) *db.Database {
+	db := &db.Database{
+		DB: jimmtest.PostgresDB(c, time.Now),
+	}
```
ale8k, line 14: Db isn't migrated?

internal/jimm/upgrade/upgrade.go
```diff
@@ -293,6 +295,68 @@ func (u *upgradeManager) MigrateAndUpgradeModel(ctx context.Context, user *openf
...
+				}
+				if mi.AgentVersion.Compare(targetVersion) >= 0 {
+					// Already at target version or newer.
+					return nil
+				}
+
+				// UpgradeModel is safe to call multiple times.
+				_, err = api.UpgradeModel(modelUUID, targetVersion, "", false, false)
+				if jujuparams.IsCodeUpgradeInProgress(err) {
+					err = errors.E("upgrade in progress")
```
SimoneDutto, line 340: from this retry i get that we are "waiting" for.

As i suggested in the other pr i would make a distinction:
- fatal errors and we rely on River to retry instead of the retry pkg
- use of the retry pkg for business logic retry

wtdy?

internal/jimm/upgrade/upgrade_test.go
```diff
@@ -429,11 +432,173 @@ func (s *upgradeManagerSuite) TestUpgradeTo_Success(c *qt.C) {
...
+		model.Controller = dbmodel.Controller{Name: "ctrl"}
+		return nil
+	})
+
+	s.dialer.EXPECT().
+		Dial(ctx, gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).
+		Return(s.api, nil)
+
+	modelInfoCalls := 0
+	s.api.EXPECT().ModelInfo(ctx, gomock.Any()).Times(2).DoAndReturn(func(ctx context.Context, mi *jujuparams.ModelInfo) error {
```
SimoneDutto, line 675: cant' we just:
```
s.api.EXPECT().ModelInfo(ctx, gomock.Any()).Times(2).DoAndReturn(func(ctx context.Context, mi *jujuparams.ModelInfo) error {
   v := oldVersion
			mi.AgentVersion = &v
			return nil
}
s.api.EXPECT().ModelInfo(ctx, gomock.Any()).Times(2).DoAndReturn(func(ctx context.Context, mi *jujuparams.ModelInfo) error {
   v := targetVersion
	mi.AgentVersion = &v
	return nil
}
```
so we avoid the counter

SimoneDutto, line 675: cant' we just:
```
s.api.EXPECT().ModelInfo(ctx, gomock.Any()).Times(2).DoAndReturn(func(ctx context.Context, mi *jujuparams.ModelInfo) error {
   v := oldVersion
			mi.AgentVersion = &v
			return nil
}
s.api.EXPECT().ModelInfo(ctx, gomock.Any()).Times(2).DoAndReturn(func(ctx context.Context, mi *jujuparams.ModelInfo) error {
   v := targetVersion
	mi.AgentVersion = &v
	return nil
}
```
so we avoid the counter

SimoneDutto (commented):
looks good

internal/river/upgradeworker.go
kian99: I'm confused by this comment, do you mean a JSON marhsal method?

internal/jimm/upgrade/upgrade.go
kian99: What's the precheck? Do you mean dry-run? I don't think it's necessary to use the dry-run here since we're already in the middle of the job. But we can make it an param in the future.

internal/jimm/upgrade/upgrade_test.go
kian99: Gonna leave this one.

internal/jimm/upgrade/upgrade_test.go
kian99: Tweaked it :+1:

internal/jimm/upgrade/upgrade_test.go
kian99: Tried it, but copying exactly as you suggested failed, tried to tweak the second call to just 1 time and that was still failing so I've just left it as is.

internal/jimm/upgrade/upgrade.go
kian99: Yeah that sounds good, and I agree. But I wouldn't overcomplicate it right now so I'll leave it as is.

internal/jimm/upgrade/upgrade_test.go
kian99: Started reverting but it was more work than keeping the suite out, so I've just kept it removed.

ale8k: Sure thing, resolving this comment.

internal/jimm/upgrade/upgrade.go
ale8k: Sorry yeah dry run 🤦 and I think it is, just in case they modify the upgrade code in the future, right now it can take a lower version. But you've made me realise something. Could someone migrate a 2.9 -> 4.0? Is this possible? If it is, then the dry run result of this would be a 3.6 version in the event they migrated from 2.9 -> 4.0 and then we'd need 2 upgrades. If not, then I guess we're ok?

internal/river/upgradeworker.go
ale8k: Yeah I wasn't sure if it implemented json.Marshaler but it does, resolving.

internal/river/upgradeworker.go
```diff
@@ -0,0 +1,45 @@
...
+}
+
+// Kind returns the kind of the job.
+func (UpgradeArgs) Kind() string { return "migration" }
```
ale8k: Upgrade?  Also shouldn't it be consistent with the other worker, so "upgrade-upgrade" (My poor mans attempt at namespacing)

internal/river/upgradeworker.go
```diff
@@ -0,0 +1,45 @@
...
+
+import (
+	"context"
+
+	"github.com/canonical/jimm/v3/internal/errors"
+	"github.com/juju/version/v2"
+	"github.com/riverqueue/river"
+)
+
+func newUpgradeWorker(migrationManager UpgradeManager) (*upgradeWorker, error) {
```
ale8k: upgradeManager?

internal/river/river.go
```diff
@@ -33,14 +36,25 @@ func StartWorkers(
 	upgradeManager UpgradeManager,
 ) error {
 	workers := river.NewWorkers()
-	w, err := newUpgradeMigrationWorker(openfgaClient, db, upgradeManager)
+
+	migrationWorker, err := newMigrationWorker(openfgaClient, db, upgradeManager)
```
ale8k, line 40: Nit: Would be nice to have a func like "setupWorkers" which creates the `workers` and just adds all the workers in a loop.

ale8k (commented):
3 more minor comments

internal/jimm/upgrade/upgrade.go
kian99: I don't follow, wouldn't calling it with dry-run be the same thing as calling it without dry-run. Since this is a non-interactive environment, if the call with dry-run fails, it's not like you can go and change anything to fix it. I.e. just run it and it will return with an error if anything fails.

internal/river/upgradeworker.go
kian99: I'll rename it to upgrade for now.

internal/river/river.go
kian99: Yeah, we can think about adding that one day.

internal/jimm/upgrade/upgrade.go
ale8k: After a chat offline, resolved. It'll be fine and we're too late anyway so we may as well proceed with what we've got. Also JIMM won't even support 2.9 controllers soon.

ale8k (approved):
lgtm! Nice work!

SimoneDutto (approved):

PR #1826, opened 2026-01-21
test: refactor import model tests

ale8k:
## Description
Refactors import model tests.  I did a similar thing as I did previously where we validate the args are passed correctly.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

### DIFF
--- a/cmd/jaas/cmd/importmodel.go
+++ b/cmd/jaas/cmd/importmodel.go
@@ -3,6 +3,7 @@
 package cmd
 
 import (
+	"errors"
 	"fmt"
 
 	"github.com/juju/cmd/v3"
@@ -12,7 +13,6 @@ import (
 	"github.com/juju/juju/jujuclient"
 	"github.com/juju/names/v5"
 
-	"github.com/canonical/jimm/v3/internal/errors"
 	"github.com/canonical/jimm/v3/pkg/api"
 	apiparams "github.com/canonical/jimm/v3/pkg/api/params"
 )
@@ -83,7 +83,7 @@ func (c *importModelCommand) Init(args []string) error {
 
 	c.req.Controller = args[0]
 	if !names.IsValidModel(args[1]) {
-		return fmt.Errorf("invalid model uuid")
+		return errors.New("invalid model uuid")
 	}
 	c.req.ModelTag = names.NewModelTag(args[1]).String()
 	return nil
@@ -97,12 +97,12 @@ func (c *importModelCommand) Run(ctxt *cmd.Context) error {
 
 	jimmAPI, err := c.jimmAPIFunc()
 	if err != nil {
-		return errors.E(err, "could not create JIMM API client")
+		return fmt.Errorf("could not create JIMM API client: %w", err)
 	}
 	defer jimmAPI.Close()
 
 	if err := jimmAPI.ImportModel(&c.req); err != nil {
-		return errors.E(err)
+		return fmt.Errorf("could not import model: %w", err)
 	}
 	return nil
 }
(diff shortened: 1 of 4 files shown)


cmd/jaas/cmd/importmodel.go
```diff
@@ -90,19 +93,32 @@ func (c *importModelCommand) Init(args []string) error {
...
-		return errors.E(err, "could not determine controller")
+	if c.jimmAPIFunc == nil {
+		c.jimmAPIFunc = c.newClient
 	}
 
-	apiCaller, err := c.NewAPIRootWithDialOpts(c.store, currentController, "", c.dialOpts)
+	jimmAPI, err := c.jimmAPIFunc()
 	if err != nil {
-		return err
+		return errors.E(err, "could not create JIMM API client")
```
alesstimec: returning `errors.E` with a string message will override any message in the underlying error.. 
i'd propose `fmt.Errorf("could not create JIMM API client: %w", err)`

cmd/jaas/cmd/importmodel.go
```diff
@@ -90,19 +93,32 @@ func (c *importModelCommand) Init(args []string) error {
...
+	if err := jimmAPI.ImportModel(&c.req); err != nil {
 		return errors.E(err)
 	}
 	return nil
 }
+
+func (c *importModelCommand) newClient() (JIMMAPI, error) {
+	currentController, err := c.store.CurrentController()
+	if err != nil {
+		return nil, errors.E(fmt.Errorf("could not determine controller: %v", err))
```
alesstimec: change `%v` to `%w`?

alesstimec: or just return `err`

cmd/jaas/cmd/importmodel_test.go
```diff
@@ -1,134 +1,116 @@
...
+	}
 
-	var model2 dbmodel.Model
-	model2.SetTag(names.NewModelTag(m.ModelUUID()))
-	err = s.JIMM.Database.GetModel(context.Background(), &model2)
-	c.Assert(err, gc.Equals, nil)
-	c.Check(model2.OwnerIdentityName, gc.Equals, "charlie@canonical.com")
-}
+	fs := gnuflag.NewFlagSet("test", gnuflag.ContinueOnError)
+	command.SetFlags(fs)
```
kian99: I believe you can delete this, it's done inside of `InitCommand`.

kian99 (approved):

cmd/jaas/cmd/importmodel_test.go
ale8k: Ah yeah, I left htis by mistake when trying to set the flags manually. Good spot!

cmd/jaas/cmd/importmodel.go
```diff
@@ -37,25 +38,27 @@ func NewImportModelCommand() cmd.Command {
 	cmd := &importModelCommand{
 		store: jujuclient.NewFileClientStore(),
 	}
+	cmd.jimmAPIFunc = cmd.newClient
```
alesstimec, line 40: this isn't needed as you have a check on L96

cmd/jaas/cmd/importmodel.go
```diff
@@ -37,25 +38,27 @@ func NewImportModelCommand() cmd.Command {
...
 
 	return modelcmd.WrapBase(cmd)
 }
 
 // importModelCommand imports a model.
 type importModelCommand struct {
 	modelcmd.ControllerCommandBase
-	store    jujuclient.ClientStore
-	dialOpts *jujuapi.DialOpts
+	store jujuclient.ClientStore
```
alesstimec: i only just noticed that `ControllerCommandBase` already has a store field and a `SetClientStore` method, so we could drop this field, right?

alesstimec (approved):
LGTM with a few comments - perhaps we need to discuss the `store` field

cmd/jaas/cmd/importmodel.go
ale8k: The others are like this, it's more for consistency / a bit of c/p.

cmd/jaas/cmd/importmodel.go
ale8k: Maybe... But all of our other tests are like this. So I think it'd be a follow up? Be nice if we could.

PR #1827, opened 2026-01-21
refactor: update jimm juju ref to 3.6.13

ale8k:
## Description
Updates JIMM to 3.6.13 to include DryRun flag of migrates. Fixed references where I could.

## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests


cmd/jaas/cmd/migratemodel.go
```diff
@@ -354,7 +354,7 @@ func (j jujuMigrateAPI) Close() error {
 // InitiateMigration implements the MigrateAPI interface.
 func (j jujuMigrateAPI) InitiateMigration(spec controllerapi.MigrationSpec) (string, error) {
 	client := controllerapi.NewClient(j.apiCaller)
-	return client.InitiateMigration(spec)
+	return client.InitiateMigration(spec, false)
```
alesstimec, line 356: why don't we add the dry run parameter to the InitiateMigration of the jujuMigrateAPI?

alesstimec (approved):
LGTM but with a question

cmd/jaas/cmd/migratemodel.go
ale8k: I guess we can as a follow up, but that'd be out of the scope of this?

internal/jimm/juju/migrationtarget_test.go
kian99, line 1: I believe you need to revert these description changes, the tests expect description v9

ale8k: Ah... But 
```go
// ModelInfo is used to report basic details about a model.
type ModelInfo struct {
	UUID                   string
	Owner                  names.UserTag
	Name                   string
	AgentVersion           version.Number
	ControllerAgentVersion version.Number
	ModelDescription       description.Model <- this is v10 now?
}
```

So what do we do here?

kian99: I'll take a look tomorrow and let you know :+1:

kian99: Although off the top of my head, I recall we use our own custom struct that handles both v9 and v10 internally. The MigratingModelInfo struct holds a `[]byte`. I believe you should be able to keep description v9 in these tests and it'll work. You can also use v10 but you probably need to make some other change in that case. I'll pull the branch and check tomorrow.

go.mod
```diff
@@ -10,7 +10,7 @@ require (
 	github.com/juju/errors v1.0.0
 	github.com/juju/gnuflag v1.0.0
 	github.com/juju/http/v2 v2.0.1
-	github.com/juju/juju v0.0.0-20251119074349-d8a727fd219f // 3.6.12
+	github.com/juju/juju v0.0.0-20260114112920-3c6b4d6fa55b // 3.6.13
```
kian99: Should this say v3.6.14 now instead?

kian99 (approved):

go.mod
ale8k: I put it to 13, this PR was pr 14

ale8k: Bumped, just tidying & will push.

PR #1829, opened 2026-01-23
refactor: adjust behaviour for bootstrap on public clouds

kian99:
## Description
This PR resolves some TODOs in the bootstrap and upgrade logic. Specifically the one that was targetted mentioned,
> // TODO: Check if it is public cloud here and set PersonalCloud accordingly.
> // For now, we're always setting it.

I checked whether our bootstrap logic supports bootstrapping onto public clouds. I was mostly able to successfully bootstrap to AWS with `./jaas bootstrap aws/us-east-1 my-test-controller 3.6.12`. I say _mostly_ because the machine was provisioned but the bootstrap failed near the end when the controller couldn't reach `https://jimm.localhost/.well-known/jwks.json` which was obviously only running locally. I'm not keen on deploying everything to AWS just to test this further as I was able to prove to myself this works and the code indicates the same.

Essentially the problem described in the TODO is that we shoudn't pass the cloud information to the StartBootstrap method if the cloud is a public cloud (i.e. AWS, Azure, etc). While it's not necessary to send the cloud info in these cases, it also doesn't do any harm, the bootstrap logic only writes these details to the filesystem if the cloud is not a public one. I did some renaming of the `PersonalCloud` field to just `Cloud` with some docstrings to hopefully make it clear (we pass these params around a lot so only some of the fields have docstrings).

In the end, the only substantial change I made was to the CLI, where we now skip sending the cloud information if it's a public cloud. This change is not required, since again, the server will decide whether it needs to use the cloud information. But I decided it was better to omit it if we know it's not needed.

I've also done some minor refactoring of the bootstrap command tests to accompany the above change.
 
Fixes [JUJU-8932](https://warthogs.atlassian.net/browse/JUJU-8932)

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

[JUJU-8932]: https://warthogs.atlassian.net/browse/JUJU-8932?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cmd/jaas/cmd/bootstrap.go
+++ b/cmd/jaas/cmd/bootstrap.go
@@ -135,8 +135,6 @@ func (c *bootstrapCommand) SetFlags(f *gnuflag.FlagSet) {
 	f.Var(&c.config, "config",
 		"Specify a configuration file, or one or more configuration options.\n    (`--config config.yaml [--config key=value ...])`")
 	f.BoolVar(&c.detach, "detach", false, "If set, the command will start the bootstrap job and return immediately with the job ID, without waiting for the job to complete.")
-	// TODO(ale8k): Support passing cloud & cloudcredential files, for now we're looking up clouds and credentials added to the store.
-	// See cmd/juju/cloud/add.go L311 on a nice way to do this and credential will be somewhere in there too.
 }
 
 // Info implements modelcmd.Command.
@@ -157,6 +155,20 @@ func (c *bootstrapCommand) Run(ctxt *cmd.Context) error {
 	bootstrapCloud, err := jujucloud.CloudByName(c.cloud)
 	if err != nil {
 		return fmt.Errorf("failed to get cloud %q: %w", c.cloud, err)
+	}
+
+	// If the cloud is a public cloud (AWS, Azure, etc), we clear bootstrapCloud to avoid sending
+	// unnecessary info, letting the server decide the cloud endpoints, etc.
+	// Regardless, the server uses its own cloud definition if it identifies the cloud is a public cloud.
+	publicClouds, _, err := jujucloud.PublicCloudMetadata(jujucloud.JujuPublicCloudsPath())
+	if err != nil {
+		return fmt.Errorf("failed to get public cloud metadata: %w", err)
+	}
+	for name := range publicClouds {
+		if name == c.cloud {
+			bootstrapCloud = nil
+			break
+		}
 	}
 
 	cloudCreds, err := c.ClientStore().CredentialForCloud(c.cloud)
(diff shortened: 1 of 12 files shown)


cmd/jaas/cmd/bootstrap.go
```diff
@@ -161,6 +159,20 @@ func (c *bootstrapCommand) Run(ctxt *cmd.Context) error {
...
+	publicClouds, _, err := jujucloud.PublicCloudMetadata(jujucloud.JujuPublicCloudsPath())
+	if err != nil {
+		return fmt.Errorf("failed to get public cloud metadata: %w", err)
+	}
+	for name := range publicClouds {
+		if name == c.cloud {
+			bootstrapCloud = nil
+			break
+		}
+	}
```
kian99, line 172: This change **can** be removed, I was on the fence about making it.

cmd/jaas/cmd/bootstrap_test.go
```diff
@@ -78,12 +78,69 @@ func TestBootstrapArgParsing(t *testing.T) {
...
+func TestBootstrapApiParams(t *testing.T) {
+	c := qt.New(t)
+	s := setupCmdMocks(t)
+
+	// Override the Juju data dir to a temp dir. The cloud-related operations
+	// are not available on the store interface.
+	tmpJujuDir := t.TempDir()
+	defer os.RemoveAll(tmpJujuDir)
+	oldJujuDir := osenv.SetJujuXDGDataHome(tmpJujuDir)
+	defer osenv.SetJujuXDGDataHome(oldJujuDir)
```
SimoneDutto, line 133: this is not concurrency safe, in the future if we see they are flaky we might need to bring the `safeClient` implementation here

SimoneDutto (approved):

cmd/jaas/cmd/bootstrap_test.go
kian99: Yeah correct, but I think we won't need to run these unit tests in parallel anyway, they're very fast as is.

cmd/jaas/cmd/bootstrap.go
alesstimec: should we move the public cloud check to the beginning of the `Run` method and avoid calling `jujucloud.CloudByName`?

alesstimec (approved):
LGTM

cmd/jaas/cmd/bootstrap.go
kian99: Not a bad idea, I've tried it now but it seems clearer in the current order so I'll leave it.

PR #1834, opened 2026-01-26
docs: add metadata for SEO

tmihoc:
One of our roadmap commitments this cycle is to get our docs in a better shape for SEO. Part of this is adding metadata to all of our docs. This PR adds them to all the non-autogenerated files.

Note 1: I used this AI prompt that we have in the documentation team for this purpose: https://github.com/canonical/AI-resources-for-docs/blob/main/docs/prompts/add-metadata-descriptions.md The results are not amazing but, imo, good enough.

Note 2: We probably don't need to worry about the autogenerated ones, as all the other docs link to them systematically as is.

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -1,3 +1,9 @@
+---
+myst:
+  html_meta:
+    description: "Learn how to contribute to JAAS documentation by opening issues, creating pull requests, and improving docs for the community."
+---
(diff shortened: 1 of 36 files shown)


ale8k (approved):
Had a scim through, looks good to me.

luci1900 (approved):
You weren't kidding about "good enough, not great".

PR #1835, opened 2026-01-27
feat: add facadeversions tool for detecting facade version diffs betwen JIMM and Juju

kian99:
## Description
This PR introduces a new tool in `cmd/` called `facadeversions`. The tool serves two purposes,
1. Generate a reference file of facade versions that JIMM supports in a format similar to Juju's apiversions.go file.
2. Compare JIMM's facade versions against Juju with options to emit a non-zero exit code when JIMM is lagging.

The code was mostly AI generated once I figured out how to generate the facade versions in JIMM. The changes include a brief README showing usage and tests for the diff logic. 

Overall this should make it easier to detect changes between facade versions in JIMM and Juju, in a follow-up I'd like to add a CI check that we are not lagging behind Juju.

Here is an example output using the `--error-on-version-lag` flag.
```
$ go run ./cmd/facadeversions compare --error-on-version-lag
ApplicationOffers: jimm=4 juju=5
Controller: jimm=11 juju=12
MigrationTarget: jimm=4 juju=6
ModelManager: jimm=9 juju=10
exit status 1
```

Partially addresses [JUJU-8964](https://warthogs.atlassian.net/browse/JUJU-8964)

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

[JUJU-8964]: https://warthogs.atlassian.net/browse/JUJU-8964?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cmd/facadeversions/generator/generator.go
kian99, line 1: This file is in a separate package. It arguably doesn't need to be but without a strong opinion on it myself, I've left it here where it was generated.

luci1900 (approved):
Sure

cmd/facadeversions/generator/generator.go
```diff
@@ -0,0 +1,92 @@
+// Copyright 2026 Canonical.
+
+package generator
```
ale8k, line 3: Looks good, but I think an additional feature is to check we're using a released version of juju, and if we aren't, the exit code should be 0 even if there's a mismatch. That way we may go bleeding edge, and should Juju have any issues, we won't accidentally think "oh I need to implement this in JIMM"? As we should only care about stable Juju?

ale8k (commented):
lgtm but 1 feature request

cmd/facadeversions/generator/generator.go
ale8k: Also FYI I briefly skimmed the code, there's a lot here...

kian99: > That way we may go bleeding edge, and should Juju have any issues, we won't accidentally think "oh I need to implement this in JIMM"?

I think for these cases we can implement a stub method in JIMM that returns nil or an error. We might even have a test that verifies verifies an error is returned (or maybe the test is just skipped in the interim), and then we'd need to revisit it. Wdyt?

ale8k: Yeah that's fine just thought maybe we could see if AI could chuck a little GH query in for us, up to you, idm.

ale8k (approved):

cmd/facadeversions/generator/generator.go
kian99: I'll leave it for now, keeps it a bit simpler.

SimoneDutto (approved):
i'm actually amazed we have to have so much code to do a diff between two maps. Maybe AI overthought this a bit or it's just golang super verbose and lacking support for simple map's operations.

Anyways, if it's useful it's just a utils

kian99:
> i'm actually amazed we have to have so much code to do a diff between two maps. Maybe AI overthought this a bit or it's just golang super verbose and lacking support for simple map's operations.
> 
> Anyways, if it's useful it's just a utils

Imo it's doing a bit more than a diff between two maps. There's the base command `go run ./cmd/facadeversions compare` which returns,
```
$ go run ./cmd/facadeversions compare
Facades only in Juju:
  - Action
  - ActionPruner
  - ... (a bunch more facades)
Facades only in JIMM:
  - JIMM
Version differences (Juju vs JIMM):
  - ApplicationOffers: missing [5]
  - Controller: missing [12]
  - MigrationTarget: missing [1, 2, 3, 5, 6]
  - ModelManager: missing [10]
```
And there's running it with the lag flag,
```
$ go run ./cmd/facadeversions compare --error-on-version-lag
ApplicationOffers: jimm=4 juju=5
Controller: jimm=11 juju=12
MigrationTarget: jimm=4 juju=6
ModelManager: jimm=9 juju=10
```
So the formatting and prettifying of the output is a good chunk of the logic.

kian99:
Closing, merged in https://github.com/canonical/jimm/pull/1837

PR #1836, opened 2026-01-27
feat: introduce supervising worker for upgrade-to

ale8k:
## Description
Introduces a supervising worker for upgrade-to within river. I had to integration test this because rivers utilities don't facilitate the use of  the river client within work functions. As such, we're running river integration style and testing the flow of the job. Namely, a straight success, failure modes after all retries on both nested jobs and finally transitive successes after N failures.

I guess the negative is that we're polling the database within the tests to wait for finalisation, but the benefit here is we're pretty darn sure it is working as expected - ultimately meaning in callers of this worker need not test the worker semantics of river.

## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests

### DIFF
--- a/internal/river/upgradetoworker.go
+++ b/internal/river/upgradetoworker.go
@@ -35,11 +35,6 @@ type upgradeToWorker struct {
 	upgradeRetries int
 }
 
-// Migrate the model
-// Wait for job - handle crash
-// Upgrade the model
-// Wait for job
-// Done
 func (w *upgradeToWorker) Work(ctx context.Context, job *river.Job[UpgradeToArgs]) error {
 	client := river.ClientFromContext[*sql.Tx](ctx)
 
--- a/internal/river/upgradetoworker.go
+++ b/internal/river/upgradetoworker.go
@@ -147,24 +147,26 @@ func waitForJobToFinalise(ctx context.Context, result *rivertype.JobInsertResult
 			if !ok {
 				return errors.New("event channel closed unexpectedly")
 			}
-			if event.Job.ID == result.Job.ID {
-				if event.Job.FinalizedAt != nil {
-					switch event.Kind {
-					// Because we've finalised, this isn't an attempt failure, but the final state.
-					case river.EventKindJobFailed:
-						// Job failed, return the last error.
-						if len(event.Job.Errors) != 0 {
-							return errors.New(event.Job.Errors[len(event.Job.Errors)-1].Error)
-						}
-						return errors.New("job failed without error details")
-					case river.EventKindJobCancelled:
-						return errors.New("job was cancelled")
-					case river.EventKindJobCompleted:
-						// Completed successfully.
-						return nil
-					}
-				}
+
+			if event.Job.ID != result.Job.ID || event.Job.FinalizedAt == nil {
+				continue
 			}
+
+			switch event.Kind {
+			// Because we've finalised, this isn't an attempt failure, but the final state.
+			case river.EventKindJobFailed:
+				// Job failed, return the last error.
+				if len(event.Job.Errors) != 0 {
+					return errors.New(event.Job.Errors[len(event.Job.Errors)-1].Error)
+				}
+				return errors.New("job failed without error details")
+			case river.EventKindJobCancelled:
+				return errors.New("job was cancelled")
+			case river.EventKindJobCompleted:
+				// Completed successfully.
+				return nil
+			}
+
 		}
 	}
 }
--- a/internal/river/upgradetoworker.go
+++ b/internal/river/upgradetoworker.go
@@ -79,7 +79,7 @@ func (w *upgradeToWorker) Work(ctx context.Context, job *river.Job[UpgradeToArgs
 		return err
 	}
 
-	if err := w.waitForJobToFinalise(migRes, eventCh); err != nil {
+	if err := w.waitForJobToFinalise(ctx, migRes, eventCh); err != nil {
 		return err
 	}
 
@@ -100,7 +100,7 @@ func (w *upgradeToWorker) Work(ctx context.Context, job *river.Job[UpgradeToArgs
 		return err
 	}
 
-	if err := w.waitForJobToFinalise(upgradeRes, eventCh); err != nil {
+	if err := w.waitForJobToFinalise(ctx, upgradeRes, eventCh); err != nil {
 		return err
 	}
 
@@ -108,7 +108,7 @@ func (w *upgradeToWorker) Work(ctx context.Context, job *river.Job[UpgradeToArgs
 	return nil
 }
 
-func (w *upgradeToWorker) waitForJobToFinalise(result *rivertype.JobInsertResult, eventCh <-chan *river.Event) error {
+func (w *upgradeToWorker) waitForJobToFinalise(ctx context.Context, result *rivertype.JobInsertResult, eventCh <-chan *river.Event) error {
 	// It may be a duplicate, so check if it has finalised. If not, wait for it to do so.
 	if result.Job.FinalizedAt != nil {
 		// It has finalised, check it's state, if it failed return error.
@@ -120,21 +120,29 @@ func (w *upgradeToWorker) waitForJobToFinalise(result *rivertype.JobInsertResult
 			return errors.New(result.Job.Errors[len(result.Job.Errors)-1].Error)
 		}
 	} else {
-		for event := range eventCh {
-			if event.Job.ID == result.Job.ID {
-				if event.Job.FinalizedAt != nil {
-					switch event.Kind {
-					// Because we've finalised, this isn't an attempt failure, but the final state.
-					case river.EventKindJobFailed:
-						// Job failed, return the last error.
-						if len(event.Job.Errors) != 0 {
-							return errors.New(event.Job.Errors[len(event.Job.Errors)-1].Error)
+		for {
+			select {
+			case <-ctx.Done():
+				return ctx.Err()
+			case event, ok := <-eventCh:
+				if !ok {
+					return nil
+				}
+				if event.Job.ID == result.Job.ID {
+					if event.Job.FinalizedAt != nil {
+						switch event.Kind {
+						// Because we've finalised, this isn't an attempt failure, but the final state.
+						case river.EventKindJobFailed:
+							// Job failed, return the last error.
+							if len(event.Job.Errors) != 0 {
+								return errors.New(event.Job.Errors[len(event.Job.Errors)-1].Error)
+							}
+						case river.EventKindJobCancelled:
+							return errors.New("job was cancelled")
+						case river.EventKindJobCompleted:
+							// Completed successfully.
+							return nil
 						}
-					case river.EventKindJobCancelled:
-						return errors.New("job was cancelled")
-					case river.EventKindJobCompleted:
-						// Completed successfully.
-						return nil
 					}
 				}
 			}
--- a/internal/river/upgradetoworker.go
+++ b/internal/river/upgradetoworker.go
@@ -1,3 +1,5 @@
+// Copyright 2026 Canonical.
+
 package river
 
 import (
@@ -10,13 +12,18 @@ import (
 	"github.com/riverqueue/river/rivertype"
 )
 
-func newUpgradeToWorker(migrateRetries int, upgradeRetries int) *upgradeToWorker {
+// waitForJobFinalisationFunc is a function that waits for a job to finalise.
+type waitForJobFinalisationFunc func(ctx context.Context, result *rivertype.JobInsertResult, eventCh <-chan *river.Event) error
+
+func newUpgradeToWorker(migrateRetries int, upgradeRetries int, finaliser waitForJobFinalisationFunc) *upgradeToWorker {
 	return &upgradeToWorker{
 		migrateRetries: migrateRetries,
 		upgradeRetries: upgradeRetries,
+		finaliser:      finaliser,
 	}
 }
 
+// UpgradeToArgs are the arguments for the upgrade-to worker.
 type UpgradeToArgs struct {
 	ModelUUID            string         `json:"model-uuid" river:"unique"`
 	TargetVersion        version.Number `json:"target-version"`
@@ -24,6 +31,7 @@ type UpgradeToArgs struct {
 	TargetControllerName string         `json:"target_controller_name"`
 }
 
+// Kind implements the [river.JobArgs] interface.
 func (UpgradeToArgs) Kind() string { return "upgrade-to" }
 
 type upgradeToWorker struct {
@@ -33,8 +41,11 @@ type upgradeToWorker struct {
 	migrateRetries int
 	// upgradeRetries is the number of times to retry the upgrade step.
 	upgradeRetries int
+	// finaliser is a function that waits for a job to finalise.
+	finaliser waitForJobFinalisationFunc
 }
 
+// Work implements the [river.Worker] interface.
 func (w *upgradeToWorker) Work(ctx context.Context, job *river.Job[UpgradeToArgs]) error {
 	client := river.ClientFromContext[*sql.Tx](ctx)
 
@@ -79,7 +90,7 @@ func (w *upgradeToWorker) Work(ctx context.Context, job *river.Job[UpgradeToArgs
 		return err
 	}
 
-	if err := w.waitForJobToFinalise(ctx, migRes, eventCh); err != nil {
+	if err := w.finaliser(ctx, migRes, eventCh); err != nil {
 		return err
 	}
 
@@ -100,7 +111,7 @@ func (w *upgradeToWorker) Work(ctx context.Context, job *river.Job[UpgradeToArgs
 		return err
 	}
 
-	if err := w.waitForJobToFinalise(ctx, upgradeRes, eventCh); err != nil {
+	if err := w.finaliser(ctx, upgradeRes, eventCh); err != nil {
 		return err
 	}
 
@@ -108,7 +119,13 @@ func (w *upgradeToWorker) Work(ctx context.Context, job *river.Job[UpgradeToArgs
 	return nil
 }
 
-func (w *upgradeToWorker) waitForJobToFinalise(ctx context.Context, result *rivertype.JobInsertResult, eventCh <-chan *river.Event) error {
+// waitForJobToFinalise waits for the job to finalise, that is, a job that will no longer
+// be retried but could have succeeded or failed after all attempts. It does so by checking
+// the event channel for updates.
+//
+// If the job has been inserted already on a previous attempt, it checks if it's finalised already,
+// and if not, waits for it to do so.
+func waitForJobToFinalise(ctx context.Context, result *rivertype.JobInsertResult, eventCh <-chan *river.Event) error {
 	// It may be a duplicate, so check if it has finalised. If not, wait for it to do so.
 	if result.Job.FinalizedAt != nil {
 		// It has finalised, check it's state, if it failed return error.
@@ -150,6 +167,4 @@ func (w *upgradeToWorker) waitForJobToFinalise(ctx context.Context, result *rive
 			}
 		}
 	}
-
-	return nil
 }
--- a/internal/river/upgradetoworker.go
+++ b/internal/river/upgradetoworker.go
@@ -12,14 +12,14 @@ import (
 	"github.com/riverqueue/river/rivertype"
 )
 
-// waitForJobFinalisationFunc is a function that waits for a job to finalise.
-type waitForJobFinalisationFunc func(ctx context.Context, result *rivertype.JobInsertResult, eventCh <-chan *river.Event) error
+// awaitCompletionFunc is a function that waits for a job to finalise.
+type awaitCompletionFunc func(ctx context.Context, result *rivertype.JobInsertResult, eventCh <-chan *river.Event) error
 
-func newUpgradeToWorker(migrateRetries int, upgradeRetries int, finaliser waitForJobFinalisationFunc) *upgradeToWorker {
+func newUpgradeToWorker(migrateRetries int, upgradeRetries int, awaitFunc awaitCompletionFunc) *upgradeToWorker {
 	return &upgradeToWorker{
-		migrateRetries: migrateRetries,
-		upgradeRetries: upgradeRetries,
-		finaliser:      finaliser,
+		migrateRetries:  migrateRetries,
+		upgradeRetries:  upgradeRetries,
+		awaitCompletion: awaitFunc,
 	}
 }
 
@@ -34,6 +34,23 @@ type UpgradeToArgs struct {
 // Kind implements the [river.JobArgs] interface.
 func (UpgradeToArgs) Kind() string { return "upgrade-to" }
 
+// InsertOpts implements the [river.JobArgsWithInsertOpts] interface.
+func (UpgradeToArgs) InsertOpts() river.InsertOpts {
+	return river.InsertOpts{
+		MaxAttempts: 3,
+		UniqueOpts: river.UniqueOpts{
+			ByArgs: true,
+			ByState: []rivertype.JobState{
+				rivertype.JobStateAvailable,
+				rivertype.JobStatePending,
+				rivertype.JobStateRunning,
+				rivertype.JobStateRetryable,
+				rivertype.JobStateScheduled,
+			},
+		},
+	}
+}
+
 type upgradeToWorker struct {
 	river.WorkerDefaults[UpgradeToArgs]
 
@@ -41,8 +58,8 @@ type upgradeToWorker struct {
 	migrateRetries int
 	// upgradeRetries is the number of times to retry the upgrade step.
 	upgradeRetries int
-	// finaliser is a function that waits for a job to finalise.
-	finaliser waitForJobFinalisationFunc
+	// awaitCompletion is a function that waits for a job to finalise.
+	awaitCompletion awaitCompletionFunc
 }
 
 // Work implements the [river.Worker] interface.
@@ -56,7 +73,7 @@ func (w *upgradeToWorker) Work(ctx context.Context, job *river.Job[UpgradeToArgs
 	)
 	defer cancel()
 
-	migRes, err := client.Insert(
+	migrateInsertResponse, err := client.Insert(
 		ctx,
 		migrationWorkerArgs{
 			Username:             job.Args.Username,
@@ -81,14 +98,14 @@ func (w *upgradeToWorker) Work(ctx context.Context, job *river.Job[UpgradeToArgs
 		return err
 	}
 
-	if err := w.finaliser(ctx, migRes, eventCh); err != nil {
+	if err := w.awaitCompletion(ctx, migrateInsertResponse, eventCh); err != nil {
 		return err
 	}
 
 	// We need not worry about a crash here and the migrate completing, as a new job will be inserted, sure,
 	// but our idempotency of the migrate service will ensure it is a no-op.
 
-	upgradeRes, err := client.Insert(
+	upgradeInsertResponse, err := client.Insert(
 		ctx,
 		upgradeArgs{
 			ModelUUID:     job.Args.ModelUUID,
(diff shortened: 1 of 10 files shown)


internal/river/upgradetoworker.go
```diff
@@ -0,0 +1,152 @@
...
+
+func newUpgradeToWorker(migrateRetries int, upgradeRetries int) *upgradeToWorker {
+	return &upgradeToWorker{
+		migrateRetries: migrateRetries,
+		upgradeRetries: upgradeRetries,
+	}
+}
+
+type UpgradeToArgs struct {
+	ModelUUID            string         `json:"model-uuid"`
```
kian99: Should this field also be marked as river: unique? I.e. we can't have 2 supervising jobs for the same model at the same time.

internal/river/upgradetoworker.go
```diff
@@ -0,0 +1,152 @@
...
+	//
+	// We could poll the job, as well as use subscriptions for quick wake ups between polls, but that adds complexity for
+	// what might be an impossible scenario for us.
+	//
+	// It would look something like: tick on X seconds, and pull from ticker & sub channel together, whichever is first determines
+	// if the job has completed. This covers us for dropped subscriptions.
+	//
+	// Open for discussion within the PR.
+	eventCh, cancel := client.Subscribe(river.EventKindJobCompleted, river.EventKindJobCancelled, river.EventKindJobFailed)
+	defer cancel()
```
kian99: I'm okay with the subscribe and the risk of dropping messages due to the size of the queue.

internal/river/upgradetoworker.go
```diff
@@ -0,0 +1,152 @@
...
+		// It has finalised, check it's state, if it failed return error.
+		//
+		// TODO: We should report all errors from all attempts and their details (i.e., timestamps)
+		// For now, simply take the last failure attempt.
+		// This should be completed in Phase 3.
+		if len(result.Job.Errors) != 0 {
+			return errors.New(result.Job.Errors[len(result.Job.Errors)-1].Error)
+		}
+	} else {
+		for {
```
kian99: Let's reduce the nesting on the code below.

internal/river/upgradetoworker.go
```diff
@@ -0,0 +1,152 @@
+package river
```
alesstimec, line 3: // Copyright

package river

internal/river/upgradetoworker.go
```diff
@@ -0,0 +1,152 @@
...
+			return errors.New(result.Job.Errors[len(result.Job.Errors)-1].Error)
+		}
+	} else {
+		for {
+			select {
+			case <-ctx.Done():
+				return ctx.Err()
+			case event, ok := <-eventCh:
+				if !ok {
+					return nil
```
kian99: Should this be an error rather?

internal/river/upgradetoworker.go
```diff
@@ -0,0 +1,152 @@
...
+)
+
+func newUpgradeToWorker(migrateRetries int, upgradeRetries int) *upgradeToWorker {
+	return &upgradeToWorker{
+		migrateRetries: migrateRetries,
+		upgradeRetries: upgradeRetries,
+	}
+}
+
+type UpgradeToArgs struct {
```
alesstimec, line 27: godoc for this and all exported methods in this file, please

internal/river/upgradetoworker.go
```diff
@@ -0,0 +1,152 @@
...
+				}
+				if event.Job.ID == result.Job.ID {
+					if event.Job.FinalizedAt != nil {
+						switch event.Kind {
+						// Because we've finalised, this isn't an attempt failure, but the final state.
+						case river.EventKindJobFailed:
+							// Job failed, return the last error.
+							if len(event.Job.Errors) != 0 {
+								return errors.New(event.Job.Errors[len(event.Job.Errors)-1].Error)
+							}
```
kian99: If the length is == 0 we don't return an error even though we're in the `EventKindJobFailed` state.

internal/river/upgradetoworker_test.go
```diff
@@ -0,0 +1,280 @@
...
+			attempt++
+			switch attempt {
+			case 1:
+				return errors.New("unexpected-error")
+			case 2:
+				return errors.New("unexpected-error")
+			case 3:
+				return errors.New("migration-failed3")
+			default:
+				return errors.New("unexpected-error")
```
kian99: Could you replace this with return fmt.Errorf("migration-failed-%d", attempt) and then lower down you can expect a string based on the retry count. That way if we change the retry count the test still works (provided the other logic remains the same).

internal/river/upgradetoworker_test.go
```diff
@@ -0,0 +1,280 @@
...
+	c.Assert(riverClient.Start(ctx), qt.IsNil)
+	c.Cleanup(func() {
+		err := riverClient.Stop(context.Background())
+		c.Check(err, qt.IsNil)
+	})
+
+	return riverClient, u.Name
+}
+
+func waitForFinalisedJob(c *qt.C, ctx context.Context, client *river.Client[*sql.Tx], jobID int64) *rivertype.JobRow {
```
kian99: Can we use river's subscribe method here to simplify things? Or maybe we'd miss the job because we'd subscribe after the job was inserted. If that's a problem, then I'd just lower the polling to like 100/200ms.

internal/river/upgradetoworker.go
```diff
@@ -0,0 +1,152 @@
...
+	// what might be an impossible scenario for us.
+	//
+	// It would look something like: tick on X seconds, and pull from ticker & sub channel together, whichever is first determines
+	// if the job has completed. This covers us for dropped subscriptions.
+	//
+	// Open for discussion within the PR.
+	eventCh, cancel := client.Subscribe(river.EventKindJobCompleted, river.EventKindJobCancelled, river.EventKindJobFailed)
+	defer cancel()
+
+	migRes, err := client.Insert(
```
kian99: If the supervising job crashes and restarts, how do we ensure it doesn't insert another migrate/upgrade job? I think we still need to handle that bit.

kian99 (commented):
Looks good, just a couple of minor things and one important one on what happens when the supervising job crashes.

internal/river/upgradetoworker.go
```diff
@@ -0,0 +1,152 @@
...
+	client := river.ClientFromContext[*sql.Tx](ctx)
+
+	// PR Note:
+	// Subscribe to job events immediately to prevent duplicate completion race as we'll catch it when inserting it
+	// in a moment.
+	//
+	// River's initial buffer size for the subscription channel is 1000 [river.subscribeChanSizeDefault].
+	// River does the following when pushing events:
+	// select {
+	// case sub.Chan <- event:
```
alesstimec: i'd only do polling, then, not both.. maybe we can trigger polling based on received events AND a ticker..

internal/river/upgradetoworker.go
```diff
@@ -0,0 +1,152 @@
...
+	if result.Job.FinalizedAt != nil {
+		// It has finalised, check it's state, if it failed return error.
+		//
+		// TODO: We should report all errors from all attempts and their details (i.e., timestamps)
+		// For now, simply take the last failure attempt.
+		// This should be completed in Phase 3.
+		if len(result.Job.Errors) != 0 {
+			return errors.New(result.Job.Errors[len(result.Job.Errors)-1].Error)
+		}
+	} else {
```
alesstimec: `return nil` after L121, then you can lose the `else` clause and reduce indent for the rest of this method

internal/river/upgradetoworker.go
```diff
@@ -0,0 +1,152 @@
...
+		for {
+			select {
+			case <-ctx.Done():
+				return ctx.Err()
+			case event, ok := <-eventCh:
+				if !ok {
+					return nil
+				}
+				if event.Job.ID == result.Job.ID {
```
alesstimec:
```
if event.Job.ID != result.JobID || event.Job.FinalizedAt == nil {
   continue
}
```
then you can reduce the indent for the `switch event.Kind` to improve readability of this method

internal/river/upgradetoworker_test.go
```diff
@@ -0,0 +1,280 @@
+package river
+
+import (
+	"context"
+	"database/sql"
+	"errors"
+	"testing"
+	"time"
+
+	"github.com/canonical/jimm/v3/internal/db"
```
alesstimec: jimm imports should go into a separate import block

internal/river/upgradetoworker_test.go
```diff
@@ -0,0 +1,280 @@
+package river
```
alesstimec, line 3: // Copyright...

package river

internal/river/upgradetoworker.go
ale8k: Oh yes, exactly my bad. Good spot!

internal/river/upgradetoworker.go
ale8k: Once it has approvals, I'll remove the comment. Will leave this comment unresolved for others to comment.

internal/river/upgradetoworker.go
ale8k: Yeeeeeep...

internal/river/upgradetoworker.go
ale8k: I'm not sure the if even makes sense after looking over this again, I think it's only failed explicitly, which is to return the error or rivers system to retry it after a crash in which case it'll append to the errors. I'll leave it is for "robustness" and add an explicit error return, but I'm pretty sure it'll never actually happen.

internal/river/upgradetoworker_test.go
ale8k: No no we can subscribe first as it is buffered up to 1k, I've just been having issues detecting completion (hence the super long polling). I did initially try subscribing but I had timing issues why I didn't figure out. I'll go back and try figure out why.

ale8k: It's working, was a small thing. Resolving!

internal/river/upgradetoworker.go
ale8k: It should be OK because of the unique insertion on the dependents, that is, it'll just run over them and subscribe again.

I did try write a test to simulate our job crashing but forcing a panic didn't work how I expected... I'll give it another crack and try get a test in to simulate the supervisor crashing.

ale8k: Think I've got an idea to write a test for this, give me a little while. Will push it tomorrow as it'll involve some trickery aha.

internal/river/upgradetoworker.go
ale8k: Please comment on the comment from kian here: https://github.com/canonical/jimm/pull/1836#discussion_r2732365508 to discuss.

internal/river/upgradetoworker.go
ale8k: Added test which covers this.

internal/river/upgradetoworker_test.go
```diff
@@ -120,7 +119,7 @@ func TestUpgradeToWorker_UpgradeFails(t *testing.T) {
 	upgradeManager := NewMockUpgradeManager(ctrl)
 
 	// Retry a few times to ensure retries work as expected and surface the LAST error.
-	riverClient, username := setupWorkers(c, ctx, database, upgradeManager, sqlDB, 1, 3)
+	riverClient, username := setupWorkers(c, ctx, database, upgradeManager, sqlDB, 1, 3, waitForJobToFinalise)
```
kian99: What do we think about passing in a struct to this `setupWorkers` function. Would aid in readability.

internal/river/upgradetoworker_test.go
```diff
@@ -0,0 +1,442 @@
...
+	)
+
+	upgradeManager.EXPECT().
+		MigrateModel(gomock.Any(), gomock.Any(), "model-uuid", "target-controller").
+		DoAndReturn(func(context.Context, *openfga.User, string, string) error {
+			// We have to prevent the migrate worker from finalising
+			// so we can see that upon restart of the supervisor, it isn't
+			// inserting a duplicate and is as such crash resillient.
+			//
+			// Wait here until the supervisor restarts to send the signal to complete.
```
kian99, line 464: That's one scenario right? But there are some other scenarios, like the migrate job finishes, supervisor starts the upgrade job, then supervisor crashes. 

I think even that scenario should be fine because the supervisor will start a new migrate job, that job will return quickly because it is idempotent and the model was already migrated, and then supervisor moves onto the upgrade job. Which will either be in progress or completed which we should be able to handle.

I'm still a little confused on one thing, but I'll elaborate separately.

internal/river/upgradetoworker_test.go
```diff
@@ -0,0 +1,442 @@
...
+		// Set max attempts to 2 so it can retry once after the crash.
+		&river.InsertOpts{MaxAttempts: 2},
+	)
+	c.Assert(err, qt.IsNil)
+
+	// The flow of what is happening here is:
+	// 1. Supervisor starts
+	// 2. Inserts migrate job which is blocked by channel.
+	// 3. Supervisor panics first time waiting on the migrate.
+	// 4. Supervisor restarts, and attempts to insert migrate job again, but it's a duplicate.
```
kian99, line 488: When a job is still running, and we try to insert a new one, river sees that it's a duplicate and the supervisor fails right? I.e. I assume that's what happens with `upgradeRes, err := client.Insert()` if an upgrade job is already running for that model UUID?

How does the supervising job get past this stage of retrying to insert a duplicate job? Does it just keep retrying the insert until the job is done, then it can sucessfully insert a new job (which will finish immediately because it's idempotent) and then the supervisor says "all done"?

If my understanding above is correct, the one issue we could face is the retrying supervisor times out while the pending job is still busy. I.e. supervisor tries 3 times to insert a job, gets "duplicate job" error and fails overall, meanwhile the upgrade job was just taking a long time to complete.

internal/river/upgradetoworker.go
alesstimec: I'm ok with subscribe, since this is only used to determine when a job is completed (or failed). Maybe we could add a simple timeout to the finaliser method to so that it polls should that timeout be reached (in case we miss the event).

internal/river/upgradetoworker.go
```diff
@@ -0,0 +1,172 @@
...
+	// what might be an impossible scenario for us.
+	//
+	// It would look something like: tick on X seconds, and pull from ticker & sub channel together, whichever is first determines
+	// if the job has completed. This covers us for dropped subscriptions.
+	//
+	// Open for discussion within the PR.
+	eventCh, cancel := client.Subscribe(river.EventKindJobCompleted, river.EventKindJobCancelled, river.EventKindJobFailed)
+	defer cancel()
+
+	migRes, err := client.Insert(
```
alesstimec: what does `migRes` stand for? can we think of a more intuitive variable name?

internal/river/upgradetoworker.go
```diff
@@ -0,0 +1,172 @@
...
+			UniqueOpts: river.UniqueOpts{
+				ByArgs: true,
+			},
+		},
+	)
+	if err != nil {
+		return err
+	}
+
+	if err := w.finaliser(ctx, migRes, eventCh); err != nil {
```
alesstimec: `finaliser` makes it sound like this is a type implementing an interface. Perhaps `awaitCompletion`?

internal/river/upgradetoworker.go
```diff
@@ -104,6 +91,13 @@ func (w *upgradeToWorker) Work(ctx context.Context, job *river.Job[UpgradeToArgs
 			MaxAttempts: w.upgradeRetries,
 			UniqueOpts: river.UniqueOpts{
 				ByArgs: true,
+				ByState: []rivertype.JobState{
```
kian99: Can you add a comment to mention that `rivertype.JobStateCompleted` has been omitted to allow this job type to be re-inserted once it has completed.

internal/river/upgradetoworker_test.go
```diff
@@ -60,6 +60,97 @@ func TestUpgradeToWorker_Success(t *testing.T) {
...
+		MigrateModel(gomock.Any(), gomock.Any(), "model-uuid", "target-controller").
+		Return(nil)
+	upgradeManager.EXPECT().
+		UpgradeModel(gomock.Any(), "model-uuid", version.MustParse("2.0.0")).
+		Return(nil)
+
+	sub, cancel := riverClient.Subscribe(river.EventKindJobCompleted)
+	c.Cleanup(cancel)
+
+	insRes, err := riverClient.Insert(ctx, UpgradeToArgs{
```
kian99, line 110: Nit, can we wrap this insert of the UpgradeTo job in a helper function. It would make reading the test a lot easier without all the river boilerplate in the way

internal/river/upgradetoworker.go
```diff
@@ -0,0 +1,166 @@
...
+	// migrateRetries is the number of times to retry the migration step.
+	migrateRetries int
+	// upgradeRetries is the number of times to retry the upgrade step.
+	upgradeRetries int
+	// finaliser is a function that waits for a job to finalise.
+	finaliser waitForJobFinalisationFunc
+}
+
+// Work implements the [river.Worker] interface.
+func (w *upgradeToWorker) Work(ctx context.Context, job *river.Job[UpgradeToArgs]) error {
```
kian99, line 76: I think a few lines of commentary on the upgradeTo worker will prove useful. It can go here on the Work function or elsewhere. Something along the lines of,
> // Each `upgradeTo` job acts as a orchestrator of several child jobs, starting them and waiting for their completion.
> // River's unique-job args are setup to ensure that in-progress jobs are not re-inserted, keyed by the model UUID and job state.
// Each child job is expected to be idempotent so that in certain edge cases where an orchestrator restart would cause re-insertion of a completed job, no changes are made.
>

internal/river/upgradetoworker.go
```diff
@@ -0,0 +1,166 @@
...
+			MaxAttempts: w.migrateRetries,
+			UniqueOpts: river.UniqueOpts{
+				ByArgs: true,
+				ByState: []rivertype.JobState{
+					rivertype.JobStateAvailable,
+					rivertype.JobStatePending,
+					rivertype.JobStateRunning,
+					rivertype.JobStateRetryable,
+					rivertype.JobStateScheduled,
+				},
```
kian99: I'm thinking we can make this list a common var to re-use between the migrate and upgrade jobs?

internal/river/upgradetoworker.go
kian99: +1, not for sounding like an interface but because I believe finaliser also has some existing meaning in Go where you can setup finalizers to run on garbage collection.
Maybe just `w.waitJob()`

internal/river/upgradetoworker.go
```diff
@@ -0,0 +1,166 @@
...
+// If the job has been inserted already on a previous attempt, it checks if it's finalised already,
+// and if not, waits for it to do so.
+func waitForJobToFinalise(ctx context.Context, result *rivertype.JobInsertResult, eventCh <-chan *river.Event) error {
+	// It may be a duplicate, so check if it has finalised. If not, wait for it to do so.
+	if result.Job.FinalizedAt != nil {
+		// It has finalised, check it's state, if it failed return error.
+		//
+		// TODO: We should report all errors from all attempts and their details (i.e., timestamps)
+		// For now, simply take the last failure attempt.
+		// This should be completed in Phase 3.
```
kian99: I'd remove this TODO, if we do it then great, if not it's probably not a big deal until someone realises it's worthwhile for real-world reasons and does it.

internal/river/upgradetoworker_test.go
```diff
@@ -0,0 +1,533 @@
...
+		ModelUUID:            "model-uuid",
+		TargetVersion:        version.MustParse("2.0.0"),
+		Username:             username,
+		TargetControllerName: "target-controller",
+	},
+		&river.InsertOpts{
+			MaxAttempts: 1,
+			UniqueOpts: river.UniqueOpts{
+				ByArgs:  true,
+				ByState: []rivertype.JobState{rivertype.JobStateAvailable, rivertype.JobStatePending, rivertype.JobStateRunning, rivertype.JobStateRetryable, rivertype.JobStateScheduled},
```
kian99: It's interesting that the upgradeTo worker needs these state args as well. Makes sense but makes you think once we call `river.Insert` from JIMM's domain, that is pushing a fair bit of "river specific knowledge" into the jimm package. Maybe we'll want to add a wrapper for inserting the upgradeTo job.

internal/river/upgradetoworker.go
```diff
@@ -0,0 +1,166 @@
...
+	migRes, err := client.Insert(
+		ctx,
+		migrationWorkerArgs{
+			Username:             job.Args.Username,
+			UUID:                 job.Args.ModelUUID,
+			TargetControllerName: job.Args.TargetControllerName,
+		},
+		&river.InsertOpts{
+			MaxAttempts: w.migrateRetries,
+			UniqueOpts: river.UniqueOpts{
```
alesstimec: Looking at https://riverqueue.com/docs/unique-jobs#unique-properties we can have `migrationWorkerArgs` implement the `JobArgsWithInsertOpts` interface, meaning that uniqueness requirements would be defined there, not here, where we are trying to insert the job.. Makes more sense to me that uniqueness be defined by a job, not when we insert it, wdyt?

so have `migrationWorkerArgs` implement
```
func (migrationWorkerArgs) InsertOps() river.InsertOpts {
  return &river.InsertOpts{
			MaxAttempts: w.migrateRetries,
			UniqueOpts: river.UniqueOpts{
				ByArgs: true,
				ByState: []rivertype.JobState{
					rivertype.JobStateAvailable,
					rivertype.JobStatePending,
					rivertype.JobStateRunning,
					rivertype.JobStateRetryable,
					rivertype.JobStateScheduled,
				},
			},
   }
}
```
(don't know if i got the brackets right, but you get the idea, rigth?)

internal/river/upgradetoworker.go
```diff
@@ -0,0 +1,166 @@
...
+		ctx,
+		upgradeArgs{
+			ModelUUID:     job.Args.ModelUUID,
+			TargetVersion: job.Args.TargetVersion,
+		},
+		&river.InsertOpts{
+			MaxAttempts: w.upgradeRetries,
+			UniqueOpts: river.UniqueOpts{
+				ByArgs: true,
+				ByState: []rivertype.JobState{
```
alesstimec: same comment about implementing the `JobArgsWithInsertOpts` for the `upgradeArgs`

internal/river/upgradeworker.go
```diff
@@ -18,25 +20,27 @@ func newUpgradeWorker(upgradeManager UpgradeManager) (*upgradeWorker, error) {
 	}, nil
 }
 
-// UpgradeArgs defines the arguments for the upgradeWorker job.
-type UpgradeArgs struct {
-	ModelUUID     string         `json:"model-uuid"`
+// upgradeArgs defines the arguments for the upgradeWorker job.
+type upgradeArgs struct {
```
alesstimec: why do we have `migrationWorkerArgs`, but here we call this type `upgradeArgs`.. i suggest using `upgradeWorkerArgs`

internal/river/upgradetoworker.go
```diff
@@ -0,0 +1,166 @@
...
+func newUpgradeToWorker(migrateRetries int, upgradeRetries int, finaliser waitForJobFinalisationFunc) *upgradeToWorker {
+	return &upgradeToWorker{
+		migrateRetries: migrateRetries,
+		upgradeRetries: upgradeRetries,
+		finaliser:      finaliser,
+	}
+}
+
+// UpgradeToArgs are the arguments for the upgrade-to worker.
+type UpgradeToArgs struct {
```
alesstimec, line 27: suggestion to make argument naming similar for all workers: `UpgradeToWorkerArgs`

alesstimec: also, does this type need any uniqueness guarantees? if so, implement the `InsertOpts` method

internal/river/upgradetoworker_test.go
```diff
@@ -0,0 +1,533 @@
...
+				continue loop
+			}
+			// We've caught the suppervising job entering "some state".
+			// Capture it, and break out.
+			supervisingJobFailureUpdate = event.Job
+			break loop
+		case <-ctx.Done():
+			c.Fatal("timed out waiting for job failed event")
+		}
+	}
```
kian99: To simplify this, what about putting it in a function like `getSupervisorJob(supervisingJobID)` and then you can do,
```go
for {
	select {
	case event := <-sub:
		if event.Job.ID == supervisingJobId {
			return event.Job
		}

	case <-ctx.Done():
		c.Fatal("timed out waiting for job failed event")
	}
}
```

kian99 (approved):

internal/river/upgradetoworker.go
ale8k: Great shout, on it!

internal/river/upgradetoworker.go
ale8k: Done

internal/river/upgradeworker.go
ale8k: Done

internal/river/upgradetoworker.go
ale8k: Done

internal/river/upgradetoworker_test.go
ale8k: I've moved to the opts now, so second arg is nil. If we wrap in a function now it would just be equivalent (and sometimes we change the arg), so I'll leave it in its new state as I think that's probably what you wanted anyways?

internal/river/upgradetoworker.go
ale8k: Modified it to be a little more explicit but added.

internal/river/upgradetoworker.go
ale8k: Fixed with Ales' comment.

internal/river/upgradetoworker.go
ale8k: Removed.

internal/river/upgradetoworker_test.go
ale8k: Fixed with Ales' comment

internal/river/upgradetoworker.go
```diff
@@ -61,51 +92,34 @@ func (w *upgradeToWorker) Work(ctx context.Context, job *river.Job[UpgradeToArgs
...
-				},
-			},
 		},
 	)
 	if err != nil {
 		return err
 	}
 
-	if err := w.finaliser(ctx, migRes, eventCh); err != nil {
+	if err := w.awaitCompletion(ctx, migrateInsertResponse, eventCh); err != nil {
```
kian99, line 101: I like the name change, awaitJob is also a nice option.

internal/river/upgradetoworker.go
```diff
@@ -61,51 +92,34 @@ func (w *upgradeToWorker) Work(ctx context.Context, job *river.Job[UpgradeToArgs
...
 		return err
 	}
 
-	upgradeRes, err := client.Insert(
+	// We need not worry about a crash here and the migrate completing, as a new job will be inserted, sure,
+	// but our idempotency of the migrate service will ensure it is a no-op.
```
kian99: You can delete this since it's covered in the docstring above

internal/river/upgradetoworker_test.go
```diff
@@ -0,0 +1,585 @@
...
+func waitForSupervisingJob(c *qt.C, ctx context.Context, sub <-chan *river.Event, supervisingJobId int64) *rivertype.JobRow {
+	for {
+		select {
+		case event := <-sub:
+			if event.Job.ID != supervisingJobId {
+				continue
+			}
+			// We've caught the suppervising job entering "some state".
+			// Capture it, and break out.
+			return event.Job
```
kian99: Can simplify further
```suggestion
			if event.Job.ID == supervisingJobId {
			  return event.Job
			}
```

internal/river/upgradetoworker_test.go
```diff
@@ -0,0 +1,585 @@
...
+
+import (
+	"context"
+	"database/sql"
+	"errors"
+	"fmt"
+	"sync"
+	"testing"
+
+	"github.com/canonical/jimm/v3/internal/db"
```
alesstimec: sort imports, please

alesstimec (approved):

PR #1838, opened 2026-01-27
fix(show-model): removes the store field from the show-model command

alesstimec:
## Description

The embedded `ControllerCommandBase` already has a `store` field, so we do not need an additional one.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

## Test instructions

1. make qa-lxd
2. juju add-model test
3. juju jaas show-model jimm-test@canonical.com/test

### DIFF
--- a/cmd/jaas/cmd/showmodel.go
+++ b/cmd/jaas/cmd/showmodel.go
@@ -38,10 +38,8 @@ The output includes the model name, model UUID, controller name, and controller 
 
 // NewShowModelCommand returns a command to display model controller information.
 func NewShowModelCommand() cmd.Command {
-	cmd := &showModelCommand{
-		store: jujuclient.NewFileClientStore(),
-	}
-
+	cmd := &showModelCommand{}
+	cmd.SetClientStore(jujuclient.NewFileClientStore())
 	return modelcmd.WrapBase(cmd)
 }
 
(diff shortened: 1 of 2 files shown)


cmd/jaas/cmd/showmodel.go
```diff
@@ -38,10 +38,8 @@ The output includes the model name, model UUID, controller name, and controller
 
 // NewShowModelCommand returns a command to display model controller information.
 func NewShowModelCommand() cmd.Command {
-	cmd := &showModelCommand{
-		store: jujuclient.NewFileClientStore(),
-	}
-
+	cmd := &showModelCommand{}
+	cmd.SetClientStore(jujuclient.NewFileClientStore())
```
kian99, line 42: Looks like this just does the same as the above,
```go
// SetClientStore implements the ControllerCommand interface.
func (c *ControllerCommandBase) SetClientStore(store jujuclient.ClientStore) {
	c.store = store
}
```

cmd/jaas/cmd/showmodel_test.go
```diff
@@ -18,9 +18,9 @@ import (
 
 func runShowModelCommand(c *qt.C, mocks *cmdMocks, args ...string) (string, error) {
 	showCmd := showModelCommand{
-		store:  mocks.store,
 		client: mocks.client,
 	}
+	showCmd.SetClientStore(mocks.store)
```
kian99, line 23: Also not needed right? Since we have access to the command base store above.

kian99 (approved):
This probably applies to all the other commands too.

SimoneDutto (approved):

cmd/jaas/cmd/showmodel.go
alesstimec: i am calling this method

PR #1843, opened 2026-01-30
refactor: improve river test retries

kian99:
## Description
This PR makes some minor improvements to our River tests and DB migrations:
1. Speeds up our River tests. 2 tests in the upgrade_test worker were taking 20s and 40s because they were retrying a job and River's default retry policy is 1s->16s->1m21s, etc. See https://riverqueue.com/docs/job-retries. Now running `go test ./internal/river -count=1` takes around 12s on my machine. Previously it was taking ~70s
2. When running the tests with verbose=true, one would see `ERROR: database "jimm_test_testmigrationworker_error_6fgoa36q" is being accessed by other users`. This was because we weren't commiting/rolling-back a tx we started with `tx, err := sqlDb.Begin()`. That's been fixed. I've also simplified the helper we use to create the test database.
3. I've removed `handleDeprecatedMigrations` which handled scenarios where a deployment was running a now very old version of JIMM that didn't use golang-migrate and then upgraded to a version of JIMM that did. We can be quite confident that no users are running this old release so the code can finally go. At the time we likely didn't have external users anyway and it was added out of an abundance of caution.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

### DIFF
--- a/internal/river/migrationworker_test.go
+++ b/internal/river/migrationworker_test.go
@@ -24,9 +24,7 @@ func TestMigrationWorker(t *testing.T) {
 
 	upgradeManager := NewMockUpgradeManager(ctrl)
 
-	db := setupTestDB(c)
-	sqlDb, err := db.SqlDB()
-	c.Assert(err, qt.IsNil)
+	db, sqlDb := setupTestDB(c)
 
 	openfgaClient := &openfga.OFGAClient{}
 	w, err := newMigrationWorker(openfgaClient, db, upgradeManager)
@@ -45,6 +43,7 @@ func TestMigrationWorker(t *testing.T) {
 
 	tx, err := sqlDb.Begin()
 	c.Assert(err, qt.IsNil)
+	defer func() { _ = tx.Rollback() }()
 
 	result, err := testWorker.Work(
 		c.Context(),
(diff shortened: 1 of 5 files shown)


luci1900 (approved):
Deleted code and rollback tests, what's not to like?

internal/river/migrationworker_test.go
```diff
@@ -45,6 +43,7 @@ func TestMigrationWorker(t *testing.T) {
 
 	tx, err := sqlDb.Begin()
 	c.Assert(err, qt.IsNil)
+	defer func() { _ = tx.Rollback() }()
```
SimoneDutto, line 46: this is the "success" test case, maybe we should `commit`?

SimoneDutto (approved):
good good

internal/river/migrationworker_test.go
kian99: Since the test works based on the data in the tx, the test still passes/works without commit so no need.
This is also a way of doing tests I've seen in tools like Django where each test runs in a transaction and then the transaction is rolled back at the end of the test, allowing you to re-use the DB quickly.

PR #1844, opened 2026-01-30
fix(middleware): added client credential login to middleware

alesstimec:
## Description

Enables client credential login to http endpoints we proxy.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

## Test instructions

1. apply the plan below
2. change resource revision to 2 and reapply the plan
3. 
```
terraform {
	required_providers {
		juju = {
			source  = "juju/juju"
			version = "1.2.0"
		}
	}
	required_version = ">= 1.5.0"
}


provider "juju" {
	client_id     = "test-client-id"
	client_secret = "2M2blFbO4GX4zfggQpivQSxwWX1XGgNf"
	controller_addresses = "jimm.localhost:443"
}

# Pull existing cloud credential for LXD from local Juju file store
resource "juju_credential" "lxd" {
	name   = "lxd-credential"

    cloud {
      name = "lxd"
    }

	auth_type = "certificate"

	attributes = {
		client-cert = <<-EOT
<REDACTED>
		EOT
		client-key  = <<-EOK
<REDACTED>
		EOK
		server-cert = <<-EOS
<REDACTED>
		EOS
	}
}


resource "juju_model" "test-model" {
	name        = "test-model"

	credential = juju_credential.lxd.name

    cloud {
      name = "lxd"
    }
}

resource "juju_application" "test" {
	name 	 = "test-app"
	model_uuid = juju_model.test-model.uuid

	charm {
		name = "juju-qa-test"
		revision = 26
		channel = "latest/stable"
	}

	resources = {
		"foo-file" = "1"
	}
}
```

### DIFF
--- a/internal/middleware/authn.go
+++ b/internal/middleware/authn.go
@@ -19,15 +19,16 @@ import (
 // identityContextKey is the unique key to extract user from context for basic-auth authentication
 type identityContextKey struct{}
 
-// JIMMAuthner is an interface that requires authentication methods from JIMM.
-type JIMMAuthner interface {
+// Authenticator is an interface that requires authentication methods from JIMM.
+type Authenticator interface {
 	AuthenticateBrowserSession(context.Context, http.ResponseWriter, *http.Request) (context.Context, error)
+	LoginClientCredentials(ctx context.Context, clientID string, clientSecret string) (*openfga.User, error)
 	LoginWithSessionToken(ctx context.Context, sessionToken string) (*openfga.User, error)
 	UserLogin(ctx context.Context, identityName string) (*openfga.User, error)
 }
 
 // AuthenticateViaCookie performs browser session authentication and puts an identity in the request's context
-func AuthenticateViaCookie(next http.Handler, jimm JIMMAuthner) http.Handler {
+func AuthenticateViaCookie(next http.Handler, jimm Authenticator) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		ctx, err := jimm.AuthenticateBrowserSession(r.Context(), w, r)
 		if err != nil {
@@ -50,7 +51,7 @@ var unauthenticatedEndpoints = map[string]struct{}{
 // verifies that the user is a JIMM admin. Note that the method needs the base
 // URL to decide if the request does not require authentication; this is to
 // safeguard against conflicting/similar endpoint names in the future.
-func AuthenticateRebac(baseURL string, next http.Handler, jimm JIMMAuthner) http.Handler {
+func AuthenticateRebac(baseURL string, next http.Handler, jimm Authenticator) http.Handler {
 	cookieAuthenticator := AuthenticateViaCookie(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		ctx := r.Context()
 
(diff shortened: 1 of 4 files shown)


SimoneDutto (approved):
rip: JIMMAuthner

luci1900 (approved):
That's nice.

internal/middleware/authn.go
```diff
@@ -89,23 +90,28 @@ func AuthenticateRebac(baseURL string, next http.Handler, jimm JIMMAuthner) http
...
+		// first try session token authentication
 		user, err := jimm.LoginWithSessionToken(ctx, password)
-		if err != nil {
-			w.WriteHeader(http.StatusUnauthorized)
-			_, _ = w.Write([]byte("error authenticating the user"))
-			return
+		if err == nil {
+			next.ServeHTTP(w, r.WithContext(withIdentity(ctx, user)))
+		}
+		// then try client credentials authentication
```
ale8k: I'd prefer an explicit method for client creds, I wouldn't consider them basic auth as that implies it is a person. It also doesn't read well as the method is called `AuthenticateWithSessionTokenViaBasicAuth`, so oyou could call it `AuthenticateWithSessionTokenViaBasicAuthOrClientCredentials` or just make a new method.

internal/middleware/authn.go
```diff
@@ -87,25 +88,30 @@ func AuthenticateRebac(baseURL string, next http.Handler, jimm JIMMAuthner) http
...
-// AuthenticateWithSessionTokenViaBasicAuth performs basic auth authentication and puts an identity in the request's context.
+// AuthenticateViaBasicAuth performs basic auth authentication and puts an identity in the request's context.
 // The basic-auth is composed of an empty user, and as a password a jwt token that we parse and use to authenticate the user.
-func AuthenticateWithSessionTokenViaBasicAuth(next http.Handler, jimm JIMMAuthner) http.Handler {
+func AuthenticateViaBasicAuth(next http.Handler, jimm Authenticator) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		ctx := r.Context()
 		// extract auth token
-		_, password, ok := r.BasicAuth()
+		username, password, ok := r.BasicAuth()
```
kian99, line 99: Agreed with Alex that we are overloading things here. It would make more sense to parse the `Authorization` header. If it's basic auth, i.e. `Basic <base64value>` we treat it as a client-id/secret, and if it's `Bearer <token>` we treat it as a session token. The auth schemes are covered [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Authentication). Unfortunately we can't change the handling for tokens since that would break backwards compatibility.

We could change the format for the ClientCredentials login provider since only JIMM parses it and clearly that doesn't work yet. I.e. on the Juju client side use a different scheme rather than `Basic`. There aren't any good schemes that cover this though. And it also takes me back to the idea that JIMM "technically" shouldn't receive a client-id/secret, the client should exchange those with the identity server for a token, then it's always bearer auth. 

A long term option would be to change the Juju client's SessionTokenLoginProvider to send a `Bearer <token>` header and support both for a time. A bunch of effort for not much gain though.

In the end, what you've done is fine. To avoid the try option 1 then try option 2, you could choose to check if `username != ""` and do the client-credentials login else do SessionToken login but that's a nit imo.

internal/middleware/authn.go
```diff
@@ -87,25 +88,30 @@ func AuthenticateRebac(baseURL string, next http.Handler, jimm JIMMAuthner) http
 	})
 }
 
-// AuthenticateWithSessionTokenViaBasicAuth performs basic auth authentication and puts an identity in the request's context.
+// AuthenticateViaBasicAuth performs basic auth authentication and puts an identity in the request's context.
 // The basic-auth is composed of an empty user, and as a password a jwt token that we parse and use to authenticate the user.
```
kian99: I'd expand the docstring to explain why we have to try one option then another.

kian99 (approved):

PR #1845, opened 2026-02-02
feat: wire river into JIMM's upgrade to handler

kian99:
## Description
This PR wires up the River client into the JIMM domain layer, allowing us to insert river jobs as necessary.

The main challenge I faced with this was the potential for circular dependencies. To resolve this I've made a package called `rivertypes`, thanks @luci1900 for the discussion on this. This package holds the job arguments that are used when inserting a job using the river client. The package go doc explains this nicely,
> // River job args are analogous to a schema for traditional queue systems, defining what
// we will insert into the DB for processing later by workers. By defining them
// in a 3rd package we avoid circular dependencies between the domain layer and the

There was one more issue with the tests, to try and summarise
- `internal/river` depends on types from `internal/jimm/upgrade` for its `UpgradeTo` interface definition (though technically the current methods only use base types, but I've added a blank import to simulate the future problem).
- That revealed an import cycle in tests. The river tests create a DB using `internal/jimmtest/` helpers. The import cycle came from the fact that `internal/jimmtest/` imports `internal/jimm/` and `internal/jimm/` imports the thing river wrapper we have to make it easier to mock in tests. 
- The solution was simply to move the DB related helpers into` internal/jimmtest/dbtest/`

I recommend reviewing the commits separately.

Fixes [JUJU-9057](https://warthogs.atlassian.net/browse/JUJU-9057)

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

[JUJU-9057]: https://warthogs.atlassian.net/browse/JUJU-9057?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


luci1900 (approved):

internal/jimm/upgrade/upgrade.go
```diff
@@ -362,18 +281,20 @@ func (u *upgradeManager) UpgradeModel(ctx context.Context, modelUUID string, tar
...
-// This currently only works with personal, non-kubernetes clouds.
-// Further work in Phase 2 is expected to be done here: https://warthogs.atlassian.net/browse/JUJU-8918
-func (u *upgradeManager) UpgradeTo(ctx context.Context, user *openfga.User, modelUUID string, targetVersion version.Number) (version.Number, error) {
-	var chosenVersion version.Number
+// This currently only works with non-kubernetes clouds.
+func (u *upgradeManager) UpgradeTo(ctx context.Context, user *openfga.User, modelUUID string, targetVersion version.Number) (int64, error) {
 	var newControllerName = fmt.Sprintf("controller-%d", time.Now().Unix())
 
+	if targetVersion == version.Zero {
+		return 0, errors.E(errors.CodeBadRequest, "target version cannot be zero")
```
alesstimec, line 294: what's the added value of wrapping the error with `errors.E` in this case (and below)

internal/jimm/upgrade/upgrade.go
```diff
@@ -53,12 +54,18 @@ type Store interface {
 	GetModel(ctx context.Context, model *dbmodel.Model) (err error)
 }
 
+// UpgradeEnqueuer defines the method to enqueue an upgrade job.
+type UpgradeEnqueuer interface {
```
alesstimec, line 58: UpgradeJobManager?

internal/jimm/upgrade/upgrade.go
kian99: None (besides adding codes in this scenario), I think everyone just sticks errors.E everywhere since that's our standard approach. Which we'll hopefully get rid of soon.

internal/jimm/upgrade/upgrade.go
kian99: I wouldn't call it a manager since that term has some usage already in JIMM associated with the business logic domains.

SimoneDutto (approved):
nice! it's very similar to what I set up originally! Plus the rivertypes.

PR #1846, opened 2026-02-03
fix(login): return CodeSessionTokenInvalid error code when the token can't be decoded

SimoneDutto:
# Description

Before the "failed to decode token" error was considered fatal, however it seems it was useful to re-trigger a login, so we set the `CodeSessionTokenInvalid` code for this error as well.
Add a test as well to prevent changing this error code in the future

# QA

`make dev-env`

`juju login jimm.localhost -c jimm-dev`
Do the login.
Change the token: `vi ~/.local/share/juju/accounts.yaml`
`juju models`
Before:
`ERROR failed to decode token; illegal base64 data at input byte 198 (fatal login error)`

Now:
it triggers a re-login:
```
Please visit http://keycloak.localhost:8082/realms/jimm/device and enter code NXHL-YQLA to log in.
```

### DIFF
--- a/internal/auth/auth.go
+++ b/internal/auth/auth.go
@@ -410,7 +410,7 @@ func (as *AuthenticationService) VerifySessionToken(token string) (_ jwt.Token, 
 		if stderrors.Is(err, jwt.ErrTokenExpired()) {
 			return nil, errors.E(errors.CodeSessionTokenInvalid, "JIMM session token expired")
 		}
-		return nil, err
+		return nil, errors.E(errors.CodeSessionTokenInvalid, fmt.Errorf("failed to parse jwt: %v", err))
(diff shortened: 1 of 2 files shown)


kian99 (approved):
What about returning the same code if the jwt failed to parse and it wasn't an expired token?

internal/auth/auth.go
SimoneDutto: @kian here?

kian99: Yeah

SimoneDutto: done

luci1900 (approved):

SimoneDutto:
Closing this pr because it's actually beneficial to use these error as fatal.

PR #1837, opened 2026-01-27
refactor: bump facade version

kian99:
## Description

Currently this PR is based on top of #1835.

This PR updates the versions of facades that JIMM implements to align with the latest that Juju 3.6 offers. Each commit covers a separate facade. I took the time to investigate what changed in each version change and each commit description documents the change and whether it was safe to bump JIMM, with reasons for adding/not-adding support for the old version.

In short, we were okay to bump all the facade versions without adding support for the old versions. This does mean in theory you can't use, say, a 3.4.x with the latest JIMM which may have worked for some facades. Instead, you must have a relatively new version of a 3.6.x client but I believe this is okay.

Fixes [JUJU-8964](https://warthogs.atlassian.net/browse/JUJU-8964)

## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests
- [ ] Covered by integration tests

[JUJU-8964]: https://warthogs.atlassian.net/browse/JUJU-8964?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1848, opened 2026-02-03
chore: re-add ModelManager v9 and Controller v11 facades

kian99:
## Description
This PR tweaks the versions of facades we support for the `ModelManager` and `Controller` facades. In https://github.com/canonical/jimm/pull/1837 we dropped support for these old versions with the thinking that all clients would have rolled forward by now but this proved not to be the case with the latest Juju dashboard. The linked PR explains that the changes to the facades were not substantive and we were already acting as the new version anyway, so I've simply re-added the old versions to support these clients.


ale8k (approved):
lgtm!

luci1900 (approved):

PR #1850, opened 2026-02-05
feat: use river for bootstrap

kian99:
## Description
This PR is a large refactor that removes our custom `JobTracker` job runner, in favor of the newly introduced River queue system.

I've kept the API the same, but a few user-facing things have been tweaked. I.e. job IDs are no longer UUIDs and are now int64 types converted to strings (keeping API backwards compatability).

I was also able to remove the bootstrap lock we were using to ensure we had limited concurrency of bootstrap operations. We can now rely on River's concurrency options. The only limitation here is that we were also limiting ourselves to 1 simultaneous bootstrap/destroy at a time due to the use of a global variable in Juju's cmd package. To address that I've added a global mutex that is shared across bootstrap and destroy, which when paired with River's concurrency should provide the same semantics as the previous bootstrap_lock.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests


internal/db/job_logs.go
luci1900, line 41: Both old and new type get coerced to string?

luci1900: Ah, you changed the schema.

luci1900 (approved):
Nice

internal/rivertypes/types.go
```diff
@@ -33,3 +34,84 @@ func (UpgradeToArgs) InsertOpts() river.InsertOpts {
...
+}
+
+// Kind implements the [river.JobArgs] interface.
+func (BootstrapArgs) Kind() string { return "bootstrap-controller" }
+
+// InsertOpts implements the [river.JobArgsWithInsertOpts] interface.
+func (BootstrapArgs) InsertOpts() river.InsertOpts {
+	return river.InsertOpts{
+		MaxAttempts: 1,
+		// Only allow 1 bootstrap job at a time.
```
luci1900, line 68: It occurs to me that we do actually end up allowing more concurrency than before. We allow bootstrap and destroy to run in parallel, as long as they're handled by different processes. 

That's probably fine.

internal/river/bootstrap_test.go
```diff
@@ -0,0 +1,206 @@
...
+			gotArgs = p
+			c.Assert(p.CloudCred.Attributes(), qt.DeepEquals, map[string]string{"attr-1": "val-1"})
+			c.Assert(p.CloudCred.AuthType(), qt.Equals, cloud.AccessKeyAuthType)
+			c.Assert(p.Cloud.Name, qt.Equals, "mycloud")
+			c.Assert(p.Cloud.Type, qt.Equals, "openstack")
+			c.Assert(p.Cloud.Regions, qt.DeepEquals, []cloud.Region{
+				{Name: "region-1", Endpoint: "region-1-endpoint"},
+			})
+			c.Assert(p.Username, qt.Equals, u.Name)
+			c.Assert(p.ControllerName, qt.Equals, "controller-name")
```
ale8k, line 59: We can compare just params no and have a matcher for the map to be deeper equal? Would reduce this

internal/river/client.go
```diff
@@ -35,7 +36,66 @@ func NewRiverClient(db *db.Database) (*Client, error) {
...
 // EnqueueUpgradeTo inserts a River job to upgrade a model to the specified version
 // by migrating and upgrading it.
-func (c *Client) EnqueueUpgradeTo(args rivertypes.UpgradeToArgs) (int64, error) {
-	job, err := c.client.Insert(context.Background(), args, nil)
+func (c *Client) EnqueueUpgradeTo(ctx context.Context, args rivertypes.UpgradeToArgs) (int64, error) {
+	job, err := c.client.Insert(ctx, args, nil)
 	return job.Job.ID, err
 }
+
+// TODO(Kian JUJU-9159): Return the isDuplicate flag so we can either return an error to callers
```
ale8k, line 45: If we have a card do we need the TODO?

internal/river/client.go
```diff
@@ -35,7 +36,66 @@ func NewRiverClient(db *db.Database) (*Client, error) {
...
+// or at least inform them that a bootstrap is in-progress and their request was ignored.
+
+// EnqueueBootstrap inserts a River job to bootstrap a new controller.
+func (c *Client) EnqueueBootstrap(ctx context.Context, args rivertypes.BootstrapArgs) (int64, error) {
```
ale8k, line 49: Why not Insert? Their client uses the language insert, but we're using enqueue?

internal/river/client.go
```diff
@@ -35,7 +36,66 @@ func NewRiverClient(db *db.Database) (*Client, error) {
...
+		return nil, err
+	}
+
+	if job.FinalizedAt != nil {
+		return job, nil
+	}
+
+	for {
+		select {
+		case event := <-subscribeChan:
```
ale8k: Nit: Perhaps a little ticker here to timeout just incase caller ever forgets to cancel context on time/withcancel.

internal/river/destroycontroller.go
```diff
@@ -0,0 +1,87 @@
...
+
+	destroyArgs := bootstrap.RunDestroyControllerArgs{
+		DestroyControllerArgs: job.Args,
+		RunnerArgs: bootstrap.RunnerArgs{
+			JujuDataDir: temp,
+			JobID:       job.ID,
+		},
+	}
+
+	if err := w.bootstrapManager.DestroyController(ctx, destroyArgs, &bootstrap.JujuCLI{}, user); err != nil {
```
ale8k, line 83: What is the purpose of JujuCLI arg here if it is empty? Could we pass nil?

internal/river/river.go
```diff
@@ -32,6 +33,11 @@ type UpgradeManager interface {
 	UpgradeModel(ctx context.Context, modelUUID string, targetVersion version.Number) error
 }
 
+type BootstrapManager interface {
```
ale8k, line 36: godoc

internal/river/utils_test.go
```diff
@@ -31,10 +37,116 @@ func setupTestDB(c *qt.C) (*db.Database, *sql.DB) {
...
+	db                   *db.Database
+	sqlDB                *sql.DB
+	riverClient          *river.Client[*sql.Tx]
+	mockUpgradeManager   *MockUpgradeManager
+	mockBootstrapManager *MockBootstrapManager
+	identity             string
+	contextCanceled      context.Context
+}
+
+func setupIntegrationTest(
```
ale8k, line 55: Not sure this change was necessary, as it bloated the PR further... If you know the PR is gonna be huge, probably leave bits like this out maybe as it can kinda throw the reviewer off thinking "what is this?"

internal/river/utils_test.go
```diff
@@ -31,10 +37,116 @@ func setupTestDB(c *qt.C) (*db.Database, *sql.DB) {
...
+}
+
+type testDeps struct {
+	db                   *db.Database
+	sqlDB                *sql.DB
+	riverClient          *river.Client[*sql.Tx]
+	mockUpgradeManager   *MockUpgradeManager
+	mockBootstrapManager *MockBootstrapManager
+	identity             string
+	contextCanceled      context.Context
```
ale8k: Pass context, don't embed it as it is considered back practice (usually anyway)

https://go.dev/blog/context-and-structs

internal/rivertypes/types.go
```diff
@@ -33,3 +34,84 @@ func (UpgradeToArgs) InsertOpts() river.InsertOpts {
 		},
 	}
 }
+
+// BootstrapArgs are the arguments for the bootstrap-controller worker.
+type BootstrapArgs struct {
```
ale8k, line 39: Why not just have the args with the worker? Why split it? And we're also calling the file types.go but it's got legitimate logic in the InsertOps so it isn't really just types?

internal/rivertypes/types.go
```diff
@@ -33,3 +34,84 @@ func (UpgradeToArgs) InsertOpts() river.InsertOpts {
...
+// InsertOpts implements the [river.JobArgsWithInsertOpts] interface.
+func (BootstrapArgs) InsertOpts() river.InsertOpts {
+	return river.InsertOpts{
+		MaxAttempts: 1,
+		// Only allow 1 bootstrap job at a time.
+		// This is used in conjuction with a global mutex
+		// in the bootstrap package to avoid issues with a
+		// global lock in Juju's cmd pkg used during bootstrap.
+		UniqueOpts: river.UniqueOpts{
+			ByState: []rivertype.JobState{
```
ale8k, line 73: Can make a global for this I think and share it across all insert ops.

internal/rivertypes/types.go
ale8k, line 68: This should go on the worker I think to explain that it has insertops on its args. Given this is implicitly called when using the args it is kind of hidden otherwise.

internal/jimm/bootstrap/bootstrap.go
```diff
@@ -8,62 +8,61 @@ import (
...
-	StopJob(ctx context.Context, jobId uuid.UUID) error
-	// Run runs a new job and returns the job ID.
-	Run(ctx context.Context, jobType string, job jobtracker.JobFunc, maxDuration time.Duration) (uuid.UUID, error)
+// JobQueue defines the method to enqueue a bootstrap/destroy-controller job.
+type JobQueue interface {
+	EnqueueBootstrap(ctx context.Context, args rivertypes.BootstrapArgs) (int64, error)
+	EnqueueDestroyController(ctx context.Context, args rivertypes.DestroyControllerArgs) (int64, error)
+	WaitForJobCompletion(ctx context.Context, jobID int64) (*rivertype.JobRow, error)
+	GetJobInfo(ctx context.Context, jobID int64) (*rivertype.JobRow, error)
+	CancelJob(ctx context.Context, jobID int64) (*rivertype.JobRow, error)
```
ale8k, line 64: This hasn't dropped in as well as I would've liked, we're like heavily exposing river details to the manager but we shouldn't need to I don't think (mainly around completion states and job rows)

internal/river/bootstrap.go
```diff
@@ -0,0 +1,87 @@
...
+
+	zapctx.Debug(ctx, "starting bootstrap-controller job")
+
+	u := &dbmodel.Identity{Name: job.Args.Username}
+	if err := w.store.FetchIdentity(ctx, u); err != nil {
+		return err
+	}
+	user := openfga.NewUser(u, w.openfgaClient)
+
+	temp, err := os.MkdirTemp("", "juju-data-dir")
```
ale8k, line 65: Should we place this in the manager call as we're splitting the logic up otherwise? Could even argue all logic should live here in the worker?

ale8k (changes requested):
Had a run over, think it'll need a few more but will do that later on. I'm not really sure this is reviewable tbh though as I'm finding myself scrolling up and down trying to critique it and missing details on each pass. I'm also worried about if it will "just work" given its size...

 Could you break it up and remove any scope creeping changes to help slim it down please?

PR #1851, opened 2026-02-05
test: refactor registercontroller into unit tests

ale8k:
## Description
Updates register controller into unit tests.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

### DIFF
--- a/cmd/jaas/cmd/registercontroller_test.go
+++ b/cmd/jaas/cmd/registercontroller_test.go
@@ -1,29 +1,22 @@
 // Copyright 2025 Canonical.
 
-package cmd_test
+package cmd
 
 import (
-	"context"
-	"os"
-	"path/filepath"
+	"bytes"
+	"testing"
 
-	"github.com/juju/cmd/v3/cmdtesting"
+	qt "github.com/frankban/quicktest"
+	"github.com/juju/juju/cmd/modelcmd"
 	"github.com/juju/juju/jujuclient"
-	gc "gopkg.in/check.v1"
-	"sigs.k8s.io/yaml"
+	"go.uber.org/mock/gomock"
 
-	"github.com/canonical/jimm/v3/cmd/jaas/cmd"
-	"github.com/canonical/jimm/v3/internal/testutils/cmdtest"
-	"github.com/canonical/jimm/v3/internal/testutils/jimmtest"
 	apiparams "github.com/canonical/jimm/v3/pkg/api/params"
 )
 
-type registerControllerDryRunSuite struct {
-}
+func TestRegisterControllerRun_DryRun_Defaults(t *testing.T) {
+	c := qt.New(t)
 
-var _ = gc.Suite(&registerControllerDryRunSuite{})
-
-func (s *registerControllerDryRunSuite) TestRegisterControllerDryRun(c *gc.C) {
 	store := jujuclient.NewMemStore()
 	store.Controllers["controller-1"] = jujuclient.ControllerDetails{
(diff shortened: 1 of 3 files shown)


cmd/jaas/cmd/registercontroller_test.go
```diff
@@ -156,140 +137,66 @@ func (s *registerControllerDryRunSuite) TestControllerInfoWithLocalFlagAndCustom
...
+	cmdMocks := setupCmdMocks(c)
 
-func (s *registerControllerSuite) TestRegisterControllerSuperuserByClientStore(c *gc.C) {
-	info := s.APIInfo(c)
-	// alice is superuser
-	bClient := s.SetupCLIAccess(c, "alice")
+	inner := &registerControllerCommand{}
+	inner.SetClientStore(cmdMocks.store)
+	inner.jimmAPIFunc = func() (JIMMAPI, error) { return cmdMocks.client, nil }
+	command := modelcmd.WrapBase(inner)
```
kian99, line 169: You don't need this `modelcmd.WrapBase` afaik, at least the other commands we've refactored don't have it.

kian99 (approved):

cmd/jaas/cmd/registercontroller_test.go
```diff
@@ -156,140 +137,66 @@ func (s *registerControllerDryRunSuite) TestControllerInfoWithLocalFlagAndCustom
...
 
-type registerControllerSuite struct {
-	cmdtest.JimmCmdSuite
-}
+func TestRegisterControllerRun_Success_FromFile(t *testing.T) {
+	c := qt.New(t)
 
-var _ = gc.Suite(&registerControllerSuite{})
-
-func (s *registerControllerSuite) TestRegisterControllerSuperuserByFile(c *gc.C) {
```
kian99, line 182: I just noticed, we don't have any integration tests in `testing/jimm_test.go` for registering a controller, the only integration tests were here in the jaas cli. Please migrate the relevant tests (register success, register not-admin, maybe others if you want).

cmd/jaas/cmd/registercontroller_test.go
luci1900: Thinking about this some more, perhaps the wrapper is for adding methods that can be mocked out. `WrapBase` does it for mocking out the client store, but we could do it for wrapping the api client.

ale8k: I used it for the SetClientStore because of Ales' change

cmd/jaas/cmd/registercontroller_test.go
ale8k: Ah ok will do!

kian99: Ignore me, I checked again while refactoring the `jaas unregister-controller` command and there are some integration tests there like `TestAddController` and several others, so all good.

cmd/jaas/cmd/registercontroller_test.go
kian99: But you already called `SetClientStore(cmdMocks.store)` on `inner`?

alesstimec (approved):

cmd/jaas/cmd/registercontroller_test.go
ale8k: As discussed, needed because we are using ClientStore earlier in the run. (possibly better ways to do this, but it's a start).

PR #1852, opened 2026-02-06
feat: add bootstrap and destroy-controller river jobs

kian99:
## Description
This PR is the first part of a set of PRs that breaks up https://github.com/canonical/jimm/pull/1850.
It introduces some changes to the River package and adds a bootstrap and destroy-controller River job.
I've done a `git checkout use-river-for-bootstrap ./internal/river ./internal/rivertypes/` so the changes don't compile as external package dependencies are missing.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests


internal/river/river.go
```diff
@@ -32,6 +34,11 @@ type UpgradeManager interface {
 	UpgradeModel(ctx context.Context, modelUUID string, targetVersion version.Number) error
 }
 
+type BootstrapManager interface {
```
ale8k, line 38: Godoc

internal/river/bootstrap.go
```diff
@@ -0,0 +1,95 @@
...
+	river.WorkerDefaults[rivertypes.BootstrapArgs]
+
+	openfgaClient    *openfga.OFGAClient
+	store            Store
+	bootstrapManager BootstrapManager
+}
+
+// Work implements the [river.Worker] interface.
+func (w *bootstrapWorker) Work(ctx context.Context, job *river.Job[rivertypes.BootstrapArgs]) error {
+	ctx = zapctx.WithFields(ctx,
```
ale8k, line 53: Do other jobs run a debug log at the beginning? Might be nice to add later

internal/rivertypes/types.go
```diff
@@ -33,3 +34,85 @@ func (UpgradeToArgs) InsertOpts() river.InsertOpts {
 		},
 	}
 }
+
+// BootstrapArgs are the arguments for the bootstrap-controller worker.
```
ale8k, line 38: Ahhh we've put this here to prevent circular dep? Is that right?

internal/river/utils_test.go
```diff
@@ -31,10 +37,115 @@ func setupTestDB(c *qt.C) (*db.Database, *sql.DB) {
...
+type testDeps struct {
+	db                   *db.Database
+	sqlDB                *sql.DB
+	riverClient          *river.Client[*sql.Tx]
+	mockUpgradeManager   *MockUpgradeManager
+	mockBootstrapManager *MockBootstrapManager
+	identity             string
+}
+
+func setupIntegrationTest(
```
ale8k, line 55: Ah I see now, you're just sharing this with the bootstrap_test. Ty for keeping ash@catchum 😆

internal/river/river.go
```diff
@@ -88,11 +107,27 @@ func newWorkers(openfgaClient *openfga.OFGAClient, store *db.Database, upgradeMa
...
+	}
+	if err := river.AddWorkerSafely(workers, bootstrapWorker); err != nil {
+		return nil, err
+	}
+
+	destroyControllerWorker, err := newDestroyControllerWorker(openfgaClient, store, bootstrapManager)
+	if err != nil {
+		return nil, err
+	}
+	if err := river.AddWorkerSafely(workers, destroyControllerWorker); err != nil {
```
ale8k, line 136: Sometimes you do simple statements like this, sometimes like line 123 - can we have them consistent please? Idm which way

internal/river/river.go
```diff
@@ -69,7 +82,13 @@ func StartWorkers(
 	return riverClient.Start(ctx)
 }
 
-func newWorkers(openfgaClient *openfga.OFGAClient, store *db.Database, upgradeManager UpgradeManager) (*river.Workers, error) {
+type workerParams struct {
```
ale8k, line 90: godoc this and explain it holds "all parameters for all workers to be registered." & Why not put the whole lot on, like store, managers etc, they're parameters too just for destroy & upgradeto?

internal/river/destroycontroller.go
```diff
@@ -0,0 +1,95 @@
...
+
+	destroyArgs := bootstrap.RunDestroyControllerArgs{
+		DestroyControllerArgs: job.Args,
+		RunnerArgs: bootstrap.RunnerArgs{
+			JujuDataDir: temp,
+			JobID:       job.ID,
+		},
+	}
+
+	if err := w.bootstrapManager.DestroyController(ctx, destroyArgs, &bootstrap.JujuCLI{}, user); err != nil {
```
ale8k, line 84: Can we use nil here? Why does it need to be a pointer to CLI? Why is JujuCLI empty here btw & what does it represent?

internal/river/destroycontroller_test.go
```diff
@@ -0,0 +1,205 @@
...
+	"github.com/riverqueue/river/rivertest"
+	"github.com/riverqueue/river/rivertype"
+	gomock "go.uber.org/mock/gomock"
+
+	"github.com/canonical/jimm/v3/internal/dbmodel"
+	"github.com/canonical/jimm/v3/internal/jimm/bootstrap"
+	"github.com/canonical/jimm/v3/internal/openfga"
+	"github.com/canonical/jimm/v3/internal/rivertypes"
+)
+
```
ale8k, line 24: Perhaps add a test for the worker constructor

internal/river/client.go
```diff
@@ -35,7 +36,66 @@ func NewRiverClient(db *db.Database) (*Client, error) {
...
 // EnqueueUpgradeTo inserts a River job to upgrade a model to the specified version
 // by migrating and upgrading it.
-func (c *Client) EnqueueUpgradeTo(args rivertypes.UpgradeToArgs) (int64, error) {
-	job, err := c.client.Insert(context.Background(), args, nil)
+func (c *Client) EnqueueUpgradeTo(ctx context.Context, args rivertypes.UpgradeToArgs) (int64, error) {
+	job, err := c.client.Insert(ctx, args, nil)
 	return job.Job.ID, err
 }
+
+// TODO(Kian JUJU-9159): Return the isDuplicate flag so we can either return an error to callers
```
ale8k, line 42: Do we need TODO or is card enough?

internal/river/client.go
```diff
@@ -35,7 +36,66 @@ func NewRiverClient(db *db.Database) (*Client, error) {
...
+// or at least inform them that a bootstrap is in-progress and their request was ignored.
+
+// EnqueueBootstrap inserts a River job to bootstrap a new controller.
+func (c *Client) EnqueueBootstrap(ctx context.Context, args rivertypes.BootstrapArgs) (int64, error) {
```
ale8k, line 46: River uses language "Insert" but we say "Enqueue", can we just say "InsertBootstrapJob"?

internal/river/client.go
```diff
@@ -35,7 +36,66 @@ func NewRiverClient(db *db.Database) (*Client, error) {
...
+	return c.client.JobGet(ctx, jobID)
+}
+
+// CancelJob cancels the specified job. It returns the final job state after cancellation.
+func (c *Client) CancelJob(ctx context.Context, jobID int64) (*rivertype.JobRow, error) {
+	return c.client.JobCancel(ctx, jobID)
+}
+
+// WaitForJobCompletion waits for the specified job to complete, returning the final job state.
+// If the job has already completed, it returns immediately.
```
ale8k: Maybe add this is expected to be used from Within a job, where an eventual timeout and subsequent context cancellation will always occur. The function waits indefinitely otherwise until an event is received.

internal/river/bootstrap_test.go
```diff
@@ -0,0 +1,207 @@
...
+	"github.com/riverqueue/river/rivertest"
+	"github.com/riverqueue/river/rivertype"
+	gomock "go.uber.org/mock/gomock"
+
+	"github.com/canonical/jimm/v3/internal/dbmodel"
+	"github.com/canonical/jimm/v3/internal/jimm/bootstrap"
+	"github.com/canonical/jimm/v3/internal/openfga"
+	"github.com/canonical/jimm/v3/internal/rivertypes"
+)
+
```
ale8k, line 25: Constructor test

internal/river/client.go
```diff
@@ -35,7 +36,66 @@ func NewRiverClient(db *db.Database) (*Client, error) {
...
+func (c *Client) WaitForJobCompletion(ctx context.Context, jobID int64) (*rivertype.JobRow, error) {
```
ale8k, line 69: This is exposing river details, yet it only ever responds to 3 specific events. We could pass a param like:

```go
type WaitForErrorResponses struct {
    cancelled string
    failed string
}
```

And for the response just the job.Errors because that's all we seem to care about here & we already have the ID? 

Then it'd just read a bit nicer

internal/river/bootstrap.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+	defer func() {
+		if err := os.RemoveAll(temp); err != nil {
+			zapctx.Error(ctx, "failed to remove temporary Juju data directory", zap.String("path", temp), zap.Error(err))
+		}
+	}()
+
+	bootstrapArgs := bootstrap.RunBootstrapArgs{
+		BootstrapArgs: job.Args,
+		RunnerArgs: bootstrap.RunnerArgs{
```
ale8k, line 78: This is the CommandRunner right?

internal/river/client.go
ale8k: Also in the other PR we call this from a manager method, but I don't think that is necessary because we end up doing:

Facade -> Manager -> River -> Manager -> CMD Runner

I think we should just do:

Facade -> River -> Manager -> CMDRUnner

Unless I'm misunderstanding the purpose of going to the manager to initiate the job start?

I only mention this because in the other PR it reads super confusing launching the job from the manager which is then using the manager. 

(Correct me if I'm wrong here)

internal/river/bootstrap.go
kian99: No, but would be nice to add. River also has middleware where we can add it. I just wanted to keep the behaviour of the previous logic.

luci1900 (approved):

internal/river/bootstrap.go
kian99: Not quite, some of these args will be used in the command runner, but no it's not that.

internal/river/bootstrap_test.go
kian99: We didn't write constructor tests for all the other workers. And it's tested in the integration syle tests, so I think it's fine.

internal/river/client.go
kian99: Doesn't matter to me either way, I'll start on that once this work lands.

internal/river/client.go
kian99: Sure, I'll rename to `InsertBootstrapJob`. Edit: I realise I've already named it Enqueue for the upgrade-to job as well and I don't want to introduce unrelated changes here. Let's stick with enqueue and rename after it's all working, should be easy enough.

You're correct on the flow. `Facade -> River -> Manager -> CMDRUnner` was an option I considered in https://github.com/canonical/jimm/pull/1845 but I kept our existing architecture where the managers are called from the facade after a nice discussion with Luci, I think that was on Monday when you were sick. We can change it at a later stage if needed.

internal/river/client.go
kian99: It doesn't need to be run from within a job. The caller just needs to set a timeout on the context, I'll add that to the godoc.

internal/river/client.go
kian99: > exposing river details

I think we're not trying to hide the River stuff. We return a river.Job type as well. I'd prefer to just leave it as is until there is a need to refactor it.

internal/river/destroycontroller.go
kian99: bootstrap.JujuCLI is the concrete implementation for the command runner. It uses a real Juju CLI. One can pass in a mock to use a fake command runner.

internal/river/destroycontroller_test.go
kian99: Same as above, we skipped it for the other workers and I think it's okay without it.

internal/river/river.go
kian99: Yeah sure, I'll put them all on it

internal/river/river.go
kian99: If there's no linter for it, stuff like this feels okay to me to have both ways.

internal/rivertypes/types.go
kian99: Yip that's correct. Explained a bit more in https://github.com/canonical/jimm/pull/1845

internal/river/river.go
luci1900: I generally use the "if-let" pattern, unless I can't because I need a value outside the if. I think naturally both will happen.

ale8k: Would be nice to just make it consistent whilst in X func but more of a nit I guess

internal/river/destroycontroller.go
ale8k: Ahhh I see what you've done in the other PR

internal/river/destroycontroller_test.go
ale8k: I think it was an oversight on my part and probably not a bad test to have but up to you

internal/river/client.go
ale8k: Sure ty!

& I don't think we should change it later, I think we should do it in this set of PRs as it really doesn't make sense to me? What was the conclusion of the discussion / merits to having it go through the manager?

internal/river/bootstrap_test.go
ale8k: I think it was an oversight on my part with the UpgradeTo and we should probably have them, esp for the pointer based ones as we have no compile time checks.

internal/river/bootstrap.go
ale8k: I mean the "RunnerArgs" specifically

internal/river/client.go
ale8k: Sure, but there's gonna be a big switch in the manager where we don't really care and the error messages can be passed in to the Wait here - it'll read much nicer?

internal/river/bootstrap_test.go
kian99: They are tested in `setupIntegrationTests` and in are larger integration tests.

internal/river/client.go
kian99: > I don't think we should change it later, I think we should do it in this set of PRs as it really doesn't make sense to me

If you're referring to the flow, I don't think the river client should be used from the facade layer. Consider a scenario where you want to associate some business logic with the insertion of a job. E.g. User John may only bootstrap a controller every X hours. It's worthwhile to keep the queue primitive in the same layer as our database, etc.

internal/river/client.go
kian99: I don't know, but I think it's fine as it is. Here's the manager code. Wdyt? We also call `WaitForJobCompletion` from a few places.
```
func (b *bootstrapManager) WaitForJobCompletion(ctx context.Context, jobId int64, config WaitConfig) error {
	maxDuration := config.MaxJobDuration
	if maxDuration == 0 {
		maxDuration = maxJobDuration
	}

	ctx, cancel := context.WithTimeout(ctx, maxDuration)
	defer cancel()

	job, err := b.jobQueue.WaitForJobCompletion(ctx, jobId)
	if err != nil {
		return fmt.Errorf("error while waiting for job completion: %w", err)
	}

	switch job.State {
	case rivertype.JobStateCompleted:
		return nil
	case rivertype.JobStateCancelled:
		return fmt.Errorf("bootstrap job was cancelled")
	case rivertype.JobStateDiscarded:
		if len(job.Errors) == 0 {
			return fmt.Errorf("bootstrap job failed")
		}
		lastErr := job.Errors[len(job.Errors)-1].Error
		return fmt.Errorf("bootstrap job failed: %s", lastErr)
	default:
		return fmt.Errorf("unexpected job state: %s", job.State)
	}
}
```

internal/river/bootstrap.go
kian99: Yip, those are not the CommandRunner args specifically, they are just the args passed from the job runner to the domain logic. I.e. it's just 
```
type RunnerArgs struct {
	JujuDataDir string
	JobID       int64
}
```
We could look to get rid of them by passing the job id through the context or by passing in some kind of job logger to the domain logic that abstracts the fact that it is writing to a DB.
The JujuDataDir could be removed if we create the temp dir inside the `Bootstrap` function, but I kept it outside to keep the same flow/behaviour as we had before.

internal/river/client.go
ale8k: Sure I guess... But then I wouldn't call that very specific method for co-ordinating the running of jobs bootstrap manager specific? 

Facade -> JobCoordinator -> River -> Manager -> Cmd Runner

Would be more what I'd expect? It feels like it's here because it's convenient then? But I more meant what we have right now isn't adding value and is more redirection for no benefit. 

& We also aren't using primitives, we have a client wrapper so I'd argue there's no difference between calling the client wrapper and a service layer (with no logic) from a facade. 

& Our testing coverage has to extend to this new method which won't do a great deal / nothing additional.

Maybe we can discuss in a call... But I don't really like this pattern and it's giving me Juju-esque nightmates 😆

internal/river/bootstrap_test.go
ale8k: Ahh good point! Yeah nvm! All good! (Given we always int test top level workers I mean)

internal/river/client.go
ale8k: Yeah I saw in other PR, hence this suggestion. We have two WaitForJobCompletion methods now, a client one - which isn't capturing the state responses and one here which really is just switching. Also we don't need the context timeout in that example because when jobs timeout the context is cancelled, so there's gonna be even less in this func.

kian99: Yeah I can remove the timeout. But what do you mean by "two WaitForJobCompletion"? Edit: Actually not sure if I can remove it, `WaitForJobCompletion` is currently being called from UpgradeTo(), the bit before we insert the River job, where we bootstrap a new controller and wait for it to be ready.

Also, how do you want me to change it? I need to wait for job success / job failure / job cancellation, so wouldn't there be a switch statement no matter what?

internal/river/bootstrap.go
```diff
@@ -0,0 +1,95 @@
...
+		return err
+	}
+
+	return nil
+}
+
+// Timeout implements the [river.Worker] interface.
+// Bootstrap operations can take a while to complete so we set a generous timeout.
+func (w *bootstrapWorker) Timeout(*river.Job[rivertypes.BootstrapArgs]) time.Duration {
+	return 60 * time.Minute
```
alesstimec, line 94: "one hour should be enough for everybody" :)   could we make this configurable?

alesstimec (approved):

internal/river/bootstrap.go
kian99: For sure, but it was hardcoded before as well so no behavioral changes there. And once we do expose a config option it adds to our growing list of config options.

internal/river/bootstrap.go
ale8k: After discussion resolved. I don't mind this how it is, would be nice to pass a little logger through or add it to JujuCLI{} but can be done in a follow up.

internal/river/bootstrap.go
ale8k: I'm sure an hour is fine, we can adjust if it causes any issues. Probably just needs documenting.

internal/river/client.go
```diff
@@ -35,7 +34,61 @@ func NewRiverClient(db *db.Database) (*Client, error) {
 
 // EnqueueUpgradeTo inserts a River job to upgrade a model to the specified version
 // by migrating and upgrading it.
-func (c *Client) EnqueueUpgradeTo(args rivertypes.UpgradeToArgs) (int64, error) {
-	job, err := c.client.Insert(context.Background(), args, nil)
+func (c *Client) EnqueueUpgradeTo(ctx context.Context, args rivertypes.UpgradeToArgs) (int64, error) {
```
ale8k, line 37: Plz change language from Enqueue haha. Also just "int" is fine and let compiler decide for return type.

internal/river/client.go
ale8k: As discussed I don't mind much, would be a nice change but it's up to you.

internal/river/river.go
```diff
@@ -69,30 +87,56 @@ func StartWorkers(
...
+	migrateRetryCount int
+	upgradeRetryCount int
```
ale8k, line 92: Ty!

internal/river/river.go
```diff
@@ -69,30 +87,56 @@ func StartWorkers(
...
+	store             *db.Database
+	upgradeManager    UpgradeManager
+	bootstrapManager  BootstrapManager
+}
+
+func newWorkers(wp workerParams) (*river.Workers, error) {
 	workers := river.NewWorkers()
 
-	migrationWorker, err := newMigrationWorker(openfgaClient, store, upgradeManager)
+	migrationWorker, err := newMigrationWorker(wp.openfgaClient, wp.store, wp.upgradeManager)
```
ale8k, line 103: Nit: If you can be bothered, it'd be nice to have this iterate over a bunch of register funcs on a global that we can update simply. Something like:
```go
type registerWorkerFunc func() error

var (
    workers = []registerWorkerFunc{
     ...so forth
    }
)

...in newowrkers
for _, rf := range() {
    err := rf()
    if err != nil {...}
}
```

Just so we don't end up hitting the eventual cognit complexity lint & it is  bit easier to know what is registered.

ale8k (approved):
Lgtm! Some mild nits but looks good to me! Nice work, a lot of effort put into this one!

PR #1853, opened 2026-02-06
refactor(revoke-audit-log-access): refactors the revoke-audit-log-access tests

alesstimec:
## Description

Refactors the revoke-audit-logs tests into unit tests.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests


luci1900 (approved):
As we discussed, this is different from the pattern in others.

But they're all a bit different, so I think we can leave the cleanup for later with a jaas base cmd.

SimoneDutto (approved):

PR #1858, opened 2026-02-09
feat: update JIMM's bootstrap logic to queue river jobs

kian99:
## Description
This PR is a follow-up to #1852 and updates JIMM's `bootstrap` package to make use of River for bootstrap/destroy operations.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests


internal/jimm/bootstrap/bootstrap.go
```diff
@@ -134,50 +129,75 @@ func NewBootstrapManager(
 	if credentialStore == nil {
 		return nil, errors.E("credential store cannot be nil")
 	}
+	if jobQueue == nil {
+		return nil, errors.E("job queue cannot be nil")
```
alesstimec, line 133: nitpicky nitpick: `errors.E("nil job queue")`

saying it cannot be nil is obviously untrue, as it is nil..

internal/jimm/bootstrap/bootstrap.go
```diff
@@ -134,50 +129,75 @@ func NewBootstrapManager(
...
-
-	job, err := b.tracker.GetJob(ctx, jobId)
+func (b *bootstrapManager) GetJobInfo(ctx context.Context, _ *openfga.User, jobId int64, offset int) (params.GetJobInfoResponse, error) {
+	job, err := b.jobQueue.GetJobInfo(ctx, jobId)
 	if err != nil {
-		return params.GetJobInfoResponse{}, errors.E("failed to get job info", err)
+		return params.GetJobInfoResponse{}, fmt.Errorf("failed to get job info: %w", err)
 	}
-
 	loggies, newOffset, err := b.store.QueryJobLog(ctx, jobId, offset)
```
alesstimec: `loggies`?? i know who named this variable ;) may i suggest a rename to `logs`?

internal/jimm/bootstrap/bootstrap.go
```diff
@@ -134,50 +129,75 @@ func NewBootstrapManager(
...
+		Error:     errorMsg.String(),
 		Logs:      loggies,
 		Watermark: newOffset,
 	}, nil
 }
 
+func toParamsJobState(ctx context.Context, state rivertype.JobState) params.JobStatus {
+	switch state {
+	case rivertype.JobStateCompleted:
+		return params.StatusSuccessful
```
alesstimec, line 172: coming back to the question whether our business logic should be returning params types..

internal/jimm/bootstrap/bootstrap.go
```diff
@@ -188,95 +208,67 @@ func (b *bootstrapManager) StopJob(ctx context.Context, user *openfga.User, jobI
...
 	}
 }
 
 // StartBootstrap starts a bootstrap job with the provided parameters.
-func (b *bootstrapManager) StartBootstrapJob(ctx context.Context, user *openfga.User, params BootstrapParams) (string, error) {
+func (b *bootstrapManager) StartBootstrapJob(ctx context.Context, user *openfga.User, params BootstrapParams) (int64, error) {
 
 	if b.jimmWellknownJWKSEndpoint == "" {
-		return "", errors.E("bootstrap login token refresh URL is not configured. Cannot proceed with bootstrap. Please configure it and try again.")
+		return 0, errors.E("bootstrap login token refresh URL is not configured. Cannot proceed with bootstrap. Please configure it and try again.")
```
alesstimec, line 245: `-1` is as invalid as a job ID can be..  `0` might actually be a valid one

internal/jimm/bootstrap/bootstrap.go
```diff
@@ -324,134 +317,73 @@ func (h commandFactory) RunWrapper(
...
-		)
-
-		// Lock the bootstrap for the same length the process is allowed to run for
-		// before being killed.
-		if err := b.store.LockBootstrap(jobCtx, jujucommands.CommandKillDelay); err != nil {
-			return errors.E(fmt.Errorf("failed to acquire bootstrap lock: %w", err))
-		}
+) error {
+	// Lock the bootstrap concurrently with destroy to avoid misuse of the store commands.
+	jujuCLILock.Lock()
```
alesstimec: this will block the goroutine for a potentially long time (up to 60min while another process bootstraps/destroys a controller).. are we OK with the wait or should we return an error that another process is in progress and the user should retry a bit later?

internal/jimm/bootstrap/bootstrap.go
```diff
@@ -571,160 +503,131 @@ func (b *bootstrapManager) tryCleanupController(ctx context.Context, jujuCmd Juj
...
 // Errors are masked but logged to avoid failing the bootstrap process.
-func (b *bootstrapManager) writeJobLog(ctx context.Context, jobId uuid.UUID, logLine string) {
+func (b *bootstrapManager) writeJobLog(ctx context.Context, jobId int64, logLine string) {
 	// Avoid storing empty log lines.
 	if logLine == "" {
 		return
 	}
 	if err := b.store.AddJobLog(ctx, jobId, logLine); err != nil {
-		zapctx.Error(ctx, "failed to write bootstrap log", zap.Error(err), zap.String("jobId", jobId.String()))
+		zapctx.Error(ctx, "failed to write bootstrap log", zap.Error(err), zap.Int64("jobId", jobId))
```
alesstimec, line 529: logging is fine. should we also have a monitored value so that alerts could be put in place at a later time?

internal/jimm/bootstrap/bootstrap.go
kian99: I'm with you, but it does match the other errors.

internal/jimm/bootstrap/bootstrap.go
kian99: Sure

internal/jimm/bootstrap/bootstrap.go
kian99: Yeah not great, but I wouldn't change it in this PR.

internal/jimm/bootstrap/bootstrap.go
kian99: I get you on this front and was thinking about you when I was writing this. I think the extra defensiveness isn't really worth it. We already have out-of-band errors, adding in-band errors feels like extra work without a real upside.

internal/jimm/bootstrap/bootstrap.go
kian99: Correct, I'm not sure whether we're okay with it or not, but it does mirror the previous behavior calling `LockBootstrap`,
```
		// This is equivalent to a SELECT FOR UPDATE in SQL,
		// which locks the row for the duration of the transaction.
		// The other select statements will wait until this lock is released.
		if err := tx.DB.Clauses(clause.Locking{Strength: "UPDATE"}).First(lock).Error; err != nil {
			return err
		}
```
But I'd be happy to follow up on this and make it non-blocking and return an error if the lock is not acquired within a few seconds.

internal/jimm/bootstrap/bootstrap.go
kian99: Yip, that's a good idea, I would make a separate card for it.

internal/jimm/bootstrap/bootstrap.go
luci1900: This is a problem for Go in general, it's entirely possible to use the value even if the error is not nil. I don't know if it's worth being extra cautious considering not all types have a possible sentinel value.

luci1900 (approved):

internal/jimm/bootstrap/bootstrap.go
alesstimec: i didn't want to comment on all of them :)

internal/jimm/bootstrap/bootstrap.go
alesstimec: TryLock is our friend :)

alesstimec (approved):

internal/jimm/bootstrap/bootstrap.go
kian99: I'll change this to use TryLock and return an error if something else has the lock.

internal/jimm/bootstrap/bootstrap.go
kian99: ha, fair enough, I'll leave it as is for now.

PR #1859, opened 2026-02-09
refactor: move to mocked unit tests for permissions cmd

luci1900:
## Description
Refactor the permissions-related commands to unit tests. There already are integration tests for the endpoints.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

### DIFF
--- a/cmd/jaas/cmd/permission.go
+++ b/cmd/jaas/cmd/permission.go
@@ -11,7 +11,6 @@ import (
 	"github.com/gosuri/uitable"
 	"github.com/juju/cmd/v3"
 	"github.com/juju/gnuflag"
-	jujuapi "github.com/juju/juju/api"
 	jujucmd "github.com/juju/juju/cmd"
 	"github.com/juju/juju/cmd/modelcmd"
 	"github.com/juju/juju/jujuclient"
@@ -159,18 +158,18 @@ List permissions where the target object and relation match
 
 // NewAddPermissionCommand returns a command to grant access.
 func NewAddPermissionCommand() cmd.Command {
-	cmd := &addPermission{}
+	cmd := &addPermissionCommand{}
 	cmd.SetClientStore(jujuclient.NewFileClientStore())
 
 	return modelcmd.WrapBase(cmd)
 }
 
-// addPermission adds permission.
-type addPermission struct {
+// addPermissionCommand adds permission.
+type addPermissionCommand struct {
 	modelcmd.ControllerCommandBase
 	out cmd.Output
 
-	dialOpts *jujuapi.DialOpts
+	jimmAPIFunc func() (JIMMAPI, error)
 
 	object       string
 	relation     string
@@ -180,7 +179,7 @@ type addPermission struct {
 }
 
 // Info implements the cmd.Command interface.
-func (c *addPermission) Info() *cmd.Info {
+func (c *addPermissionCommand) Info() *cmd.Info {
 	return jujucmd.Info(&cmd.Info{
 		Name:     "add-permission",
 		Args:     "<object> <relation> <target_object>",
@@ -191,7 +190,7 @@ func (c *addPermission) Info() *cmd.Info {
 }
 
 // Init implements the cmd.Command interface.
-func (c *addPermission) Init(args []string) error {
+func (c *addPermissionCommand) Init(args []string) error {
 	if c.filename != "" {
 		return nil
 	}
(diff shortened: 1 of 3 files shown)


cmd/jaas/cmd/permission.go
```diff
@@ -450,6 +472,20 @@ func (c *checkPermissionCommand) Run(ctxt *cmd.Context) error {
 	return nil
 }
 
+func (c *checkPermissionCommand) newClient() (JIMMAPI, error) {
```
ale8k, line 475: We could embed a struct with the client on right and not need to make newClient 4 times?

cmd/jaas/cmd/permission_test.go
```diff
@@ -1,659 +1,247 @@
...
-	"github.com/canonical/jimm/v3/internal/dbmodel"
-	"github.com/canonical/jimm/v3/internal/openfga"
-	ofganames "github.com/canonical/jimm/v3/internal/openfga/names"
-	"github.com/canonical/jimm/v3/internal/testutils/cmdtest"
-	apiparams "github.com/canonical/jimm/v3/pkg/api/params"
-	jimmnames "github.com/canonical/jimm/v3/pkg/names"
+	"path"
+	"testing"
+
+	"github.com/canonical/jimm/v3/pkg/api/params"
```
ale8k: we usually do

internal
external
pkg 

ordering

cmd/jaas/cmd/permission_test.go
```diff
@@ -1,659 +1,249 @@
...
-	c.Assert(db.DB.Create(u1).Error, gc.IsNil)
+	command := &addPermissionCommand{
+		jimmAPIFunc: func() (JIMMAPI, error) {
+			return cmdMocks.client, nil
+		},
+	}
+	command.SetClientStore(cmdMocks.store)
+
+	tmpDir := os.TempDir()
+	defer os.Remove(tmpDir)
```
ale8k: c.Cleanup would be better here.

cmd/jaas/cmd/permission_test.go
```diff
@@ -1,659 +1,249 @@
...
-	context, err = cmdtesting.RunCommand(c, cmd.NewListPermissionsCommandForTesting(s.ClientStore(), bClient), "--format", "tabular")
-	c.Assert(err, gc.IsNil)
-	var builder strings.Builder
-	err = cmd.FormatRelationsTabular(&builder, &expectedData)
-	c.Assert(err, gc.IsNil)
-	c.Assert(cmdtesting.Stdout(context), gc.Equals, builder.String())
+	command.SetClientStore(cmdMocks.store)
+
+	tmpDir := os.TempDir()
+	defer os.Remove(tmpDir)
```
ale8k: c.Cleanup again

And thinking about it, you could make a lil test func to create the file, add the c.Cleanup and return it's name and use it here and AddPermission too to remove the duplication.

cmd/jaas/cmd/permission_test.go
```diff
@@ -1,659 +1,249 @@
...
-	}
-	err = db.SetCloudCredential(ctx, &cred)
-	c.Assert(err, gc.IsNil)
-
-	model := dbmodel.Model{
-		Name: petname.Generate(2, "-"),
-		UUID: sql.NullString{
-			String: id.String(),
-			Valid:  true,
+func TestCheckPermission(t *testing.T) {
```
ale8k, line 170: Could you add 2 more tests please

1. 1 to run check permission with the "smart" format
2. Another to check the tabular formatting of list

Noticed these two aren't tested and they'd be quick enough to do and nice additions.

cmd/jaas/cmd/permission_test.go
```diff
@@ -1,659 +1,249 @@
...
+	cmdMocks := setupCmdMocks(c)
+
+	command := &listPermissionsCommand{
+		jimmAPIFunc: func() (JIMMAPI, error) {
+			return cmdMocks.client, nil
+		},
+	}
+	command.SetClientStore(cmdMocks.store)
+
+	initCommand(c, command, "--target", "group-mygroup", "--relation", "member")
```
ale8k: We're only testing 2 of 4 flags here, it needs to do:

--object, --relation, --target and --resolve 

If we don't do --object, you may send a tuple with no --object & resolve is needed to make sure we can turn this on when we need it.

ale8k:
I also checked all this is covered in our e2e tests and it is so after the addition flags tests this looks good to me.

cmd/jaas/cmd/permission.go
luci1900: That's the case for all of these, we have well over a dozen instances of this function.

I plan to do that as a follow-up for all of the commands.

cmd/jaas/cmd/permission_test.go
luci1900: Smart is the default formatter for check

ale8k (approved):
LGTM!

kian99 (approved):

PR #1860, opened 2026-02-09
refactor: move to mocked unit tests for roles cmd

luci1900:
## Description
Refactor the role-related commands to unit tests. There already are integration tests for the endpoints.

## Engineering checklist
- [ ] Documentation updated
- [x] Covered by unit tests

### DIFF
--- a/cmd/jaas/cmd/role_test.go
+++ b/cmd/jaas/cmd/role_test.go
@@ -1,145 +1,134 @@
 // Copyright 2025 Canonical.
 
-package cmd_test
+package cmd
 
 import (
-	"context"
-	"fmt"
-	"strings"
+	"bytes"
+	"testing"
 
-	"github.com/juju/cmd/v3/cmdtesting"
-	gc "gopkg.in/check.v1"
-	"gopkg.in/yaml.v3"
-
-	"github.com/canonical/jimm/v3/cmd/jaas/cmd"
-	"github.com/canonical/jimm/v3/internal/dbmodel"
-	"github.com/canonical/jimm/v3/internal/testutils/cmdtest"
 	"github.com/canonical/jimm/v3/pkg/api/params"
+	qt "github.com/frankban/quicktest"
+	"go.uber.org/mock/gomock"
 )
 
-type roleSuite struct {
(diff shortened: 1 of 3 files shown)


cmd/jaas/cmd/role_test.go
```diff
@@ -1,145 +1,134 @@
...
-	c.Assert(err, gc.IsNil)
-	c.Assert(roleEntry.UUID, gc.Not(gc.Equals), "")
+	ctx := newTestContext(c)
 
-	_, err = cmdtesting.RunCommand(c, cmd.NewRenameRoleCommandForTesting(s.ClientStore(), bClient), "test-role", "renamed-role")
-	c.Assert(err, gc.IsNil)
+	cmdMocks.client.EXPECT().
+		AddRole(gomock.Any()).
+		DoAndReturn(func(req *params.AddRoleRequest) (params.AddRoleResponse, error) {
+			c.Assert(req.Name, qt.Equals, "myrole")
```
kian99: Prefer c.Check() inside the mock functions. If the assert fails it stops the routine and then you end up waiting for the test to timeout rather than seeing the failure immediately (at least that's what memory tells me).

Same comment applies to https://github.com/canonical/jimm/pull/1859/ but it's not a big deal.

kian99 (approved):

cmd/jaas/cmd/role_test.go
luci1900: I really like tests that fail instantly, I hadn't realised the callback might reverse that.

kian99: I went to double check and in our cmd tests at least, both fail quickly. I realised I'm remembering behaviour from Juju, where they wait for a worker launched in a separate go routine to complete and in those cases if an assertion fails, the routine stopped and then the test would take a while instead of failing instantly.

Probably still better to prefer the Check() when you're inside a callback and you don't know how the code is calling it, but yeah not a big deal here.

luci1900: Yeah, it seemed to make no practical difference after I changed them.

ale8k (approved):
lgtm!

PR #1861, opened 2026-02-10
feat: cleanup unused code and wire up bootstrap via river

kian99:
## Description
This PR is the 3rd in a series of PRs to move bootstrapping to use River. It follows on from #1858.
This PR does the final wiring up and removes unused code among other bits. In full it,
- Adds auth checks for registering/unregistering a controller to the API layer, since we moved the auth checks out of the domain layer. This highlighted that unit tests for these methods in the API layer were missing so I've added some.
- Adds a DB migration to remove the FK relation between job logs and the old job tracker table.
- Adds a DB migration to remove unused tables for the job tracker and bootstrap lock.
- Modifies the job_logs table and logic to use a river `int64` ID.
- Removes unused files for the job tracker and bootstrap lock.
- Completes wiring up the bootstrap manager to support the use of River.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests


internal/jujuapi/jimm.go
```diff
@@ -211,6 +212,9 @@ func (r *controllerRoot) AddModelToController(ctx context.Context, req apiparams
 // AddController allows adds a controller to the pool of controllers
 // available to JIMM.
 func (r *controllerRoot) AddController(ctx context.Context, req apiparams.AddControllerRequest) (apiparams.ControllerInfo, error) {
+	if !r.user.JimmAdmin {
```
ale8k, line 215: Just moving them up a layer?

internal/jujuapi/jimm_unit_test.go
```diff
@@ -139,12 +269,12 @@ func (s *jimmUnitTestSuite) TestPrepareModelMigration_InvalidUserMapping(c *gc.C
...
 func (s *jimmUnitTestSuite) TestBootstrapStatus(c *gc.C) {
 	ctx := context.Background()
-	uuidGenerated := uuid.New()
+	expectedJobID := int64(1)
 	jimm := &jimmtest.JIMM{
 		BootstapManager_: func() jimm.BootstrapManager {
 			return &mocks.BootstapManager{
-				GetJobInfo_: func(ctx context.Context, user *openfga.User, jobId uuid.UUID, offset int) (apiparams.GetJobInfoResponse, error) {
-					if jobId != uuidGenerated {
+				GetJobInfo_: func(ctx context.Context, user *openfga.User, jobId int64, offset int) (apiparams.GetJobInfoResponse, error) {
```
ale8k, line 276: Everywhere we use jobId we're using int64, but int is better because it can adapt to 32vs64 bit systems. Why explicitly 64?

ale8k (approved):
lgtm, 1 minor comment. Nice!

internal/jujuapi/jimm_unit_test.go
kian99: It's the type that River's insert returns. If I used `int` where we pass the int64 then I'd have to do a cast `int(value)` and add a lint ignore rule for the potential overflow.

internal/jujuapi/jimm.go
kian99: Yip

internal/dbmodel/sql/postgres/034_recreate_job_logs.up.sql
```diff
@@ -0,0 +1,14 @@
+--- Drop and recreate job_logs to match the current JobLog model.
+---
+--- The table was originally introduced as bootstrap_logs (028) with a UUID job_id
+--- referencing job_tracker_entries, then renamed to job_logs (033). The new job
+--- identifier used by JIMM for these logs is River's job ID (BIGINT).
+
+DROP TABLE IF EXISTS job_logs;
```
luci1900, line 7: We do this conditional drop/create in several places, but I don't like it.

To be addressed separately if I remember.

kian99: Wdym by several places?

internal/jujuapi/jimm_unit_test.go
luci1900: I don't know if `int` is better either, types varying in size based on architecture tends to lead to obscure bugs.

luci1900 (approved):

internal/dbmodel/sql/postgres/034_recreate_job_logs.up.sql
luci1900: Several of our migrations do this, not just this new one.

kian99: I did it like this here because the table schema changed fairly drastically, it no longer has a foreign-key relation to the `job_tracker_entries` table, since I was removing the primary I think the most straightforward approach was to drop the table and recreate it.

kian99: > Several of our migrations do this, not just this new one.

Hmm yeah there might be one or two other that do it. But it's not too common. We can't change those old migrations anyway though as we've already released.

luci1900: I mean `IF EXISTS` and `IF NOT EXISTS`. Without those bits, it's more likely the migration would fail if applied in the wrong order. And if they're applied in the right order, it shouldn't be necessary.

PR #1862, opened 2026-02-10
refactor: switch most commands to a shared base

luci1900:
## Description
Switch most commands to a shared base with a method to set and get a JIMM client.

The two exceptions are `addmodel` and `migratemodel`, since they need more than one client internally and even the JIMM client is handled differently.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests


kian99 (approved):
Very nice! Ty for this

SimoneDutto (commented):
just a question

cmd/jaas/cmd/base.go
SimoneDutto, line 1: looks good but do we need for it to be public? Since we are doing whitebox testing everywhere?

luci1900: I suppose not, you're right. I don't think we even need the interface for it.

SimoneDutto: another question: i don't think the methods have to be public since you are embedding the thing? I could be wrong

SimoneDutto (approved):
looks good in general

cmd/jaas/cmd/base.go
luci1900: True, I can make the methods private as well. Although it doesn't matter much with the struct private.

luci1900: That worked fine.

SimoneDutto: yeah true, but if they are private is more explicit if you don't mind

PR #1864, opened 2026-02-11
chore: update copyright header

kian99:
## Description
Update the copyright header of all files to 2026. Avoids issues with the linter from constantly updating them during dev.


alesstimec (changes requested):
I think we've been doing this wrong.. the copyright year should not be updated. it should reflect the year when the file was created so that it serves as the basis of the copyright claim.

PR #1865, opened 2026-02-13
feat: return error when job is duplicate

kian99:
## Description
This PR ensures that when a client requests a bootstrap/controller-destroy/upgrade-to job, we return an error if the job inserted was a duplicate (i.e. if a job was still busy running).

As a drive-by I also made 2 other changes,
- Added timeouts to the upgrade/migrate/upgrade-to River workers (to avoid the default timeout of 1m).
- Added a check to the migrate logic that returns early if the migration failed.

Fixes [JUJU-9159](https://warthogs.atlassian.net/browse/JUJU-9159)

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

[JUJU-9159]: https://warthogs.atlassian.net/browse/JUJU-9159?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/jimm/upgrade/upgrade.go
+++ b/internal/jimm/upgrade/upgrade.go
@@ -7,10 +7,9 @@ package upgrade
 import (
 	"context"
 	"database/sql"
+	stderrors "errors"
 	"fmt"
 	"time"
-
-	stderrors "errors"
 
 	"github.com/juju/clock"
 	jujuerrors "github.com/juju/errors"
@@ -20,6 +19,7 @@ import (
 	"github.com/juju/retry"
 	"github.com/juju/version/v2"
 	"github.com/juju/zaputil/zapctx"
+	"github.com/riverqueue/river/rivertype"
 	"go.uber.org/zap"
 
 	"github.com/canonical/jimm/v3/internal/dbmodel"
@@ -55,7 +55,7 @@ type Store interface {
 
 // UpgradeEnqueuer defines the method to enqueue an upgrade job.
 type UpgradeEnqueuer interface {
-	EnqueueUpgradeTo(ctx context.Context, args rivertypes.UpgradeToArgs) (int64, error)
+	EnqueueUpgradeTo(ctx context.Context, args rivertypes.UpgradeToArgs) (*rivertype.JobInsertResult, error)
 }
 
 // upgradeManager provides a means to manage controller upgrades within JIMM.
@@ -312,7 +312,7 @@ func (u *upgradeManager) UpgradeTo(ctx context.Context, user *openfga.User, mode
 		return 0, errors.E(fmt.Errorf("failed to clone controller: %w", err))
 	}
 
-	id, err := u.enqueuer.EnqueueUpgradeTo(ctx, rivertypes.UpgradeToArgs{
+	job, err := u.enqueuer.EnqueueUpgradeTo(ctx, rivertypes.UpgradeToArgs{
 		ModelUUID:            modelUUID,
 		TargetVersion:        targetVersion,
 		Username:             user.Name,
@@ -321,8 +321,11 @@ func (u *upgradeManager) UpgradeTo(ctx context.Context, user *openfga.User, mode
 	if err != nil {
 		return 0, errors.E(fmt.Errorf("failed to enqueue model migration and upgrade job: %w", err))
 	}
+	if job.UniqueSkippedAsDuplicate {
+		return 0, errors.E("an upgrade job for this model is already in progress", errors.CodeInProgress)
+	}
 
-	return id, nil
+	return job.Job.ID, nil
 }
 
 // MigrateModel migrates a model to a new controller without upgrading the model's agent.
@@ -367,7 +370,7 @@ func (u *upgradeManager) MigrateModel(ctx context.Context, user *openfga.User, m
 			Attempts: 30,
 			Delay:    10 * time.Second,
 			Func: func() error {
-				_, err = u.jujuManager.ModelInfo(ctx, user, mt)
+				mi, err := u.jujuManager.ModelInfo(ctx, user, mt)
 				if err != nil {
 					return err
 				}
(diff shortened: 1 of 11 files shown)


luci1900 (approved):

internal/jimm/upgrade/upgrade.go
```diff
@@ -321,8 +321,11 @@ func (u *upgradeManager) UpgradeTo(ctx context.Context, user *openfga.User, mode
 	if err != nil {
 		return 0, errors.E(fmt.Errorf("failed to enqueue model migration and upgrade job: %w", err))
 	}
+	if job.UniqueSkippedAsDuplicate {
+		return 0, errors.E("an upgrade job for this model is already in progress", errors.CodeInProgress)
```
ale8k, line 325: Destroy is "a bootstrap job is already in progress - please wait for it to complete before starting a new one"`

Shouldn't we make these a bit consistent? Perhaps a func like "duplicateJobErr(jobName)"?

internal/jimm/upgrade/upgrade.go
```diff
@@ -377,12 +380,16 @@ func (u *upgradeManager) MigrateModel(ctx context.Context, user *openfga.User, m
...
-				// It hasn't migrated yet, so error out.
-				if m.Controller.Name != targetControllerName {
-					return modelNotMigratedErr
+				if m.Controller.Name == targetControllerName {
+					return nil
+				}
+
+				if mi.Migration != nil && mi.Migration.End != nil {
+					endUTC := mi.Migration.End.UTC().Format(time.RFC3339)
+					return fmt.Errorf("model migration failed: migration ended at %s with status %s", endUTC, mi.Migration.Status)
```
ale8k, line 389: Would be better as a nested fmt.Errorf, otherwise it's reading like two errors despite being one.

internal/jimm/upgrade/upgrade_test.go
```diff
@@ -418,6 +421,55 @@ func TestMigrateModel_Success(t *testing.T) {
 	c.Assert(err, qt.IsNil)
 }
 
+func TestMigrateModel_EndsEarly(t *testing.T) {
```
ale8k, line 424: Nice!

internal/river/migrationworker.go
```diff
@@ -86,3 +87,8 @@ func (w *migrationWorker) Work(ctx context.Context, job *river.Job[migrationWork
 
 	return nil
 }
+
+// Timeout implements the [river.Worker] interface.
+func (w *migrationWorker) Timeout(*river.Job[migrationWorkerArgs]) time.Duration {
+	return 10 * time.Minute
```
ale8k, line 93: We think the max time of a migration is 10 minutes?

internal/river/upgradetoworker.go
```diff
@@ -35,6 +36,11 @@ type upgradeToWorker struct {
 	awaitCompletion awaitCompletionFunc
 }
 
+// Timeout implements the [river.Worker] interface.
+func (w *upgradeToWorker) Timeout(*river.Job[rivertypes.UpgradeToArgs]) time.Duration {
+	return 20 * time.Minute
```
ale8k, line 43: Maybe some godocs on the rationale behind the timeouts would be good. I get you've just done 10+10 here but I don't know why we picked these numbers.

internal/jimm/upgrade/upgrade.go
kian99: I don't think there's a need, the message is also intentional different. A bootstrap operation is "global" per se, while the model migrate+upgrade is per model so the error reflects that if that makes sense.

internal/jimm/upgrade/upgrade.go
kian99: Can you expand on what you mean? What would a nested fmt.Errorf look like?

internal/river/migrationworker.go
kian99: I asked IS and if the migration succeeds apparently its under 1 minute. If it fails it can take a long time (e.g. some unit agent is in an error state, I've seen this locally when implementing migrations to JIMM). Now this timeout is should be viewed in conjunction with what the migrate is doing which is retrying a call to ModelInfo every 10s for 30 attempts for a total of 5 minutes. We don't want the river worker to time out earlier than that and cancel the context.

internal/river/upgradetoworker.go
kian99: Adding, technically this timeout should be longer since the upgrade / migrate have retries but 20 minutes seems like decent place to start, wdyt? If the supervisor job times out, we should also probably cancel any job it is currently waiting for, but I'd add that work in a separate PR.

internal/jimm/upgrade/upgrade.go
ale8k: Yeah ok sure thing, Idm just thought you might've missed consistency but makes sense to me.

internal/jimm/upgrade/upgrade.go
ale8k: Shouldn't this:
```go
		return fmt.Errorf("model migration failed: migration ended at %s with status %s", endUTC, mi.Migration.Status)
```
Be:
```go
		return fmt.Errorf("model migration failed: %w", fmt.Errorf(migration ended at %s with status %s", endUTC, mi.Migration.Status))
```
Incase we ever wanna unwrap? 

Not a big deal more of a nit.

kian99: I think it's okay, not sure what we'd be looking to unwrap.

internal/river/migrationworker.go
ale8k: I'm with ya, I was more thinking "give people chance to resolve the states". But we'll validate everything is ready to be migrated (running states) ahead of time anyway with the dryrun stuff. Fine with me.

internal/river/upgradetoworker.go
ale8k: Seems good to me! And yes 100%, a simple little context poll in a new routine when we start the supervisor which receives jobIds on a channel perhaps?

ale8k (approved):

internal/jimm/upgrade/upgrade.go
ale8k: Yeah sure sure, just reads more correct when we wrap is all.

PR #1866, opened 2026-02-16
chore(jujuclient): remove Ping because it's unused

SimoneDutto:
## Description
I was migrating Ping to the jujuclient, but searching for its use i couldn't find it. So i've just removed it


kian99 (commented):
While you're removing this, I see there this code in `internal/jujuclient/dial.go`
```go
		if err := client.Call(ctx, "Pinger", 1, "", "Ping", nil, nil); err != nil {
			zapctx.Error(ctx, "connection failed", zap.Error(err))
			return false
		}
		return true
```
We can replace this with usage of the Juju api package.

SimoneDutto:
> While you're removing this, I see there this code in `internal/jujuclient/dial.go`
> 
> ```go
> 		if err := client.Call(ctx, "Pinger", 1, "", "Ping", nil, nil); err != nil {
> 			zapctx.Error(ctx, "connection failed", zap.Error(err))
> 			return false
> 		}
> 		return true
> ```
> 
> We can replace this with usage of the Juju api package.

there is not client for the pinger method. At least that i could fin

kian99:
> there is not client for the pinger method. At least that i could fin

Huh you're right, that's interesting. Guess this is all we can do.

kian99 (approved):
Gonna try and merge https://github.com/canonical/jimm/pull/1633 before this btw.

ale8k (approved):
Ah ya lgtm.

PR #1867, opened 2026-02-16
chore: rename bootstrap facades and params

kian99:
## Description
This PR renames the bootstrap/destroy-controller job methods/params/CLI to separate the older bootstrap job logic from the river job management methods that we want to introduce. Bootstrap jobs were quite specific in how they worked and with the introduction of River we want the new approach to be generic while leaving the bootstrap behaviours as-is.

Note that this is a breaking change, but the bootstrap logic is not expected to have adoption yet (without a formal designation we could say it's not a GA feature).

The one pain point in this PR is the change in the `job-status` command and facade method. As mentioned above, the logic backing this was very specific to controller bootstrap/destroy. What I've opted to do is create a `juju jaas bootstrap-status` command with an alias for `juju jaas destroy-status`. The facade has changed from `GetJobInfo` to `BootstrapInfo` which is a bit unfortunate as it is used for bootstrap status **and** destroy-controller status.

Note that I haven't renamed the types/functions in the BootstrapManager. I can do that if desired but it is less problematic since its names are scoped to that package.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests


cmd/jaas/cmd/bootstrap.go
```diff
@@ -232,26 +232,26 @@ func (c *bootstrapCommand) Run(ctxt *cmd.Context) error {
...
 			resp.JobID,
 		)
 	} else {
 		fmt.Printf(`
-Starting bootstrap job.
+Starting bootstrap.
 
-Should you cancel this process, you can track the progress via job-status with the job ID:
-	juju [jaas] job-status %s
+Should you cancel this process, you can track the progress via bootstrap-status with the job ID:
```
ale8k, line 253: Still says job ID on the end, should we call it a "bootstrap-status-id"? 😆

cmd/jaas/cmd/bootstrapstatus.go
```diff
@@ -17,27 +17,28 @@ import (
 )
 
 const (
-	jobStatusCommandDoc = `
-Displays logs for a job.
+	bootstrapStatusCommandDoc = `
+Displays logs for a bootstrap or destroy-controller job.
```
ale8k, line 21: We call it a job again, but the example is just <id>, perhaps we just say "id"?..

ale8k (approved):
looks fine to me, naming is kinda hard here as I think we should call them job ids still despite the command being renamed but idk... It's up to you aha. Leaving my approval anyways.

SimoneDutto (approved):

cmd/jaas/cmd/bootstrap.go
kian99: Yeah I was also unsure, I think some mention of "job" is okay since that's probably a better use of naming then adding "operation" or something else everywhere.

PR #1869, opened 2026-02-17
chore(jujuclient): use jujuclient instead of raw facades for modelsummarywatcher

SimoneDutto:
## Description
This is a little different from the others because the client returns a full watcher you can call `Next` and `Stop`.
So i've added an interface to enable mock testing in the jimm manager layer.


internal/jimm/juju/generate.go
kian99, line 1: I'd suggest renaming this to generate.go

internal/jujuclient/modelsummarywatcher.go
```diff
@@ -17,40 +15,16 @@ func (c Connection) SupportsModelSummaryWatcher() bool {
 	return c.hasFacadeVersion("Controller", 11)
 }
 
+// SummaryWatcher defines the interface for watching model summaries.
+type SummaryWatcher interface {
+	Stop() error
+	Next() ([]params.ModelAbstract, error)
+}
```
kian99, line 22: Would we be able to define this interface in `internal/jimm/juju` inside the API definition?

SimoneDutto: i've tried, but it seems it's not possible because:
- we can't return the concrete type here because it needs to satisfy the API interface in the Dialer
- we can't use the interface from `internal/jimm/juju` because it's a cycle
- we can't define two interfaces with the same methods because the types don't match

internal/jimm/juju/generate.go
SimoneDutto: we put the go generate in package.go usually right?

kian99: We used to have a package_test.go for defining the entrypoint for gocheck and people put the go gen comments there. Without gocheck I'd put the generate commands in a generate.go file so it's easier to find.

SimoneDutto: i'm fine with it

internal/jujuclient/modelsummarywatcher.go
kian99: Thanks for taking a look, sounds good.

internal/jimm/juju/watcher_test.go
```diff
@@ -144,7 +144,7 @@ func TestModelSummaryWatcher(t *testing.T) {
 
 			mockWatcher.EXPECT().Stop().Do(func() error {
 				return nil
-			})
+			}).AnyTimes()
```
kian99, line 147: Why any times?

kian99 (approved):

internal/jimm/juju/watcher_test.go
SimoneDutto: we have a ticker in ms that is creating the watcher unless it has been created already, and i think it's racy and sometimes the watcher is created twice. It wasn't a problem before because we hadn't a check around how many times the stop is called

SimoneDutto: the code we have for this watcher is very odd btw

luci1900 (approved):

PR #1871, opened 2026-02-17
refactor: improve domain layout

kian99:
## Description
In this PR we refactor where certain interfaces are defined. Currently we define several interfaces in `internal/jimm/` that define what each of the domain packages exposes i.e. we have a `GroupManager` interface that defines the methods the `group.groupManager` struct exposes (notice that the struct itself is not exported). This works but ideally we'd define the interfaces where they are consumed instead in `internal/jujuapi` and `internal/jimmhttp/rebac_admin`. 

This PR moves those interface definitions to `internal/jujuapi` and the `rebac_admin` package re-uses them. Functionally not much has changed, but possibly this provides more clarity when navigating the codebase.


internal/jujuapi/interface.go
```diff
@@ -1,27 +1,338 @@
...
+	BootstrapManager() BootstrapManager
+	UpgradeManager() UpgradeManager
+	JobManager() JobManager
 
 	ResourceTag() names.ControllerTag
 	PubSubHub() *pubsub.Hub
 }
+
+// RoleManager provides a means to manage roles within JIMM.
+type RoleManager interface {
```
ale8k, line 56: Nice :)

internal/jujuapi/jimm_adapter.go
```diff
@@ -0,0 +1,72 @@
+// Copyright 2026 Canonical.
```
luci1900: I still don't like this, but at least it's in one place.

luci1900 (approved):
Sure.

I felt we were overusing package privacy anyway.

alesstimec (approved):
LGTM, but without the copyright year changes

PR #1873, opened 2026-02-17
feat: use juju client for storage

ale8k:
## Description

Switches our storage client to use the juju client. Cleaned up a bit of the status parser and usage of context too.


ale8k:
closing and reimplementing to keep context.

PR #1874, opened 2026-02-18
feat: use Juju api pkg for application offers

kian99:
## Description
In this PR we move to Juju's api package for application offer related operations. The key change is that we no longer return jujuparams.ApplicationOfferAdminDetailsV5 types and instead return crossmodel.ApplicationOfferDetails types.

A few things worth noting,
1. ~~This PR is a draft until we land https://github.com/juju/juju/pull/21801 and update to that version of Juju~~. This resolves an issue where certain API calls for offer details don't include the offer UUID forcing an extra round-trip to our database. The only issue is that the offer UUID was present on the params type but was missing on the crossmodel type.
2. I've added some unit tests to the `jujuapi` layer for app offers, there were none before. These were mostly AI generated and then tweaked.

Fixes [JUJU-9128](https://warthogs.atlassian.net/browse/JUJU-9128)

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

[JUJU-9128]: https://warthogs.atlassian.net/browse/JUJU-9128?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


SimoneDutto (approved):

internal/jimm/juju/applicationoffer.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2025 Canonical.
+// Copyright 2026 Canonical.
```
alesstimec, line 1: no :)

internal/jujuapi/applicationoffers.go
```diff
@@ -189,9 +183,6 @@ func (r *controllerRoot) modifyOfferAccess(ctx context.Context, change jujuparam
 		if err := r.jimm.PermissionManager().RevokeOfferAccess(ctx, r.user, change.OfferURL, ut, change.Access); err != nil {
 			return errors.E(err)
 		}
-		if err := r.jimm.JujuManager().RevokeOfferAccessOnController(ctx, r.user, ut, change.OfferURL, change.Access); err != nil {
```
alesstimec, line 192: i see we are no longer revoking/granting offer access on the actual controller. Which seems ok, IFF nobody is using Candid anymore. If somebody attempted to migrate a model from a JAAS controller to a standalone controller connected to Candid AND Candid was configured to use the same IdP (know the same set of users), then this offers become unusable.. An edge case, but we need to be aware of it.

alesstimec (approved):
LGTM with a comment (read it!)

internal/jimm/juju/applicationoffer.go
kian99: I'll fix this in a follow-up to avoid the long-running CI. Hopefully that's okay. I've disabled the header linter so the mistake shouldn't creep in anymore either.

internal/jujuapi/applicationoffers.go
kian99: Btw I think we'd have other issues as well since we no longer grant permissions to models and cloud, etc.

PR #1878, opened 2026-02-19
feat(list-jobs): add ListJobs facade for JIMM

SimoneDutto:
## Description
Add ListJobs To JIMM facade.
<strike>
A bunch of maybe controversial decisions I've made:
- don't handle the pagination: `.First()` accepts from 1 to 10000 so i don't think there is a world where we actually need 2 pages of jobs. We can always query from the CLI the maximum.
- No defaults: the default Kind, Status and Count will be set from the CLI, and the server will just validate it's the right thing.
</strike>

ListJobs is paginated with a continuation token (it doesn't support limit and offset), and the token is the riverCursor struct base64 encoded.
ListJobs accepts filters, but if empty it will just return all jobs of all kinds and error state

I've added a way to do actual integration testing with river, because I wasn't sure the mock tests were enough for this pagination feature.
To accomplish that I need to add `func (c *Client) SetClient(client *river.Client[*sql.Tx])` because i need the `internal/river:Client` struct to initialize the JobManager, but i also need the actual riverClient to interact with river inserting jobs, waiting, etc.


internal/jimm/jobs/jobs.go
```diff
@@ -55,3 +57,57 @@ func (j *JobManager) GetJobInfo(ctx context.Context, jobID int64) (JobInfo, erro
...
+	if err != nil {
+		return nil, err
+	}
+	kind, err := convertKind(params.Kind)
+	if err != nil {
+		return nil, err
+	}
+	if params.Count <= 0 || params.Count > 10000 {
+		return nil, errors.E("Count must be between 1 and 10_000.")
+	}
```
kian99: Without pagination it will be annoying for the web team to build UIs. They could do paginate on the client side but if it's not too hard to accept an offset or something of the sort, I think we should.

SimoneDutto: is the web team implementing the listing of jobs from the UI? Because in that case this very explicit API where you need to specify everything is not very handy.

Because to get all jobs of all kinds and all statuses you need to call the same facade multiple times.

If we want some more UI friendly maybe we can make the filters optional, and return everything, not just the ids.

SimoneDutto: what do you prefer? I wasn't sure if the UI wants to list the jobs, but maybe it is a requirement

kian99: >is the web team implementing the listing of jobs from the UI? 

I assume we'll end up there so you can view the progress of your upgrades, bootstraps, etc. Like we disucssed today.

It will likely be the case that the GetJobDetail() for getting the details of a single job will be used when you want to zoom in on one job and see the logs/status/failures/etc.

For listing jobs we should return enough info that one could build a sensible summary view, i.e. current-attempt-number, total attempts, status, is-done, etc.

internal/jimm/jobs/jobs.go
```diff
@@ -55,3 +57,132 @@ func (j *JobManager) GetJobInfo(ctx context.Context, jobID int64) (JobInfo, erro
...
+			riverStates = append(riverStates, rivertype.JobStateAvailable, rivertype.JobStateScheduled)
+		default:
+			return nil, errors.E("invalid job status: %s", status)
+		}
+	}
+
+	return riverStates, nil
+}
+
+func convertKinds(kinds []string) ([]string, error) {
```
kian99: I think we should avoid checking if the string is a valid kind, I don't think we gain anything.

internal/jimm/jobs/jobs_integration_test.go
```diff
@@ -0,0 +1,272 @@
...
+	c.Assert(err, qt.IsNil)
+
+	return jobManager, riverClient
+}
+
+func TestListJobs_Pagination(t *testing.T) {
+	c := qt.New(t)
+	ctx := c.Context()
+
+	jobManager, client := setupJobsIntegrationTest(c)
```
kian99, line 138: Very nice!

internal/jimm/jobs/jobs_integration_test.go
```diff
@@ -0,0 +1,272 @@
...
+
+	// Test filtering by failed status
+	resp, err := jobManager.ListJobs(ctx, apiparams.ListJobsRequest{
+		Statuses: []apiparams.JobStatus{apiparams.StatusFailed},
+		Count:    100,
+	})
+	c.Assert(err, qt.IsNil)
+	c.Assert(len(resp.Jobs), qt.Equals, 3, qt.Commentf("Expected 3 failed jobs, got %d", len(resp.Jobs)))
+
+	// Verify all returned jobs are in cancelled state (failed jobs get cancelled)
```
kian99: What does it mean that failed jobs get cancellled?

pkg/api/params/params.go
```diff
@@ -716,3 +716,38 @@ type JobInfoResponse struct {
...
+	// Cursor is the pagination cursor to continue from a previous query.
+	Cursor string `json:"cursor,omitempty"`
+}
+
+// ListJobInfo holds summary information about a job.
+type ListJobInfo struct {
+	// ID is the unique identifier for the job.
+	ID int64 `json:"id" yaml:"id"`
+	// Status is the current status of the job.
+	Status string `json:"status" yaml:"status"`
```
kian99: We should probably replace this type from `string` -> `JobStatus` and same in the struct I created, `JobInfoResponse`, to indicate that it is not any arbitrary string.

kian99 (approved):
This is great, ty!

internal/jimm/jobs/jobs.go
kian99: Unless we add a way to list what the valid kinds are. But for now I'd leave out the check and we'll iterate on it in the future.

internal/jimm/jobs/jobs_integration_test.go
SimoneDutto: yeah this comment is a thing to help me and the AI, i can remove it

pkg/api/params/params.go
SimoneDutto: yep, let me do it

internal/jimm/jobs/jobs.go
SimoneDutto: why? it's good to make sure you don't put the wrong string in

SimoneDutto: i didn't create the enum because it seemed a little too much, but at least some validation is nice

kian99: Unlike with `status` field which has known values ahead of time defined in the API params, the `Kind` field has no known values ahead of time. I want to avoid clients (like the dashboard) from hard-coding values on their end like "bootstrap-job" and "upgrade-to" because they want to avoid errors from validation, that would cause mismatches between older dashboard versions and new versions of JIMM. 

I would leave this field without validation unless/until we also have a way to provide clients with a list of valid `kind` strings, possibly via another endpoint.

SimoneDutto: This validation doesn't change the fact they'd still need to hard-code the type string somewhere if they want to filter jobs.
This validation is just an harness on our side to make sure they job string they are searching does actually exist on our end.

What you are describing is a different problem about how do we communicate the list of job types

SimoneDutto: we can put the them just like JobStatus as an enum in the params pkg

kian99: > we can put the them just like JobStatus as an enum in the params pkg

I wouldn't do that because they can change, it's dynamic based on what we move to River.

> This validation doesn't change the fact they'd still need to hard-code the type string somewhere if they want to filter jobs.

They should avoid that because it will mean if JIMM moves a new thing to River, you can't search for that from an old version of the client when the API really hasn't changed.

My biggest advocating reason for leaving out this validation is that whenever we add a new job kind we have to update this as well. Just avoid the pain and don't validate it and return an empty list if it's not a valid job kind. Then we can implement a way for clients to retrieve all valid job kinds. If you really want the validation, then we should add that method to "get all valid job kinds" but until then just keep it simple imo.

SimoneDutto: ok, removed then

internal/jimm/jobs/jobs.go
```diff
@@ -47,11 +48,123 @@ func (j *JobManager) GetJobInfo(ctx context.Context, jobID int64) (JobInfo, erro
...
+		case apiparams.StatusRunning:
+			riverStates = append(riverStates, rivertype.JobStateRunning)
+		case apiparams.StatusFailed:
+			riverStates = append(riverStates, rivertype.JobStateCancelled)
+		case apiparams.StatusSuccessful:
+			riverStates = append(riverStates, rivertype.JobStateCompleted)
+		case apiparams.StatusPending:
+			riverStates = append(riverStates, rivertype.JobStateAvailable, rivertype.JobStateScheduled)
+		default:
+			return nil, errors.E("invalid job status: %s", status)
```
alesstimec: i think you should add `errors.CodeBadRequest` here

internal/jimm/jobs/jobs.go
```diff
@@ -47,11 +48,123 @@ func (j *JobManager) GetJobInfo(ctx context.Context, jobID int64) (JobInfo, erro
...
+	if err != nil {
+		return apiparams.ListJobsResponse{}, err
+	}
+	// Set default count if not provided
+	count := req.Count
+	if count <= 0 {
+		count = 100
+	}
+	if count > 10000 {
+		return apiparams.ListJobsResponse{}, errors.E("Count must be between 1 and 10_000.")
```
alesstimec: i think our errors strings begin with a lower case character.. also, `100` and `10000` should be consts.. why did you put it as `10_000` in the error string?

internal/jimm/jobs/jobs_integration_test.go
```diff
@@ -0,0 +1,278 @@
+// Copyright 2025 Canonical.
```
alesstimec: 2026?

internal/testutils/jimmtest/mocks/jimm_jobmanager_mock.go
```diff
@@ -0,0 +1,30 @@
+// Copyright 2025 Canonical.
```
alesstimec: 2026?

alesstimec (approved):
LGTM with a few comments

internal/jimm/jobs/jobs.go
SimoneDutto: 10_000 is the max accepted by the river's API

SimoneDutto: i can put them in a const

PR #1876, opened 2026-02-18
feat: use juju client for storage

ale8k:
## Description
Uses juju client for storage. Leaves the context in for 4.0 client changes.

PR #1879, opened 2026-02-20
test: use new integration test suite for dial tests

kian99:
## Description
This PR updates the dial tests in the `jujuclient` to use the new test suite that uses real Juju controllers and removes the final usages of Juju's ConnSuite (but doesn't remove the ConnSuite itself just yet).

I've also removed the `jujuclient` package's test files `cloud_test.go` and `storage_test.go` which are being removed in https://github.com/canonical/jimm/pull/1876 and another pending PR for switching clouds over to the Juju API client.

It's worth noting that tests like `TestDial` are taking > 5s because the suite creates 3 models and then tears them down at the end. In a follow-up PR I'll look at making this model setup optional.

## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests


internal/jujuclient/dial_test.go
```diff
@@ -6,97 +6,56 @@ import (
...
+	"github.com/juju/version/v2"
 	gc "gopkg.in/check.v1"
 
 	"github.com/canonical/jimm/v3/internal/dbmodel"
-	"github.com/canonical/jimm/v3/internal/jimm/juju"
 	"github.com/canonical/jimm/v3/internal/jujuclient"
 	"github.com/canonical/jimm/v3/internal/testutils/jimmtest"
 )
 
-type jujuclientSuite struct {
```
SimoneDutto, line 22: yeaaaaah

SimoneDutto (approved):

kian99:
@SimoneDutto One tricky thing with this PR, I've had to move the `dial_test.go` file from `jujuclient` to `testing` because the CI is setup to only run the tests in `testing` as integration tests with real controllers. Maybe in the future we can simplify and just run all the tests in one job, the unit tests should be pretty fast in any case compared to the integration tests so the difference will hopefully be negligible and then you can put tests anywhere as needed.

SimoneDutto:
> @SimoneDutto One tricky thing with this PR, I've had to move the `dial_test.go` file from `jujuclient` to `testing` because the CI is setup to only run the tests in `testing` as integration tests with real controllers. Maybe in the future we can simplify and just run all the tests in one job, the unit tests should be pretty fast in any case compared to the integration tests so the difference will hopefully be negligible and then you can put tests anywhere as needed.

yeah, we can try. At the time of setting up these things unit tests were still half e2e and half unit.

testing/dial_test.go
```diff
@@ -0,0 +1,115 @@
+// Copyright 2025 Canonical.
```
alesstimec, line 1: 2026?

testing/dial_test.go
```diff
@@ -0,0 +1,115 @@
...
+	"github.com/canonical/jimm/v3/internal/testutils/jimmtest"
+)
+
+type dialSuite struct {
+	jimmtest.WebsocketE2ESuite
+}
+
+var _ = gc.Suite(&dialSuite{})
+
+func (s *dialSuite) TestDial(c *gc.C) {
```
alesstimec, line 25: we could take it a step further and get rid of gc in favor of qt.. but..

alesstimec (approved):

testing/dial_test.go
kian99: I was about to say "Lol good one" until I realised this was a legitimate comment. I just moved the file and made some tweaks but I guess git is seeing that as too many changes to count as a rename - see `internal/jujuclient/dial_test.go` which was removed in this PR.

testing/dial_test.go
kian99: Currently working on that :+1:

PR #1883, opened 2026-02-24
feat: remove gocheck from test suites

kian99:
## Description
This PR does a few things, particularly it:
1. Removes the JujuConnSuite, this was a simple ~50 line removal since nothing was using it.
2. Refactors the test suites away from gocheck to use quicktest instead. 
3. Refactors the test suites away to move away from the suite model in line with the above move away from gocheck.

You will notice any test that relied on gocheck and our integration suites has had very similar changes.
- All checkers like `gc.Equal` are now `qt.Equal`.
- Structs like `type fooSuite struct` have been removed if unneeded or adapted into a small `SetupFooTests()` function that is called at the start of each test that needs the custom setup.
- Refactored tests now call `c := qt.New(t)` and `s := jimmtest.SetupJimmEnv(c)` or some variant at the start of each test to setup the integration test environment.

There are still some trace uses of gocheck around the codebase used in unit tests that I will get rid of in a follow-up.

Fixes [JUJU-9240](https://warthogs.atlassian.net/browse/JUJU-9240)

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

[JUJU-9240]: https://warthogs.atlassian.net/browse/JUJU-9240?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/jimmhttp/rebac_admin/identities_integration_test.go
```diff
@@ -21,14 +21,15 @@ import (
 	jimmnames "github.com/canonical/jimm/v3/pkg/names"
 )
 
-type identitiesSuite struct {
-	jimmtest.JIMMSuite
+func SetupIdentitiesTest(c *qt.C) (s jimmtest.JIMMEnv) {
```
alesstimec: why do we need a separate function for this? why not just call jimmtest.SetupJimmEnv?

internal/jimmhttp/rebac_admin/resources_integration_test.go
```diff
@@ -3,25 +3,23 @@
...
 	"github.com/canonical/jimm/v3/internal/common/utils"
 	"github.com/canonical/jimm/v3/internal/jimmhttp/rebac_admin"
 	"github.com/canonical/jimm/v3/internal/jujuapi"
 	"github.com/canonical/jimm/v3/internal/testutils/jimmtest"
 )
 
-type resourcesSuite struct {
-	jimmtest.JIMMSuite
+func SetupResourcesTest(c *qt.C) (s jimmtest.JIMMEnv) {
+	return jimmtest.SetupJimmEnv(c)
```
alesstimec: same question here..

alesstimec (approved):

internal/jimmhttp/rebac_admin/identities_integration_test.go
kian99: This a leftover, I started doing it this way then realised what you wrote was simpler and changed it but forgot to come back to these two. Fixing them.

internal/jimmhttp/rebac_admin/resources_integration_test.go
kian99: Fixed

internal/testutils/jimmtest/setup_e2e.go
```diff
@@ -205,9 +209,9 @@ type DeployApplicationParams struct {
 }
 
 // DeployApplication deploys a charm in the specified model as the given user.
-func (s *WebsocketE2ESuite) DeployApplication(c *gc.C, user *openfga.User, modelTag names.Tag, params DeployApplicationParams) {
+func (s *WebsocketE2ESuite) DeployApplication(c *qt.C, user *openfga.User, modelTag names.Tag, params DeployApplicationParams) {
```
SimoneDutto: rename WebsocketE2ESuite to something not mentioning suite anymore

internal/testutils/jimmtest/setup_e2e.go
```diff
@@ -280,66 +274,51 @@ func (s *WebsocketE2ESuite) OpenNoAssert(c *gc.C, d LoginDetails, modelTag *name
...
 	return conn
 }
 
-// E2ESuite is a suite that initialises a JIMM with
-// an externally bootstrapped controller.
-// It creates cloud credential, and creates a model.
-type E2ESuite struct {
-	JIMMSuite
-	LoggingSuite
+// JimmWithControllers is a environemnt that initialises a JIMM svc
```
SimoneDutto: environment

internal/testutils/jimmtest/setup_e2e.go
```diff
@@ -280,66 +274,51 @@ func (s *WebsocketE2ESuite) OpenNoAssert(c *gc.C, d LoginDetails, modelTag *name
...
 }
 
-func (s *E2ESuite) SetUpTest(c *gc.C) {
-	s.cleanupFuncs = nil
-	s.UseHardcodedJWKS(c)
-	s.JIMMSuite.SetUpTest(c)
-	s.LoggingSuite.SetUpTest(c)
+func SetupJimmWithControllers(c *qt.C, opts ...SetupOption) JimmWithControllers {
+	jimmEnv := SetupJimmEnv(c, append([]SetupOption{WithHardcodedJWKS()}, opts...)...)
+	// SetupTestLogger(c)
```
SimoneDutto: why?

internal/testutils/jimmtest/setup_e2e.go
```diff
@@ -348,75 +327,62 @@ func (s *E2ESuite) SetUpTest(c *gc.C) {
...
 	s.Model = new(dbmodel.Model)
 	s.Model.SetTag(mt)
-	err = s.JIMM.Database.GetModel(ctx, s.Model)
-	c.Assert(err, gc.Equals, nil)
-}
+	err = jimmEnv.JIMM.Database.GetModel(ctx, s.Model)
+	c.Assert(err, qt.Equals, nil)
 
-// Cleanup registers a function to be called during TearDownTest.
-func (s *E2ESuite) Cleanup(f func()) {
```
SimoneDutto, line 371: leaving a comment to make sure we have a cleanup

local/jimm/setup-controller.sh
```diff
@@ -40,5 +40,5 @@ if [[ -n "${AGENT_VERSION:-}" ]]; then
 fi
 
 echo "Bootstrapping controller"
-JUJU_DEV_FEATURE_FLAGS=ssh-jump juju bootstrap "${BOOTSTRAP_ARGS[@]}"
+JUJU_DEV_FEATURE_FLAGS=ssh-jump /snap/juju/current/bin/juju bootstrap "${BOOTSTRAP_ARGS[@]}"
```
SimoneDutto: it's a problem in corporate laptops right?

SimoneDutto (commented):
i think we can row get rid of `juju-db` from the sys-deps in the makefile right?

Are we are missing the cleanup for each test? Or i've just missed it?

local/jimm/setup-controller.sh
kian99: Ah yeah, I'll revert this, I use it locally.

internal/testutils/jimmtest/setup_e2e.go
kian99: I need to delete that, it's already called inside of `SetupJimmEnv`.

internal/testutils/jimmtest/setup_e2e.go
kian99: Cleanup is way easier now. Whenever you create something that you need to cleanup you just call `c.Cleanup()` and the ordering is already handled because it acts like defer except across the entire test.

kian99:
@SimoneDutto The change to use `--replay` fixed that debug-log test. Only 1 more test is failing now, `TestOffer`, where it's hitting a timeout after 5 minutes trying to cleanup the model. Not sure why though.

SimoneDutto:
@kian99 one question:
I see every test starts with jimmtest.SetupWebsocketEnv(c) but i don't see the corresponding cleanup. I believe that each test creates a model, and cloud credentials and other stuff on the backing controller we never clean up?

kian99:
> @kian99 one question: I see every test starts with jimmtest.SetupWebsocketEnv(c) but i don't see the corresponding cleanup. I believe that each test creates a model, and cloud credentials and other stuff on the backing controller we never clean up?

When we create a model in the suite, we call `c.Cleanup(<cleanup-call>)` so the cleanup gets registered for the end of the test. Essentially you write the cleanup as you write the setup, with `c.Cleanup()` acting like a defer statement.

SimoneDutto:
> > @kian99 one question: I see every test starts with jimmtest.SetupWebsocketEnv(c) but i don't see the corresponding cleanup. I believe that each test creates a model, and cloud credentials and other stuff on the backing controller we never clean up?
> 
> When we create a model in the suite, we call `c.Cleanup(<cleanup-call>)` so the cleanup gets registered for the end of the test. Essentially you write the cleanup as you write the setup, with `c.Cleanup()` acting like a defer statement.

aaaa, thanks!

SimoneDutto (approved):
thanks for the effort!

PR #1884, opened 2026-02-24
fix(paths): move internal paths on a different port

### LINKED ISSUE #1872: /metrics endpoint shouldn't be exposed on the same port as jimm's service
right now we have the `/metrics` endpoint exposed on the same port as JIMM, this is an issue because when an ingress is created for jimm it exposes the metrics endpoints as well.

SimoneDutto:
## Description
Move internal paths under a different path.

> fly by increase ReadTimeout

Fix: #1872
## QA

`make dev-env`

Visit:
`http://jimm.localhost:17090/debug/status`
`http://jimm.localhost:17090/metrics`


cmd/jimmsrv/service/service.go
```diff
@@ -660,6 +649,28 @@ func NewServiceFromDependencies(ctx context.Context, deps *ServiceDependencies)
...
 }
 
+// NewInternalService builds the internal-only HTTP server serving /metrics and /debug.
+func NewInternalService(addr string, corsAllowedOrigins []string) *http.Server {
+	mux := chi.NewRouter()
+	corsOpts := cors.New(cors.Options{
+		AllowedOrigins:   corsAllowedOrigins,
+		AllowedMethods:   []string{"GET"},
+		AllowCredentials: true,
+	})
```
kian99, line 659: We don't need to worry about CORS for these endpoints unless we think the dashboard is going to request them.

cmd/jimmsrv/service/service.go
```diff
@@ -660,6 +649,28 @@ func NewServiceFromDependencies(ctx context.Context, deps *ServiceDependencies)
 	return s, nil
 }
 
+// NewInternalService builds the internal-only HTTP server serving /metrics and /debug.
```
ale8k: Don't put /metrics and /debug in godoc as it is likely subject to add more

cmd/jimmsrv/main.go
```diff
@@ -187,6 +187,12 @@ func start(ctx context.Context, s *service.Service) error {
 		}
 	}
 
+	internalAddr := os.Getenv("JIMM_INTERNAL_LISTEN_ADDR")
+	if internalAddr == "" {
+		internalAddr = ":9090"
```
ale8k, line 192: Prometheus default is 9090 isn't it? Maybe use 8080?

cmd/jimmsrv/main.go
```diff
@@ -246,7 +252,7 @@ func start(ctx context.Context, s *service.Service) error {
 	httpsrv := &http.Server{
 		Addr:              addr,
 		Handler:           jimmsvc,
-		ReadHeaderTimeout: time.Second * 5,
+		ReadHeaderTimeout: time.Second * 30,
```
ale8k, line 255: I think we can remove this possibly. I don't remember why I added it but I'm not worried?

ale8k (approved):
Some minor comments, but lgtm, nice!

cmd/jimmsrv/main.go
SimoneDutto: it doesn't matter anyways because we are mapping this port in docker, and in the charm

cmd/jimmsrv/main.go
SimoneDutto: if you remove it:
```
[10:38:02] ➜  jimm git:(separate-internal-paths) ✗ golangci-lint run      
cmd/jimmsrv/main.go:252:14: G112: Potential Slowloris Attack because ReadHeaderTimeout is not configured in the http.Server (gosec)
        httpsrv := &http.Server{
                Addr:    addr,
                Handler: jimmsvc,
        }
1 issues:
* gosec: 1
```

luci1900 (approved):
I'm not entirely convinced the separate port is necessary, but even just the separation of all internal paths registered in one place is useful.

cmd/jimmsrv/service/service.go
SimoneDutto: i've added it because they might need to ask for the version or something to show on the dashboard is already there

PR #1886, opened 2026-02-24
feat: implement state based jwt rotation

ale8k:
## Description
We have an issue in our current implementation where we immediately rotate the key, this is problematic as any controller's out there with the previous JWK cached will be unable to verify the authenticity of newer JWTs until the cache control TTL elapses. 

This is the first phase in a multiphase piece of work to rewrite our JWKS generation, rotation, and presentation. 
This PR is heavily inspired by the ideas explained [here](https://engineering.zalando.com/posts/2025/01/automated-json-web-key-rotation.html).

We now support rotating with a grace period between the N sets, dependent on the grace period configuration.

This PR specifically is purely the state machine and does not include the JWT generation not is it plugged it into JIMM yet.

I've additionally moved old rotate func up in the jwk file so that you can read only from the v2 for an easier review. (See my comment in jwks.go)

This implements:
- Implements V2 JWKS rotation state machine with metadata persistence and legacy migration.
- Ensures fresh installs create an active key immediately & & rotations use pre-active + grace overlap.
- Adds rotation tests using), with deterministic timing so we can be pretty sure this is sound (synctest)
- Crash tolerance (due to the metadata retrieval at each rotate)

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

### DIFF
--- a/internal/jimmjwx/jwks.go
+++ b/internal/jimmjwx/jwks.go
@@ -7,6 +7,7 @@ import (
 	"crypto/rand"
 	"crypto/rsa"
 	"crypto/x509"
+	"encoding/json"
 	"encoding/pem"
 	"fmt"
 	"time"
@@ -18,6 +19,7 @@ import (
 	"go.uber.org/zap"
 
 	"github.com/canonical/jimm/v3/internal/errors"
+	"github.com/canonical/jimm/v3/internal/vault"
 )
 
 // CredentialStore defines the interface for a store that can manage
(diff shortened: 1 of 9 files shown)


internal/jimmjwx/jwks.go
```diff
@@ -117,39 +210,42 @@ func rotateJWKS(ctx context.Context, credStore CredentialStore, initialExpiryTim
...
 
-// StartJWKSRotator starts a simple routine which checks the vaults TTL for the JWKS on a ticker.C.
-// It is expected that this routine will be cleaned up alongside other background services sharing
-// the same cancellable context.
-//
-// TODO(ale8k)[possibly?]:
-// For now, there's a single key, and this is probably OK. But possibly extend
-// this to contain many at some point differentiated by KIDs.
+// StartJWKSRotatorParams holds the params to start the JWKS rotator.
+type StartJWKSRotatorParams struct {
```
ale8k, line 214: Read from here, ignore top bit. Did that purely for reviewability

internal/jimmjwx/jwks.go
```diff
@@ -7,6 +7,7 @@ import (
 	"crypto/rand"
 	"crypto/rsa"
 	"crypto/x509"
+	"encoding/json"
```
ale8k, line 10: Read down to comment explaining why this was moved and where to review from.

ale8k:
Closing as it is being done in the charm.

PR #1887, opened 2026-02-25
test: simplify model creation in tests

kian99:
## Description
This PR does 2 things,
1. It simplifies the test environments we have, merging `JimmWithControllers` and `WebsocketEnv`. There were no usages of the former except within the latter, so I've merged them, keeping the name `JimmWithControllers` as I think it's clearer.
2. Removes any model creation from within the environment setup. Any tests that need to create a model must now create them within the test.

I had AI do some of the refactoring and it made what I think is a nice change. It created helper functions `CreateModelForBob` and `CreateModelForCharlie`. Our tests always used these users and setup permissions for them so I've called out their roles (and Alice's role) in a godoc. Charlie and Bob have add-model permissions on all controllers and Alice is a superuser.

The tests, I think, are easy to read when you have the above in mind.

As an interesting note, creation of a model + teardown consumes about 3s.

## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests

### DIFF
--- a/internal/testutils/jimmtest/setup_e2e.go
+++ b/internal/testutils/jimmtest/setup_e2e.go
@@ -131,18 +131,17 @@ func (info *ControllerInfo) ToAPIInfo() *api.Info {
 	}
 }
 
-// WebsocketEnv is a suite that initialises a JIMM with
+// JimmWithControllers is a suite that initialises a JIMM with
 // externally bootstrapped controller(s), and provides
 // methods to open websocket connections to the JIMM API.
-type WebsocketEnv struct {
-	JimmWithControllers
+type JimmWithControllers struct {
+	JIMMEnv
 
 	Params jujuapi.Params
 	HTTP   *httptest.Server
 
-	Credential2 *dbmodel.CloudCredential
-	Model2      *dbmodel.Model
-	Model3      *dbmodel.Model
+	BobCredential     *dbmodel.CloudCredential
+	CharlieCredential *dbmodel.CloudCredential
 }
 
 type LoginDetails struct {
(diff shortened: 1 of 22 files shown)


internal/testutils/jimmtest/setup_e2e.go
```diff
@@ -152,52 +151,94 @@ type LoginDetails struct {
...
-	jimmEnv := jimmWithControllers.JIMMEnv
-	s := WebsocketEnv{
-		JimmWithControllers: jimmWithControllers,
+// SetupJimmWithControllers sets up a JIMM environment with externally bootstrapped controllers defined in the config file.
+// The config file path is specified in the JIMM_BACKING_CONTROLLER_CONFIG environment variable.
+//
+// The environment includes some test users by convention:
+// - bob@canonical.com has add-model permission on the controllers and cloud-credentials.
+// - charlie@canonical.com has add-model permission on the controllers and cloud-credentials.
+// - alice@canonical.com is an admin.
```
luci1900: Mention alice first, I guess.

internal/testutils/jimmtest/setup_e2e.go
```diff
@@ -131,18 +131,17 @@ func (info *ControllerInfo) ToAPIInfo() *api.Info {
 	}
 }
 
-// WebsocketEnv is a suite that initialises a JIMM with
+// JimmWithControllers is a suite that initialises a JIMM with
```
ale8k: It isn't a suite anymore, nor are any of the tests - should we just say it is a test helper or something? Idk wdyt?

internal/testutils/jimmtest/setup_e2e.go
```diff
@@ -152,52 +151,94 @@ type LoginDetails struct {
...
-	err = jimmEnv.JIMM.Database.GetModel(ctx, s.Model3)
+	s.CharlieCredential = new(dbmodel.CloudCredential)
+	s.CharlieCredential.SetTag(cct)
+	err = jimmEnv.JIMM.Database.GetCloudCredential(ctx, s.CharlieCredential)
 	c.Assert(err, qt.Equals, nil)
 
+	return s
+}
+
+func (s *JimmWithControllers) CreateModelForCharlie(c *qt.C) *dbmodel.Model {
```
ale8k, line 228: godoc all the public members. And just explain what it does and how it is expected to be used please.

internal/testutils/jimmtest/setup_e2e.go
```diff
@@ -152,52 +151,94 @@ type LoginDetails struct {
...
+		Owner:  names.NewUserTag("charlie@canonical.com"),
+		Name:   petname.Generate(2, "-"),
+		Cloud:  names.NewCloudTag(TestE2ECloudName),
+		Region: TestE2ECloudRegionName,
+		Cred:   s.CharlieCredential.ResourceTag(),
+	}
+	return s.CreateModel(c, args)
+}
+
+func (s *JimmWithControllers) CreateModelForCharlieWithBobReadAccess(c *qt.C) *dbmodel.Model {
```
ale8k, line 243: I'm not really a fan of this, I'd prefer we didn't put users in the function names and just pass a user. 

I.e.,

CreateModelFull(c, email) (terrible name I know but just something that handles the cred too) and it can create the credential too. .
Then maybe another like SetModelAccess(c, email, model).
Etc.

More of a nit I guess and not a huge deal though.

ale8k (approved):
LGTM nice work on this cleans it up a lot, I would prefer we remove the names inside of function signatures though and just pass a user.

internal/testutils/jimmtest/setup_e2e.go
luci1900: I actually like the hard-coded names. Tests are often too verbose, making a sort of DSL out of common actions is often helpful.

kian99: So the AI made these helpers and I was on the path of changing them just like you suggested to `CreateModel(c, email)` until I went to look at the usages, and it all fit very nicely into our test scenario. I.e. create a model owned by person A, verify it can't be seen by person B. Create a model by person A and one by person B, verify the admin can see both, etc. If you read the tests now it's just very clear.

ale8k: I did, and it is clear, I just don't really like having usernames within the funcs as they shouldn't hold any meaning but they do now, and any other tests ideally they should hold the same meaning but might not if we aren't careful and think it would be just as clear specifying the user and carefully naming variables within the test. If that makes sense...?

kian99: Discussed offline that we'll keep them and then see how they fare.

luci1900 (approved):

PR #1889, opened 2026-02-26
feat: use jujuclient for clouds

ale8k:
## Description
This PR updates the cloud client to use the juju client.

The primary changes of this PR are:
- Updates our JIMM dep to the tip of 3.6 to include the new CheckCredentialsModels client method
- Use jujuclient in clouds.go on all methods
- Rename UpdateCredential -> UpdateCloudsCredentialForce to reflect what it actually is doing
- Update client method CheckCredentialsModels godoc so it is clear why we have it and don't just call UpdateCloudsCredential without the force flag (as it performs this check for a single model)
- Change return type of UpdateCloudsCredentialForce and  CheckCredentialsModels to return the clients return type and not the nested Model field's type for two reasons:
    - Should we ever wish to remove the client, it will be easier as we handle the logic of parsing the error outside the package
    - It was duplicated across the two methods, now it is in a single location
- Change Clouds() to return the type, for the same reason as above (_should we ever wish to remove the client_)

There's a little remaining work to refactor `updateControllerCloudCredential`, UpdateCloudsCredentialForce and CheckCredentialsModels actually take different parameters and we need this in our client right now. It would be preferable we don't have this in our client (again, in the event we wish to just delete the file). 

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests


internal/jimm/juju/cloudcredential.go
```diff
@@ -237,25 +237,39 @@ func (j *JujuManager) updateCredential(ctx context.Context, credential *dbmodel.
...
 			Attributes: attr,
 		},
 	})
+
 	if err != nil {
-		return models, errors.E(err)
+		return nil, errors.E(err)
+	}
+
+	// Shouldn't happen, the Juju client presumes that
```
luci1900, line 260: If we assume it won't happen, we should probably at least log if it does. Worth considering panicking as well.

internal/jujuclient/cloud.go
```diff
@@ -5,156 +5,86 @@ package jujuclient
...
-		return nil, errors.E(jujuerrors.Cause(err))
-	}
-	if out.Results[0].Error != nil {
-		return out.Results[0].Models, errors.E(out.Results[0].Error)
-	}
-	return out.Results[0].Models, nil
+// UpdateCloudsCredentialForce updates the given credential on the controller.
+// It calls UpdateCloudsCredentials, and adapts it to:
+// 1. Take a single credential
+// 2. Always force the update (force=true).
```
luci1900, line 33: Why?

internal/jimm/juju/cloudcredential.go
ale8k: Yeah happy to add, the juju client code doesn't care about this but I wanted to just add a little precaution.

internal/jujuclient/cloud.go
ale8k: So AFAICT, our previous code does the check + update separately to enable batch checking across controllers, as we update the credential for all controllers currently holding it.

internal/jimm/juju/cloudcredential.go
luci1900: The other option is to remove the check entirely and let the index below panic. Up to you.

luci1900 (approved):

SimoneDutto (approved):
i'm not fully clued up on this piece of work, but looks good to me

ale8k:
> i'm not fully clued up on this piece of work, but looks good to me

Before merging, I'll explain tomorrow in standup so it is clearer

PR #1892, opened 2026-03-03
fix(jaas-usage): use sed to replace `jaas` with `juju jaas` in the index.md doc

SimoneDutto:
## Description

In the doc right now we document jaas commands as:
```
Usage
jaas show-model [options] <model>
```

This is not correct because you can't run jaas commands by itself, and you need to run them via juju.

I've tried to use:
```
jaasCmd := jujucmd.NewSuperCommand(jujucmd.SuperCommandParams{
		Name: "jaas",
		Doc:  jaasDoc,
               	UsagePrefix: "juju"
	})
```

the problem is that `UsagePrefix` is not used to generate documentation, so the easier thing was just to amend our generate script to fix the doc.


luci1900 (approved):
I suppose.

kian99 (approved):

PR #1895, opened 2026-03-06
fix: address panic in response handling and better handling for ModelStatus api method

kian99:
## Description
This PR fixes two issues that are causing test failures between the Terraform provider and the latest version of JIMM.
1. JIMM would panic in the `toFullModelInfo` because we were not checking if the `Error` field was nil before using it.
2. The client method `ModelStatus` returns the status of a single model but the underlying API request is a bulk one, so any errors are not surfaced in the `err` value and instead captured as a field on the list of responses.

To test the fix for the latter I've added an integration style test in the `testing` package. Ideally we would have the dial_test.go file inside of the `jujuclient` package and allow integration style tests from any package but our CI is currently setup to only run integration tests from the `testing` package. So this will need some further thought and work.

## Engineering checklist
- [ ] Documentation updated
- [x] Covered by unit tests

### DIFF
--- a/internal/jujuapi/types.go
+++ b/internal/jujuapi/types.go
@@ -325,7 +325,7 @@ func toFullModelInfo(modelInfo jujuclient.ModelInfo) jujuparams.ModelInfo {
 
 	var secretBackendResults []jujuparams.SecretBackendResult
 	for _, sb := range modelInfo.SecretBackends {
-		secretBackendResults = append(secretBackendResults, jujuparams.SecretBackendResult{
+		res := jujuparams.SecretBackendResult{
 			Result: jujuparams.SecretBackend{
 				Name:                sb.Result.Name,
 				BackendType:         sb.Result.BackendType,
@@ -336,12 +336,15 @@ func toFullModelInfo(modelInfo jujuclient.ModelInfo) jujuparams.ModelInfo {
 			NumSecrets: sb.NumSecrets,
 			Status:     sb.Status,
 			Message:    sb.Message,
-			Error: &jujuparams.Error{
+		}
+		if sb.Error != nil {
+			res.Error = &jujuparams.Error{
 				Message: sb.Error.Error(),
 				Code:    string(errors.ErrorCode(sb.Error)),
 				Info:    errors.ErrorInfo(sb.Error),
-			},
-		})
+			}
+		}
+		secretBackendResults = append(secretBackendResults, res)
 	}
 	modelInfoParams.SecretBackends = secretBackendResults
 
--- a/internal/jujuclient/modelmanager.go
+++ b/internal/jujuclient/modelmanager.go
@@ -142,6 +142,9 @@ func (c Connection) ModelStatus(ctx context.Context, modelTag names.ModelTag) (b
 	if len(statuses) == 0 {
 		return base.ModelStatus{}, errors.E("no status returned for model")
 	}
+	if statuses[0].Error != nil {
+		return base.ModelStatus{}, statuses[0].Error
+	}
 	return statuses[0], nil
 }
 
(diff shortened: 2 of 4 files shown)


internal/jujuapi/types.go
```diff
@@ -336,12 +336,15 @@ func toFullModelInfo(modelInfo jujuclient.ModelInfo) jujuparams.ModelInfo {
 			NumSecrets: sb.NumSecrets,
 			Status:     sb.Status,
 			Message:    sb.Message,
-			Error: &jujuparams.Error{
+		}
+		if sb.Error != nil {
```
SimoneDutto, line 340: shouldn't we check the error before populating all of these fields?

internal/jujuclient/modelmanager.go
```diff
@@ -142,6 +142,9 @@ func (c Connection) ModelStatus(ctx context.Context, modelTag names.ModelTag) (b
 	if len(statuses) == 0 {
 		return base.ModelStatus{}, errors.E("no status returned for model")
 	}
+	if statuses[0].Error != nil {
+		return base.ModelStatus{}, statuses[0].Error
```
SimoneDutto, line 146: would this be `error.E`ed later on? Otherwise we could loose the "code"

testing/dial_test.go
```diff
@@ -82,6 +84,29 @@ func TestDialWithJWT(t *testing.T) {
...
+
+	api, err := s.JIMM.Dialer.Dial(context.Background(), &ctl, names.ModelTag{}, nil, nil)
+	c.Assert(err, qt.Equals, nil)
+	defer api.Close()
+
+	_, err = api.ModelStatus(context.Background(), names.NewModelTag(uuid.NewString()))
+	c.Assert(err, qt.Not(qt.IsNil))
+	c.Check(errors.ErrorCode(err), qt.Equals, errors.CodeNotFound)
+	c.Check(err, qt.ErrorMatches, `model .* not found.*`)
+}
```
SimoneDutto, line 108: i see the code is here, so disregard my previous comment

internal/jujuclient/modelmanager.go
kian99: ErrorCode traverses the error tree for the code, that's why the test still works.

internal/jujuapi/types.go
kian99: I don't follow? We're checking the error is not nil

SimoneDutto: `if sb.Error` should be checked before creating res. Right?

kian99: But the error fields are on res? Unless you mean to avoid filling in the other fields, better to fill in everything that we can.

SimoneDutto: yeah, i'm suggesting to avoid filling the other things. You'd never know if `sb.Result` is nil when there is an error.
And it would cause another panic.

kian99: sb.Result is not a pointer so as it is there is there shouldn't be any more room for panics.

SimoneDutto: that's fine then

SimoneDutto (approved):

internal/jujuapi/types.go
```diff
@@ -336,12 +336,15 @@ func toFullModelInfo(modelInfo jujuclient.ModelInfo) jujuparams.ModelInfo {
...
+			res.Error = &jujuparams.Error{
```
ale8k, line 341: We always get caught by this (Incoming AI prompt: "create me a linter that verifies all nested errors in Juju responses are properly handled") 😆

ale8k (approved):

PR #1903, opened 2026-03-18
feat: add new error methods

kian99:
## Description
This PR proposes new methods to the internal `errors` package that would help us move away from the `errors.E` variadic method.

The goal is to provide type safe alternatives to the variadic `errors.E` that exposes where we wrap errors, add error codes/map-info, etc. The API shape is likely still far from ideal but is intended to keep the existing behaviour while making it easier to refactor going forward.

There are 3 new functions that return our existing `Error` type. These are `New()`, `Newf()`, `Wrap()`. 
Additional methods on the `Error` type have also been introduced: `Wrap()`, `WithCode()`, `WithInfo()`, `WithMessage()`, and `WithMessagef()`. See their usages below. Note that `Error` reflects our struct type and `error` reflects the error interface.

New functions:
- `New(msg string) Error` - Creates Error from message.
- `Newf(format string, args ...any) Error` - Creates Error using formatted message.
- `Wrap(err error) Error` - Create Error from existing error.

New Methods on `Error` type:
- `Wrap(err error) Error` - Sets the given error as the underlying error.
- `WithCode(code Code) Error` - Sets the given code on the error.
- `WithInfo(info map[string]any) Error` - Sets the given info on the error.

### Migration Patterns
Our errors package is only needed if you need to annotate an error with a code or info map. Otherwise, we can use `fmt.Errorf()`.

```go
// Before → After
errors.E(err)                              → err
errors.E("msg")                            → fmt.Errorf("msg") or stderrors.New("msg")
errors.E(fmt.Errorf("msg: %w", err))       → fmt.Errorf("msg: %w", err)
errors.E(errors.CodeNotFound)              → errors.New("").WithCode(errors.CodeNotFound)
errors.E(errors.CodeBadRequest, "msg")     → errors.New("msg").WithCode(errors.CodeBadRequest)
errors.E(err, "msg")                       → errors.New("msg").Wrap(err) // Keep the underlying error code/map but use a new message.
errors.E(err, fmt.Sprintf("msg %s", val))  → errors.Newf("msg %s", val).Wrap(err) // Keep the underlying error code/map but use a new message.
errors.E(err, errors.CodeUnauthorized)     → errors.Wrap(err).WithCode(errors.CodeUnauthorized) // Add a code/map onto an existing error.
```

One pattern that may emerge is below,
```go
err := fmt.Errorf("something happened: %w, err")
return errors.Wrap(err).WithCode(errors.CodeUnauthorized)
```

This lets you annotate an error with a new message and keep the original in the chain, then wraps the new error to allow for setting an error code.

If you don't need to keep the error in the chain then a simplified version is below.
```go
return errors.Newf("something happened: %s", err).WithCode(errors.CodeUnauthorized)
```

The methods `errors.New("msg").Wrap(err)` and `errors.Newf("msg %s", val).Wrap(err)` are two patterns we likely want to minimise and/or evaluate removing. These mean we are trying to keep the underlying error code while masking the underlying error message with a new one. Hopefully we can come up with some pattern in the future to avoid this.

This PR is a draft while I gather feedback on the new methods.

Fixes [JUJU-8836](https://warthogs.atlassian.net/browse/JUJU-8836)

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

[JUJU-8836]: https://warthogs.atlassian.net/browse/JUJU-8836?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


luci1900 (approved):
I like the API.

I think `New` and `Newf` with `Wrap` are easy to use incorrectly, especially when compared to `fmt.Errorf`. Under the assumption that we'd attempt to gradually remove their usage, I think that's an acceptable tradeoff.

internal/errors/errors.go
```diff
@@ -31,7 +31,7 @@ type Error struct {
 }
 
 // Error implements the error interface.
-func (e *Error) Error() string {
+func (e Error) Error() string {
```
SimoneDutto, line 34: removing the pointer here is potentially impacting `errors.Is` or `errors.As`. We need to be careful because I've fiddled a bit around errors in the terraform provider and we tend to do funky business

SimoneDutto (commented):
one question:
- why don't we use `juju/errors` with const errors? it's very clean to handle error codes, and we might even just get rid of our wrapper all together?
- tou can mark a func to be deprecated, if we want to discourage people from using it: // Deprecated: The rule Title uses for word boundaries does not handle Unicode              
// punctuation properly. Use golang.org/x/text/cases instead.

internal/errors/errors.go
luci1900: It's an interface, pointers to it should be redundant. If we do rely on the difference, it's really subtle and we should stop that.

SimoneDutto:
> one question:
> 
> - why don't we use `juju/errors` with const errors? it's very clean to handle error codes, and we might even just get rid of our wrapper all together?
> 
> - tou can mark a func to be deprecated, if we want to discourage people from using it: // Deprecated: The rule Title uses for word boundaries does not handle Unicode              
> 
> // punctuation properly. Use golang.org/x/text/cases instead. 

One question -> two questions

kian99:
> one question:
> 
> * why don't we use `juju/errors` with const errors? it's very clean to handle error codes, and we might even just get rid of our wrapper all together?
> * tou can mark a func to be deprecated, if we want to discourage people from using it: // Deprecated: The rule Title uses for word boundaries does not handle Unicode
>   // punctuation properly. Use golang.org/x/text/cases instead.

- The Juju errors package has been moved from `github.com/juju/errors` to the internal folder of the `juju/juju` repo so that's not really an option. We could c/p any bits from their package that we need though.
- We can mark errors.E as deprecated for sure. One thing to note is that since we're not developing a library it's not as critical and we'll get rid of it entirely in probably a few weeks, once we settle on a new errors api.

internal/errors/errors.go
```diff
@@ -110,7 +110,46 @@ func E(args ...interface{}) error {
...
+// Newf creates a new Error with a formatted message.
+func Newf(format string, args ...any) Error {
+	return Error{Message: fmt.Sprintf(format, args...)}
+}
+
+// Wrap creates a new Error with err as the underlying error.
+//
+// Useful for wrapping an error with a custom message, code, or info,
+// while preserving the original error in the chain.
+func Wrap(err error) Error {
```
luci1900, line 130: Maybe this could be called `From`.

PR #1904, opened 2026-03-19
feat: add DB controller profile CRUD

kian99:
## Description
Implements the DB methods for the controller profiles work. The new DB table holds a flat representation of all the fields that a controller profile contains. The create/read/update/remove methods are all straightforward, but there is a `Version` field that provides optimistic update handling to guard against concurrent updates.

Resolves [JUJU-9424](https://warthogs.atlassian.net/browse/JUJU-9424)

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

[JUJU-9424]: https://warthogs.atlassian.net/browse/JUJU-9424?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/dbmodel/sql/postgres/036_add_controller_profiles.up.sql
```diff
@@ -0,0 +1,31 @@
+-- Add controller profiles for reusable, non-secret bootstrap settings.
+
+CREATE TABLE IF NOT EXISTS controller_profiles (
+    id SERIAL PRIMARY KEY,
+    created_at TIMESTAMP WITH TIME ZONE DEFAULT NOW(),
+    updated_at TIMESTAMP WITH TIME ZONE,
+    name TEXT NOT NULL UNIQUE,
```
ale8k: Should we varchar some of these?

internal/db/controllerprofile.go
```diff
@@ -0,0 +1,158 @@
...
+// an existing one with the same name.
+//
+// New profiles must be created with Version set to 0. Replacing an existing
+// profile requires Version to match the last value read by the caller.
+func (d *Database) CreateOrReplaceControllerProfile(ctx context.Context, profile *dbmodel.ControllerProfile) (err error) {
+	const op = "db.CreateOrReplaceControllerProfile"
+	if err := d.ready(); err != nil {
+		return errors.E(err)
+	}
+	if profile.Name == "" {
```
ale8k: Why not do this in validate?

ale8k (approved):
LGTM. Not 100% sure how I feel on the CreateOrReplace of a profile though.

internal/db/controllerprofile.go
kian99: Good point, changed.

internal/dbmodel/sql/postgres/036_add_controller_profiles.up.sql
kian99: Yeah I'll do 255 on some of them.

luci1900: Why not `TEXT` for all of them?

kian99: It seems reasonable to put a cap on all of them except description.
We don't do validation on inputs at the API layer which we would ideally.

internal/db/controllerprofile.go
```diff
@@ -0,0 +1,158 @@
...
+		return err
+	}
+
+	durationObserver := servermon.DurationObserver(servermon.DBQueryDurationHistogram, op)
+	defer durationObserver()
+	defer servermon.ErrorCounter(servermon.DBQueryErrorCount, &err, op)
+
+	err = d.Transaction(func(d *Database) error {
+		current := dbmodel.ControllerProfile{Name: profile.Name}
+		err := d.ForUpdate().GetControllerProfile(ctx, &current)
```
luci1900, line 51: This looks like it should work fine.

The other way to do it is to update with a `WHERE` that includes the current version and `RETURNING`. You'd then have to check whether an update actually happened.

I don't think we need to optimise for minimal round-trips, though.

luci1900 (approved):

kian99:
Merging into the feature branch, fixing the failing integration test in https://github.com/canonical/jimm/pull/1909 which will be merged into the feature branch once it lands.

PR #1905, opened 2026-03-19
Bump migrationtarget modelconfig appoffer facades for 4

ale8k:
## Description
Whilst investigating all facade methods in JIMM and checking what is missing in Juju 4. I realised the parameters on these specific methods have not changed. 

We still require ModelManager and Controller to be bumped, but I thougtht we may as well put the work in for these one now as they seem sensible to do.

I've simply just made a "v4" variant of each method, just in case there's ever an implementation change. For now it appears there isn't. 

Facades bumped and why:
- ApplicationOffers (6): [Bumped here](https://github.com/juju/juju/pull/19955) to introduce model qualifiers. Wire wise, we appear OK.
- MigrationTarget (7): [Bumped here](https://github.com/juju/juju/pull/20240/changes#diff-315db6f39093f9f3b2dbe2b788bff266a7ee8f674faabeb889c7a9bee3ae5631), Migration Target 6 was done by @kian99 to handle redirects. However I'm not actually sure why it was bumped to 7 in this case as the reference PRs in the merge forward from 3.6 don't include the migration target to 7. The doc on this PR does state it is to handle model qualifiers though (cosmetic apparently?)
- ModelConfig (4): [Bumped here](https://github.com/juju/juju/pull/18754) Appears to be just when plugging in service & conn to DQLite (possibly migration target (7) is the same ordeal?)

### DIFF
--- a/internal/jujuapi/applicationoffers.go
+++ b/internal/jujuapi/applicationoffers.go
@@ -28,6 +28,7 @@ func init() {
 		findOffersMethod := rpc.Method(r.FindApplicationOffers)
 		applicationOffersMethod := rpc.Method(r.ApplicationOffers)
 
+		// Juju 3
 		r.AddMethod("ApplicationOffers", 5, "Offer", offerMethod)
 		r.AddMethod("ApplicationOffers", 5, "GetConsumeDetails", getConsumeDetailsMethod)
 		r.AddMethod("ApplicationOffers", 5, "ListApplicationOffers", listOffersMethod)
@@ -36,8 +37,22 @@ func init() {
 		r.AddMethod("ApplicationOffers", 5, "FindApplicationOffers", findOffersMethod)
 		r.AddMethod("ApplicationOffers", 5, "ApplicationOffers", applicationOffersMethod)
 
-		return []int{5}
+		// Juju 4
+		r.AddMethod("ApplicationOffers", 6, "Offer", offerMethod)
+		r.AddMethod("ApplicationOffers", 6, "GetConsumeDetails", getConsumeDetailsMethod)
+		r.AddMethod("ApplicationOffers", 6, "ListApplicationOffers", listOffersMethod)
+		r.AddMethod("ApplicationOffers", 6, "ModifyOfferAccess", modifyOfferAccessMethod)
+		r.AddMethod("ApplicationOffers", 6, "DestroyOffers", destroyOffersMethod)
+		r.AddMethod("ApplicationOffers", 6, "FindApplicationOffers", findOffersMethod)
+		r.AddMethod("ApplicationOffers", 6, "ApplicationOffers", applicationOffersMethod)
+
+		return []int{5, 6}
 	}
+}
+
+// Offer creates a new ApplicationOffer.
+func (r *controllerRoot) OfferV4(ctx context.Context, args jujuparams.AddApplicationOffers) (jujuparams.ErrorResults, error) {
+	return r.Offer(ctx, args)
 }
 
 // Offer creates a new ApplicationOffer.
@@ -73,6 +88,12 @@ func (r *controllerRoot) offer(ctx context.Context, args jujuparams.AddApplicati
 		return errors.E(err)
 	}
 	return nil
+}
+
+// GetConsumeDetailsV4 implements the GetConsumeDetails procedure of the
+// ApplicationOffers facade.
+func (r *controllerRoot) GetConsumeDetailsV4(ctx context.Context, args jujuparams.ConsumeOfferDetailsArg) (jujuparams.ConsumeOfferDetailsResults, error) {
+	return r.GetConsumeDetails(ctx, args)
 }
 
 // GetConsumeDetails implements the GetConsumeDetails procedure of the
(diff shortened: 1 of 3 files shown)


internal/jujuapi/migrationtarget.go
```diff
@@ -50,10 +51,37 @@ func init() {
...
+		r.AddMethod("MigrationTarget", 7, "Prechecks", preChecks)
+		r.AddMethod("MigrationTarget", 7, "CACert", caCert)
+		r.AddMethod("MigrationTarget", 7, "Activate", activate)
+		r.AddMethod("MigrationTarget", 7, "AdoptResources", adoptResources)
+		r.AddMethod("MigrationTarget", 7, "Abort", abort)
+		r.AddMethod("MigrationTarget", 7, "CheckMachines", checkMachines)
+		r.AddMethod("MigrationTarget", 7, "Import", importMethod)
+		r.AddMethod("MigrationTarget", 7, "LatestLogTime", latestLogTime)
+
+		return []int{6, 7}
```
ale8k, line 64: 67777777

kian99 (commented):
Could you update the PR description or commit description to point to the Juju PRs where the facades have been bumped? Similar to https://github.com/canonical/jimm/pull/1837/commits, it will help us reviewers understand what the reasons for the facade bump are.

internal/jujuapi/applicationoffers.go
```diff
@@ -28,6 +28,7 @@ func init() {
 		findOffersMethod := rpc.Method(r.FindApplicationOffers)
 		applicationOffersMethod := rpc.Method(r.ApplicationOffers)
 
+		// Juju 3
```
SimoneDutto, line 31: do we need to keep implementing the 3 version, in the spec we've specified that once we release a version of jimm supporting juju 4, we will always require a 4 client to talk with that version of JIMM

kian99: This is a good point. 

This is targeting jimm v3. We should either avoid supporting these facades until we make a jimm v4 branch or support both (for consumers like the dashboard or existing Juju 3 clients) if desired/necessary.

ale8k:
> Could you update the PR description or commit description to point to the Juju PRs where the facades have been bumped? Similar to https://github.com/canonical/jimm/pull/1837/commits, it will help us reviewers understand what the reasons for the facade bump are.

Done, I'm actually unsure on the MigrationTarget, but it appears to be the same as ModelConfig (Ian switched some bits to hit the service), but there's a lot in that merge forward to scour through... It also mentions model qualifiers on the mig target. 
```go
		// v7 handles requests with a model qualifier instead of a model owner.
		registry.MustRegisterForMultiModel("MigrationTarget", 7, func(stdCtx context.Context, ctx facade.MultiModelContext) (facade.Facade, error) {
			api, err := makeFacade(stdCtx, ctx, requiredMigrationFacadeVersions)
			if err != nil {
				return nil, errors.Errorf("making migration target version 7: %w", err)
			}
			return api, nil
```
Maybe I'm missing something?

kian99:
> * MigrationTarget (7): [Bumped here](https://github.com/juju/juju/pull/20240/changes#diff-315db6f39093f9f3b2dbe2b788bff266a7ee8f674faabeb889c7a9bee3ae5631), Migration Target 6 was done by @kian99 to handle redirects. However I'm not actually sure why it was bumped to 7 in this case as the reference PRs in the merge forward from 3.6 don't include the migration target to 7. The doc on this PR does state it is to handle model qualifiers though (cosmetic apparently?)

If you look at the merge forward, it seems that v7 was added because Juju 4 was not released yet but had already introduced v6 of the migration target facade. Then Juju 3 bumped it (for the changes that I made) and so during the merge forward, Juju 4's existing v6 was bumped up to v7. Juju 4's v7 is just for handling the model qualifier instead of model owner. C/p from that PR.
> // v7 handles requests with a model qualifier instead of a model owner. 

> * ModelConfig (4): [Bumped here](https://github.com/juju/juju/pull/18754) Appears to be just when plugging in service & conn to DQLite (possibly migration target (7) is the same ordeal?)

I don't see the api/facadeversions.go file changed in that PR.

ale8k:
> > * MigrationTarget (7): [Bumped here](https://github.com/juju/juju/pull/20240/changes#diff-315db6f39093f9f3b2dbe2b788bff266a7ee8f674faabeb889c7a9bee3ae5631), Migration Target 6 was done by @kian99 to handle redirects. However I'm not actually sure why it was bumped to 7 in this case as the reference PRs in the merge forward from 3.6 don't include the migration target to 7. The doc on this PR does state it is to handle model qualifiers though (cosmetic apparently?)
> 
> If you look at the merge forward, it seems that v7 was added because Juju 4 was not released yet but had already introduced v6 of the migration target facade. Then Juju 3 bumped it (for the changes that I made) and so during the merge forward, Juju 4's existing v6 was bumped up to v7. Juju 4's v7 is just for handling the model qualifier instead of model owner. C/p from that PR.
> 
> > // v7 handles requests with a model qualifier instead of a model owner.
> 
> > * ModelConfig (4): [Bumped here](https://github.com/juju/juju/pull/18754) Appears to be just when plugging in service & conn to DQLite (possibly migration target (7) is the same ordeal?)
> 
> I don't see the api/facadeversions.go file changed in that PR.

yeah I correct myself after (maybe refresh) sorry.

It is updated here in that PR (model config): https://github.com/juju/juju/pull/18754/changes#diff-86efc276f23f14177002e04aee8f2ac9d907d4672ee1ad0a9b73acd92db4ab56R25

kian99:
> > I don't see the api/facadeversions.go file changed in that PR.
> 
> yeah I correct myself after (maybe refresh) sorry.
> 
> It is updated here in that PR (model config): https://github.com/juju/juju/pull/18754/changes#diff-86efc276f23f14177002e04aee8f2ac9d907d4672ee1ad0a9b73acd92db4ab56R25

Strange, I'm still not seeing it in that link. I took a look and found it in this PR https://github.com/juju/juju/pull/17702 where I see the change was to pull secret-backend config out of model config, replacing `juju model-config secret-backend` and `juju model-config secret-backend=myvault` with dedicated methods. I believe Ian mentioned this yesterday in our x-team meet.

PR #1908, opened 2026-03-19
feat(supported-versions): add SupportedVersions facade method

SimoneDutto:
## Description

`SupportedVersions` returns the list of Juju versions that JAAS/JIMM considers supported.
It calls the GitHub API to fetch the releases, it does filtering, and returns the list.

## API shape
I like the idea of the optional contextualVersion (stealing the concept from `ContextualTuples`  in Openfga, so if they send it to us we can potentially build some logic around suggested upgrade, etc.

### Previous implementation
Instead of querying GitHub at runtime from the facade, we now generate a JSON file at build time (via `make generate`) and embed it into the `jimm` binary (and also keep the generated JSON in-repo for visibility / traceability).

### The why

What i like:
- No dependency on the GitHub API at runtime (no rate limiting, no outages).
- Works in air‑gapped environments.
- Every JIMM release carries an explicit snapshot of “known/supported Juju versions”.
  - Example: if someone runs **JIMM 3.3.19** with **Juju 3.6.25**, we can check the generated `release.json` shipped with 3.3.19 and confirm that version was never tested against that version of JIMM.

What i don't like too much:
- When Juju publishes a new release, the CI in the very next pr may start failing because the generated JSON is out of date.
- Fix is intentional and simple: run `make generate` and commit the updated JSON.

This is acceptable and even useful because it makes Juju version bumps explicit, and it aligns with how we run e2e:
- Vitaly said that when a Juju release is created on GitHub, the same Jenkins job also publishes the corresponding `stable` snap track update.
- Therefore, whenever there is a new Juju release, it also becomes available in the snap channel used by our e2e suite, so updating the generated list keeps our “supported versions” aligned with what e2e will exercise.

Let me know if you like it or not.


luci1900 (dismissed):
Good idea.

We could also add a jaas command that calls this endpoint.

kian99 (commented):
The only downside with this approach is that JIMM needs to be updated, released and deployed before a user will ever see new versions of Juju from the supported versions endpoint.
That decision seems like something we should resolve beforehand.

SimoneDutto:
> The only downside with this approach is that JIMM needs to be updated, released and deployed before a user will ever see new versions of Juju from the supported versions endpoint. That decision seems like something we should resolve beforehand.

I've discussed it briefly in the standup after you were gone, i'll bring it up today again

internal/jujuapi/jimm.go
```diff
@@ -840,3 +843,7 @@ func (r *controllerRoot) ListJobs(ctx context.Context, req apiparams.ListJobsReq
 
 	return r.jimm.JobManager().ListJobs(ctx, req)
 }
+
+func (r *controllerRoot) SupportedJujuVersions(ctx context.Context, req apiparams.SupportedJujuVersionsRequest) (apiparams.SupportedJujuVersionsResponse, error) {
```
kian99, line 848: Godoc

internal/jimm/juju/supportedversions.go
```diff
@@ -0,0 +1,129 @@
...
+	if err != nil {
+		return params.SupportedJujuVersionsResponse{}, errors.E(err)
+	}
+	return params.SupportedJujuVersionsResponse{Versions: releases}, nil
+}
+
+// fetchReleasesFromGitHub queries the GitHub API for juju/juju releases and returns
+// a filtered, sorted slice of stable releases >= minSupportedVersion.
+// If contextualVersion is non-nil, only releases strictly greater than it are included.
+func fetchReleasesFromGitHub(ctx context.Context, contextualVersion *version.Number) ([]params.VersionElem, error) {
```
kian99: Can we rename `contextualVersion` to `minVersion` and can we do that specific filtering outside of this function.

internal/jimm/juju/supportedversions_test.go
kian99, line 1: These are pure integration tests with Github right? 
One integration test seems fine, but I'd also like to see a test where we return a mock response and have a clear assertion that the results are being sorted, etc.

Something like,
```go
mockResp := []string{v3.3.1, v3.1.3, v3.5.6}
client := mockClient{mockResp}
resp := fetchReleases(client)
c.Assert(resp, qt.DeepEquals, []string{
   "v3.5.6",
   "v3.3.1",
   "v3.1.3",
})
```

pkg/api/params/params.go
```diff
@@ -751,3 +751,25 @@ type ListJobsResponse struct {
 	// If empty, there are no more results.
 	NextCursor string `json:"next_cursor,omitempty" yaml:"next_cursor,omitempty"`
 }
+
+// SupportedJujuVersionsResponse holds the response for a SupportedJujuVersions call.
+type SupportedJujuVersionsRequest struct {
+	// ContextualVersion is an optional version string that can be used to provide context
+	// for the supported versions response.
+	// For example, this could be used to indicate the version of the client making the request,
+	// allowing the server to tailor the supported versions response accordingly.
```
kian99: Isn't this too vague? We currently use it to set a minimum version, but the way this docstring is phrased seems like its too open-ended.

internal/jimm/juju/supportedversions.go
```diff
@@ -0,0 +1,129 @@
...
+	ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
+	defer cancel()
+
+	client := github.NewClient(nil)
+	minVersion := version.MustParse(minSupportedVersion)
```
alesstimec: we shouldn't be using `MustParse` in production code..   can't we just do:
```
minVersion := version.Number{
  Major: x,
  Minor: y, 
  Patch: z.
}
```

alesstimec (commented):
looks good, once all comments are addressed

internal/jimm/juju/supportedversions.go
SimoneDutto: ok, for the rename. 

For the filtering i'd left it here to avoid looping two times around the slice

pkg/api/params/params.go
SimoneDutto: i'm in favor of leaving the `ContextualVersion` in because we could change the filtering mechanism not to be `minVersion`. That's why it's vague, to avoid doing breaking changes to the facade if we change the filtering mechanism.

internal/jimm/juju/supportedversions.go
SimoneDutto: in this case it's fine because minSupportedVersion is a const string. If this panics, it will panic in tests as well

internal/jujuapi/jimm.go
SimoneDutto: added

internal/jimm/juju/supportedversions_test.go
SimoneDutto: added some unit and integration tests in separate files

internal/jimm/juju/supportedversions.go
kian99: Sure

pkg/api/params/params.go
kian99: It would still be a breaking change in behavior, even if the field doesn't change.
I strongly feel we should make the API clear and easy to use and if we identify that something must change, we should do a facade bump or make a new method/add a param for the new functionality.

alesstimec: my vote goes to `minVersion`

SimoneDutto: implemented with minVersion

internal/jimm/juju/supportedversions_test.go
```diff
@@ -0,0 +1,124 @@
...
+	for _, v := range resp.Versions {
+		c.Assert(v.Date, qt.Not(qt.Equals), "")
+		c.Assert(v.LinkToRelease, qt.Not(qt.Equals), "")
+	}
+}
+
+func TestSupportedVersions_MinVersion(t *testing.T) {
+	c := qt.New(t)
+
+	t.Run("filters to versions strictly greater than minVersion", func(t *testing.T) {
```
kian99: Minor
```suggestion
	c.Run("filters to versions strictly greater than minVersion", func(c * qt.C) {
```
That way you get a slightly different error trace if the test fails I believe.

pkg/api/params/params.go
```diff
@@ -751,3 +751,24 @@ type ListJobsResponse struct {
 	// If empty, there are no more results.
 	NextCursor string `json:"next_cursor,omitempty" yaml:"next_cursor,omitempty"`
 }
+
+// SupportedJujuVersionsResponse holds the response for a SupportedJujuVersions call.
+type SupportedJujuVersionsRequest struct {
+	// MinVersion is an optional lower-bound version filter.
+	// When set, only versions strictly greater than MinVersion are included in the response.
+	// Versions equal to or below MinVersion are excluded.
+	MinVersion *string `json:"min-version,omitempty"`
```
kian99, line 760: Ty

pkg/api/params/params.go
```diff
@@ -751,3 +751,24 @@ type ListJobsResponse struct {
...
+}
+
+// VersionElem represents a single supported Juju version.
+type VersionElem struct {
+	// Version is the "x.x.x" version string.
+	Version       string `json:"version"`
+	Date          string `json:"date"`
```
kian99: We could docstring `Date` as well. From above, it's the published date.

kian99 (approved):

internal/jimm/juju/supportedversions.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+const minSupportedVersion = "3.6.5"
+
+// blacklistedVersions lists specific releases that should be excluded even if they
+// otherwise meet the stable release criteria.
+var blacklistedVersions = []version.Number{
+	version.MustParse("4.0.0"),
+	version.MustParse("4.0.1"),
+	version.MustParse("4.0.2"),
```
alesstimec, line 32: missing 4.0.3?

internal/jimm/juju/supportedversions.go
```diff
@@ -0,0 +1,139 @@
...
+	if minVersion != nil {
+		v, err := version.Parse(*minVersion)
+		if err != nil {
+			return params.SupportedJujuVersionsResponse{}, fmt.Errorf("invalid min version %q: %w", *minVersion, err)
+		}
+		parsedMinVersion = &v
+	}
+
+	client := j.GitHubClient
+	if client == nil {
```
alesstimec, line 49: i'm not a fan of this approach. i'd rather see the client passed in every time and fail if it is not, but ymmv

alesstimec (approved):

internal/jimm/juju/supportedversions.go
SimoneDutto: tbh, this was just to showcase the blacklist. I guess the full list will go on until 4.1?

PR #1911, opened 2026-03-20
refactor: replace redundant dial permissions with superuser

luci1900:
## Description

Remove the ability to explicitly pass permissions when dialling a controller, since in the vast majority of cases it wasn't used. The couple cases that did use it didn't need it.

Always set `superuser` permissions when dialling, whether as an `admin` user or as an openfga user.

Since the user wasn't passed in the vast majority of cases anyway, the net effect of these changes on the dialling login request is zero.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests


ale8k (approved):
Nice! I was surprised at how little the changes there were. Had a look around too to make sure this is our only use of perm dial and all looks good to me. Ty for this, much cleaner!

kian99 (approved):

SimoneDutto (approved):

PR #1912, opened 2026-03-20
refactor: pass in user to dial

luci1900:
## Description
Explicitly pass in the user tag when dialling the underlying Juju controller, in all cases where we're directly acting on behalf of the user.

Some cases remain where we dial as `admin`, like the watcher.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests


SimoneDutto (approved):
i'm shocked that it just works, but good news!

SimoneDutto:
@nvinuesa will be happy about this patch

kian99 (approved):

PR #1915, opened 2026-03-24
chore: replace errors.E(err) with err

kian99:
## Description
Ran a find and replace of `errors.E(err)` -> `err`. A minor refactoring that gets rid of the unnecessary wrapping except for one usage in `internal/errors/errors_test.go`

I had to fix one file at `internal/jimm/juju/cloudcredential.go`

PR #1916, opened 2026-03-24
chore: remove more redundant uses of errors.E

kian99:
## Description
This PR follows on from #1915 to remove more unnecessary uses of `errors.E()`. It also introduces a simple `errors.New()` which calls the stdlib `errors.New()` so that we can remove more instances of `errors.E`. These changes include,
- `errors.E(fmt.Errorf(...))` -> `fmt.Errorf()`
- `errors.E(dbError(err))` -> `dbError(err)`
- `errors.E("msg")` -> `errors.New("msg")`
- `errors.E(dbError(result.Error))` -> `dbError(result.Error)`

Any uses of errors.E that inject error codes, etc are left in-tact.

### DIFF
--- a/internal/db/applicationoffer.go
+++ b/internal/db/applicationoffer.go
@@ -51,7 +51,7 @@ func (d *Database) GetApplicationOffer(ctx context.Context, offer *dbmodel.Appli
 	case offer.URL != "":
 		db = db.Where("url = ?", offer.URL)
 	default:
-		return errors.E("missing offer UUID or URL")
+		return errors.New("missing offer UUID or URL")
 	}
 
 	db = db.Preload("Model").Preload("Model.Controller")
--- a/internal/db/applicationoffer.go
+++ b/internal/db/applicationoffer.go
@@ -118,7 +118,7 @@ func (d *Database) FindApplicationOffersByModel(ctx context.Context, modelName, 
 		offer := offer
 		err := d.GetApplicationOffer(ctx, &offer)
 		if err != nil {
-			return nil, errors.E(dbError(err))
+			return nil, dbError(err)
 		}
 		offers[i] = offer
 	}
--- a/internal/db/applicationoffer.go
+++ b/internal/db/applicationoffer.go
@@ -26,7 +26,7 @@ func (d *Database) AddApplicationOffer(ctx context.Context, offer *dbmodel.Appli
 
 	result := db.Create(offer)
 	if result.Error != nil {
-		return errors.E(dbError(result.Error))
+		return dbError(result.Error)
 	}
 	return nil
 }
@@ -81,7 +81,7 @@ func (d *Database) DeleteApplicationOffer(ctx context.Context, offer *dbmodel.Ap
 
 	result := db.Delete(offer)
 	if result.Error != nil {
-		return errors.E(dbError(result.Error))
+		return dbError(result.Error)
 	}
 	return nil
 }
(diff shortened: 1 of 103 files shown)


internal/db/applicationoffer.go
```diff
@@ -26,7 +26,7 @@ func (d *Database) AddApplicationOffer(ctx context.Context, offer *dbmodel.Appli
 
 	result := db.Create(offer)
 	if result.Error != nil {
-		return errors.E(dbError(result.Error))
+		return dbError(result.Error)
```
luci1900, line 29: Is this one ok? I presume we don't need to check the code up the chain?

luci1900: Same for the other `dbError` ones.

kian99: The code will continue to be available since the `errors.ErrorCode` function traverses the chain of errors.

luci1900 (approved):

internal/errors/errors.go
```diff
@@ -59,6 +59,12 @@ func (e *Error) ErrorInfo() map[string]any {
 	return e.Info
 }
 
+// New constructs a new error with the given message.
+// It is a wrapper around the stdlib errors.New function.
+func New(text string) error {
+	return stderr.New(text)
```
alesstimec, line 65: is there really any value to this? can't we just call `stderr.New` or `fmt.Errorf`?

internal/jimm/bootstrap/bootstrap_test.go
```diff
@@ -500,7 +500,7 @@ func (s *bootstrapManagerSuite) TestBootstrapJob_BinaryStoreGetFails(c *qt.C) {
 		gomock.Any(),
 		&dbmodel.Controller{Name: p.ControllerName},
 	).Return(
-		errors.E(errors.CodeNotFound, errors.E("test err")),
+		errors.E(errors.CodeNotFound, errors.New("test err")),
```
alesstimec: isn't this equivalent to:
```suggestion
		errors.E(errors.CodeNotFound, "test err"),
```

alesstimec (commented):
i'm not really in fond of:
```
errors.E(<error code>, errors.New("<some message>"))
```
which i think is equivalent to 
```
errors.E(<error code>, "<some message")

otherwise i'm ok with this change..

File: internal/errors/errors.go (hunk 2 above)
Comment (kian99) [reply internal/errors/errors.go:65]: It avoids importing `errors` and `stderr "errors"` everywhere this is needed.

File: internal/jimm/bootstrap/bootstrap_test.go (hunk 3 above)
Comment (kian99) [reply internal/jimm/bootstrap/bootstrap_test.go]: Yip, my bad, just a leftover I didn't catch from the find and replace.

File: internal/jimm/bootstrap/bootstrap_test.go [hunk 4]
```diff
@@ -538,7 +538,7 @@ func (s *bootstrapManagerSuite) TestBootstrapJob_ExecutorFails(c *qt.C) {
...
```
alesstimec: hmmm.. i wonder if we could introduce a shorthand for this.. something like:
```
errors.NewNotFound("test err")
```

wdyt?

alesstimec (approved):
LGTM.. and later on we can think about additional methods to create different errors..

internal/jimm/bootstrap/bootstrap_test.go
kian99: Yeah there's a lot of these kinds of errors we could introduce. I won't do it here though.

PR #1917, opened 2026-03-24
feat: remove errors.E

kian99:
## Description
Another attempt at removing `errors.E`. The difference between this PR and https://github.com/canonical/jimm/pull/1903 is that this one offers a much simpler alternative for error handling. <s>
2 new error functions are introduced to replace `errors.E` namely,
- `ErrWithCode(err error, code Code) error` this function takes an error and a code and returns an error that has the specified code.
- `MsgWithCode(msg string, code Code) error` this function takes a string and a code and returns an error that has the specified code.
</s>

1 new error function has been introduced to replace `errors.E`,
- `Codef(code Code, format code Code, format string, args ...any) error`

This calls `fmt.Errorf` underneath and returns an error with the appropriate code. It turns out this was all we needed to cover all our error use cases. The only caveat is that if you want to carry an error's code through the stack but replace its message you need to either know the code and do something like,
```
errors.Codef(errors.CodeNotFound, "resource %q not found", resourceName)
```
Or one can extract the error code beforehand with `errors.ErrorCode(err)`. 

In many cases where we were checking the error code e.g. in the DB we often check the gorm error and if it was NotFound error we mask the underlying DB's "record not found" error and just return "<resource> was not found". That's cleanly done with the above pattern.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

### DIFF
--- a/internal/auth/auth.go
+++ b/internal/auth/auth.go
@@ -174,7 +174,7 @@ func NewAuthenticationService(ctx context.Context, params AuthenticationServiceP
 
 	provider, err := oidc.NewProvider(ctx, params.IssuerURL)
 	if err != nil {
-		return nil, errors.ErrWithCode(fmt.Errorf("failed to create oidc provider: %v", err), errors.CodeServerConfiguration)
+		return nil, errors.Codef(errors.CodeServerConfiguration, "failed to create oidc provider: %v", err)
 	}
 
 	authSvc := &AuthenticationService{
@@ -397,7 +397,7 @@ func (as *AuthenticationService) VerifySessionToken(token string) (_ jwt.Token, 
 	}()
 
 	if len(token) == 0 {
-		return nil, errors.MsgWithCode("no token presented", errors.CodeSessionTokenInvalid)
+		return nil, errors.Codef(errors.CodeSessionTokenInvalid, "no token presented")
 	}
 
 	decodedToken, err := base64.StdEncoding.DecodeString(token)
@@ -408,13 +408,13 @@ func (as *AuthenticationService) VerifySessionToken(token string) (_ jwt.Token, 
 	parsedToken, err := jwt.Parse(decodedToken, jwt.WithKey(as.signingAlg, []byte(as.jwtSessionKey)))
 	if err != nil {
 		if stderrors.Is(err, jwt.ErrTokenExpired()) {
-			return nil, errors.MsgWithCode("JIMM session token expired", errors.CodeSessionTokenInvalid)
+			return nil, errors.Codef(errors.CodeSessionTokenInvalid, "JIMM session token expired")
 		}
 		return nil, err
 	}
 
 	if _, err = mail.ParseAddress(parsedToken.Subject()); err != nil {
-		return nil, errors.MsgWithCode("failed to parse email", errors.CodeSessionTokenInvalid)
+		return nil, errors.Codef(errors.CodeSessionTokenInvalid, "failed to parse email")
 	}
 
 	return parsedToken, nil
@@ -472,7 +472,7 @@ func (as *AuthenticationService) VerifyClientCredentials(ctx context.Context, cl
 
 	_, err = cfg.Token(ctx)
 	if err != nil {
-		return errors.ErrWithCode(fmt.Errorf("invalid client credentials: %v", err), errors.CodeUnauthorized)
+		return errors.Codef(errors.CodeUnauthorized, "invalid client credentials: %v", err)
 	}
 	return nil
 }
@@ -534,7 +534,7 @@ func (as *AuthenticationService) AuthenticateBrowserSession(ctx context.Context,
 
 	identityId, ok := session.Values[SessionIdentityKey]
 	if !ok {
-		return ctx, errors.MsgWithCode("session is missing identity key", errors.CodeForbidden)
+		return ctx, errors.Codef(errors.CodeForbidden, "session is missing identity key")
 	}
 
 	err = as.validateAndUpdateAccessToken(ctx, identityId)
--- a/internal/auth/auth.go
+++ b/internal/auth/auth.go
@@ -174,7 +174,7 @@ func NewAuthenticationService(ctx context.Context, params AuthenticationServiceP
 
 	provider, err := oidc.NewProvider(ctx, params.IssuerURL)
 	if err != nil {
-		return nil, errors.E(errors.CodeServerConfiguration, fmt.Errorf("failed to create oidc provider: %v", err))
+		return nil, errors.ErrWithCode(fmt.Errorf("failed to create oidc provider: %v", err), errors.CodeServerConfiguration)
 	}
 
 	authSvc := &AuthenticationService{
@@ -223,7 +223,7 @@ func (as *AuthenticationService) AuthCodeURL() (string, string, error) {
 	b := make([]byte, 8)
 	_, err := rand.Read(b)
 	if err != nil {
-		return "", "", errors.E(fmt.Sprintf("failed to generate state secret: %s", err.Error()))
+		return "", "", errors.New(fmt.Sprintf("failed to generate state secret: %s", err.Error()))
 	}
 	state := base64.RawURLEncoding.EncodeToString(b)
 	return as.oauthConfig.AuthCodeURL(state), state, nil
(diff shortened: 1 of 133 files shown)


internal/errors/errors.go
```diff
@@ -65,58 +62,20 @@ func New(text string) error {
...
-			e.Info = v
-		default:
-			zapctx.Default.DPanic("unknown type passed to errors.E", zap.String("type", fmt.Sprintf("%T", arg)), zap.Any("value", arg))
-			return fmt.Errorf("unknown type (%T) passed to errors.E", arg)
-		}
-	}
-	if setCode {
-		return &e
+// ErrWithCode constructs an error with an explicit code.
+func ErrWithCode(err error, code Code) error {
```
alesstimec: how about something like:
```
func New(err error) *Error {
  return &Error{
      Err: err,
  }
}

func (e *Error) WithCode(code Code) *Error {
  return &Error{
     Err: e,
     Code: code
  }
}

func (e *Error) WithMessage(msg string) *Error {
  return &Error{
     Err: e,
     Message: msg,
  }
}
   
```

But then we'd need functions to get the top most error code, etc..

kian99: Replaced with `errors.Codef` as discussed.

internal/auth/auth.go
```diff
@@ -174,7 +174,7 @@ func NewAuthenticationService(ctx context.Context, params AuthenticationServiceP
 
 	provider, err := oidc.NewProvider(ctx, params.IssuerURL)
 	if err != nil {
-		return nil, errors.E(errors.CodeServerConfiguration, fmt.Errorf("failed to create oidc provider: %v", err))
+		return nil, errors.Codef(errors.CodeServerConfiguration, "failed to create oidc provider: %v", err)
```
luci1900, line 177: Does `%w` not work?

kian99: It works great, the tests have an example of its usage. And there are some usages around the code.

internal/jimm/juju/controller_test.go
```diff
@@ -1316,7 +1315,7 @@ func TestInitiateMigration(t *testing.T) {
 			},
 		},
 		initiateMigrationResults: []result{{}},
-		expectedError:            "unauthorized access",
+		expectedError:            "unauthorized",
```
kian99, line 1318: This change is because of a change above,
```diff
-		return result, errors.E(errors.CodeUnauthorized)
+		return result, errors.Codef(errors.CodeUnauthorized, "unauthorized")
```
With `errors.E` if there was no message set, the message would be the stringified error code, which in this case is "unauthorized access". I could have left the string part of `Codef` empty, which would have kept the same behaviour as before or I could do `string(errors.CodeUnauthorized)`. But I ended up changing it to "unauthorized" with no strong reason except that in many other places I saw `errors.E(errors.CodeUnauthorized, "unauthorized")`, so there was already some inconsistency.

The same change is also elsewhere.

internal/jimm/permissions/tagresolver.go
kian99, line 1: This file is an example of where we were masking the underlying error. Ideally we would check if the db returns a 'not found' error (and/or other relevant errors) and return an appropriate message/code for that. But to keep this PR simple, we simply adjust the error include the underlying DB error.

internal/rpcproxy/rpcproxy.go
```diff
@@ -828,11 +828,11 @@ func (p *clientProxy) handleLegacyLogin(ctx context.Context, msg *message) (*mes
...
-		errRedirect := errors.E(
-			errors.CodeRedirect,
-			"redirection to alternative server required",
-			info,
-		)
+		errRedirect := &errors.Error{
+			Code:    errors.CodeRedirect,
+			Message: "redirection to alternative server required",
+			Info:    info,
+		}
```
kian99, line 835: There was 1 case where we needed to specify an info map in the error. Instead of creating a function for this 1 use-case, we can just use the `Error` struct directly, giving fine-grained access to the code/message/info/underlying-error.

internal/auth/auth.go
SimoneDutto: i think Luci was asking why didn't you use %w here, and i'm curious as well

internal/auth/auth.go
```diff
@@ -223,7 +223,7 @@ func (as *AuthenticationService) AuthCodeURL() (string, string, error) {
 	b := make([]byte, 8)
 	_, err := rand.Read(b)
 	if err != nil {
-		return "", "", errors.E(fmt.Sprintf("failed to generate state secret: %s", err.Error()))
+		return "", "", fmt.Errorf("failed to generate state secret: %s", err.Error())
```
SimoneDutto, line 226: what do we prefer?
`fmt.Errorf("failed to generate state secret: %s", err.Error())`
or
`fmt.Errorf("failed to generate state secret: %w", err)`

Is it the same output?

internal/auth/auth.go
```diff
@@ -472,7 +472,7 @@ func (as *AuthenticationService) VerifyClientCredentials(ctx context.Context, cl
 
 	_, err = cfg.Token(ctx)
 	if err != nil {
-		return errors.E(errors.CodeUnauthorized, fmt.Errorf("invalid client credentials: %v", err))
+		return errors.Codef(errors.CodeUnauthorized, "invalid client credentials: %v", err)
```
SimoneDutto, line 475: %w?

internal/db/cloud.go
```diff
@@ -30,7 +29,7 @@ func (d *Database) AddCloud(ctx context.Context, c *dbmodel.Cloud) (err error) {
 	if err := db.Create(c).Error; err != nil {
 		err := dbError(err)
 		if errors.ErrorCode(err) == errors.CodeAlreadyExists {
-			return errors.E(fmt.Sprintf("cloud %q already exists", c.Name), err)
+			return errors.Codef(errors.CodeAlreadyExists, "cloud %q already exists", c.Name)
```
SimoneDutto, line 32: since the whole intent of this error rehaul was to avoid masking the error, in these instances can you add a zapctx.Error or something to make sure we don't loose the err?

SimoneDutto (commented):
Before I review the whole things i have a few questions

internal/auth/auth.go
kian99: It was already %v, so I didn't change anything. For the most part, I tried not to change any behaviour and just focused on the syntax for the errors api.

internal/auth/auth.go
kian99: It's the same output but %w will wrap the underlying error and it means it will carry any error code along with the wrapping. %s will result in an identical error string but no wrapping of the error.

I'm not interested in changing it from %s to %w in this PR though. I want to keep the behaviour the same as much as possible and just remove errors.E.

internal/auth/auth.go
kian99: Not for this PR.
Edit: Unless necessary (i.e. we spot that its a bug).

internal/db/cloud.go
kian99: In this case because the error is an "already exists" error (and we've checked that), I don't think you need to worry about losing the error. The DB error will just be some constant thing like "record already exists" or similar.

luci1900 (approved):

internal/db/cloud.go
alesstimec: since we check on L31 that the error code is already `CodeAlreadyExists` we could just do `fmt.Errorf("cloud %q already exists: %w", c.Name, err)`, right?

kian99: > since we check on L31 that the error code is already `CodeAlreadyExists` we could just do `fmt.Errorf("cloud %q already exists: %w", c.Name, err)`, right?

Yes, but that changes the error message from what it originally was. The full error message would then become something like 'cloud "foo" already exists: record already exists', which is a bit redundant.

internal/db/cloudcredential.go
```diff
@@ -25,7 +24,7 @@ func (d *Database) SetCloudCredential(ctx context.Context, cred *dbmodel.CloudCr
 	defer servermon.ErrorCounter(servermon.DBQueryErrorCount, &err, op)
 
 	if cred.CloudName == "" || cred.OwnerIdentityName == "" || cred.Name == "" {
-		return errors.E(errors.CodeBadRequest, fmt.Sprintf("invalid cloudcredential tag %q", cred.CloudName+"/"+cred.OwnerIdentityName+"/"+cred.Name))
+		return errors.Codef(errors.CodeBadRequest, "invalid cloudcredential tag %q", cred.CloudName+"/"+cred.OwnerIdentityName+"/"+cred.Name)
```
alesstimec, line 27: this seems like an error that shouldn't be coming out of the database type as we don't pass in the cloudcredential tag.. it should be coming from a level above

alesstimec: an multiple errors in this file return a string that assumes the format of the cloudcredential tag.. we should address this in the future

alesstimec (approved):

internal/db/cloudcredential.go
kian99: Yeah you're right, it's weird for the database to have this knowledge and construct the error here.

kian99: But resolving as it's unrelated to the change.

PR #1918, opened 2026-03-25
feat: accept JWKS as environment variables

kian99:
## Description
This change removes the logic that was creating a JWKS and rotating it periodically from within JIMM's application code and instead we expect a path to two files, one holding the JWKS and another holding the signing key. The paths are passed in as environment variables and JIMM periodically reads the files to allow the operator to change their values without requiring a service restart. 

The benefits of this approach are:
1. We no longer use our Vault for credentials related to the operation of JIMM (only for dynamic application-related values like user credentials and controller credentials).
2. The operator can decide whether to rotate secrets or use a static JWKS.
3. The application code is simplified.

The design is fairly simple, 
- As mentioned, we introduce 2 environment variables that represent,
  1. The JWKS - a path to a file holding a set of public keys that JIMM will advertise.
  2. The private key - a path to a file holding a single private key used to mint new JWT tokens.
- We read from the file every 5 minutes which we assume to be << (much lower) than the frequency with which the keys are rotated. This makes it easier to verify correctness, provided that the assumption is met.
- In the corresponding charm PR, we will be introducing a new public key ~7 days before the existing one is set to expire, then roughly 1 hour - 1 day (tbd) later, update the signing key, ensuring the public key has had time to propagate to controllers. 

This change doesn't keep backwards compatibility with the previous approach and leaves the old data in Vault for manual cleanup. This also implies that there will be a period of time, approx. ~cache-interval, where JIMM will receive a new JWKS from the operator and communication with backing controllers will be disrupted. This seems like a better alternative than the complexity of keeping around both approaches, especially since the last time the key rotation took place there was no significant fallout.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests


internal/db/secrets.go
```diff
@@ -21,16 +19,6 @@ const (
...
-	// These constants are used to create the appropriate identifiers for JWKS related data.
-	jwksKind          = "jwks"
-	jwksPublicKeyTag  = "jwksPublicKey"
-	jwksPrivateKeyTag = "jwksPrivateKey"
-	jwksExpiryTag     = "jwksExpiry"
-	oauthKind         = "oauth"
-	oauthKeyTag       = "oauthKey"
-	//nolint:gosec // Thinks credentials hardcoded.
-	oauthSessionStoreSecretTag = "oauthSessionStoreSecret"
 )
```
kian99, line 22: Just a note, I also removed some of these unused variables and related functions around oauth secrets. We used to generate an oauth secret in JIMM and store it in Vault but that was moved to the charm a long while ago but some tests and unused functions were not removed.

docker-compose.common.yaml
```diff
@@ -24,6 +24,80 @@ services:
       JIMM_LISTEN_ADDR: "0.0.0.0:80"
       JIMM_TEST_PGXDSN: "postgresql://jimm:jimm@db/jimm"
       JIMM_JWT_EXPIRY: 30s
+      JIMM_JWKS: |-
```
kian99: I'm wondering if we should base64 encode this value so that it's easier to pass. Passing JSON into an env var requires a lot of escaping shenanigans.

luci1900: If we have to, I'm not sure.

The values are already base64. And certificates are similarly annoying.

luci1900 (dismissed):
I like this.

cmd/jimmsrv/service/service.go
```diff
@@ -170,6 +170,15 @@ type Params struct {
...
 	JWTExpiryDuration time.Duration
 
+	// JWKS is the JSON Web Key Set document served by JIMM.
+	JWKS string
+
+	// JWKSPrivateKey is the PEM encoded RSA private key used to sign controller JWTs.
+	JWKSPrivateKey string
+
+	// JWKSCacheMaxAge is the Cache-Control max-age, in seconds, for /.well-known/jwks.json.
+	JWKSCacheMaxAge string
```
alesstimec: why isn't it an `int64`, then?

internal/jimmhttp/wellknown_handler.go
```diff
@@ -3,34 +3,39 @@
...
 
 	"github.com/canonical/jimm/v3/internal/errors"
-	"github.com/canonical/jimm/v3/internal/jimm/credentials"
 )
 
 const (
 	JwksEndpoint = "/jwks.json"
 )
 
+type JWKSProvider interface {
```
alesstimec, line 24: godoc, please

internal/jimmjwx/jwks.go
```diff
@@ -4,208 +4,151 @@ package jimmjwx
...
-	PutJWKSPrivateKey(ctx context.Context, pem []byte) error
+	"github.com/canonical/jimm/v3/internal/errors"
+)
 
-	// PutJWKSExpiry sets the expiry time for the current JWKS within the store.
-	PutJWKSExpiry(ctx context.Context, expiry time.Time) error
+type JWKSServiceParams struct {
+	JWKS          string
+	PrivateKeyPEM string
+	CacheMaxAge   string
```
alesstimec: `int64`

internal/jimmjwx/jwks.go
```diff
@@ -4,208 +4,151 @@ package jimmjwx
...
-// and a PEM encoded private key for JWT signing.
-func generateJWK(ctx context.Context) (jwk.Set, []byte, error) {
-
-	// Due to the sensitivity of controllers, it is best we allow a larger encryption bit size
-	// and accept any negligible wire cost.
-	keySet, err := rsa.GenerateKey(rand.Reader, 4096)
-	if err != nil {
-		return nil, nil, err
+func (jwks *JWKSService) Get(_ context.Context) (jwk.Set, error) {
+	if jwks == nil {
```
alesstimec: wouldn't this call panic, if `jwks == nil`?

internal/jimmjwx/jwks.go
```diff
@@ -4,208 +4,151 @@ package jimmjwx
...
-// It will return a jwk.Set containing the public key
-// and a PEM encoded private key for JWT signing.
-func generateJWK(ctx context.Context) (jwk.Set, []byte, error) {
-
-	// Due to the sensitivity of controllers, it is best we allow a larger encryption bit size
-	// and accept any negligible wire cost.
-	keySet, err := rsa.GenerateKey(rand.Reader, 4096)
-	if err != nil {
-		return nil, nil, err
+func (jwks *JWKSService) Get(_ context.Context) (jwk.Set, error) {
```
alesstimec: godoc, please

internal/jimmjwx/jwks.go
```diff
@@ -4,208 +4,151 @@ package jimmjwx
...
+	return jwks.set, nil
+}
 
-	privateKeyPEM := pem.EncodeToMemory(
-		&pem.Block{
-			Type:  "RSA PRIVATE KEY",
-			Bytes: x509.MarshalPKCS1PrivateKey(keySet),
-		},
-	)
+func (jwks *JWKSService) CacheMaxAge() int64 {
```
alesstimec: godoc, please

internal/jimmjwx/jwks.go
```diff
@@ -4,208 +4,151 @@ package jimmjwx
...
+	if jwks == nil {
```
alesstimec: same about panic question

internal/jimmjwx/jwks.go
```diff
@@ -4,208 +4,151 @@ package jimmjwx
...
+		return 0
+	}
+	return jwks.cacheMaxAge
+}
 
-	// We also use the same methodology of generating UUIDs for our KID
-	kid, err := uuid.NewRandom()
-	if err != nil {
-		return nil, nil, err
+func (jwks *JWKSService) SigningKey() jwk.Key {
```
alesstimec: godoc, please

internal/jimmjwx/utils_test.go
```diff
@@ -3,88 +3,86 @@
...
+func newJWKSService(c *qt.C) (*jimmjwx.JWKSService, jwk.Set) {
+	c.Helper()
+	params, set, _ := newJWKSServiceParams(c)
+	service, err := jimmjwx.NewJWKSService(params)
+	c.Assert(err, qt.IsNil)
+	return service, set
+}
 
-	return store
+func newJWTService(c *qt.C, expiry time.Duration) (*jimmjwx.JWTService, jwk.Set) {
```
alesstimec: suggestion to call this `mustNewJWTService`

internal/jimmjwx/utils_test.go
```diff
@@ -3,88 +3,86 @@
...
+		JWKS:          string(rawJWKS),
+		PrivateKeyPEM: string(privateKey),
+		CacheMaxAge:   "600",
+	}, set, privateKey
 }
 
-// setupCredentialStore sets up an in-memory credential store for testing.
-func setupCredentialStore(_ context.Context, _ *qt.C) credentials.CredentialStore {
-	store := jimmtest.NewInMemoryCredentialStore()
+func newJWKSService(c *qt.C) (*jimmjwx.JWKSService, jwk.Set) {
```
alesstimec: suggestion to call this `mustNewJWKSService`

docker-compose.common.yaml
alesstimec: i would base64 encode it, but ymmv

alesstimec (commented):
looking good, but a few comments

internal/jimmjwx/jwks.go
kian99: Not quite, methods are the same as a function with the first param being the receiver. 
I.e.
```go
func (jwks *JWKSService) Get(_ context.Context) (jwk.Set, error)
```
Is equivalent to,
```go
func Get(jwks *JWKSService, _ context.Context) (jwk.Set, error)
```

internal/jimmjwx/utils_test.go
kian99: This is in a test helper so I don't think it needs the extra `Must`.

internal/jimmjwx/utils_test.go
kian99: This is also a test helper so I don't think it needs the `must`.

internal/jimmjwx/jwks.go
kian99: I started changing this but stopped halfway. It seems better to only fetch the env var in `main.go` and then let this package decide on the parsing and error message. What do you think? It also makes it easier to test since we actually don't have any tests for `main.go`.

internal/jimmjwx/utils_test.go
alesstimec: `must` prefix to me means that we're internally checking errors and might panic if an error occurs

kian99: Yeah but this won't panic, it will fail the test, which is a slightly different but important difference.

internal/jimmjwx/jwks.go
```diff
@@ -3,209 +3,236 @@
...
+	for {
+		select {
+		case <-ticker.C:
+			jwks.refresh(ctx)
+		case <-ctx.Done():
+			zapctx.Info(ctx, "exiting jwks refresh polling")
+			return
 		}
+	}
+}
```
kian99, line 125: For this method, there was an initial approach where the JWKSService had a `Close()`/`Stop()` (take your pick) method that would use channels to signal the refresh loop to stop and wait for it to exit. That was fine and allowed the caller to be sure the refresh loop had stopped.

I thought it would be a bit easier to rely on the caller to pass a context in that gets stopped on shutdown. It's definitely less explicit though. I'm okay with either approach. This current one is also tricky to test, you'll notice I ended up testing it by checking for the log so that's kinda funky.

cmd/jimmsrv/main.go
```diff
@@ -192,6 +192,14 @@ func start(ctx context.Context, s *service.Service) error {
...
 	}
 	internalSrv := jimmsvc.NewInternalService(internalAddr, corsAllowedOrigins)
+	jwksPath := os.Getenv("JIMM_JWKS_PATH")
+	if jwksPath == "" {
+		jwksPath = os.Getenv("JIMM_JWKS")
+	}
+	jwksPrivateKeyPath := os.Getenv("JIMM_JWKS_PRIVATE_KEY_PATH")
+	if jwksPrivateKeyPath == "" {
+		jwksPrivateKeyPath = os.Getenv("JIMM_JWKS_PRIVATE_KEY")
+	}
```
SimoneDutto, line 202: this is amazing, because now we can just fix a key from docker compose and we don't have to redo `qa-lxd` everytime we restart our machines.

docker-compose.yaml
```diff
@@ -29,6 +29,9 @@ services:
     container_name: jimm-test
     ports:
       - 17070:80
+    volumes:
+      - ./internal/testutils/jimmtest/testdata/jwks.json:/etc/jimm/jwks.json:ro
+      - ./internal/testutils/jimmtest/testdata/jwks_private_key.pem:/etc/jimm/jwks_private_key.pem:ro
```
SimoneDutto, line 34: can you add this to the developer doc? I think we need these files now to start dev jimm, right?

cmd/jimmsrv/service/service_test.go
```diff
@@ -43,11 +43,16 @@ func TestMain(m *testing.M) {
...
+	c.Assert(err, qt.IsNil)
+
 	return jimmsvc.Params{
-		DSN:            testdb.CreateEmptyDatabase(c),
-		ControllerUUID: "6acf4fd8-32d6-49ea-b4eb-dcb9d1590c11",
-		PrivateKey:     "ly/dzsI9Nt/4JxUILQeAX79qZ4mygDiuYGqc2ZEiDEc=",
-		PublicKey:      "izcYsQy3TePp6bLjqOo3IRPFvkQd2IKtyODGqC6SdFk=",
+		DSN:                testdb.CreateEmptyDatabase(c),
+		ControllerUUID:     "6acf4fd8-32d6-49ea-b4eb-dcb9d1590c11",
+		PrivateKey:         "ly/dzsI9Nt/4JxUILQeAX79qZ4mygDiuYGqc2ZEiDEc=",
```
SimoneDutto, line 53: i'm surprised the linter doesn't complain about this

internal/jimmhttp/wellknown_handler.go
```diff
@@ -52,59 +57,25 @@ func (wkh *WellKnownHandler) SetupMiddleware() {
...
 // The purpose of this is to allow juju controllers to retrieve the public key from JIMM
 // and decode the presented JWT.
-//
-// The JWKS is expected to be cached by the client, where the expiry time
-// is the expiry time persisted for this set in our credential store.
 func (wkh *WellKnownHandler) JWKS(w http.ResponseWriter, r *http.Request) {
 
 	ctx := r.Context()
-	if wkh == nil || wkh.CredentialStore == nil {
+	if wkh == nil || wkh.JWKSProvider == nil {
```
SimoneDutto, line 63: is this even possible now? Imo JIMM should have failed earlier:
- if JWKSProvider is nil it should fail in `NewWellKnownHandler`
- wkh should never be nil, right?

internal/jimmhttp/wellknown_handler.go
```diff
@@ -52,59 +57,25 @@ func (wkh *WellKnownHandler) SetupMiddleware() {
...
-		remaining := int64(expiry.Sub(now).Seconds())
-		if remaining < maxAgeSeconds {
-			actualMaxAge = remaining
-		}
-	} else {
-		actualMaxAge = 10
-	}
-	w.Header().Set("Cache-Control", fmt.Sprintf("must-revalidate, max-age=%d", actualMaxAge))
+	// Cache-control of 10 minutes
+	w.Header().Set("Cache-Control", "must-revalidate, max-age=600")
```
SimoneDutto: do we need must-revalidate? From what I got, it seems it's needed to make sure nobody can use the cache value if the origin is unreachable.

internal/jimmjwx/jwks.go
```diff
@@ -3,209 +3,236 @@
...
+//
+// The file paths are expected to be managed by the operator, e.g. a Juju charm or otherwise.
+// JIMM reads from the files with a relatively high frequency (every 5 minutes) compared to
+// the expected frequency of updates (hours/days) to allow for changes to be picked up in a
+// timely manner without requiring a restart of the service.
+//
+// The provided context is used to manage the lifecycle of the refresh loop; cancelling the
+// context will stop the loop and must be done on cleanup.
+func NewJWKSService(ctx context.Context, p JWKSServiceParams) (*JWKSService, error) {
+	if ctx == nil {
```
SimoneDutto: we have never done this, do we need it?

internal/jimmjwx/jwks.go
```diff
@@ -3,209 +3,236 @@
...
+		return nil, err
+	}
 
-	// PutJWKSExpiry sets the expiry time for the current JWKS within the store.
-	PutJWKSExpiry(ctx context.Context, expiry time.Time) error
+	service := &JWKSService{
+		jwksPath:       p.JWKSPath,
+		privateKeyPath: p.PrivateKeyPath,
+		cached:         material,
+	}
```
SimoneDutto, line 75: i don't know if it's too much to ask for, but i would like us to handle the jwks and the private key in two separate key in two separate services. So we make sure we have different level of alert on what we are doing with them. One is public and not important, the other is super important and private.
Wdyt?

internal/jimmjwx/jwks.go
```diff
@@ -3,209 +3,236 @@
...
-// generateJWKS generates a new set of JWK using RSA256[4096]
-//
-// It will return a jwk.Set containing the public key
-// and a PEM encoded private key for JWT signing.
-func generateJWK(ctx context.Context) (jwk.Set, []byte, error) {
+	signingKey, err := jwk.FromRaw(privateKey)
+	if err != nil {
+		return cachedJWKS{}, fmt.Errorf("create jwks signing key: %w", err)
+	}
+	if err := signingKey.Set(jwk.AlgorithmKey, jwa.RS256); err != nil {
```
SimoneDutto, line 155: can we derive the algo from the key? Or we just hard-code this and it's fine?

internal/jimmjwx/jwks.go
```diff
@@ -3,209 +3,236 @@
...
+		return nil, errors.New("jwks private key is not rsa")
 	}
+	return rsaKey, nil
+}
 
-	ks := jwk.NewSet()
-	err = ks.AddKey(jwks)
-	if err != nil {
-		return nil, nil, err
+func matchingPublicKey(set jwk.Set, signingPublicKey *rsa.PublicKey) (jwk.Key, error) {
```
SimoneDutto, line 212: what is this method doing?

SimoneDutto (commented):
a few comments here and there

cmd/jimmsrv/service/service_test.go
kian99: 🤷

internal/jimmhttp/wellknown_handler.go
kian99: It's likely impossible as the code is written now. But without a strong opinion myself, it seems fine to leave the checks.

internal/jimmhttp/wellknown_handler.go
kian99: That's a fair point, I re-read the must-revalidate section on https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Cache-Control and I do agree we can probably remove it so I'll do that.

internal/jimmjwx/jwks.go
kian99: Ah yeah agreed, I'll remove this

internal/jimmjwx/jwks.go
kian99: I don't really follow you here. I don't see anything stopping us from doing things with them separately.

internal/jimmjwx/jwks.go
kian99: Not sure if we can derive it, but even if we can't we could add metadata to the key file. But for now, we just expect a specific type and we can extend it later if the need arises.

internal/jimmjwx/jwks.go
kian99: It checks that the key we are signing with has a matching public key in the JWKS. We don't want to sign tokens with a key that doesn't have a corresponding public key in our JWKS.

internal/jimmjwx/jwks.go
```diff
@@ -3,209 +3,233 @@
...
+	}
 
-	// PutJWKSExpiry sets the expiry time for the current JWKS within the store.
-	PutJWKSExpiry(ctx context.Context, expiry time.Time) error
+	service := &JWKSService{
+		jwksPath:       p.JWKSPath,
+		privateKeyPath: p.PrivateKeyPath,
+		cached:         material,
+	}
+	go service.refreshLoop(ctx)
```
alesstimec, line 76: we could do refresh on read as simone did in the supported versions cache.. fewer goroutines

internal/jimmjwx/jwks.go
```diff
@@ -3,209 +3,233 @@
...
+	return service, nil
 }
 
-// JWKSService handles the creation, rotation and retrieval of JWKS for JIMM.
-// It utilises the underlying credential store currently in effect.
-type JWKSService struct {
-	credentialStore CredentialStore
+// Get returns the JWKS set to be served at /.well-known/jwks.json.
+func (jwks *JWKSService) Get(_ context.Context) (jwk.Set, error) {
+	if jwks == nil {
```
alesstimec: i don't think this check is needed, because we could add it to every method with a pointer receiver.. if we reach L83 we deserve to panic :)

internal/jimmjwx/jwks.go
alesstimec: as i wrote in the comment above, i don't think we need a refresh loop - we can refresh on `Get`, if enough time has passed since the last refresh.

alesstimec (approved):
LGTM with a few comments.. but how do you see upgrade working for current deploys? would we need an intermediate version that would still fetch jwks from vault AND combine that with a jwks from the cham, then upgrade to a purely charm-based version?

internal/jimmjwx/jwks.go
SimoneDutto: don't we have a std lib method for this? This method feels very manual and error prone

internal/jimmjwx/jwks.go
SimoneDutto: i don't like the idea of handling these two values, that serves two different purposes and two different level of security associated to it, in the same service

internal/jimmjwx/jwks.go
SimoneDutto: yeah, i agree it would be better to do it lazily

internal/jimmjwx/jwks.go
kian99: Sure, I'll remove this.

internal/jimmjwx/jwks.go
kian99: That's how it was initially but was a bit harder to read as it wasn't immediately clear what would happen if the read from disk failed. I think the current approach is a lot easier to read but that does come at the cost of the extra go routine. I don't think an extra go routine is a problem if it adds clarity.

internal/jimmjwx/jwks.go
SimoneDutto: after a discussion offline, i get them being together because it's less code, less watchers for the files. We would just need to thread the two values with different level of awareness

internal/jimmjwx/jwks.go
kian99: Ptal, made a tweak to use a comparator function from the jwk lib. From a quick search the difference now is that we compare by computing a "thumbprint" of the keys which is agnostic to the algorithm of the key whereas before the comparison of the 'exponent' and 'modulus' was specific to RSA keys.

SimoneDutto: this feels better

SimoneDutto (approved):
happy to approve

internal/jimmjwx/jwks.go
```diff
@@ -3,209 +3,227 @@
...
-	// JWTs from JIMM.
-	putJwks := func(expires time.Time) error {
-		set, key, err := generateJWK(ctx)
-		if err != nil {
-			return err
-		}
+func (jwks *JWKSService) refresh(ctx context.Context) {
+	material, err := loadJWKS(jwks.jwksPath, jwks.privateKeyPath)
+	if err != nil {
+		zapctx.Error(ctx, "failed to refresh jwks, using cached value", zap.Error(err))
```
alesstimec, line 104: we should add monitoring around this error so that we can potentially add alerting

kian99: Discussed that alerting on logs is already an option in tools like Loki.
Made a Jira card to add it to our default set of alert rules in the charm https://warthogs.atlassian.net/browse/JUJU-9583.

PR #1923, opened 2026-03-31
feat: add additional flags to jaas bootstrap command

kian99:
## Description
This change adds additional flags to the `jaas bootstrap` command to give it similar capabilities to `juju bootstrap`, allowing a user to specify the bootstrap base, bootstrap constraints, model constraints, model defaults, storage-pool, config.

The constraint parsing logic was largely lifted from the Juju codebase.

Fixes: [JUJU-9427](https://warthogs.atlassian.net/browse/JUJU-9427)

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

[JUJU-9427]: https://warthogs.atlassian.net/browse/JUJU-9427?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cmd/jaas/cmd/bootstrap.go
+++ b/cmd/jaas/cmd/bootstrap.go
@@ -143,8 +143,8 @@ func (c *bootstrapCommand) SetFlags(f *gnuflag.FlagSet) {
 		"json": cmd.FormatJson,
 	})
 	f.StringVar(&c.credentialName, "credential", "", "The name of the cloud credential to use for bootstrapping. Only required if more than one credential is available for the cloud.")
-	f.StringVar(&c.bootstrapBase, "bootstrap-base", "", "Specify the base of the bootstrap machine")
-	f.Var(&c.bootstrapCons, "bootstrap-constraints", "Specify bootstrap machine constraints")
+	f.StringVar(&c.bootstrapBase, "bootstrap-base", "", "Specify the base of the bootstrap machine.")
+	f.Var(&c.bootstrapCons, "bootstrap-constraints", "Specify bootstrap machine constraints.")
 	f.Var(&c.constraints, "constraints", "Set model constraints")
 	f.Var(&c.config, "config",
 		"Specify a configuration file, or one or more configuration options.\n    (`--config config.yaml [--config key=value ...])`")
--- a/cmd/jaas/cmd/bootstrap.go
+++ b/cmd/jaas/cmd/bootstrap.go
@@ -427,6 +427,7 @@ func parseConstraintFlag(ctxt *cmd.Context, values []string, flagName string) (m
 	if len(values) == 0 {
 		return nil, nil
 	}
+	// ParseWithAliases requires only spaces and name=value pairs.
 	joined := strings.Join(values, " ")
 	_, aliases, err := constraints.ParseWithAliases(joined)
 	common.WarnConstraintAliases(ctxt, aliases)
--- a/cmd/jaas/cmd/bootstrap.go
+++ b/cmd/jaas/cmd/bootstrap.go
@@ -12,6 +12,8 @@ import (
 	jujucmd "github.com/juju/juju/cmd"
 	"github.com/juju/juju/cmd/juju/common"
 	"github.com/juju/juju/cmd/modelcmd"
+	corebase "github.com/juju/juju/core/base"
+	"github.com/juju/juju/core/constraints"
 	"github.com/juju/juju/jujuclient"
 	jujuparams "github.com/juju/juju/rpc/params"
 	"github.com/juju/names/v5"
@@ -46,10 +48,11 @@ file contents.
 These config options must match the config options supported by the Juju CLI
 for the version of Juju being bootstrapped. See the Juju documentation for
 the version specified for the full list of supported bootstrap config
-options.
+options. Additional bootstrap settings can be supplied with --bootstrap-base,
+--bootstrap-constraints, --constraints, --model-default, and --storage-pool,
+these align with the corresponding Juju CLI options.
 
-Note that some config options may not be specified as they will automatically
-be set.
+Note that some config options will be automatically set but can be overriden.
 These are:
 
 - login-token-refresh-url
@@ -73,6 +76,8 @@ Note that JIMM will internally do the following:
 	juju [jaas] bootstrap <cloud[/region]> <controller name> <controller version>
 	juju [jaas] bootstrap mycloud/region mycontroller 3.6.8
 	juju [jaas] bootstrap mycloud/region mycontroller 3.6.8 --config controller-service-type=loadbalancer
+	juju [jaas] bootstrap mycloud/region mycontroller 3.6.8 --bootstrap-base ubuntu@24.04 --bootstrap-constraints mem=8G --constraints arch=amd64
+	juju [jaas] bootstrap mycloud/region mycontroller 3.6.8 --storage-pool name=controller-pool --storage-pool type=ebs --config audit-log-enabled=true
 `
 )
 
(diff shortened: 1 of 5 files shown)


luci1900 (approved):
That's some confusing parsing and formatting, but it should be fine if it works.

kian99:
> That's some confusing parsing and formatting, but it should be fine if it works.

I agree, but it gets the job done.

cmd/jaas/cmd/bootstrap.go
```diff
@@ -128,8 +143,15 @@ func (c *bootstrapCommand) SetFlags(f *gnuflag.FlagSet) {
 		"json": cmd.FormatJson,
 	})
 	f.StringVar(&c.credentialName, "credential", "", "The name of the cloud credential to use for bootstrapping. Only required if more than one credential is available for the cloud.")
+	f.StringVar(&c.bootstrapBase, "bootstrap-base", "", "Specify the base of the bootstrap machine")
+	f.Var(&c.bootstrapCons, "bootstrap-constraints", "Specify bootstrap machine constraints")
```
ale8k:
```suggestion
	f.Var(&c.bootstrapCons, "bootstrap-constraints", "Specify bootstrap machine constraints.")
```

cmd/jaas/cmd/bootstrap.go
```diff
@@ -128,8 +143,15 @@ func (c *bootstrapCommand) SetFlags(f *gnuflag.FlagSet) {
 		"json": cmd.FormatJson,
 	})
 	f.StringVar(&c.credentialName, "credential", "", "The name of the cloud credential to use for bootstrapping. Only required if more than one credential is available for the cloud.")
+	f.StringVar(&c.bootstrapBase, "bootstrap-base", "", "Specify the base of the bootstrap machine")
```
ale8k:
```suggestion
	f.StringVar(&c.bootstrapBase, "bootstrap-base", "", "Specify the base of the bootstrap machine.")
```

cmd/jaas/cmd/bootstrap.go
```diff
@@ -310,3 +321,146 @@ func cloudToParams(cloudName, regionName string, cloud *jujucloud.Cloud) apipara
 	}
 	return paramsCloud
 }
+
+func (c *bootstrapCommand) bootstrapOptions(ctxt *cmd.Context) (apiparams.BootstrapOptions, error) {
```
ale8k, line 325: buildBootstrapOptions would be better or something similar

cmd/jaas/cmd/bootstrap.go
```diff
@@ -310,3 +321,146 @@ func cloudToParams(cloudName, regionName string, cloud *jujucloud.Cloud) apipara
...
+	return values, nil
+}
+
+func stringifyScalar(value interface{}) (string, error) {
+	if value == nil {
+		return "", fmt.Errorf("nil value")
+	}
+	switch value.(type) {
+	case string,
+		bool,
```
ale8k, line 416: This means we'd need to maintain this right? The types Juju uses?... I mean for now it is basically everything but should they switch we'd have an issue?

Second Q: Do we need to do this? Why not just fmt.Sprint?

cmd/jaas/cmd/bootstrap.go
```diff
@@ -310,3 +321,146 @@ func cloudToParams(cloudName, regionName string, cloud *jujucloud.Cloud) apipara
...
+		strValue, err := stringifyScalar(value)
+		if err != nil {
+			return nil, fmt.Errorf("%s value for %q must be a scalar, got %T", flagName, key, value)
+		}
+		values[key] = strValue
+	}
+	return values, nil
+}
+
+func stringifyScalar(value interface{}) (string, error) {
```
ale8k, line 410: Isn't this func more: "RejectComplexTypesAndParseValuesIntoStringsThatPassTheGate"? It isn't just stringifying single vals, it is acting as a gate?

cmd/jaas/cmd/bootstrap.go
```diff
@@ -310,3 +321,146 @@ func cloudToParams(cloudName, regionName string, cloud *jujucloud.Cloud) apipara
...
+	return result, nil
+}
+
+// splitEscapedFields splits a whitespace-delimited constraint string while
+// preserving spaces that were escaped as `\ ` inside an individual field.
+func splitEscapedFields(value string) []string {
+	if value == "" {
+		return nil
+	}
+	normalized := strings.ReplaceAll(value, `\ `, "\x00")
```
ale8k, line 459: Is this needed? Can we not just send '\ '? Does it work without this?

cmd/jaas/cmd/bootstrap_test.go
```diff
@@ -412,3 +469,87 @@ func TestBootstrapSpecifiedCredentialWithDefault(t *testing.T) {
...
+			expected: nil,
+		},
+		{
+			name:     "plain fields",
+			value:    "mem=8G arch=amd64",
+			expected: []string{"mem=8G", "arch=amd64"},
+		},
+		{
+			name:     "escaped spaces",
+			value:    `tags=prod\ blue zones=us-east-1a`,
```
ale8k, line 492: Yeah I'm curious to know if it works with "\ "

cmd/jaas/cmd/bootstrap.go
kian99: I don't think there's a risk of any new base types being introduced so that's fine.

And we could use `fmt.Sprintf` if we're sure there is no way for types like `[]string` but because you can point the flag to a yaml file we could get types like `[]string` we forbid that. Otherwise we'd need to know how to convert that `[]string` into a plain string based on the Juju's parsing of that key-value.

cmd/jaas/cmd/bootstrap.go
kian99: I believe we do need it. Otherwise the space in `\ ` would get removed below. Since it's escaped, we convert it to a special char, split on spaces using `strings.Fields` and then convert that special char back to a space.

But maybe there's a simpler way to do it that I'm not aware of.

I tried commenting out the line but then the test fails with,
```
        got:
          []string{"tags=prod\\", "blue", "zones=us-east-1a"}
        want:
          []string{"tags=prod blue", "zones=us-east-1a"}
```
It seems that `strings.Fields()` doesn't respect the escaping of `\ `.

cmd/jaas/cmd/bootstrap.go
kian99: Yeah stringifyScalarOrError I guesss, but I'll probably keep the name as is.

cmd/jaas/cmd/bootstrap.go
ale8k: Yeah I see the issue :/... But if we're only allowing non-complex types, then I don't think it is unreasonable to expect non-complex and if they provide a complex by file - we could just attempt it through and let Juju handle it?

cmd/jaas/cmd/bootstrap.go
ale8k: No no it's ok, was just curious, I think this is necessary

ale8k (approved):
LGTM, maybe we can just allow users to send complex types though and remove the scalar func to reduce the amount of parsing we're doing.

cmd/jaas/cmd/bootstrap.go
kian99: Yeah that would be okay too I think, let the server handle it. I'll leave as is for now but fair point.

PR #1925, opened 2026-03-31
fix(context): t.Context() can't be used in t.Cleanup()

SimoneDutto:
# Description

t.Context() is cancelled before calling t.Cleanup(), so we need to use context.Background() instead.

```
// Context returns a context that is canceled just before
// Cleanup-registered functions are called.
//
// Cleanup functions can wait for any resources
// that shut down on [context.Context.Done] before the test or benchmark completes.
func (c *common) Context() context.Context {
	c.checkFuzzFn("Context")
	return c.ctx
}
```

# QA

No e2e test in caas-modelmanager_test.go should fail.


luci1900 (approved):
`context.Background()` is ok in tests, although technically cleaning up should get its own context that can be cancelled/timed out separately.

kian99 (approved):

PR #1926, opened 2026-04-01
fix(e2e-tests): fix controller facade e2e testing

SimoneDutto:
## Description

fix by using the controller client instead of raw facade calls, and reformat not-implement messages

# QA
no tests in controller_test.go or controlerroot_test.go should fail.

### DIFF
--- a/testing/controllerroot_test.go
+++ b/testing/controllerroot_test.go
@@ -7,8 +7,8 @@ import (
 
 	qt "github.com/frankban/quicktest"
 	"github.com/juju/juju/api"
+	"github.com/juju/juju/core/semversion"
 	jujuparams "github.com/juju/juju/rpc/params"
-	"github.com/juju/version/v2"
 
 	"github.com/canonical/jimm/v3/internal/testutils/jimmtest"
 )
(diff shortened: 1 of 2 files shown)


luci1900 (approved):
More client good

testing/controllerroot_test.go
```diff
@@ -53,5 +53,5 @@ func TestUnimplementedRootFails(t *testing.T) {
 	defer conn.Close()
 	var resp jujuparams.RedirectInfoResult
 	err := conn.APICall(t.Context(), "NoSuch", 1, "", "Method", nil, &resp)
-	c.Assert(err, qt.ErrorMatches, `(?s).*no such request - method NoSuch\(1\).Method is not implemented \(not implemented\).*`)
+	c.Assert(err, qt.ErrorMatches, `(?s).*unknown method "Method" at version 1 for facade type "NoSuch" \(not implemented\).*`)
```
kian99, line 56: Did this change because the conn.APICall error is different in Juju 4 or something else? 
If the error is from the controller then it's a bit tricky because this test needs to work with Juju 4 and Juju 3 controllers right (but we're still testing with only Juju 3 iirc so I doubt it's this)?

kian99 (approved):

testing/controllerroot_test.go
SimoneDutto: it's the way the client enriches a `not-implemented`, so no api server changes and it should work with both 3 and 4 with this new regex

PR #1928, opened 2026-04-01
fix(e2e test): fix modelmanager_test by using the client instead of raw facade

SimoneDutto:
## Description

- Update the juju from tip of 4.0
- use the modelmanager client

## QA

no modelmanager_test.go tests should fail anymore.


internal/dbmodel/model.go
```diff
@@ -138,7 +138,6 @@ func (m Model) ToJujuModel() jujuparams.Model {
 	jm.Name = m.Name
 	jm.UUID = m.UUID.String
 	jm.Qualifier = m.OwnerIdentityName
-	jm.Type = m.CloudRegion.Cloud.Type
```
SimoneDutto, line 141: this was a wrong thing i've put because i wasn't understanding that we had an actual bug in jimm v3

testing/modelmanager_test.go
```diff
@@ -269,119 +270,101 @@ func TestCreateModel(t *testing.T) {
...
 		c.Assert(mi.UUID, qt.Not(qt.Equals), "")
-		c.Assert(mi.Qualifier, qt.Equals, test.owner)
+		c.Assert(mi.Qualifier, qt.Equals, coremodel.Qualifier(test.owner))
 		c.Assert(mi.ControllerUUID, qt.Equals, jimmtest.ControllerUUID)
 		c.Assert(mi.Users, qt.Not(qt.HasLen), 0)
-		if test.credentialTag == "" {
-			c.Assert(mi.CloudCredentialTag, qt.Not(qt.Equals), "")
+		emptyCred := names.CloudCredentialTag{}
+		if test.credentialTag == emptyCred {
+			c.Assert(mi.CloudCredential, qt.Not(qt.Equals), "")
```
kian99, line 360: What's going on here? If empty then miCloudCredential != ""?

testing/modelmanager_test.go
```diff
@@ -269,119 +270,101 @@ func TestCreateModel(t *testing.T) {
...
 		} else {
-			tag, err := names.ParseCloudCredentialTag(mi.CloudCredentialTag)
-			c.Assert(err, qt.Equals, nil)
-			c.Assert(tag.String(), qt.Equals, test.credentialTag)
+			c.Assert(mi.CloudCredential, qt.Equals, test.credentialTag.Id())
 		}
-		if test.cloudTag == "" {
-			c.Assert(mi.CloudTag, qt.Equals, names.NewCloudTag(jimmtest.TestE2ECloudName).String())
+		if test.cloud == "" {
+			c.Assert(mi.Cloud, qt.Equals, jimmtest.TestE2ECloudName)
```
kian99, line 365: I'd prefer if we were explicit in the tests, but I recall we accept an empty cloud and if JIMM only has 1 cloud, then we select that one.

testing/modelmanager_test.go
kian99, line 1: Can you add to the PR description a summary of what you needed to change here?

testing/modelmanager_test.go
```diff
@@ -486,88 +473,56 @@ func TestModifyModelAccessErrors(t *testing.T) {
...
-		},
-		expectError: `invalid action "not-an-action"`,
+		about:       "invalid username",
+		user:        "not-a-user-tag",
+		access:      "read",
+		modelUUID:   model.UUID.String,
+		expectError: `unsupported local user; if this is a service account add @serviceaccount domain`,
 	}, {
-		about: "invalid access",
-		modifyModelAccess: jujuparams.ModifyModelAccess{
```
kian99, line 550: How come these `modifyModelAccess` was removed?

testing/modelmanager_test.go
SimoneDutto: As a note, i've just refactor the tests a bit because now we are using the client.

I think the original intent was: "if the cred is not specified, then we check another cred has been used (not empty)"

testing/modelmanager_test.go
SimoneDutto: that's exactly it

testing/modelmanager_test.go
SimoneDutto: i didn't change much, it's just type refactoring basically, the original intent of the test is the same

testing/modelmanager_test.go
SimoneDutto: it's because we were always "granting" something, and using the client it's just easier to refactor the type to grant an access to a user, without this jujuparams struct.

It's just a refactor of the test type to make it more readable.

kian99 (approved):

testing/modelmanager_test.go
```diff
@@ -269,119 +270,101 @@ func TestCreateModel(t *testing.T) {
...
+		mi, err := client.CreateModel(t.Context(), test.name, names.NewUserTag(test.owner), test.cloud, test.region, test.credentialTag, test.config)
 		if test.expectError != "" {
 			c.Assert(err, qt.ErrorMatches, test.expectError)
 			continue
 		}
 		c.Assert(err, qt.Equals, nil)
 		c.Assert(mi.Name, qt.Equals, test.name)
 		c.Assert(mi.UUID, qt.Not(qt.Equals), "")
-		c.Assert(mi.Qualifier, qt.Equals, test.owner)
+		c.Assert(mi.Qualifier, qt.Equals, coremodel.Qualifier(test.owner))
```
alesstimec, line 355: can't we 
```
c.Assert(string(mi.Qualifier), qt.Equals, test.owner)
```
and save ourselves an import (coremodel)?

alesstimec (approved):
LGTM (with a comment)

testing/modelmanager_test.go
SimoneDutto: yeah, but what do we gain from not importing the core pkg?  We do it in many other places

SimoneDutto: merging, but happy to remove the import if we need it

PR #1933, opened 2026-04-02
fix(remove-cloud-spec): remove CloudSpec in favor of ControllerModelSummary

SimoneDutto:
# Description

implement ControllerModelSummary by querying all models and taking only the controller. In this way we can avoid using CloudSpec in JIMM, since it's causing troubles.

And now we can do `make qa-lxd` againt the `4.0` branch in juju.

> fly-by update delve

# QA

`git checkout 4.0` in juju

in jimm `make qa-lxd`, and it should work.

### DIFF
--- a/internal/jujuclient/modelmanager.go
+++ b/internal/jujuclient/modelmanager.go
@@ -116,6 +116,22 @@ func (c Connection) DumpModelDB(ctx context.Context, tag names.ModelTag) (map[st
 	return modelmanager.NewClient(&c).DumpModelDB(ctx, tag)
 }
 
+// ControllerModelSummary retrieves the ModelSummary for the controller
+// model. ControllerModelSummary uses the ListModelSummaries procedure on
+// the ModelManager facade.
+func (c Connection) ControllerModelSummary(ctx context.Context) (base.UserModelSummary, error) {
+	modelSummaries, err := modelmanager.NewClient(&c).ListModelSummaries(ctx, c.user.ResourceTag().String(), true)
+	if err != nil {
+		return base.UserModelSummary{}, err
+	}
+	for _, r := range modelSummaries {
+		if r.IsController {
+			return r, nil
+		}
+	}
+	return base.UserModelSummary{}, errors.E("controller model not found", errors.CodeNotFound)
+}
+
 // ListModelSummaries retrieves the list of model summaries from the controler
 func (c Connection) ListModelSummaries(ctx context.Context, ms jujuparams.ModelSummariesRequest) ([]base.UserModelSummary, error) {
 	return modelmanager.NewClient(&c).ListModelSummaries(ctx, c.user.ResourceTag().String(), ms.All)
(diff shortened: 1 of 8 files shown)


internal/testutils/jimmtest/api.go
```diff
@@ -238,11 +237,11 @@ func (a *API) Clouds(ctx context.Context) (map[names.CloudTag]jujucloud.Cloud, e
 	return a.Clouds_(ctx)
 }
 
-func (a *API) CloudSpec(ctx context.Context) (cloudspec.CloudSpec, error) {
```
kian99, line 241: I thought we were also using CloudSpec for controller bootstrap/destroy. Specifically when destroying a controller?

SimoneDutto: image


we don't

internal/jujuclient/modelmanager.go
```diff
@@ -116,6 +116,22 @@ func (c Connection) DumpModelDB(ctx context.Context, tag names.ModelTag) (map[st
 	return modelmanager.NewClient(&c).DumpModelDB(ctx, tag)
 }
 
+// ControllerModelSummary retrieves the ModelSummary for the controller
+// model. ControllerModelSummary uses the ListModelSummaries procedure on
+// the ModelManager facade.
+func (c Connection) ControllerModelSummary(ctx context.Context) (base.UserModelSummary, error) {
+	modelSummaries, err := modelmanager.NewClient(&c).ListModelSummaries(ctx, c.user.ResourceTag().String(), true)
```
ale8k, line 123: A normal ListModelSummaries and then extracting the controller model summary is fine where it is called no? And we already have it?

ale8k (commented):
I don't think we need to add a new method called ControllerModelSummary in our client, we already have ListModelSummaries

internal/jujuclient/modelmanager.go
SimoneDutto: having the jujuclient package is exactly to do this kind of things in my opinion

kian99: Yeah I think it's fine like this, iirc this is how it was anyway a few weeks/months ago, before I switched it to CloudSpec.

internal/testutils/jimmtest/api.go
kian99: Right, maybe it was in the Juju CLI that it was using CloudSpec or maybe we just removed it.

SimoneDutto: Ian said there is no use of CloudSpec bar the jaas one

SimoneDutto: ah no, it's terraform:
https://github.com/SimoneDutto/terraform-provider-juju/blob/use-juju-4-edge/internal/juju/controllers.go#L828

internal/jujuclient/modelmanager.go
ale8k: Ah ok sure idm

ale8k (approved):
lgtm, nice!

kian99 (approved):

PR #1937, opened 2026-04-03
ci(juju-4): test jaas with a backing juju 4 controller

SimoneDutto:
## Description

e2e tests against 3 and 4 backing controllers.

### DIFF
--- a/.github/workflows/integration-test.yaml
+++ b/.github/workflows/integration-test.yaml
@@ -60,7 +60,7 @@ jobs:
         id: jaas
         with:
           jimm-version: dev
-          juju-channel: "3.6/stable"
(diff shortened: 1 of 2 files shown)


.github/workflows/integration-test.yaml
kian99, line 1: We're not going to run this with both?

SimoneDutto: yeah, right now this version is just setting the juju CLI, which we require 4.0 in the feature branch, but it still bootstrap a 3.6.x backing controller.

We could expand this to bootstrap a 4.0 controller in another pr

alesstimec (approved):

kian99 (approved):

PR #1945, opened 2026-04-14
feat: rename admin user used when dialling as self

luci1900:
## Description
This ensures JIMM always dials Juju controllers as a Juju external user, whether dialling on behalf of a user that in turn dialled JIMM or not.


kian99 (approved):

alesstimec (approved):
LGTM, but let's wait for APAC team feedback on the Juju PR before landing this.

PR #1946, opened 2026-04-15
fix(vault-dockerfile): fix vault image

SimoneDutto:
# Description

Remove installation of bash/jq.

I think it was a left-over to do some initialization, but it's not used anymore and it was returning the error for the v2 image.
```
[2/4] RUN apk add --no-cache bash jq:
0.471 ERROR: Unable to open log: Permission denied
```
Plus, pin vault to `1.21` for the time being.

# QA
If the tests pass it should be ok.

### DIFF
--- a/local/vault/Dockerfile
+++ b/local/vault/Dockerfile
@@ -1,4 +1,4 @@
-FROM hashicorp/vault:latest
+FROM hashicorp/vault:1.21
 
 # Add our policy and entrypoint
 COPY policy.hcl                   /vault/policy.hcl
--- a/local/vault/Dockerfile
+++ b/local/vault/Dockerfile
@@ -1,9 +1,5 @@
 FROM hashicorp/vault:latest
 
-# Add jq to make scripting the calls a bit easier
-# ref: https://stedolan.github.io/jq/
-RUN apk add --no-cache bash jq
-
 # Add our policy and entrypoint
 COPY policy.hcl                   /vault/policy.hcl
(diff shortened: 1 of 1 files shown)


local/vault/Dockerfile
```diff
@@ -1,8 +1,4 @@
-FROM hashicorp/vault:latest
-
-# Add jq to make scripting the calls a bit easier
-# ref: https://stedolan.github.io/jq/
-RUN apk add --no-cache bash jq
+FROM hashicorp/vault:1
```
kian99: I didn't see this tag at https://hub.docker.com/r/hashicorp/vault/tags and,
```
docker pull hashicorp/vault:1
Error response from daemon: manifest for hashicorp/vault:1 not found: manifest unknown: manifest unknown
```

ale8k (approved):

local/vault/Dockerfile
SimoneDutto: yeah, it is 1.21

SimoneDutto: they don't track the major 1, sadly

kian99 (approved):

local/vault/Dockerfile
ale8k: Why not 2.0.0? Does it have some breaking changes? Nvm see Huw's message

SimoneDutto: because the vault charm doesn't even have a v2 track, so it makes sense to continue testing and devving with v1

PR #1947, opened 2026-04-15
chore: ran go fix ./...

kian99:
## Description
Ran `go fix ./...`, see https://go.dev/blog/gofix on how the new functionality in Go 1.26 works, c/p from there,
> Go fix uses a suite of algorithms to identify opportunities to improve your code, often by taking advantage of more modern features of the language and library.

Most of the changes simply change `interface{}` -> `any`, but others are more elaborate and cool to see.

It's also worth noting this related blog post https://go.dev/blog/inliner on the use of `//go:fix inline` which we can use to deprecate a function in favor of another and use `go fix` to make the change, preserving behavior due to re-ordered arguments and other subtle behaviors.

### DIFF
--- a/internal/jimmjwx/jwks_test.go
+++ b/internal/jimmjwx/jwks_test.go
@@ -47,8 +47,7 @@ func TestGenerateJWKS(t *testing.T) {
 // As such, it will retry 60 times on a 500ms basis.
 func TestStartJWKSRotatorWithNoJWKSInTheStore(t *testing.T) {
 	c := qt.New(t)
-	ctx, cancelCtx := context.WithCancel(context.Background())
-	defer cancelCtx()
+	ctx := t.Context()
 
 	store := newStore(c)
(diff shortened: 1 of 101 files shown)


internal/jimmjwx/jwks_test.go
```diff
@@ -47,8 +47,7 @@ func TestGenerateJWKS(t *testing.T) {
 // As such, it will retry 60 times on a 500ms basis.
 func TestStartJWKSRotatorWithNoJWKSInTheStore(t *testing.T) {
 	c := qt.New(t)
-	ctx, cancelCtx := context.WithCancel(context.Background())
-	defer cancelCtx()
+	ctx := t.Context()
```
luci1900, line 50: Huh, it does cancel by itself https://pkg.go.dev/testing#T.Context

luci1900 (approved):
I like the generics, `any` and `wg.Go`. The rest I don't mind.

PR #1950, opened 2026-04-16
fix: switch to external admin user

luci1900:
## Description
Change the user set as subject in the JWT when dialling to a Juju controller as self/admin.

Previously it was `admin`, now it is `jaas-<jimm-controller-uuid>@external`.

This ensures always we always dial as a Juju external user when using a JWT.

Improved version of https://github.com/canonical/jimm/pull/1945

## Engineering checklist
- [ ] Documentation updated
- [x] Covered by unit tests

### DIFF
--- a/internal/jujuclient/dial.go
+++ b/internal/jujuclient/dial.go
@@ -36,10 +36,6 @@ import (
 	jimmversion "github.com/canonical/jimm/v3/version"
 )
 
-const (
-	adminUser = "admin"
-)
-
 // A ControllerCredentialsStore is a store for controller credentials.
 type ControllerCredentialsStore interface {
 	// GetControllerCredentials retrieves the credentials for the given controller from a vault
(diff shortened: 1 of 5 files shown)


internal/jujuclient/dial.go
```diff
@@ -145,6 +142,12 @@ func (d *Dialer) Dial(ctx context.Context, ctl *dbmodel.Controller, modelTag nam
 	}, nil
 }
 
+// Generate the admin username, based on the JIMM controller UUID.
+// This is a Juju external user, just like the JIMM users.
+func (d *Dialer) makeAdminUserName() string {
+	return fmt.Sprintf("jaas-%s@external", d.ControllerUUID)
```
ale8k: As it shouldn't change anyway, we can do it in an init function and remove the call to this function multiple times

kian99: Or you could use sync.Once, but it's such a minor thing I wouldn't even worry about optimising it.

I wonder if it would work in the init function since we receive the uuid via an env var, so at program startup we won't have a value just yet.

ale8k: You can pull env from init, so it should work but sync.Once is fine too

alesstimec (approved):

SimoneDutto (approved):
this is just a back port of the feature/juju-4 right?

luci1900:
> this is just a back port of the feature/juju-4 right?

Almost, this also includes the controller UUID in the username for better debugging/audit. We'd have to merge this into 4 to gain the same benefit.

Issue #1942, opened 2026-04-09
Dangling Juju offers in JAAS environments


kunalmohan-work:
Deleting a deployment in JAAS env leaves some stale juju offers behind. When trying to create a new deployment, juju complains that the offer name already exists. If I try to delete the offer, it says that offer does not exists. If I rename the offer, deployment succeeds.

## Details

I've been working on deploying COS to K8s cluster in JAAS environment. The deployment creates juju offers (grafana, loki, tempo, mimir and pyroscope) for other applications to consume to exports the logs and metrics to COS.
Since I'm still figuring out the correct terraform templates, deployment errors out sometimes due to some issues. To avoid issues from stale state, I delete the terraform state from s3 and remove applications from the environment when I want to do a fresh deployment from scratch. What I delete:

- Terraform state in s3 bucket
- Any application data in s3 bucket
- juju applications
- Juju storage
- Juju secrets

I clear everything without the `--force` flag. I use the flag only if its stuck on deleting.

I cross verify from kubectl that all pods and volumes are also deleted.

But somewhere the dangling juju offers persist. When I do a fresh deployment, I get an error that the offer name already exists and I should use a different name. When I try to delete the offer, it says that no offer with this name exists. If I create the same offer with a new name, it succeeds. I've come across this issue twice in the same environment. This issue has persisted across days so it's probably not an issue with sync frequency between JIMM and Juju.

moisesbenzan:
Saw this same issue happening with the mimir charm in a COS deployment. For what is worth, my deployment was deleted and recreated using the juju terraform provider.

luci1900:
We've merged a potential fix for this problem https://github.com/canonical/jimm/pull/1951, it will be included in the next release.

Assuming the dangling offer happens due to a timeout between JIMM and the Juju controller when attempting to delete the offer, we now clean up the offer from JIMM's database when it is being fetched and cannot be found in Juju.

This doesn't fix the root cause, but it should address this and other causes of offers dangling.

luci1900:
We've released v3.3.23, which includes this fix https://github.com/canonical/jimm/releases/tag/v3.3.23

You should be able to use it by refreshing the `3/edge` charm channel.

luci1900:
If you still encounter this issue after updating, please re-open the issue.

kian99:
[...]

PR #1951, opened 2026-04-16
fix: delete dangling juju offer on fetch

luci1900:
## Description
When `GetApplicationOffer` is called, if the Juju controller reports there is no such offer, delete it from JIMM's database.

This is dealing with a situation where a previous call to `DestroyOffer` had managed to call into the Juju controller, but the response timed out or similar. The endpoint then didn't end up deleting the offer from JIMM's database.

This is a speculative fix for https://github.com/canonical/jimm/issues/1942.

## Engineering checklist
- [ ] Documentation updated
- [ ] Covered by unit tests


internal/jimm/juju/applicationoffer.go
```diff
@@ -384,27 +385,24 @@ func (j *JujuManager) GetApplicationOffer(ctx context.Context, user *openfga.Use
 	return offerDetails, nil
 }
 
+func (j *JujuManager) maybeCleanupOffer(ctx context.Context, errFromAPI error, offer *dbmodel.ApplicationOffer) error {
```
ale8k: LGTM, maybe name this cleanupOfferIfNotFound though?

ale8k (commented):
LGTM, was thinking about the poll loop we have, I think what we have here is probably OK because to try do it in the loop you'd have to list all apps and offers right?

internal/jimm/juju/applicationoffer.go
luci1900: I named it after `maybeCleanupModel`

internal/jimm/juju/applicationoffer.go
```diff
@@ -384,27 +385,29 @@ func (j *JujuManager) GetApplicationOffer(ctx context.Context, user *openfga.Use
 	return offerDetails, nil
 }
 
+// cleanupOfferIfDangling checks if the offer has been deleted on the controller.
+// If it has, it deletes the offer from the JIMM database.
+// This deals with timeouts to Juju in the JIMM DestroyOffer facade.
+func (j *JujuManager) cleanupOfferIfDangling(ctx context.Context, errFromAPI error, offer *dbmodel.ApplicationOffer) {
+	// Expect a not found error if the offer is already deleted on the controller.
+	isNotFoundErr := errors.ErrorCode(errFromAPI) == errors.CodeNotFound
```
luci1900: Is this actually the correct check? Do we get an error with this code from the Juju client?

ale8k: So we type the RPC errors here: https://github.com/canonical/jimm/blob/c940cfb44803034cab74f34b2143b97e9c106ab5/internal/rpc/client.go#L205-L211

Because our RPCError has `func (e *Error) ErrorCode() string`

and our ErrorCode func checks for it via:
```go
		if v, ok := err.(interface{ ErrorCode() string }); ok {
			if code := v.ErrorCode(); code != "" {
				return Code(code)
			}
		}
```
Meaning it doesn't have to come from the client, they're quite literally sent in JSON over the wire (not really sure why, but they are lol)

Our GetApplicationOffer method ultimately calls `ApplicationOffers`.

In here: https://github.com/juju/juju/blob/7bd7ebf73be53d9ca68f8487ffe7d9f9f44527db/apiserver/facades/client/applicationoffers/applicationoffers.go#L402-L411 we set not found on a ServerError and the server error type is responsible for setting codes here: https://github.com/juju/juju/blob/7bd7ebf73be53d9ca68f8487ffe7d9f9f44527db/apiserver/errors/errors.go#L158-L167 which just takes first unwrap and switches it into that big switch which should set not found.

This is the part I'm not 100% on  though but I **think** it works this way, when we do a batched error response, and I think in Juju's RPC they extract batched errors into the JSON of the msg. Meaning you should see it on any param type with a params.Error inside if it is populated. But yeah I'm not 100% on how it is ultimately set on the message type sent over the wire. Pretty sure it does work that way though.

Hope this helps understand Juju's crazy RPC error coding haha.

testing/applicationoffers_test.go
```diff
@@ -654,3 +654,31 @@ func TestApplicationOffers(t *testing.T) {
...
+
+	url := "bob@canonical.com/" + model.Name + ".test-offer1"
+
+	// Simulate dangling offer
+	offer := dbmodel.ApplicationOffer{
+		ModelID: model.ID,
+		URL:     url,
+		Name:    "test-offer1",
+	}
+	err := s.JIMM.Database.AddApplicationOffer(context.Background(), &offer)
```
ale8k: I don't think we need this test, the app,ication_offer_test you did I think covers this nicely. Was this added to try and cover the uncertainty around error codes?

internal/jimm/juju/applicationoffer.go
```diff
@@ -413,6 +416,23 @@ func (j *JujuManager) DestroyOffer(ctx context.Context, user *openfga.User, offe
...
+	if err := j.Database.DeleteApplicationOffer(ctx, offer); err != nil {
+		return fmt.Errorf("failed to delete application offer from database: %w", err)
+	}
+	if err := j.OpenFGAClient.RemoveApplicationOffer(
+		ctx,
+		offer.ResourceTag(),
+	); err != nil {
+		zapctx.Error(
+			ctx,
+			"cannot remove application offer",
```
ale8k: from database (just so it is consistent with our error message)

internal/jimm/juju/applicationoffer.go
```diff
@@ -413,6 +416,23 @@ func (j *JujuManager) DestroyOffer(ctx context.Context, user *openfga.User, offe
...
+// offerDelete removes an application offer from the database and OpenFGA.
+func (j *JujuManager) offerDelete(ctx context.Context, offer *dbmodel.ApplicationOffer) error {
+	if err := j.Database.DeleteApplicationOffer(ctx, offer); err != nil {
+		return fmt.Errorf("failed to delete application offer from database: %w", err)
+	}
+	if err := j.OpenFGAClient.RemoveApplicationOffer(
+		ctx,
+		offer.ResourceTag(),
+	); err != nil {
+		zapctx.Error(
```
ale8k, line 423: I'd bubble the error up and log in L376, that way when we can have a really specific log line and when we have some kind of stack tracing, it is clear where it came from and what happened. 

Also it does indeed have a tangible error that another caller could use after all, it just so happens in our case of 

```go
	offerDetails, err := api.GetApplicationOffer(ctx, offerURL)
	if err != nil {
		j.cleanupOfferIfDangling(ctx, err, &offer)
		return nil, err
	}
```

We're not using it.

ale8k (approved):
lgtm! just some minor comments nice work!

internal/jimm/juju/applicationoffer.go
luci1900: So in short, yes. Cheers!

In the meantime I fixed the e2e test, verifying this specific aspect.

internal/jimm/juju/applicationoffer.go
luci1900: No, that would mean changing the error message in delete endpoint. I think we should keep it unchanged.

testing/applicationoffers_test.go
luci1900: I definitely want to test the client machinery does what is expected. It might even change accidentally 3->4.

internal/jimm/juju/applicationoffer.go
luci1900: I re-read the original code, I had missed that *only* the openfga delete is what we're trying to swallow.

I'm leaving the zap log in the helper method, but otherwise returning the error up the chain. The `DestroyOffer` endpoint should behave exactly as before.

testing/applicationoffers_test.go
ale8k: Yeah it can, which is why the conn suite was good. But now we have an integration test and controller test which are doing the same thing. Perhaps we delete the integration test?

internal/jimm/juju/applicationoffer.go
```diff
@@ -373,7 +373,7 @@ func (j *JujuManager) GetApplicationOffer(ctx context.Context, user *openfga.Use
 
 	offerDetails, err := api.GetApplicationOffer(ctx, offerURL)
 	if err != nil {
-		return nil, err
+		return nil, j.cleanupOfferIfDangling(ctx, err, &offer)
```
ale8k: Ty

testing/applicationoffers_test.go
```diff
@@ -654,3 +654,49 @@ func TestApplicationOffers(t *testing.T) {
...
+func TestApplicationOfferDangling(t *testing.T) {
+	c := qt.New(t)
+	s, model := SetupAppOfferTest(c)
+	ctx := context.Background()
+
+	conn := s.Open(c, nil, "bob@canonical.com", nil)
+	defer conn.Close()
+	client := applicationoffers.NewClient(conn)
+
+	url := "bob@canonical.com/" + model.Name + ".test-offer1"
```
ale8k, line 667: Everyone @ external?  As we're setting everyone below. Or alteratnively create a bob.

luci1900: What do you mean?

The actual offer URL doesn't matter except for being able to fetch the offer by it.

testing/applicationoffers_test.go
```diff
@@ -654,3 +654,49 @@ func TestApplicationOffers(t *testing.T) {
...
+	err = s.JIMM.OpenFGAClient.AddModelApplicationOffer(ctx, model.ResourceTag(), offer.ResourceTag())
+	c.Assert(err, qt.IsNil)
+
+	everyoneIdentity := &dbmodel.Identity{Name: ofganames.EveryoneUser}
+	everyoneUser := openfga.NewUser(everyoneIdentity, s.JIMM.OpenFGAClient)
+	err = everyoneUser.SetApplicationOfferAccess(ctx, offer.ResourceTag(), ofganames.ReaderRelation)
+	c.Assert(err, qt.IsNil)
+
+	// Trigger cleanup
+	_, err = client.ApplicationOffer(url)
```
ale8k, line 689: Maybe doc that because our endpoint is synchronous, the subsequent call is safe from races

testing/applicationoffers_test.go
ale8k: I know, just so it reads a bit better, I read bob, and saw everyone is all aha.

internal/jimm/juju/applicationoffer.go
alesstimec: i don't like this approach, where we wrap the returned error in a call to `cleanupOfferIfDangling`.. 
why not do 
```
if errors.ErrorCode(err) == errors.CodeNotFound {
    if derr := j.deleteApplicationOffer(ct, offer); derr != nil {
        zapctx.Error(ctx, "error cleaning up a dangling offer", zap.Error(derr), zap.String(offer.UUID))
    }
    return errors.Codef(errors.CodeNotFound, "application offer not found")
}
return err
```

internal/jimm/juju/applicationoffer.go
```diff
@@ -413,6 +418,20 @@ func (j *JujuManager) DestroyOffer(ctx context.Context, user *openfga.User, offe
 	return nil
 }
 
+// deleteApplicationOffer removes an application offer from the database and OpenFGA.
+func (j *JujuManager) deleteApplicationOffer(ctx context.Context, offer *dbmodel.ApplicationOffer) error {
+	if err := j.Database.DeleteApplicationOffer(ctx, offer); err != nil {
+		return err
```
alesstimec, line 420: perhaps log the error and offer UUID.. that way we can look up these errors in logs to see what is really happening and how often

internal/jimm/juju/applicationoffer.go
alesstimec: /me moans about `maybeCleanupModel`

alesstimec (commented):
A few comments

internal/jimm/juju/applicationoffer.go
luci1900: I was assuming we'd call the same thing in a poller, hence the function signature. I suppose I can inline it now.

testing/applicationoffers_test.go
luci1900: We're setting everyone as reader, as is normal for the endpoint. We aren't setting bob as admin at all, since it's not needed for the test.

internal/jimm/juju/applicationoffer_test.go
```diff
@@ -653,6 +653,111 @@ func TestGetApplicationOffer(t *testing.T) {
...
+		},
+		OwnerIdentityName: u.Name,
+		ControllerID:      controller.ID,
+		CloudRegionID:     cloud.Regions[0].ID,
+		CloudCredentialID: cred.ID,
+	}
+	err = j.Database.AddModel(ctx, &model)
+	c.Assert(err, qt.IsNil)
+
+	offer := dbmodel.ApplicationOffer{
```
alesstimec, line 724: delete this app offer and its access

alesstimec (approved):

PR #1956, opened 2026-04-17
Add errorlint to golangci-lint config

larrasket:
## Description
Add errorlint to golangci-lint config

  Adds the errorlint linter, which flags fmt.Errorf calls that use %v or %s to format an error as these produce opaque strings and break errors.Is/errors.As for callers.

  Also fixes a bunch of missing imports that surfaced once typecheck could actually see the errors.As/errors.Is calls from the previous fixes, plus two actual bugs uncovered in the
  process:

   - vault.go: status code check was dropped when converting a type assertion to errors.As, meaning all Vault response errors during delete were silently swallowed instead of just
  404s
   - main.go: logic was inverted when converting err.(*service.SignalError) — the server was exiting on signal shutdown and staying up on real errors
## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -17,6 +17,7 @@ linters:
     - govet
     - importas
     - ineffassign
+    - errorlint
     - nilerr
     - promlinter
     - reassign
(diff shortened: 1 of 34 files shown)


ale8k:
Ty for this, had a skim... Looks sensible to me. Without going through each in detail, I can't say if the `.As(...)` is right in *all* cases (some may require the stricter `.(*err)`.

Unfortunately we're only source available, and not open source. So we cannot accept this contribution, thank you anyway though.

PR #1957, opened 2026-04-17
internal/db: close sql.Conn after migration

larrasket:
## Description
migrateFromSource acquired a dedicated *sql.Conn via sqlDB.Conn() but never closed it.

Add defer conn.Close() immediately after the connection is acquired.

## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests


ale8k:
As with the previous PR, we cannot merge this (see [here](https://github.com/canonical/jimm/blob/v3/LICENSE)). Thank you for the contribution though.

PR #1959, opened 2026-04-20
feat: add cli commands for controller profiles

ale8k:
## Description
Adds controller profile CLI commands. Discussed with @kian99 this PR is more just boiler plate / presumed implementation details as they aren't fully worked out in the spec yet.

Spec gaps / uncertainties
- The spec does not define the exact CLI UX for CRUD commands so the current implementation uses a YAML file or stdin for add and update.
- The spec says save is replace-by-name, the current implementation exposes separate add-controller-profile and update-controller-profile commands on top of that API behaviour.
- The spec requires version-aware updates - the current implementation autofetches the current version during update if the input omits it. 
- The broader spec includes reusing saved profiles for bootstrap, but the current implementation only adds CRUD management commands. This PR does not include the bootstrap work.

The list above is more for reference when we return to adjust the implementation details as we figure out those details in the spec.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests


cmd/jaas/cmd/controllerprofile.go
```diff
@@ -0,0 +1,390 @@
...
+	"github.com/juju/juju/cmd/modelcmd"
+	"github.com/juju/juju/jujuclient"
+	"sigs.k8s.io/yaml"
+
+	apiparams "github.com/canonical/jimm/v3/pkg/api/params"
+)
+
+const (
+	addControllerProfileDoc = `
+Adds a saved controller profile to JIMM.
```
alesstimec:
```suggestion
Adds a controller profile to JIMM.
```

cmd/jaas/cmd/controllerprofile.go
```diff
@@ -0,0 +1,390 @@
...
+
+The controller profile definition is read from a YAML file or from stdin.
+`
+	addControllerProfileExample = `
+    juju jaas add-controller-profile my-profile --file ./profile.yaml
+    cat profile.yaml | juju jaas add-controller-profile my-profile --file -
```
alesstimec, line 31: do we have to use the `file` flag since it's an expected parameter?

cmd/jaas/cmd/controllerprofile.go
```diff
@@ -0,0 +1,390 @@
...
+// addControllerProfileCommand adds a saved controller profile.
+type addControllerProfileCommand struct {
+	jaasCommandBase
+	out cmd.Output
+
+	file cmd.FileVar
+	name string
+}
+
+func (c *addControllerProfileCommand) Info() *cmd.Info {
```
alesstimec, line 89: missing godocs for exported methods

cmd/jaas/cmd/controllerprofile.go
```diff
@@ -0,0 +1,390 @@
...
+}
+
+func (c *addControllerProfileCommand) SetFlags(f *gnuflag.FlagSet) {
+	c.CommandBase.SetFlags(f)
+	c.out.AddFlags(f, "yaml", map[string]cmd.Formatter{
+		"yaml": cmd.FormatYaml,
+		"json": cmd.FormatJson,
+	})
+	c.file.StdinMarkers = stdinMarkers
+	f.StringVar(&c.file.Path, "file", "", "Specify a file-path for the controller profile, use '-' to read from stdin.")
```
alesstimec, line 107: again, i wouldn't use a flag for this since it's a required parameter

cmd/jaas/cmd/controllerprofile.go
```diff
@@ -0,0 +1,390 @@
...
+		return fmt.Errorf("could not create JIMM client: %w", err)
+	}
+	defer client.Close()
+
+	if req.Version == 0 {
+		current, err := client.GetControllerProfile(&apiparams.GetControllerProfileRequest{Name: req.Name})
+		if err != nil {
+			return err
+		}
+		req.Version = current.Version
```
alesstimec, line 209: what does a version mean for controller profiles? and why are we copying the same controller version as the one that already exists? i'm pretty sure we should be incrementing it.. 

but then again, incrementing the profile version on the client side is wrong: the server should be keeping track of profile version..

cmd/jaas/cmd/controllerprofile.go
```diff
@@ -0,0 +1,390 @@
...
+	out cmd.Output
+
+	name string
+}
+
+func (c *showControllerProfileCommand) Info() *cmd.Info {
+	return jujucmd.Info(&cmd.Info{
+		Name:     "show-controller-profile",
+		Args:     "<name>",
+		Purpose:  "Show a saved controller profile from jimm.",
```
alesstimec: i don't think we should be emphasizing that these controller profiles come "from jimm". And also we should agree once and for all whether it is spelled "JIMM" or "jimm" (i am in favor of JIMM)

cmd/jaas/cmd/controllerprofile.go
```diff
@@ -0,0 +1,390 @@
...
+		Doc:      listControllerProfilesDoc,
+		Examples: listControllerProfilesExample,
+	})
+}
+
+func (c *listControllerProfilesCommand) SetFlags(f *gnuflag.FlagSet) {
+	c.CommandBase.SetFlags(f)
+	c.out.AddFlags(f, "yaml", map[string]cmd.Formatter{
+		"yaml": cmd.FormatYaml,
+		"json": cmd.FormatJson,
```
alesstimec: tabular is also a very nice option when listing things.. which would mean implementing the tabular formatter, but that's relatively easy

alesstimec (dismissed):

cmd/jaas/cmd/controllerprofile.go
ale8k: C/p sorry haha

cmd/jaas/cmd/controllerprofile.go
ale8k: Well... The behaviour isn't entirely defined yet and I wanted to keep it consistent with `registercontroller` which has a `--flag`. wdyt?

cmd/jaas/cmd/controllerprofile.go
ale8k: The server does, we're just fetching what the server says the version is so we can update that version
https://github.com/ale8k/jimm/blob/1cb16f2ad83a6c37b02f91dc19de0cd886d72ab5/internal/db/controllerprofile.go#L76 

You'll also see we provide 0, server saves as 1. It is defined in the spec to work this way too

cmd/jaas/cmd/controllerprofile.go
ale8k: Ah ok sure, I'll remove and yeah me too "JIMM" is better. Sorry I just typed "something" here really.

cmd/jaas/cmd/controllerprofile.go
ale8k: done

cmd/jaas/cmd/controllerprofile.go
```diff
@@ -0,0 +1,439 @@
...
+	if req.Name != "" && req.Name != c.name {
+		return apiparams.SaveControllerProfileRequest{}, fmt.Errorf("provided controller profile name doesn't match, %s != %s", c.name, req.Name)
+	}
+	req.Name = c.name
+	req.CreatedAt = ""
+	req.UpdatedAt = ""
+	// This may look odd, but if a user accidentally puts a version here,
+	// the server will reject it with a profile does not exist error which
+	// would be confusing to users.
+	req.Version = 0
```
SimoneDutto, line 163: how can the user accidentally put a version here?

SimoneDutto (approved):
lgtm

luci1900 (approved):

cmd/jaas/cmd/controllerprofile.go
ale8k: If they set version `	// Version must be provided when updating an existing profile.
	Version uint `json:"version" yaml:"version"`` on the file when creating. 
We should probably use an explicit schema tbf when reading the file that doesn't have version, I was just being lazy here lol. But in our API ^ we autoincrement it so idk. Maybe Kian initially wanted it to be set b user and switched to API, idk.

PR #1960, opened 2026-04-20
ci(e2e): skip tests relying on validating cloud creds before updates

SimoneDutto:
## Description

Skip the tests relying on cloud credentials validation in Juju 4.
In Juju 4 this is not implemented: see https://github.com/juju/juju/issues/22273 and for the time being we can just skip this.

We can't skip it conditionally on the backing controller version because in a scenario with mixed deployment juju 4 and juju 3, we don't know during the test which controller jimm will try to add the controller to.

### DIFF
--- a/testing/cloud_test.go
+++ b/testing/cloud_test.go
@@ -201,6 +201,11 @@ func TestUpdateCloudCredentialsErrors(t *testing.T) {
 }
 
 func TestUpdateCloudCredentialsForce(t *testing.T) {
+	t.Skip(`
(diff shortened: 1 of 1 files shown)


testing/cloud_test.go
```diff
@@ -548,6 +557,10 @@ func TestRevokeCredentialsCheckModels(t *testing.T) {
 }
 
 func TestAddCloudError(t *testing.T) {
```
ale8k, line 559: This shouldn't touch credentials? I think this is wrong?

ale8k (approved):
lgtm bar addclouderror test, prob needs investigation

luci1900 (approved):

PR #1964, opened 2026-04-23
feat(upgradeto): upgradeto call now returns the job id

alesstimec:
## Description

Adds job ID to the response of the UpgradeTo call.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests


luci1900 (approved):

SimoneDutto (approved):
sgtm

ale8k:
We don't want to expose job ids?

SimoneDutto:
> We don't want to expose job ids?

This is needed, so the web team can start a migration and monitor the status of the job via its job id instead of polling model status.
Ultimately is good that we, as the job manager, have the ability of marking a job as successful or not, and giving the job id to them is a way to convey that info.

See: https://github.com/alesstimec/jimm/blob/upgrade-to-return-job-id/internal/jimm/upgrade/upgrade.go#L256

kian99:
> > We don't want to expose job ids?
> 
> This is needed, so the web team can start a migration and monitor the status of the job via its job id instead of polling model status. Ultimately is good that we, as the job manager, have the ability of marking a job as successful or not, and giving the job id to them is a way to convey that info.
> 
> See: https://github.com/alesstimec/jimm/blob/upgrade-to-return-job-id/internal/jimm/upgrade/upgrade.go#L256

Fwiw I was also under the impression that after the latest round of discussions we didn't want to expose job ids and rather put the info under the model info. I know the web team needs something, but it seems that our end of the work isn't complete.

SimoneDutto:
> > > We don't want to expose job ids?
> > 
> > 
> > This is needed, so the web team can start a migration and monitor the status of the job via its job id instead of polling model status. Ultimately is good that we, as the job manager, have the ability of marking a job as successful or not, and giving the job id to them is a way to convey that info.
> > See: https://github.com/alesstimec/jimm/blob/upgrade-to-return-job-id/internal/jimm/upgrade/upgrade.go#L256
> 
> Fwiw I was also under the impression that after the latest round of discussions we didn't want to expose job ids and rather put the info under the model info. I know the web team needs something, but it seems that our end of the work isn't complete.

I think adding this is fine, and it's unblocking the web team on this specific item.
We can then improve w/e we want next cycle, this is just adding an id, and letting them call a facade we already have (JobInfo).

Plus we are already sort of exposing job ids in the list method:
https://github.com/SimoneDutto/jimm/blob/merge-main-10/pkg/api/params/params.go#L867-L878

right?

ale8k:
> > > > We don't want to expose job ids?
> > > 
> > > 
> > > This is needed, so the web team can start a migration and monitor the status of the job via its job id instead of polling model status. Ultimately is good that we, as the job manager, have the ability of marking a job as successful or not, and giving the job id to them is a way to convey that info.
> > > See: https://github.com/alesstimec/jimm/blob/upgrade-to-return-job-id/internal/jimm/upgrade/upgrade.go#L256
> > 
> > 
> > Fwiw I was also under the impression that after the latest round of discussions we didn't want to expose job ids and rather put the info under the model info. I know the web team needs something, but it seems that our end of the work isn't complete.
> 
> I think adding this is fine, and it's unblocking the web team on this specific item. We can then improve w/e we want next cycle, this is just adding an id, and letting them call a facade we already have (JobInfo).
> 
> Plus we are already sort of exposing job ids in the list method: https://github.com/SimoneDutto/jimm/blob/merge-main-10/pkg/api/params/params.go#L867-L878
> 
> right?

Ya, but phase 4 was aiming to remove that (we have a card for it). Idm merging this, but it'll need:

- clearly explaining to web that phase 4 will change this next cycle and job ids won't exist anymore?
- an alternative method to help them

kian99:
> I think adding this is fine, and it's unblocking the web team on this specific item. We can then improve w/e we want next cycle, this is just adding an id, and letting them call a facade we already have (JobInfo).
> 
> Plus we are already sort of exposing job ids in the list method: https://github.com/SimoneDutto/jimm/blob/merge-main-10/pkg/api/params/params.go#L867-L878
> 
> right?

I mostly wonder whether the web team should be implementing things using these APIs if we want to (potentially?) get rid of them in favor of show-model, etc.

SimoneDutto:
> > I think adding this is fine, and it's unblocking the web team on this specific item. We can then improve w/e we want next cycle, this is just adding an id, and letting them call a facade we already have (JobInfo).
> > Plus we are already sort of exposing job ids in the list method: https://github.com/SimoneDutto/jimm/blob/merge-main-10/pkg/api/params/params.go#L867-L878
> > right?
> 
> I mostly wonder whether the web team should be implementing things using these APIs if we want to (potentially?) get rid of them in favor of show-model, etc.

from the discussion I had with @andogq, he is fine with adding this as an interim solution and then we can move away from it no problem.

Atm we are exposing job-ids and it's part of our API. If we want to move away from this paradigm, it would require a breaking change regardless, right?

ale8k:
> > > I think adding this is fine, and it's unblocking the web team on this specific item. We can then improve w/e we want next cycle, this is just adding an id, and letting them call a facade we already have (JobInfo).
> > > Plus we are already sort of exposing job ids in the list method: https://github.com/SimoneDutto/jimm/blob/merge-main-10/pkg/api/params/params.go#L867-L878
> > > right?
> > 
> > 
> > I mostly wonder whether the web team should be implementing things using these APIs if we want to (potentially?) get rid of them in favor of show-model, etc.
> 
> from the discussion I had with @andogq, he is fine with adding this as an interim solution and then we can move away from it no problem.
> 
> Atm we are exposing job-ids and it's part of our API. If we want to move away from this paradigm, it would require a breaking change regardless, right?

It would ya, so again idm. And given they're happy with it as an interim, I think this is ok then.

andogq:
Thanks for getting this merged everyone! An interim solution works well for us now, and we can round out the feature when an alternate API is released in the future!

PR #1966, opened 2026-04-28
fix: integration tests

SimoneDutto:
## Description

Fix integration tests:
- bootstrap: in juju 4 CLI when we add-k8s the storage-class is not set anymore, so k8s can't be bootstrap
- upgrade/migrate: for now we need to skip because we can't test this feature with juju 4 backing controllers. And our test scripts are not build to use the Juju 4 CLI required to interact with JIMM and being able to bootstrap 3 controllers. 
- fly-by: i've notice i could do `jaas add-model` and i was reminded our add-model was a c/p from juju 3.6, so i've c/p from juju 4 since this version of jimm would work with juju 4 client moving forward.

In a separate PR we should work on finding a nicer way to setup JAAS with 3 backing controllers. Maybe using jaas bootstrap.

### DIFF
--- a/cmd/jaas/cmd/addmodel.go
+++ b/cmd/jaas/cmd/addmodel.go
@@ -6,10 +6,10 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"slices"
 	"sort"
 	"strings"
 
-	"github.com/canonical/lxd/shared/logger"
 	"github.com/juju/errors"
 	"github.com/juju/gnuflag"
 	"github.com/juju/juju/api"
@@ -23,7 +23,6 @@ import (
 	coremodel "github.com/juju/juju/core/model"
 	"github.com/juju/juju/core/output"
 	"github.com/juju/juju/environs"
-	"github.com/juju/juju/environs/config"
 	"github.com/juju/juju/rpc/params"
 	"github.com/juju/names/v6"
 
@@ -57,10 +56,10 @@ type addModelCommand struct {
 	jimmAPIFunc  func(api.Connection) JIMMAPI
 	cloudAPIFunc func(api.Connection) AddModelCloudAPI
 
-	modelOwner  string
-	apiRoot     api.Connection
-	jimmClient  JIMMAPI
-	cloudClient AddModelCloudAPI
+	apiRoot      api.Connection
+	jimmClient   JIMMAPI
+	cloudClient  AddModelCloudAPI
+	modelCreator string
 
 	Name             string
 	Owner            string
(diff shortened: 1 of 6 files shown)


luci1900 (approved):

cmd/jaas/cmd/addmodel.go
```diff
@@ -6,10 +6,10 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"slices"
 	"sort"
 	"strings"
 
-	"github.com/canonical/lxd/shared/logger"
 	"github.com/juju/errors"
```
kian99, line 13: Not a big fan of using juju/errors, since that's no longer updated and Juju team no uses github.com/juju/internal/errors.
I know this is copy-pasted from the Juju repo, can you just remove the use of juju/errors?

tests/upgrades/upgrade.sh
```diff
@@ -1,5 +1,10 @@
 #!/bin/bash
 
+echo "Test skipped because Juju 4 doesn't support migration yet, so we could still test the internal migration with JIMM and 3 controllers." \
```
kian99, line 3: I don't want to see this test completely skipped because it's our only e2e test for the automated model upgrade story.

It seems like we need a way to be able to bootstrap Juju 3 and Juju 4 controllers for these tests, or re-think these tests into a different form.

SimoneDutto: i agree! i'm working on re-adding it back in another pr with a little tweak to our setup-controller.sh script

cmd/jaas/cmd/addmodel.go
```diff
@@ -327,11 +322,11 @@ func (c *addModelCommand) getCloudRegion(ctx *cmd.Context) (cloudTag names.Cloud
 		// User specified "cloud/region".
 		cloudName, cloudRegion = c.CloudRegion[:sep], c.CloudRegion[sep+1:]
 		if !names.IsValidCloud(cloudName) {
-			return fail(ctx, fmt.Errorf("invalid cloud name %q", cloudName))
```
SimoneDutto, line 330: @kian99 what about these specific error where we need the code to have the right formatting of the error

kian99: I don't imagine this code is being used to pass information higher up the chain, we're already at the edge in the CLI. It's probably just for formatting the message as "cloud name '' not valid" which we can write out by hand. Wdyt?

SimoneDutto: not a big fan of this approach, we are still using juju/errors in terraform as well, so i don't see much harm in keeping it in this file as well.

But if you want i can remove it as well

kian99: Yeah it's not a big deal so we can leave it.

kian99 (approved):

PR #1967, opened 2026-04-28
ci: add a docker cache because sometimes pulling from docker is super slow

SimoneDutto:
Sometimes the docker layer takes a long time to pull images for JIMM (2 -> 15 minutes)

This cache is a simple addition to make the tests a little faster, now it's basically instant.


.github/workflows/e2e-test.yaml
```diff
@@ -18,6 +18,11 @@ jobs:
       - name: Checkout JIMM repo
         uses: actions/checkout@v4
 
+      - name: Cache Docker images
```
kian99: Should we add [read-only](https://github.com/ScribeMD/docker-cache#read-only) to these usages outside of the cache.yaml workflow?

kian99 (commented):
Seems like a good idea. We could also do this a different way and push our custom test images to a registry and additionally use the Canonical docker registry cache.

.github/workflows/e2e-test.yaml
SimoneDutto: yeah, i was trying to understand how much we can actually shave

SimoneDutto:
> Seems like a good idea. We could also do this a different way and push our custom test images to a registry and additionally use the Canonical docker registry cache.

this approach feels simpler, no? Don't need to interact with other stuff, and it's leveraging github caching

luci1900 (approved):
Nice.

Cache key looks good, should be fine.

.github/workflows/e2e-test.yaml
SimoneDutto: it also helps when you are proposing a pr with some changes to the Dockerfile and you do multiple runs of the tests. I guess it would hurt to leave the save in the workflows as well, wdyt?

.github/workflows/cache.yaml
```diff
@@ -6,6 +6,35 @@ on:
...
+        uses: actions/cache@v4
+        with:
+          path: /tmp/docker-images.tar
+          key: docker-${{ runner.os }}-${{ hashFiles('docker-compose.yaml', 'docker-compose.common.yaml', 'local/*/Dockerfile') }}
+
+      - name: Load Docker images from cache
+        if: steps.docker-cache.outputs.cache-hit == 'true'
+        run: docker load -i /tmp/docker-images.tar
+
+      - name: Pull all Docker Compose images
```
ale8k, line 28: Do we need this? Can't we just build and it'll also pull missing images?

ale8k (approved):
lgtm, idk if we need a separate pull step (it's harmless anyway so up to you)

.github/workflows/e2e-test.yaml
kian99: Just laying out some thoughts.
When we save the cache in one of the workflows that runs on PRs, that cache will only apply to that branch.
If we populate the cache during e.g. e2e-test.yaml, that cache will not contain all the images we use across all jobs (assuming e2e-test.yaml doesn't use all the images).

Looking at https://github.com/canonical/jimm/actions/caches, we see the message "Approaching total cache storage limit (9.99 GB of 10 GB Used)".

Based on all that, I think we should only cache on the v3 branch, and PRs should just use that cache if its a hit, otherwise avoid taking up cache space. That assumes PRs are not frequently changing the images we use, which I think is a fair assumption.

.github/workflows/release-binaries.yaml
```diff
@@ -35,7 +35,7 @@ jobs:
           fetch-depth: 0
 
       - name: Set up Go
-        uses: actions/setup-go@v5
+        uses:  actions/setup-go@v6
         with:
           go-version-file: 'go.mod'
           cache: true
```
kian99: false?

.github/workflows/cache.yaml
```diff
@@ -18,7 +47,7 @@ jobs:
           fetch-depth: 0
 
       - name: Setup Go
-        uses: actions/setup-go@v4
+        uses:   actions/setup-go@v6
         with:
           go-version-file: 'go.mod'
```
kian99, line 54: Maybe set `cache: true` just to be explicit.

.github/workflows/release-binaries.yaml
SimoneDutto: yep

.github/workflows/cache.yaml
SimoneDutto: sure!

kian99 (approved):

PR #1970, opened 2026-04-29
test: renable integration tests by adding the option to bootstrap a juju 3.6 controller

SimoneDutto:
## Description

I've originally disabled this test because it was required to bootstrap a juju 3 controller to perform the upgrade, or migration. And our `setup-controller.sh` script wasn't ready for it.
Now i've made the necessary changes to enable these tests.

### DIFF
--- a/.github/workflows/integration-test.yaml
+++ b/.github/workflows/integration-test.yaml
@@ -64,6 +64,13 @@ jobs:
           go-version-file: 'go.mod'
           cache: false
 
+      - name: Download Juju 3.6 binary
+        run: |
+          snap download juju --channel=3.6/stable --basename=juju-3.6
+          unsquashfs -d /tmp/juju-3.6 juju-3.6.snap
+          sudo install -m 755 /tmp/juju-3.6/bin/juju /usr/local/bin/juju-3.6
+          rm -f juju-3.6.snap juju-3.6.assert
+
       - name: Go vendor to speed up docker build
         run: go mod vendor
 
@@ -85,6 +92,7 @@ jobs:
         env:
           BACKING_CONTROLLER_NAME: ${{ steps.jaas.outputs.backing-controller-name }}
           JIMM_CONTROLLER_NAME: ${{ steps.jaas.outputs.jimm-controller-name }}
+          JUJU_BOOTSTRAP_BINARY: /usr/local/bin/juju-3.6
       
       - name: Stop the application container
         run: docker stop jimm
--- a/.github/workflows/integration-test.yaml
+++ b/.github/workflows/integration-test.yaml
@@ -21,12 +21,12 @@ jobs:
         id: docker-cache
         uses: actions/cache/restore@v5
         with:
-          path: /tmp/docker-images.tar
+          path: docker-images.tar
           key: docker-${{ runner.os }}-${{ hashFiles('docker-compose.yaml', 'docker-compose.common.yaml', 'local/*/Dockerfile') }}
 
       - name: Load Docker images from cache
         if: steps.docker-cache.outputs.cache-hit == 'true'
-        run: docker load -i /tmp/docker-images.tar
+        run: docker load -i docker-images.tar
 
       - name: Set up Docker Compose
         uses: docker/setup-compose-action@v1
--- a/internal/jimm/juju/migrationtarget.go
+++ b/internal/jimm/juju/migrationtarget.go
@@ -210,6 +210,7 @@ func (j *JujuManager) Prechecks(ctx context.Context, user *openfga.User, model M
 	if err != nil {
 		return fmt.Errorf("failed to serialize model description: %w", err)
 	}
+
 	err = api.Prechecks(ctx, jujuparams.MigrationModelInfo{
 		UUID:                   model.UUID,
 		Qualifier:              model.Owner,
@@ -355,6 +356,30 @@ func modifyModelDescription(modelDescription description.Model, userMapping dbmo
 			return fmt.Errorf("no external user mapping found for cloud credential local user %q", modelDescription.Owner().Id())
 		}
 		ownerTag = namesv5.NewUserTag(newOwner)
+	}
+
+	for _, app := range modelDescription.Applications() {
+		for _, offer := range app.Offers() {
+			acl := offer.ACL()
+			for user, access := range acl {
+				if user == ofganames.EveryoneUser {
+					continue
+				}
+				userTag := namesv5.NewUserTag(user)
+				if userTag.IsLocal() {
+					newUser, ok := userMapping[userTag.Id()]
+					if !ok {
+						return fmt.Errorf("no external user mapping found for local user %q with %s access to offer %q", userTag.Id(), access, offer.OfferName())
+					}
+					// Adding a key to a map while iterating over doesn't ensure that the new key will be iterated over.
+					// However, in this case it is safe to modify the ACL map in place, because the keys
+					// we add are for non-local users, which we don't process in the loop, so it doesn't matter
+					// if they are iterated over or not.
+					delete(acl, user)
+					acl[newUser] = access
+				}
+			}
+		}
 	}
 
 	modelDescription.SetCloudCredential(description.CloudCredentialArgs{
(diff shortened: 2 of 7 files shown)


.github/workflows/integration-test.yaml
```diff
@@ -64,6 +64,13 @@ jobs:
           go-version-file: 'go.mod'
           cache: false
 
+      - name: Download Juju 3.6 binary
+        run: |
+          snap download juju --channel=3.6/stable --basename=juju-3.6
+          unsquashfs -d /tmp/juju-3.6 juju-3.6.snap
+          sudo install -m 755 /tmp/juju-3.6/bin/juju /usr/local/bin/juju-3.6
+          rm -f juju-3.6.snap juju-3.6.assert
```
ale8k, line 72: Why not just get the binary the way we do in the bootstrap? Would prefer ungrzipped > unsquashing an fs since we aren't caring for the sqashfs here right

.github/workflows/integration-test.yaml
ale8k, line 15: I'd prefer if we had a matrix here and kept the comment about Juju 4 migrations not working. 

Then we can skip if 4. But at least the matrix is there.

.github/workflows/integration-test.yaml
SimoneDutto: this is easier to get the right base/os version

ale8k: Isn't it just a case of arch as musl,dqlite,etc are bundled? Like we don't rely on any sys libs in Juju (or we do?...)

.github/workflows/integration-test.yaml
SimoneDutto: for now we only have one option, to migrate 3 models to 4 controllers, so it doesn't make sense to have a matrix now

ale8k: It'd be nice to have it so we can keep the comment to track it is more what I meant

kian99:
@SimoneDutto Re-running the integration test today is failing because of infra outage and DDoS, I can merge anyway but I'll wait until next week.

kian99:
Infra is back, re-ran the test and it's failing with,
```
Checking if the token change appears on the sink model
jimm:jimm-test@canonical.com/internal-migration-model-2ox2 -> source-controller:admin/sink-model
Token migration failed, expected 'def' but got 'abc' after 10 tries.
```

tests/upgrades/upgrade.sh
```diff
@@ -49,11 +43,6 @@ model_uuid="$(echo "$model_info" | jq -r ".[\"$UPGRADING_MODEL_NAME\"].\"model-u
 current_model_version="$(echo "$model_info" | jq -r ".[\"$UPGRADING_MODEL_NAME\"].\"agent-version\"")"
 echo "Current model version is $current_model_version"
 
-if [ "$current_model_version" != "$SOURCE_CONTROLLER_VERSION" ]; then
-    echo "Model should be at version $SOURCE_CONTROLLER_VERSION to perform upgrade test, but is at $current_model_version"
-    exit 1
-fi
```
kian99, line 55: Why was this removed?

tests/upgrades/upgrade.sh
```diff
@@ -30,7 +24,7 @@ if [[ "$controller_exists" -gt 0 ]]; then
     echo "Controller $UPGRADE_CONTROLLER already exists, skipping setup."
 else
     echo "Setting up $UPGRADE_CONTROLLER"
-    AGENT_VERSION="$SOURCE_CONTROLLER_VERSION" CONTROLLER_NAME="$UPGRADE_CONTROLLER" local/jimm/setup-controller.sh
+    CONTROLLER_NAME="$UPGRADE_CONTROLLER" local/jimm/setup-controller.sh
```
kian99, line 27: If we're no longer specifying the controller version, what controller version are starting from and upgrading to?

internal/jimm/juju/migrationtarget.go
```diff
@@ -210,6 +210,7 @@ func (j *JujuManager) Prechecks(ctx context.Context, user *openfga.User, model M
 	if err != nil {
 		return fmt.Errorf("failed to serialize model description: %w", err)
 	}
+	zapctx.Error(ctx, string(serializedModel))
```
kian99: dd, debug line

internal/jimm/juju/migrationtarget.go
```diff
@@ -357,6 +358,28 @@ func modifyModelDescription(modelDescription description.Model, userMapping dbmo
...
+						return fmt.Errorf("no external user mapping found for local user %q with %s access to offer %q", userTag.Id(), access, offer.OfferName())
+					}
+					delete(acl, user)
+					acl[newUser] = access
+				}
+			}
+		}
+	}
+
+	zapctx.Error(context.Background(), fmt.Sprintf("%+v", modelDescription.Applications()))
```
kian99: dd, debug line

tests/modelmigration/migrate-providing-model.sh
```diff
@@ -149,16 +156,14 @@ echo
...
 echo "Switching to $JIMM_CONTROLLER_NAME and waiting for the model to appear"
 juju switch "$JIMM_CONTROLLER_NAME"
-# For some reason, wait-for model when the model doesn't exist yet just
-# hangs until you run another Juju CLI command but since we can't do that
-# in CI, we will just loop until the model appears.
 # Loop for up to 5 minutes (30 iterations of 10 seconds each)
 success=0
 for _ in {1..30}; do
-    if juju wait-for model "$PROVIDER_MODEL_NAME" --timeout=10s; then
+    if juju show-model "$PROVIDER_MODEL_NAME" > /dev/null 2>&1; then
```
kian99, line 162: Sending the output to /dev/null might make debugging this a little harder then just printing the output but no big deal.

kian99 (approved):

tests/upgrades/upgrade.sh
SimoneDutto: the check was there before because we were bootstrapping from a lower juju 3 version to latest, now we don't need this check anymore, because we bootstrap from 3 to 4 and it's obvious the versions are not the same

tests/upgrades/upgrade.sh
SimoneDutto: latest 3/stable -> latest in 4/edge

tests/modelmigration/migrate-providing-model.sh
SimoneDutto: i've sent it to null to avoid floading the output of these tests, at the end of the day it's either found or not-found, not very useful to debug

tests/modelmigration/migrate-providing-model.sh
```diff
@@ -195,12 +200,24 @@ echo "Token migration successful, got '$new_token' as expected."
 
 echo
 echo "Ensuring the dummy-sink app is still running with an active state"
-juju wait-for application dummy-sink --timeout=5m
+success=0
+for _ in {1..60}; do
+    app_status=$(juju status -m admin/"$CONSUMER_MODEL_NAME" dummy-sink --format json | jq -r '.applications["dummy-sink"]["application-status"].current')
+    if [[ "$app_status" == "active" ]]; then
+        success=1
```
ale8k, line 207: I wouldn't done success = 0 or != active, and then `exit "$success"` otherwise this reads a bit confusingly setting exit to 1.

ale8k (approved):
lgtm bar exiting

tests/modelmigration/migrate-providing-model.sh
SimoneDutto: runners are struggling today, if i get green CI i'll merge ignoring this comment. Sorry!

internal/jimm/juju/migrationtarget.go
```diff
@@ -357,6 +358,26 @@ func modifyModelDescription(modelDescription description.Model, userMapping dbmo
...
+					continue
+				}
+				userTag := namesv5.NewUserTag(user)
+				if userTag.IsLocal() {
+					newUser, ok := userMapping[userTag.Id()]
+					if !ok {
+						return fmt.Errorf("no external user mapping found for local user %q with %s access to offer %q", userTag.Id(), access, offer.OfferName())
+					}
+					delete(acl, user)
+					acl[newUser] = access
```
alesstimec, line 379: we're modifying the map we're ranging over?

SimoneDutto: yes

alesstimec (approved):

PR #1971, opened 2026-04-29
feat: allow use of controller profiles during bootstrap

kian99:
## Description
This PR does a few things to close out the work on controller profiles.
1. It adds a flag `--profile` to `juju jaas bootstrap` allowing you to specify a profile, this sets the default values for bootstrap. These defaults can be modified by supplying additional flags to `juju jaas bootstrap` and additional config options are merged with those from the profile. The merging logic is a bit of ugly/tedious code, but it seems necessary if we want to allow this capability.
2. It updates the help text for the `juju jaas add-controller-profile` command to describe the schema for a controller profile. Currently there is no way to discern, besides reading the code, the supported fields for creating a controller profile.
3. It updates the format ingested by the `juju jaas add-controller-profile` command so that we don't read fields that are set by the server / command args. E.g. the command is `juju jaas add-controller-profile my-profile --file ./profile.yaml` where the name `my-profile` is a required argument but _could_ also be specified in `profile.yaml` and the code would error if they didn't match. Now the file format ingested is only the relevant fields.
4. It updates our how-to manage controllers doc to add a section on the controller profile feature.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

### DIFF
--- a/cmd/jaas/cmd/bootstrap.go
+++ b/cmd/jaas/cmd/bootstrap.go
@@ -373,6 +373,7 @@ func mergeBootstrapCloud(profile, explicit apiparams.BootstrapCloud) apiparams.B
 	return merged
 }
 
+// mergeBootstrapCloudRegion merges parameters from a profile with those explicitly provided for the cloud region.
 func mergeBootstrapCloudRegion(profile, explicit apiparams.BootstrapCloudRegion) apiparams.BootstrapCloudRegion {
 	merged := profile
 	if explicit.Name != "" {
@@ -425,6 +426,7 @@ func (c *bootstrapCommand) bootstrapOptions(ctxt *cmd.Context) (apiparams.Bootst
 	}, nil
 }
 
+// mergeBootstrapOptions merges parameters from a profile with those explicitly provided for the bootstrap options.
 func mergeBootstrapOptions(profile, explicit apiparams.BootstrapOptions) apiparams.BootstrapOptions {
 	merged := apiparams.BootstrapOptions{
 		BootstrapBase:         profile.BootstrapBase,
@@ -451,8 +453,7 @@ func mergeMaps[M1 ~map[K]V, M2 ~map[K]V, K comparable, V any](base M1, override 
 	if len(base) == 0 && len(override) == 0 {
 		return nil
 	}
-	merged := make(map[K]V, len(base)+len(override))
-	maps.Copy(merged, base)
+	merged := maps.Clone(base)
 	maps.Copy(merged, override)
 	return merged
 }
--- a/cmd/jaas/cmd/bootstrap.go
+++ b/cmd/jaas/cmd/bootstrap.go
@@ -4,6 +4,7 @@ package cmd
 
 import (
 	"fmt"
+	"maps"
 	"strings"
 
 	"github.com/juju/cmd/v3"
@@ -26,6 +27,10 @@ const (
 Requests the JIMM server to bootstrap a Juju controller.
 The controller will be created asychronously on the specificed
 cloud and region.
+
+Saved controller profiles can be applied with --profile. The profile's saved
+cloud definition and bootstrap options are used as defaults, and any explicit
+command line arguments or flags override those saved values.
 
 By default the command will wait for the bootstrap job to complete
 while printing the job logs. Note that the logs will not follow the
@@ -75,6 +80,7 @@ Note that JIMM will internally do the following:
 	bootstrapExamples = `
 	juju [jaas] bootstrap <cloud[/region]> <controller name> <controller version>
 	juju [jaas] bootstrap mycloud/region mycontroller 3.6.8
+	juju [jaas] bootstrap mycloud/region mycontroller 3.6.8 --profile production-aws
 	juju [jaas] bootstrap mycloud/region mycontroller 3.6.8 --config controller-service-type=loadbalancer
 	juju [jaas] bootstrap mycloud/region mycontroller 3.6.8 --bootstrap-base ubuntu@24.04 --bootstrap-constraints mem=8G --constraints arch=amd64
 	juju [jaas] bootstrap mycloud/region mycontroller 3.6.8 --storage-pool name=controller-pool --storage-pool type=ebs --config audit-log-enabled=true
@@ -95,6 +101,7 @@ type bootstrapCommand struct {
 
 	credentialName string
 	detach         bool
+	profileName    string
 	bootstrapBase  string
 	constraints    common.ConstraintsFlag
 	bootstrapCons  common.BootstrapConstraintsFlag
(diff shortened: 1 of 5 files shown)


cmd/jaas/cmd/bootstrap.go
```diff
@@ -27,6 +28,10 @@ Requests the JIMM server to bootstrap a Juju controller.
 The controller will be created asychronously on the specificed
 cloud and region.
 
+Saved controller profiles can be applied with --profile. The profile's saved
```
ale8k, line 31: The wording here could be shorter I think

WB:
```
If you'd wish to use a controller profile for bootstrapping, you may use "--profile". Explicitly provided command line arguments or flags take precedence.  ```

File: cmd/jaas/cmd/bootstrap.go [hunk 2]
```diff
@@ -322,6 +340,56 @@ func cloudToParams(cloudName, regionName string, cloud *jujucloud.Cloud) apipara
...
+		merged.Endpoint = explicit.Endpoint
+	}
+	if explicit.HostCloudRegion != "" {
+		merged.HostCloudRegion = explicit.HostCloudRegion
+	}
+	merged.Region = mergeBootstrapCloudRegion(profile.Region, explicit.Region)
+	return merged
+}
+
+func mergeBootstrapCloudRegion(profile, explicit apiparams.BootstrapCloudRegion) apiparams.BootstrapCloudRegion {
```
ale8k, line 377: We go doc'd mergeBootstrapCloud but not this one

cmd/jaas/cmd/bootstrap.go
```diff
@@ -322,6 +340,56 @@ func cloudToParams(cloudName, regionName string, cloud *jujucloud.Cloud) apipara
 	return paramsCloud
 }
 
+// mergeBootstrapCloud merges parameters from a profile with those explicitly provided.
+// All fields are copied to avoid re-using the underlying data structures from the profile.
+func mergeBootstrapCloud(profile, explicit apiparams.BootstrapCloud) apiparams.BootstrapCloud {
+	merged := profile
+	if explicit.Name != "" {
```
ale8k, line 347: Shame we gotta do this, was thinking we could convert profile and explicit into maps, then loop over but the slices part kinda makes it ugly.

cmd/jaas/cmd/bootstrap.go
```diff
@@ -357,6 +425,49 @@ func (c *bootstrapCommand) bootstrapOptions(ctxt *cmd.Context) (apiparams.Bootst
 	}, nil
 }
 
+func mergeBootstrapOptions(profile, explicit apiparams.BootstrapOptions) apiparams.BootstrapOptions {
```
ale8k, line 430: Same for docco here

cmd/jaas/cmd/bootstrap.go
```diff
@@ -357,6 +425,49 @@ func (c *bootstrapCommand) bootstrapOptions(ctxt *cmd.Context) (apiparams.Bootst
...
+	return merged
+}
+
+// mergeMaps merges two maps of the same type.
+// See maps.Copy where the generic type signatures were lifted from.
+func mergeMaps[M1 ~map[K]V, M2 ~map[K]V, K comparable, V any](base M1, override M2) map[K]V {
+	if len(base) == 0 && len(override) == 0 {
+		return nil
+	}
+	merged := make(map[K]V, len(base)+len(override))
```
ale8k: Couldn't we simplify this to:
```go
merged := maps.Clone(base)
maps.Copy(merged, override)
return merged
```

cmd/jaas/cmd/bootstrap.go
```diff
@@ -357,6 +425,49 @@ func (c *bootstrapCommand) bootstrapOptions(ctxt *cmd.Context) (apiparams.Bootst
...
+	maps.Copy(merged, base)
+	maps.Copy(merged, override)
+	return merged
+}
+
+func cloneStoragePool(pool *apiparams.BootstrapStoragePool) *apiparams.BootstrapStoragePool {
+	if pool == nil {
+		return nil
+	}
+	return &apiparams.BootstrapStoragePool{
```
ale8k, line 465: We can just dereference here right? 
```go
	clone := *pool
	clone.Attributes = maps.Clone(pool.Attributes)
	rturn &clone
```

cmd/jaas/cmd/bootstrap_test.go
```diff
@@ -261,6 +263,101 @@ func TestBootstrapApiParams(t *testing.T) {
...
+			BootstrapConstraints: map[string]string{
+				"mem":   "4G",
+				"cores": "2",
+			},
+			ModelDefault: map[string]string{
+				"logging-config": "<root>=INFO",
+			},
+			BootstrapConfig: map[string]string{
+				"bootstrap-timeout":       "30",
+				"controller-service-type": "loadbalancer",
```
ale8k, line 328: Nice, I was wondering if it is worth just updating our bootstrap bash test to use profiles too.
Up to you if you wanna, but meh. I think this is enough.

ale8k (approved):
Nice! Some minor godoc / refactor comments other lgtm.

cmd/jaas/cmd/bootstrap.go
ale8k: Unless we convert the slices into arrays ahead of time? That way you could add/remove keys here and it'd just always work?

cmd/jaas/cmd/bootstrap.go
kian99: Yeah looks good, changed.

cmd/jaas/cmd/bootstrap_test.go
kian99: Agreed that this is enough.

cmd/jaas/cmd/bootstrap.go
kian99: Yeah I added a godoc to that as I thought that one was a bit more confusing.

cmd/jaas/cmd/bootstrap.go
kian99: I think it's fine as is, okay to keep it?

cmd/jaas/cmd/bootstrap.go
kian99: I'll add one here.

cmd/jaas/cmd/bootstrap.go
kian99: Added

cmd/jaas/cmd/bootstrap.go
kian99: I think this would work too, but I'll leave it as is unless there is a strong reason to change.

cmd/jaas/cmd/bootstrap.go
kian99: Leaving it as is for now, maybe Dora also has a say.

luci1900 (approved):

kian99:
@TeodorPt I'm merging this for now, if you have any input on the doc changes, please feel free to review them and I'll address them in a follow up.

PR #1973, opened 2026-05-04
fix: fix cache key in github workflow

SimoneDutto:
# Description

Apparently we can't unpack to /tmp but we need to use github workflow home. 

The reason why just changing the `key` is not enough to fix the cache is:

----------

Cache version is a way to stamp a cache with metadata of the path and the compression tool used while creating the cache. This ensures that the consuming workflow run uniquely matches a cache it can actually decompress and use. For more information, see [Cache Version](https://github.com/actions/cache#cache-version) in the Actions Cache documentation.

-------------

What we were seeing together in call is consistent with this because if you change the path it produces a new "version" of the cache, and it's why i could fetch the key when the path was "/tmp" and not "workspace".

Hopefully this should fix the caching

> i've added the v2, so i can check the caching is actually working properly without worry about the previous ghost versions

### DIFF
--- a/.github/workflows/cache.yaml
+++ b/.github/workflows/cache.yaml
@@ -18,12 +18,12 @@ jobs:
         id: docker-cache
         uses: actions/cache@v5
         with:
-          path: /tmp/docker-images.tar
-          key: docker-${{ runner.os }}-${{ hashFiles('docker-compose.yaml', 'docker-compose.common.yaml', 'local/*/Dockerfile') }}
+          path: docker-images.tar
+          key: docker-v2-${{ runner.os }}-${{ hashFiles('docker-compose.yaml', 'docker-compose.common.yaml', 'local/*/Dockerfile') }}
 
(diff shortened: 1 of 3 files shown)


ale8k (approved):

.github/workflows/cache.yaml
```diff
@@ -18,12 +18,12 @@ jobs:
         id: docker-cache
         uses: actions/cache@v5
         with:
-          path: /tmp/docker-images.tar
-          key: docker-${{ runner.os }}-${{ hashFiles('docker-compose.yaml', 'docker-compose.common.yaml', 'local/*/Dockerfile') }}
+          path: docker-images.tar
```
ale8k, line 21: So this would be root of current workspace dir if I'm understanding it right?

SimoneDutto: it should, yes

kian99 (approved):

PR #1974, opened 2026-05-05
fix: cache

SimoneDutto:
# Description

1.
The cache wasn't populated on the feature branches, upon research the regex was wrong.

2.
the previous implementation of caching was wrong, because it was caching the output of the build
process, and that is not correct for 2 reasons. 1. at built time we could potentially change the
images with something we weren't computing the hash for (e.g.  vault/policy.hcl). 2. we were caching
the jimm image, which shouldn't be cached because everytime we NEED to build it from scratch.

So right now we are caching just the image we pull with docker compose pull, we loose the ability to cache the openfga image because it's built from scracth everytime.


kian99 (dismissed):

luci1900 (approved):
tsk tsk

kian99 (approved):
Just putting these thoughts down after our call on this,
1. Our cache is fragile since any change to the docker-compose.yaml file will invalidate it, even when the images are still valid. A better option for caching images that we only pull might be to use a docker registry cache, then as long as we fully specify the image tags they will come from our fast registry. But that requires us to host such a registry or use the Canonical one, which may require switching to self-hosted runners.
2. For caching images that we build, there is some cool supports in docker buildx for using different cache backends, one of them is for Github actions - see https://docs.docker.com/build/cache/backends/. That might be useful for other bits if we want to look into it more in the future.

.github/workflows/cache.yaml
```diff
@@ -19,17 +19,14 @@ jobs:
         uses: actions/cache@v5
         with:
           path: docker-images.tar
-          key: docker-v2-${{ runner.os }}-${{ hashFiles('docker-compose.yaml', 'docker-compose.common.yaml', 'local/*/Dockerfile') }}
+          key: docker-v2-${{ runner.os }}-${{ hashFiles('docker-compose.yaml', 'docker-compose.common.yaml') }}
```
kian99, line 22: I would suggest changing this to v3 but we are removing something from the list of hashed files anyway.

SimoneDutto: yeah, right, that's why i didn't do the v3

PR #1977, opened 2026-05-05
feat: (almost) always use jwt

kian99:
## Description
There were several instances in JIMM where we would use a controller's username/password to make calls instead of a JWT, usually because the controller didn't previously support the use of a JWT in these instances. Recent versions of Juju support a JWT for _almost_ all use-cases (which exact version I still need to verify).

The 2 places that still use a username/password:
- When destroying a controller, we populate the CLI store with the controller's admin username/password. This can probably be replaced with a JWT but needs to be verified.
- When migrating a model internally, one controller authenticates to another. This is done using basic auth though there is room for JWT based auth as the struct already has a `token` field. The `token` field is for authenticating with JIMM i.e. controller 1 -> JIMM -> controller 2. We would need to grow functionality in Juju to allow one controller to login to another with a JWT minted by JIMM. Most of the primitives for this are there.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests


cmd/jimmsrv/service/service.go
```diff
@@ -641,11 +641,11 @@ func NewServiceFromDependencies(ctx context.Context, deps *ServiceDependencies)
 	s.mux.Handle("/api", websocketCors.Handler(jujuapi.APIHandler(ctx, s.jimm, params)))
 	s.mux.Handle("/model/*", websocketCors.Handler(http.StripPrefix("/model", jujuapi.ModelHandler(ctx, s.jimm, params))))
 	// Uploading local charms (s3 compatible endpoint and legacy HTTP endpoint, respectively)
-	proxyHandler := jimmhttp.NewHTTPProxyHandler(s.jimm.LoginManager, s.jimm.JujuManager)
+	proxyHandler := jimmhttp.NewHTTPProxyHandler(s.jimm.LoginManager, s.jimm.JujuManager, s.jimm.JWTService)
```
kian99: I'm not sure if I'm keen on passing services into the proxy handler. 

The way the shape usually looks for our websocket handlers is,
1. Services (DB, JWT minting, OpenFGA) fed into,
2. Domain (JujuManager, GroupManager, etc) fed into,
3. Handlers (JujuRPC, Rebac API)

Here we subvert that but it may be okay because the proxy handler is just doing proxy logic and it may be overkill / weird to create a proxy domain for what constitutes as infra/plumbing.

internal/jimmhttp/httpproxy_handler.go
```diff
@@ -89,13 +98,24 @@ func (hph *HTTPProxyHandler) ProxyHTTP(w http.ResponseWriter, req *http.Request)
 		return
 	}
 
+	requestHeaders, err := jujuclient.NewControllerAuthorizationHeader(
```
kian99: Here for example we could push the call to `jujuclient.NewControllerAuthorizationHeader` into a domain method. Ala,
```go
requestHeaders, err := hph.jujuManager.NewControllerAuthHeader(...)
```
Then we don't need to provide the JWT service to the proxy handler.

luci1900:
Looks surprisingly light.

As long as we figure out which Juju versions this can work with, it's fine by me.

kian99:
@luci1900 After looking back on the earliest version that supports JWT beyond just the websocket handler, I finally unearthed the history.

https://github.com/juju/juju/pull/15322 (March 2023) was the first PR that introduced JWT support in Juju which was 3.2 era work. That only added support for JWTs in the websocket path. Later in https://github.com/juju/juju/pull/15495 (May 2023) there was a large change to refactor auth and that partially introduced JWT support for HTTP handlers, but the JWT auth was just a stub and returned a "not implemented" error. Only in https://github.com/juju/juju/pull/18052 (Sep 2024) did that get implemented. That final change landed in v3.5.5 and subsequently all v3.6 versions.

I wanted to try running our CI against an early v3.6 version of Juju but the latest Snap cannot bootstrap old versions of Juju due to the fix for MongoBleed. I'll have to use an old Juju binary and test locally if we want to take this route, but I'm confident all versions of 3.6 are safe with this change. Wdyt?

luci1900 (approved):
That sounds like it's worth releasing it and seeing what happens, like with the other dial changes.

internal/jimmhttp/httpproxy_handler.go
```diff
@@ -89,13 +104,23 @@ func (hph *HTTPProxyHandler) ProxyHTTP(w http.ResponseWriter, req *http.Request)
...
+	mt := names.NewModelTag(modelUUID)
+	ct := names.NewControllerTag(controllerDetails.ControllerUUID)
+	jwt, err := hph.loginTokenProvider.NewLoginToken(ctx, mt, ct, user)
+	if err != nil {
+		writeError(ctx, w, http.StatusInternalServerError, err, "failed to generate login token")
+		return
+	}
+
+	requestHeaders := make(http.Header)
+	requestHeaders.Set("Authorization", "Bearer "+base64.StdEncoding.EncodeToString(jwt))
```
SimoneDutto, line 116: isn't there a better way of adding the bearer token? interpolaing string seems weird (e.g. very simple to forget the white space)

internal/jujuclient/dial.go
```diff
@@ -312,6 +309,26 @@ func (c *Connection) Context() context.Context {
 	return c.ctx
 }
 
+func (c *Connection) authorizationHeader(modelTag names.ModelTag, extraHeaders http.Header) (http.Header, error) {
```
SimoneDutto, line 312: maybe give it a better name, it's adding the authorizationHeader plus all the others

internal/rpc/httpproxy.go
```diff
@@ -95,12 +93,17 @@ func ProxyHTTP(ctx context.Context, ctl ConnectionDetails, w http.ResponseWriter
...
-			// We just need to set up basic auth here
-			pr.Out.SetBasicAuth(adminUsername, ctl.Password)
+			// while JIMM replaces the client's auth header with controller auth.
+			pr.Out.Header.Del("Authorization")
+			for key, vals := range ctl.RequestHeaders {
+				pr.Out.Header.Del(key)
+				for _, val := range vals {
+					pr.Out.Header.Add(key, val)
+				}
+			}
```
SimoneDutto, line 106: what could we possibly have in ctl.RequestHeaders? other than the Bearer?

SimoneDutto (commented):
a few question, lgtm

internal/jimmhttp/httpproxy_handler.go
kian99: Not in the stdlib - https://github.com/golang/go/issues/70522

internal/jujuclient/dial.go
kian99: Played around with different names like `appendAuthorizationHeader` and `connectionHeader` but I think this is the best I see rn. Very minimal usage too so not too much of an issue.

internal/rpc/httpproxy.go
kian99: Right now it's just auth. Maybe the change was too general but it could be useful in the future. I can revert to just passing the token to this function and it adds the `Authorization` header.

kian99: One other point is that this code doesn't need to know what the header should look like.

internal/jimmhttp/httpproxy_handler.go
luci1900: I've been annoyed at this in the past.

SimoneDutto (approved):
lgtm

PR #1982, opened 2026-05-19
feat(modelupgrader): add ModelUpgrader facade to support model upgrade

SimoneDutto:
## Description
Add ModelUpgrader.

It's quite simple:
- it calls the juju api client method
- the permission is following juju's https://github.com/juju/juju/blob/3.6/apiserver/facades/client/modelupgrader/upgrader.go#L86-L99, so model writer or controller superuser. 

## QA

e2e testing

### DIFF
--- a/internal/jujuapi/modelupgrader_test.go
+++ b/internal/jujuapi/modelupgrader_test.go
@@ -1,0 +1,137 @@
+// Copyright 2026 Canonical.
+
+package jujuapi_test
+
+import (
+	"context"
+	"testing"
+
+	qt "github.com/frankban/quicktest"
+	jujuparams "github.com/juju/juju/rpc/params"
+	"github.com/juju/names/v5"
+	"github.com/juju/version/v2"
+
+	"github.com/canonical/jimm/v3/internal/errors"
+	"github.com/canonical/jimm/v3/internal/jujuapi"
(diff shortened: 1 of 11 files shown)


internal/jujuapi/modelupgrader_test.go
```diff
@@ -0,0 +1,137 @@
...
+	root := newTestControllerRoot(jimm, "alice@canonical.com", false)
+
+	result, err := root.UpgradeModel(ctx, jujuparams.UpgradeModelParams{
+		ModelTag: names.NewModelTag(modelUUID).String(),
+	})
+	c.Assert(errors.ErrorCode(err), qt.Equals, errors.CodeUnauthorized)
+	c.Assert(result, qt.DeepEquals, jujuparams.UpgradeModelResult{})
+}
+
+func TestUpgradeModel_Success(t *testing.T) {
```
ale8k, line 65: Do we need this? We have an integration test doing happy path

ale8k (approved):
lgtm but prob remove the mock happy path

internal/jujuapi/modelupgrader_test.go
SimoneDutto: well, it's just a mock test

kian99 (approved):

PR #1983, opened 2026-05-20
feat: insert controller into db during bootstrap

kian99:
## Description
This change ensures that we keep a record of which controllers are being bootstrapped in the database during bootstrap. We do this by creating a new table called `controller_bootstraps` and inserting a row when we create a controller at the start of the bootstrap operation. This proved a lot more feasible than putting an incomplete controller into the `controllers` table since that table holds many invariants around containing a full-fledged controller. 

Keeping a record of controllers in the bootstrap process will be useful in the future when paired with some other changes as an alternative to our global bootstrap lock and will be useful to list the controllers that are still busy being created.

There is one other fix this PR implements for a bug that was present. While a controller was being created, it was technically possible to make a request to our `AddController` endpoint to add an existing controller with the same name as the one being bootstrapped, "stealing" the name before the bootstrapping controller could complete.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests


internal/jimm/bootstrap/bootstrap.go
```diff
@@ -248,6 +253,40 @@ func (b *BootstrapManager) StartBootstrapJob(ctx context.Context, user *openfga.
...
+	cleanupReservation := true
+	defer func() {
+		if !cleanupReservation {
+			return
+		}
+		cleanupCtx := context.WithoutCancel(ctx)
+		if err := b.store.DeleteControllerBootstrap(cleanupCtx, bootstrapReservation); err != nil && errors.ErrorCode(err) != errors.CodeNotFound {
+			zapctx.Error(cleanupCtx, "failed to remove controller bootstrap reservation", zap.Error(err), zap.String("controller", params.ControllerName))
+		}
+	}()
```
kian99, line 289: Heads up, we can't delete the controller_bootstrap entry at the end of this `StartBootstrapJob()` as this only enques the river job.

internal/jimm/bootstrap/bootstrap.go
```diff
@@ -355,6 +403,22 @@ func (b *BootstrapManager) BootstrapController(
...
+		if errors.ErrorCode(err) == errors.CodeNotFound {
+			return errors.Codef(errors.CodeNotFound, "controller bootstrap %q not found", p.ControllerName)
+		}
+		return fmt.Errorf("failed to get controller bootstrap reservation: %w", err)
+	}
+	defer func() {
+		cleanupCtx := context.WithoutCancel(ctx)
+		if err := b.store.DeleteControllerBootstrap(cleanupCtx, bootstrapReservation); err != nil && errors.ErrorCode(err) != errors.CodeNotFound {
+			zapctx.Error(cleanupCtx, "failed to remove controller bootstrap reservation", zap.Error(err), zap.String("controller", p.ControllerName))
+		}
```
kian99, line 419: We can finally delete the controller bootstrap entry once the actual bootstrap and add logic is complete. Note that we also call `DeleteControllerBootstrap()` from within JIMM's `AddController()` logic if the context contains a job ID, meaning that we are bootstrapping the controller. We avoid the error log here if the error code is "not found".

internal/db/controller_bootstrap.go
```diff
@@ -0,0 +1,117 @@
...
+		return err
+	}
+
+	durationObserver := servermon.DurationObserver(servermon.DBQueryDurationHistogram, op)
+	defer durationObserver()
+	defer servermon.ErrorCounter(servermon.DBQueryErrorCount, &err, op)
+
+	db := d.DB.WithContext(ctx)
+	switch {
+	case bootstrap.ID != 0:
```
luci1900, line 43: This is the main reason I don't like this API. If you populate two struct fields, you won't know what the query is.

It's not new, though.

internal/jimm/bootstrap/bootstrap.go
```diff
@@ -248,6 +253,40 @@ func (b *BootstrapManager) StartBootstrapJob(ctx context.Context, user *openfga.
...
+	}
+
+	bootstrapReservation := &dbmodel.ControllerBootstrap{
+		Name:      params.ControllerName,
+		CloudName: params.Cloud.Name,
+	}
+	if len(params.Cloud.Regions) > 0 {
+		bootstrapReservation.CloudRegion = params.Cloud.Regions[0].Name
+	}
+	if err := b.store.AddControllerBootstrap(ctx, bootstrapReservation); err != nil {
```
luci1900, line 274: If we want the benefit of colliding even if the JIMM process crashed after kicking off the bootstrap, we'd have to commit at some point after this.

luci1900 (approved):

internal/db/controller_bootstrap.go
kian99: Yeah true

internal/jimm/bootstrap/bootstrap.go
kian99: This line is committing. There is no broader transaction at play.

internal/db/controller_bootstrap.go
```diff
@@ -0,0 +1,117 @@
+// Copyright 2025 Canonical.
```
alesstimec: 2026 :)

internal/db/controller_bootstrap_test.go
```diff
@@ -0,0 +1,58 @@
+// Copyright 2025 Canonical.
```
alesstimec: 2026

internal/dbmodel/controller_bootstrap.go
```diff
@@ -0,0 +1,40 @@
+// Copyright 2025 Canonical.
```
alesstimec: 2026

internal/dbmodel/controller_bootstrap_test.go
```diff
@@ -0,0 +1,33 @@
+// Copyright 2025 Canonical.
```
alesstimec: 2026

alesstimec (approved):
LGTM, but we are in 2026 :)

internal/jujuapi/jimm.go
```diff
@@ -289,6 +289,18 @@ func (r *controllerRoot) ListControllers(ctx context.Context) (apiparams.ListCon
...
 	}
+	if r.user.JimmAdmin {
+		bootstraps, err := r.jimm.JujuManager().ListControllerBootstraps(ctx)
+		if err != nil {
+			return apiparams.ListControllersResponse{}, err
+		}
+		for _, bootstrap := range bootstraps {
+			controllersInfo = append(controllersInfo, bootstrap.ToAPIControllerInfo())
+		}
+		slices.SortFunc(controllersInfo, func(a, b apiparams.ControllerInfo) int {
```
SimoneDutto, line 300: why sorting them here?

Sort them out the `if` so it's always sorted? or do they come sorted from `ListControllers`

internal/jimm/juju/controller.go
```diff
@@ -203,6 +227,18 @@ func addControllerTx(ctx context.Context, j *JujuManager, jujuClouds []dbmodel.C
...
+			return errors.Codef(errors.CodeAlreadyExists, "controller %q already exists", ctl.Name)
+		} else if errors.ErrorCode(err) != errors.CodeNotFound {
+			return err
+		}
+		if err := j.Database.GetControllerBootstrap(ctx, &dbmodel.ControllerBootstrap{Name: ctl.Name}); err == nil {
+			return errors.Codef(errors.CodeInProgress, "controller %q is bootstrapping", ctl.Name)
+		} else if errors.ErrorCode(err) != errors.CodeNotFound {
+			return err
+		}
+	}
```
SimoneDutto, line 245: wouldn't it be cleaner to just have a method called `isControllerNameUsed?`. Feels like this code shouldn't be here

internal/dbmodel/controller_bootstrap.go
```diff
@@ -0,0 +1,40 @@
...
+// ControllerBootstrap reserves a controller name while a bootstrap job is in progress.
+type ControllerBootstrap struct {
+	ID        uint `gorm:"primarykey"`
+	CreatedAt time.Time
+	UpdatedAt time.Time
+
+	Name        string `gorm:"not null;uniqueIndex"`
+	CloudName   string
+	CloudRegion string
+	JobID       sql.NullInt64 `gorm:"column:job_id;uniqueIndex"`
```
SimoneDutto, line 24: why do we need the JobID in this line?

kian99:
@luci1900 Apologies for the delayed reply, I forgot to click submit.

internal/jimm/bootstrap/bootstrap.go
```diff
@@ -248,6 +253,40 @@ func (b *BootstrapManager) StartBootstrapJob(ctx context.Context, user *openfga.
...
+	if err := b.store.GetController(ctx, &dbmodel.Controller{Name: params.ControllerName}); err == nil {
+		return 0, errors.Codef(errors.CodeAlreadyExists, "controller %q already exists", params.ControllerName)
+	} else if errors.ErrorCode(err) != errors.CodeNotFound {
+		return 0, fmt.Errorf("failed to check if controller exists: %w", err)
+	}
+	if err := b.store.GetControllerBootstrap(ctx, &dbmodel.ControllerBootstrap{Name: params.ControllerName}); err == nil {
+		return 0, errors.Codef(errors.CodeInProgress, "controller %q is bootstrapping", params.ControllerName)
+	} else if errors.ErrorCode(err) != errors.CodeNotFound {
+		return 0, fmt.Errorf("failed to check if controller bootstrap exists: %w", err)
+	}
```
SimoneDutto, line 265: same comment about `isControllerNameUsed`

internal/jimm/bootstrap/bootstrap.go
kian99: Which, reading this again, may surface an issue where the `AddControllerBootstrap()` call succeeds but the river insertion fails and now the controller name is reserved but no bootstrap is scheduled.

internal/jimm/bootstrap/bootstrap.go
SimoneDutto, line 419: we could potentially delete the bootstrapping controller when we add the other one in the `db` layer, so we can do a proper transaction without having the defer at play.
Wdyt?

internal/jimm/juju/controller.go
```diff
@@ -153,6 +153,30 @@ func (act *addControllerTransactor) setCloudRegionControllerPriorities(cloud dbm
...
+		jobID, ok := bootstrapJobIDFromContext(ctx)
+		if !ok || !bootstrapReservation.JobID.Valid || bootstrapReservation.JobID.Int64 != jobID {
+			return errors.Codef(errors.CodeInProgress, "controller %q is bootstrapping", act.controller.Name)
+		}
+		if err := act.tx.DeleteControllerBootstrap(ctx, &bootstrapReservation); err != nil {
+			return err
+		}
+	case errors.ErrorCode(err) != errors.CodeNotFound:
+		return err
+	}
```
SimoneDutto, line 182: can you add some godoc here? it would be difficult to understand what's going on in a few months

internal/dbmodel/controller_bootstrap.go
kian99: As dicussed used when bootstrap is finished in internal/jimm/juju/controller.go

internal/jimm/bootstrap/bootstrap.go
kian99: As above, I'll leave as is since it's more a style choice.

internal/jimm/bootstrap/bootstrap.go
kian99: As discussed, we also do the deletion in the business logic layer in a tx when we add the controller to JIMM. I initially mentioned we could technically delete this but that was wrong, if the call to `runBootstrap` fals, we need to cleanup the controller_bootstraps table so it's good that this is here.

internal/jimm/juju/controller.go
kian99: More of a style choice for sure, I'll leave this as is as discussed.

internal/jujuapi/jimm.go
kian99: Yeah ListControllers does a DB query with `Order("name asc")`.

PR #1985, opened 2026-05-21
feat: add support for description v11

kian99:
## Description
This PR adds support in JIMM for description v11, introduced in Juju 3.6.23. Without this change our integration test that imports a model from a standalone Juju controller into JIMM fails.

I've updated the main test to make it easier to add a test case for a new version as the current test was hard-coding test logic for v9 and v10.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

### DIFF
--- a/internal/description/description.go
+++ b/internal/description/description.go
@@ -3,12 +3,20 @@
 // Package description provides version-agnostic wrappers around major versions
 // of `github.com/juju/description` which hold Juju model descriptions, to
 // facilitate migrations between different Juju versions.
+// When deserializing, the wrapper will select the appropriate description version
+// based on the target controller version being migrated to. When serializing,
+// the wrapper will use the same description version as was used for deserialization.
+//
+// When updating this package to accomodate a new major version of `github.com/juju/description`,
+// a new wrapper struct should be added that implements the Model interface, and the
+// Deserialize and migrationDescriptionVersion functions should be updated to support the new version.
 package description
 
 import (
 	"fmt"
 
 	descriptionv10 "github.com/juju/description/v10"
+	descriptionv11 "github.com/juju/description/v11"
 	descriptionv9 "github.com/juju/description/v9"
 	"github.com/juju/juju/environs/config"
 	"github.com/juju/names/v5"
@@ -17,7 +25,7 @@ import (
 	"github.com/canonical/jimm/v3/internal/errors"
 )
 
-const latestDescriptionVersion = 10
+const latestDescriptionVersion = 11
 
 // CloudCredentialArgs holds the arguments needed to set a cloud credential.
 type CloudCredentialArgs struct {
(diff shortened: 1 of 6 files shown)


internal/description/v11.go
```diff
@@ -0,0 +1,99 @@
...
+	app descriptionv11.Application
+}
+
+// Name returns the application name.
+func (a *applicationV11) Name() string {
+	return a.app.Name()
+}
+
+// Offers returns the application offers.
+func (a *applicationV11) Offers() []ApplicationOffer {
```
SimoneDutto, line 92: i've now realised i've landed my fix for ACL in the feature branch for juju 4

kian99: That ACL issue only affected Juju 4 controllers I believe, so no biggie.

SimoneDutto: yeah yeah, just one to merge the feature branch into v3, so we can finally develop on a single branch

ale8k (approved):

internal/description/description.go
```diff
@@ -17,7 +25,12 @@ import (
 	"github.com/canonical/jimm/v3/internal/errors"
 )
 
-const latestDescriptionVersion = 10
+const latestDescriptionVersion = 11
+
+// latestDescriptionTargetVersion is intentionally higher than any released
+// controller version so migrationDescriptionVersion resolves to the newest
+// supported description wrapper.
+var latestDescriptionTargetVersion = version.MustParse("999.0.0")
```
SimoneDutto, line 33: why do you need that? can't you just return an error if it doesn't match?

testing/modelupgrader_test.go
```diff
@@ -33,33 +36,49 @@ func TestUpgradeModel(t *testing.T) {
...
+		lowerVersion = version.Number{
+			Major: 3,
+			Minor: 6,
+			Patch: 20,
+		}
+	case 4:
+		lowerVersion = version.Number{
+			Major: 4,
+			Minor: 0,
+			Patch: 4,
```
SimoneDutto: wasn't 404 burned?

SimoneDutto: i think 406 is an actual release

SimoneDutto (commented):
a few comments and lgtm

testing/modelupgrader_test.go
kian99: The latest snap on 4.0/stable is 4.0.5. There is a 4.0.10 in 4.0/candidate. Not sure what is best here.

internal/description/description.go
kian99: I don't follow your logic? This is just for allowing us to use the `Deserialize(raw []byte, targetControllerVersion version.Number) (Model, error)` function, and always use the latest model description by passing in a "large" controller version.

testing/modelupgrader_test.go
SimoneDutto: but 4.0.4 exists? I think it doesn't and this test would fail, no?

kian99: You're right it doesn't exist - https://streams.canonical.com/juju/tools/agent/
I'll use 4.0.6

internal/description/description.go
SimoneDutto: yeah, but it seems like a workaround.

You use this to call `Deserialize` -> `migrationDescriptionVersion` so it falls back to the default case and use `latestDescriptionVersion`.

Can't you just have `func Deserialize(raw []byte, useLatest bool) (Model, error)`?

Just by navigating the pr i can't see any other use of `Deserialize` with a version for a controller, other than the tests. I could be wrong though

kian99: Deserialise is used elsewhere, see `internal/jimm/juju/migrationtarget.go`

SimoneDutto (approved):

internal/description/description.go
SimoneDutto: got you, i understand the need of this "large" version

PR #1987, opened 2026-05-26
feat: add params for job status

kian99:
## Description
This change adds some new fields to the `ControllerInfo` and `ModelControllerInfo` structs in our API package representing the status of the bootstrap job and upgrade-to jobs respectively.

Note the doc string for `BootstrapJobStatus` states that the status will only be shown if the controller is being bootstrapped, while the doc string for `UpgradeToJobStatus` is not as explicit. It's not currently clear to me when we will stop showing the upgrade-to status.

Fixes [JUJU-9544](https://warthogs.atlassian.net/browse/JUJU-9544)

[JUJU-9544]: https://warthogs.atlassian.net/browse/JUJU-9544?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/api/params/params.go
+++ b/pkg/api/params/params.go
@@ -184,6 +184,10 @@ type ControllerInfo struct {
 	// Status contains the current status of the controller. The status
 	// will either be "available", "deprecated", or "unavailable".
 	Status jujuparams.EntityStatus `json:"status"`
+
+	// BootstrapJobStatus holds the status of a bootstrap job if this
+	// controller is currently being bootstrapped.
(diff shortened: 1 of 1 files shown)


alesstimec (approved):
:shipit:

SimoneDutto (approved):
i can't believe we didn't have these before

kian99:
@alesstimec @SimoneDutto The jaas cmd tests were failing because we were missing yaml tags. Added those.

PR #1988, opened 2026-05-26
feat: allow multiple model-uuids for the upgrade-to facade

SimoneDutto:
## Description

UpgradeTo facede accepts multiple model-uuids. It returns a slice of `Results` to be closer to Juju's way of handling batches

### DIFF
--- a/cmd/jaas/cmd/upgradeto.go
+++ b/cmd/jaas/cmd/upgradeto.go
@@ -88,10 +88,9 @@ func (c *upgradeToCommand) Run(ctxt *cmd.Context) error {
 	}
 	defer client.Close()
 
-	modelTag := names.NewModelTag(c.modelUUID)
 	resp, err := client.UpgradeTo(&apiparams.UpgradeToRequest{
 		TargetControllerName: c.controllerName,
-		ModelTag:             modelTag.String(),
+		ModelUUIDs:           []string{c.modelUUID},
 	})
 	if err != nil {
 		return err
--- a/cmd/jaas/cmd/upgradeto_test.go
+++ b/cmd/jaas/cmd/upgradeto_test.go
@@ -25,7 +25,7 @@ func TestUpgradeTo(t *testing.T) {
 	}
 
 	s.client.EXPECT().UpgradeTo(upgradeToParams).Return(apiparams.UpgradeToResponse{
-		Results: []apiparams.UpgradeToResult{{Success: true}},
+		Results: []apiparams.UpgradeToResult{{}},
 	}, nil)
 	s.client.EXPECT().Close().Return(nil)
 
@@ -37,7 +37,9 @@ func TestUpgradeTo(t *testing.T) {
 	ctx := newTestContext(c)
 	err := upgradeToCmd.Run(ctx)
 	c.Assert(err, qt.IsNil)
-	c.Assert(ctx.Stdout.(*bytes.Buffer).String(), qt.Matches, "results:\n- success: true\n")
+	c.Assert(ctx.Stdout.(*bytes.Buffer).String(), qt.Matches, `results:
+- \{\}
+`)
 }
 
 func TestUpgradeToWithFailureResponse(t *testing.T) {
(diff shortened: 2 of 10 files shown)


testing/jimm_test.go
```diff
@@ -1097,11 +1096,15 @@ func TestUpgradeTo_InvalidModelTag(t *testing.T) {
...
 		TargetControllerName: model.Controller.Name,
 	}
-	_, err := client.UpgradeTo(&req)
-	c.Assert(err, qt.ErrorMatches, `(invalid model tag "invalid-model-tag": )?"invalid-model-tag" is not a valid tag \(bad request\)`)
+	resp, err := client.UpgradeTo(&req)
+	c.Assert(err, qt.IsNil)
+	c.Assert(resp.Results, qt.HasLen, 1)
+	c.Assert(resp.Results[0].Success, qt.IsFalse)
+	c.Assert(resp.Results[0].Error, qt.IsNotNil)
+	c.Assert(resp.Results[0].Error.Message, qt.Equals, `invalid model UUID "invalid-model-uuid"`)
```
kian99: Not something you did, but we should check the error code returned. The code is a part of the API contract.

pkg/api/params/params.go
```diff
@@ -389,16 +389,21 @@ type RemoveControllerProfileRequest struct {
...
 // UpgradeToResponse holds the response for phase 1 of an automated upgrade.
 type UpgradeToResponse struct {
-	Success bool  `json:"success" yaml:"success"`
-	JobID   int64 `json:"job-id" yaml:"job-id"`
+	Results []UpgradeToResult `json:"results" yaml:"results"`
+}
+
+// UpgradeToResult holds the result for a single model in an UpgradeTo request.
+type UpgradeToResult struct {
+	Success bool              `json:"success" yaml:"success"`
```
kian99: I don't think I've commonly seen this pattern in Juju's API, normally if there is no error it's a success.

kian99 (commented):
lgtm, my only item before approving is whether we need the `Success` field.

pkg/api/params/params.go
SimoneDutto: yeah, i can just remove it

testing/jimm_test.go
SimoneDutto: done

pkg/api/params/params.go
SimoneDutto: done

cmd/jaas/cmd/upgradeto_test.go
```diff
@@ -37,7 +37,9 @@ func TestUpgradeTo(t *testing.T) {
 	ctx := newTestContext(c)
 	err := upgradeToCmd.Run(ctx)
 	c.Assert(err, qt.IsNil)
-	c.Assert(ctx.Stdout.(*bytes.Buffer).String(), qt.Matches, "results:\n- success: true\n")
+	c.Assert(ctx.Stdout.(*bytes.Buffer).String(), qt.Matches, `results:
+- \{\}
+`)
```
kian99: It's unfortunate that the output is so poor here. Maybe we can revisit this in the future, but we'd need to print something that isn't just the plain output.

kian99 (approved):

cmd/jaas/cmd/upgradeto_test.go
SimoneDutto: i think we can report something like:
```
upgrade
<model-uuid>: success
<model-uuid2>: error
```

luci1900 (approved):
Sure.

Don't think we can fix the poor UX.

cmd/jaas/cmd/upgradeto.go
```diff
@@ -88,10 +88,9 @@ func (c *upgradeToCommand) Run(ctxt *cmd.Context) error {
 	}
 	defer client.Close()
 
-	modelTag := names.NewModelTag(c.modelUUID)
 	resp, err := client.UpgradeTo(&apiparams.UpgradeToRequest{
```
ale8k, line 91: There's error per each, so we could print the erroneous ones in their own block? 

Also might we worth adding a tabular formatter now that we have multiple (see other commands on how to) and utilise the error field

ale8k: Also if there's any erroneous ones, we should return a none 0 exit code I think

ale8k (approved):
Spot on ty! But I'd add tabular formatter, and only print the erroneous results perhaps.

cmd/jaas/cmd/upgradeto.go
SimoneDutto: yeah, we can do it in a follow-up

PR #1989, opened 2026-05-26
feat: store child jobs on supervisor job output

kian99:
## Description
This PR adds River ["output"](https://riverqueue.com/docs/recorded-output) information to the parent/supervisor upgrade-to job. Specifically the output information holds details on the child job IDs that will allow for reporting on their status later.

Fixes [JUJU-9547](https://warthogs.atlassian.net/browse/JUJU-9547)

## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests

[JUJU-9547]: https://warthogs.atlassian.net/browse/JUJU-9547?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/river/upgradetoworker.go
+++ b/internal/river/upgradetoworker.go
@@ -5,7 +5,9 @@ package river
 import (
 	"context"
 	"database/sql"
+	"encoding/json"
 	"errors"
+	"fmt"
 	"time"
 
 	"github.com/riverqueue/river"
@@ -56,6 +58,10 @@ func (w *upgradeToWorker) Timeout(*river.Job[rivertypes.UpgradeToArgs]) time.Dur
 // of the migration job and the insertion of the upgrade job).
 func (w *upgradeToWorker) Work(ctx context.Context, job *river.Job[rivertypes.UpgradeToArgs]) error {
 	client := river.ClientFromContext[*sql.Tx](ctx)
+	supervisorOutput, err := loadUpgradeToSupervisorOutput(job)
+	if err != nil {
+		return err
+	}
 
 	eventCh, cancel := client.Subscribe(
 		river.EventKindJobCompleted,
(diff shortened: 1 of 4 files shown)


internal/river/upgradetoworker.go
```diff
@@ -105,6 +121,61 @@ func (w *upgradeToWorker) Work(ctx context.Context, job *river.Job[rivertypes.Up
...
+
+func withUpgradeJobID(output rivertypes.UpgradeToSupervisorOutput, jobID int64) rivertypes.UpgradeToSupervisorOutput {
+	upgradeJobID := jobID
+	output.UpgradeJobID = &upgradeJobID
+	output.UpdatedAt = time.Now().UTC()
+	return output
+}
+
+func persistUpgradeToSupervisorOutput(ctx context.Context, client *river.Client[*sql.Tx], jobID int64, output rivertypes.UpgradeToSupervisorOutput) error {
+	if _, err := client.JobUpdate(ctx, jobID, &river.JobUpdateParams{Output: output}); err != nil {
```
ale8k, line 178: Record output this should be no?

internal/river/upgradetoworker.go
```diff
@@ -105,6 +121,61 @@ func (w *upgradeToWorker) Work(ctx context.Context, job *river.Job[rivertypes.Up
 	return nil
 }
 
+func loadUpgradeToSupervisorOutput(job *river.Job[rivertypes.UpgradeToArgs]) (rivertypes.UpgradeToSupervisorOutput, error) {
```
ale8k, line 129: prob needs godoccing why we do this so it's really obvious

ale8k (approved):
Looks mostly right to me, just needs the right call to the client

internal/river/upgradetoworker.go
kian99: [The docs](https://riverqueue.com/docs/recorded-output#how-to-record-job-output) state that recordOutput only save the output once the job is complete. So if the supervisor crashes before both child jobs are completed, it would not have written the child jobs to the DB. That's why this uses JobUpdate and passes the output directly.

internal/river/upgradetoworker.go
kian99: Added

internal/river/upgradetoworker.go
ale8k: Ahh I'm with ya.

luci1900 (approved):
Looks fine, if difficult to follow.

FWIW this is the sort of thing that would be much easier to do without River, just a state machine in a table with a loop. But I won't grumble further about that :)

PR #1990, opened 2026-05-26
feat: add bootstrap metadata to be queried later

SimoneDutto:
## Description

Fixes https://warthogs.atlassian.net/browse/JUJU-9552

Add bootstrap name to river job metadata, so later-on can be queried.

Althougth this seems very simple the godoc for `Metadata` is:
```
// Metadata is a JSON object blob of arbitrary data that will be stored with
// the job. Users should not overwrite or remove anything stored in this
// field by River.
Metadata []byte
```

@ale8k are we sure we are supposed to use this field?

Reading other features:
[ResumableStep](https://pkg.go.dev/github.com/riverqueue/river#ResumableStep) wraps a named unit of work. On the first attempt, all steps run in order. If a step returns an error, River records the last successfully completed step in the job's metadata

Are we sure we are supposed to write in here? are we at risk of compromising other features?

Edit: after a bit of digging through the code, the insertion is fine because no other metadata are written yet.
However if you collide with River metadata they will be overriden, so it would be better to have very specific names for metadata field.
In our case "controller-name" i think it's perfectly fine

### DIFF
--- a/internal/jimm/bootstrap/bootstrap.go
+++ b/internal/jimm/bootstrap/bootstrap.go
@@ -7,6 +7,7 @@ package bootstrap
 import (
 	"context"
 	"database/sql"
+	"encoding/json"
 	"fmt"
 	"net/url"
 	"runtime"
@@ -62,11 +63,15 @@ type Store interface {
 
 // JobQueue defines the method to enqueue a bootstrap/destroy-controller job.
 type JobQueue interface {
-	EnqueueBootstrap(ctx context.Context, args rivertypes.BootstrapArgs) (*rivertype.JobInsertResult, error)
+	EnqueueBootstrap(ctx context.Context, args rivertypes.BootstrapArgs, metadata []byte) (*rivertype.JobInsertResult, error)
 	EnqueueDestroyController(ctx context.Context, args rivertypes.DestroyControllerArgs) (*rivertype.JobInsertResult, error)
 	WaitForJobCompletion(ctx context.Context, jobID int64) (*rivertype.JobRow, error)
 	GetJobInfo(ctx context.Context, jobID int64) (*rivertype.JobRow, error)
 	CancelJob(ctx context.Context, jobID int64) (*rivertype.JobRow, error)
+}
+
+type bootstrapJobMetadata struct {
+	ControllerName string `json:"controller-name"`
 }
 
 // JujuManager defines the juju manager methods required by the job.
@@ -312,7 +317,12 @@ func (b *BootstrapManager) StartBootstrapJob(ctx context.Context, user *openfga.
 			ControllerModelConfig: params.BootstrapOptions.ControllerModelConfig,
 		},
 	}
-	job, err := b.jobQueue.EnqueueBootstrap(ctx, bootstrapArgs)
+	metadata, err := json.Marshal(bootstrapJobMetadata{ControllerName: params.ControllerName})
+	if err != nil {
+		return 0, fmt.Errorf("failed to marshal bootstrap job metadata: %w", err)
+	}
+
+	job, err := b.jobQueue.EnqueueBootstrap(ctx, bootstrapArgs, metadata)
 	if err != nil {
 		return 0, fmt.Errorf("failed to enqueue bootstrap job: %w", err)
 	}
--- a/internal/jimm/bootstrap/bootstrap_test.go
+++ b/internal/jimm/bootstrap/bootstrap_test.go
@@ -4,6 +4,7 @@ package bootstrap_test
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"math"
 	"runtime"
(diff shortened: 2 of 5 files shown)


internal/jimm/bootstrap/bootstrap.go
```diff
@@ -62,13 +63,17 @@ type Store interface {
...
 type JobQueue interface {
-	EnqueueBootstrap(ctx context.Context, args rivertypes.BootstrapArgs) (*rivertype.JobInsertResult, error)
+	EnqueueBootstrap(ctx context.Context, args rivertypes.BootstrapArgs, metadata []byte) (*rivertype.JobInsertResult, error)
 	EnqueueDestroyController(ctx context.Context, args rivertypes.DestroyControllerArgs) (*rivertype.JobInsertResult, error)
 	WaitForJobCompletion(ctx context.Context, jobID int64) (*rivertype.JobRow, error)
 	GetJobInfo(ctx context.Context, jobID int64) (*rivertype.JobRow, error)
 	CancelJob(ctx context.Context, jobID int64) (*rivertype.JobRow, error)
 }
 
+type bootstrapJobMetadata struct {
```
ale8k, line 73: I'd godoc this, and any types we use for metadata to explain their purpose - I know it is obviously the controller name, but saying why we need it might be helpful

ale8k (approved):
Looks spot on to me!

kian99 (commented):
I don't think we'll need the metadata on the bootstrap job. Since we have the `controller_bootstraps` table which is used for reserving a controller name during the bootstrap process, and because it contains the job ID. We can just query that table.

ale8k:
> I don't think we'll need the metadata on the bootstrap job. Since we have the `controller_bootstraps` table which is used for reserving a controller name during the bootstrap process, and because it contains the job ID. We can just query that table.

Yeah this works too. Idm which approach we choose to find the job. Up to you @SimoneDutto. I personally would just do as the spec states, and then later improve but it's up to you.

SimoneDutto:
> > I don't think we'll need the metadata on the bootstrap job. Since we have the `controller_bootstraps` table which is used for reserving a controller name during the bootstrap process, and because it contains the job ID. We can just query that table.
> 
> Yeah this works too. Idm which approach we choose to find the job. Up to you @SimoneDutto. I personally would just do as the spec states, and then later improve but it's up to you.

honestly, i would rather drop this metadata search all together or at least keep it consistent with the other search for model
```
Where(
            "metadata->>'model-uuid' = @model_uuid",
```

Otherwise it just looks confusing

ale8k:
> > > I don't think we'll need the metadata on the bootstrap job. Since we have the `controller_bootstraps` table which is used for reserving a controller name during the bootstrap process, and because it contains the job ID. We can just query that table.
> > 
> > 
> > Yeah this works too. Idm which approach we choose to find the job. Up to you @SimoneDutto. I personally would just do as the spec states, and then later improve but it's up to you.
> 
> honestly, i would rather drop this metadata search all together or at least keep it consistent with the other search for model
> 
> ```
> Where(
>             "metadata->>'model-uuid' = @model_uuid",
> ```
> 
> Otherwise it just looks confusing

Sounds good to me then, happy if we switch it now.

SimoneDutto:
> > > > I don't think we'll need the metadata on the bootstrap job. Since we have the `controller_bootstraps` table which is used for reserving a controller name during the bootstrap process, and because it contains the job ID. We can just query that table.
> > > 
> > > 
> > > Yeah this works too. Idm which approach we choose to find the job. Up to you @SimoneDutto. I personally would just do as the spec states, and then later improve but it's up to you.
> > 
> > 
> > honestly, i would rather drop this metadata search all together or at least keep it consistent with the other search for model
> > ```
> > Where(
> >             "metadata->>'model-uuid' = @model_uuid",
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Otherwise it just looks confusing
> 
> Sounds good to me then, happy if we switch it now.

i'm saying we can keep this querying method for now, because querying per model-uuid without these metadata would require us to store the job-id somewhere else.

So let's keep this approach

SimoneDutto:
@kian99 what do you think?

kian99:
> i'm saying we can keep this querying method for now, because querying per model-uuid without these metadata would require us to store the job-id somewhere else.

The job ID is available on the `controller_bootstraps` table, that's why I'm suggesting you could close this PR.

kian99:
The linked Jira card is incorrect btw, it should be https://warthogs.atlassian.net/browse/JUJU-9551, I left a comment to the same effect.

SimoneDutto:
> > i'm saying we can keep this querying method for now, because querying per model-uuid without these metadata would require us to store the job-id somewhere else.
> 
> The job ID is available on the `controller_bootstraps` table, that's why I'm suggesting you could close this PR.

yeah, but that means that sometimes we query per metadata (to find the models) and sometimes we use a table?
It seems pretty inconstistent to me

kian99 (approved):

kian99:
> yeah, but that means that sometimes we query per metadata (to find the models) and sometimes we use a table?
> It seems pretty inconstistent to me

Yeah fair enough

PR #1991, opened 2026-05-27
test: fix flaky test by allowing multiple calls to `Stop()`

SimoneDutto:
## Description

The reason the test is flaky is that it can happen a watcher is instanciate in the WatchAllModelSummaries for loop before it is cancelled.

# QA

Run the test with `stress` and now it's not flaky anymore.

### DIFF
--- a/internal/jimm/juju/watcher_test.go
+++ b/internal/jimm/juju/watcher_test.go
@@ -273,7 +273,7 @@ func TestWatcherClearsControllerUnavailable(t *testing.T) {
 		<-ctx.Done()
 		return nil, ctx.Err()
 	})
-	mockWatcher.EXPECT().Stop().Return(nil)
+	mockWatcher.EXPECT().Stop().Return(nil).AnyTimes()
 
 	w := juju.Watcher{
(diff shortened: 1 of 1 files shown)


luci1900 (approved):

kian99 (approved):

kian99 (commented):
This might be papering over the problem you stated i.e.
> a watcher is instantiated in the WatchAllModelSummaries for loop before it is cancelled.

There might be a more correct fix in that we shouldn't instantiate a watcher if stop/cancel was called. I looked at this a long while ago. But this seems fine.

SimoneDutto:
> This might be papering over the problem you stated i.e.
> 
> > a watcher is instantiated in the WatchAllModelSummaries for loop before it is cancelled.
> 
> There might be a more correct fix in that we shouldn't instantiate a watcher if stop/cancel was called. I looked at this a long while ago. But this seems fine.

another fix is to avoid instanciating the watcher with a 1ms ticker, in production we use 10 minutes.

But i figured this was a better fix after-all

PR #1992, opened 2026-05-27
feat: modify upgrade-to CLI to support more models, and have a tabular format

SimoneDutto:
## Description
Support multiple models in the upgrade-to cmd.
Output in tabular format by default.

An example: 
```
Model UUID                            Status   Error
93608db4-f1cb-4da5-9926-8233981aef0a  success  
2cb433a6-04eb-4ec4-9567-90426d20a004  failed   failed to run upgrade
```


luci1900 (approved):

cmd/jaas/cmd/upgradeto.go
kian99: Update the help text too please.

cmd/jaas/cmd/upgradeto.go
```diff
@@ -88,18 +90,46 @@ func (c *upgradeToCommand) Run(ctxt *cmd.Context) error {
...
 
-	err = c.out.Write(ctxt, resp)
+	resp, err := client.UpgradeTo(req)
 	if err != nil {
-		return err
+		return fmt.Errorf("upgrade-to request failed: %w", err)
+	}
+	if len(resp.Results) != len(req.ModelUUIDs) {
+		return fmt.Errorf("invalid upgrade-to response: got %d results for %d model UUIDs", len(resp.Results), len(req.ModelUUIDs))
+	}
```
kian99, line 104: Should we print the actual response somewhere? Even if you specify yaml / json, you can't see the actual response in this scenario.

kian99 (approved):

cmd/jaas/cmd/upgradeto.go
SimoneDutto: doesn't it get printed in the cmd?

SimoneDutto: just like the `fmt.Errorf("missing required arguments: controller name and at least one model UUID")` gets printed, right?

kian99: Pretty it wouldn't, how/where would it get printed?

SimoneDutto: or do you mean "printing the response when there is an error?"
But that is not supposed to be a problem either, because if there is an error the response is empty

kian99: The test `TestUpgradeToFailsWhenResponseLengthDoesNotMatchModelUUIDs` seems to confirm that it doesn't get printed.

kian99: I mean printing the response when `len(resp.Results) != len(req.ModelUUIDs)`.

kian99: We could debug log the response or print it. Not sure which is best.

SimoneDutto: image


it does get printed not in `Stdout` but in `Stderr` like so

SimoneDutto: in case they mismatch something has gone very wrong in JIMM, because it should not happen at all. My check is more of a safety net, but it should be impossible.

The errors get printed

kian99: I mean the actual response.

kian99: I suppose in this case there is nothing very interesting in the response to help with debugging so resolving.

PR #1994, opened 2026-05-28
feat: add show-controller api method

kian99:
## Description
This PR implements a new `ShowController` API method. The method allows a user to view the details on a single controller. It follows the same permissions as ListControllers, showing users any controllers they have AddModel access to and only shows bootstrapping controllers to admins.

I'm not particularly happy with the incosistent locations of our permission checks. E.g. in the API layer we check if the user is an admin to determine whether they can see bootstrapping controllers, but we check in the service layer whether the user has AddModel permissions on the controller they are trying to view.

Fixes [JUJU-9552](https://warthogs.atlassian.net/browse/JUJU-9552)

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

[JUJU-9552]: https://warthogs.atlassian.net/browse/JUJU-9552?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/jimm/juju/jimm.go
+++ b/internal/jimm/juju/jimm.go
@@ -55,15 +55,32 @@ func (j *JujuManager) forEachController(ctx context.Context, controllers []dbmod
 }
 
 // ControllerInfo returns info about a controller connected to JIMM.
-func (j *JujuManager) ControllerInfo(ctx context.Context, name string) (*dbmodel.Controller, error) {
-
+func (j *JujuManager) ControllerInfo(ctx context.Context, user *openfga.User, name string) (*dbmodel.Controller, error) {
 	ctl := dbmodel.Controller{
 		Name: name,
 	}
 	if err := j.Database.GetController(ctx, &ctl); err != nil {
 		return nil, err
 	}
+
+	canAddModel, err := user.IsAllowedAddModelToController(ctx, ctl.ResourceTag())
+	if err != nil {
+		zapctx.Error(ctx, "error checking user permissions for controller", zap.String("controller", ctl.Name), zap.Error(err))
+		return nil, errors.New("error checking user permissions for controller")
+	}
+	if !canAddModel {
+		return nil, errors.Codef(errors.CodeUnauthorized, "unauthorized")
+	}
 	return &ctl, nil
+}
+
+// GetControllerBootstrap returns the pending bootstrap reservation for a controller.
+func (j *JujuManager) GetControllerBootstrap(ctx context.Context, name string) (*dbmodel.ControllerBootstrap, error) {
+	bootstrap := dbmodel.ControllerBootstrap{Name: name}
+	if err := j.Database.GetControllerBootstrap(ctx, &bootstrap); err != nil {
+		return nil, err
+	}
+	return &bootstrap, nil
 }
 
 // ListControllerBootstraps returns the currently pending controller bootstraps.
--- a/internal/jujuapi/jimm.go
+++ b/internal/jujuapi/jimm.go
@@ -77,6 +77,7 @@ func init() {
 		upgradeToMethod := rpc.Method(r.UpgradeTo)
 		listUserCloudsMethod := rpc.Method(r.ListUserClouds)
 		modelControllerInfoMethod := rpc.Method(r.ModelControllerInfo)
+		showControllerMethod := rpc.Method(r.ShowController)
 		jobInfoMethod := rpc.Method(r.JobInfo)
 		listJobsMethod := rpc.Method(r.ListJobs)
 		supportedVersionMethd := rpc.Method(r.SupportedJujuVersions)
@@ -99,6 +100,7 @@ func init() {
 		r.AddMethod("JIMM", 4, "RemoveController", removeControllerMethod)
 		r.AddMethod("JIMM", 4, "RevokeAuditLogAccess", revokeAuditLogAccessMethod)
 		r.AddMethod("JIMM", 4, "SetControllerDeprecated", setControllerDeprecatedMethod)
+		r.AddMethod("JIMM", 4, "ShowController", showControllerMethod)
 		r.AddMethod("JIMM", 4, "UpdateMigratedModel", updateMigratedModelMethod)
 
 		// JIMM ReBAC RPC
@@ -313,7 +315,7 @@ func (r *controllerRoot) RemoveController(ctx context.Context, req apiparams.Rem
 		return apiparams.ControllerInfo{}, errors.Codef(errors.CodeUnauthorized, "unauthorized")
 	}
 
-	ctl, err := r.jimm.JujuManager().ControllerInfo(ctx, req.Name)
+	ctl, err := r.jimm.JujuManager().ControllerInfo(ctx, r.user, req.Name)
 	if err != nil {
 		return apiparams.ControllerInfo{}, err
 	}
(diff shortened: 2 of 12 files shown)


luci1900 (approved):
I'm a little concerned by the amount of queue querying we're doing, it's generally an antipattern for a reason.

But I can't think of a better way to do it and it looks fine otherwise.

internal/jujuapi/jimm.go
```diff
@@ -868,6 +870,36 @@ func (r *controllerRoot) ModelControllerInfo(ctx context.Context, req apiparams.
...
 }
 
+// ShowController returns information about a controller or an in-progress bootstrap reservation.
+func (r *controllerRoot) ShowController(ctx context.Context, req apiparams.ShowControllerRequest) (apiparams.ControllerInfo, error) {
+	controller, err := r.jimm.JujuManager().ControllerInfo(ctx, r.user, req.ControllerName)
+	if err != nil && errors.ErrorCode(err) != errors.CodeNotFound {
+		return apiparams.ControllerInfo{}, err
+	}
+	if err == nil {
+		return controller.ToAPIControllerInfo(), nil
```
ale8k, line 880: The responses are inconsistent here right? Maybe I'm wrong but:

1. We'll return controller api info here which is a full info persisted in our controller db.
2. On line 893, we convert the bootstrap controller into the same response but it isn't hydrated? 

Also I was under the impression, we'd return very specific JIMM only details on this show-controller and not dump everything we store about the controller?

kian99: Inconsistent how?
I suggest checking out the code to see it more easily.

Returning here we have the controller in the DB.

Returning later, we convert the minimal details we have for the bootstrapping controller into the API response, hydrate with `response.BootstrapJobStatus = status`.

As for the parameters we return, there is an existing `apiparams.ControllerInfo` struct, there was no mention of changing that.

ale8k: I was doing a quick in browser "." when I commented earlier.

But uhm, when we run bootstrap.ToControllerAPIInfo, we're gonna have

```go
// ToAPIControllerInfo converts a pending bootstrap entry to controller info for list/show APIs.
func (c ControllerBootstrap) ToAPIControllerInfo() apiparams.ControllerInfo {
	ci := apiparams.ControllerInfo{
		Name:        c.Name,
		CloudRegion: c.CloudRegion,
		Status: jujuparams.EntityStatus{
			Status: "bootstrapping",
		},
	}
	if c.CloudName != "" {
		ci.CloudTag = names.NewCloudTag(c.CloudName).String()
	}
	return ci
}
```

Which doesn't include the same details as line 880 will, because line 880 is sourcing it from the controller table? And at no point after 893 do we hydrate it further?

Again maybe I'm wrong but it looks like it isn't returning the same hydrated apiparams.ControllerInfo which is an inconsistent response, dependent on if you are admin.

At the least I would've expected the bootstrap details to be merged with the controller info from db right?

internal/jimm/juju/jimm.go
```diff
@@ -55,17 +55,34 @@ func (j *JujuManager) forEachController(ctx context.Context, controllers []dbmod
 }
 
 // ControllerInfo returns info about a controller connected to JIMM.
-func (j *JujuManager) ControllerInfo(ctx context.Context, name string) (*dbmodel.Controller, error) {
-
+func (j *JujuManager) ControllerInfo(ctx context.Context, user *openfga.User, name string) (*dbmodel.Controller, error) {
```
alesstimec, line 58: interesting that we get controller info by controller's name.. this means that we need to do a DB lookup before we do the actual authz check. 

would getting controller info by its UUID make sense? that way we could do the check without touching the DB (JIMM's DB, we still need to check with OpenFGA)

internal/jimm/juju/jimm.go
```diff
@@ -55,17 +55,34 @@ func (j *JujuManager) forEachController(ctx context.Context, controllers []dbmod
...
+	if !canAddModel {
+		return nil, errors.Codef(errors.CodeUnauthorized, "unauthorized")
+	}
 	return &ctl, nil
 }
 
+// GetControllerBootstrap returns the pending bootstrap reservation for a controller.
+func (j *JujuManager) GetControllerBootstrap(ctx context.Context, name string) (*dbmodel.ControllerBootstrap, error) {
+	bootstrap := dbmodel.ControllerBootstrap{Name: name}
+	if err := j.Database.GetControllerBootstrap(ctx, &bootstrap); err != nil {
```
alesstimec, line 80: no authorization check required here?

internal/jujuapi/jimm.go
```diff
@@ -868,6 +870,36 @@ func (r *controllerRoot) ModelControllerInfo(ctx context.Context, req apiparams.
...
+	}
+	// If the user is not a JIMM admin, return the not found error.
+	// Only JIMM admins can see controllers undergoing bootstrap.
+	if !r.user.JimmAdmin {
+		return apiparams.ControllerInfo{}, err
```
alesstimec, line 885: can't we just return barebones controller info here (and below)?

internal/jujuapi/jimm.go
kian99: Ah but we don't have any other info until after bootstrap. Like the api addresses, CA cert, etc.

internal/jujuapi/jimm.go
kian99: I don't follow? That's what is happening, we get the barebones info from `GetControllerBootstrap`, add the job status details and then return.

internal/jujuapi/jimm.go
SimoneDutto: it reads a bit confusing that we exit early on if err == nil, because it's uncommon in go.

I would have preferred two different paths, so it's easier to read in a single go

SimoneDutto (approved):
besides the method not being of my exact liking it's alright

internal/jimm/juju/jimm.go
kian99: For now the only place this is used is in the ShowController api handler which guards it by requiring the user to be admin. Ultimately we don't have a clear cut place to put the authz checks.

internal/jimm/juju/jimm.go
kian99: Yeah indeed, I was thinking the same.

It would need to start at the ListControllers method, that one does the same thing of fetching the controller from JIMM's DB then checking authz against OpenFGA. And with OpenFGA it's not ideal for doing the kind of query that says, "Give me all the resource of type X that user Foo has access to" iirc.

internal/jujuapi/jimm.go
ale8k: Yeah but we could populate with something like "Currently bootstrapping..." or idk - something of the sort and agreed with Simone

kian99: Open to suggestions on this, but it's like this because of the needed check for NotFound errors. Initially it was,
```go
	controller, err := r.jimm.JujuManager().ControllerInfo(ctx, r.user, req.ControllerName)
	if err == nil {
		return controller.ToAPIControllerInfo(), nil
	}
	if errors.ErrorCode(err) != errors.CodeNotFound {
		return apiparams.ControllerInfo{}, err
	}
```
But I tried to make it a bit more idiomatic though still not perfect.

And regarding,
> Yeah but we could populate with something ...

The api addresses are best left empty no? What fields do you suggest we fill in?

ale8k: Honestly I'm not sure :shrug:. I think, return minimal details for both calls, the idea we had was that we'll have JIMM specific details only in the responses here. cc @alesstimec you was in the call IIRC with Joe, this is right - right?

ale8k (approved):

PR #1995, opened 2026-05-29
feat: add ListModels to list models per users and their upgrade-to status

SimoneDutto:
## Description

The implementation is:
- list models per user
- check if river contains any of the model-uuid in metadata of jobs
- add a simple "in-progress" status for said models, it contains less info than `show-model` but it's a lighter query, and you can always gather more details targeting a specific model

I've added an integration test for this multiple-metadata query to make sure it was actually working as expected.

### DIFF
--- a/internal/jimm/jobs/jobs.go
+++ b/internal/jimm/jobs/jobs.go
@@ -6,7 +6,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"time"
 
 	"github.com/riverqueue/river"
 	"github.com/riverqueue/river/rivertype"
@@ -161,9 +160,9 @@ func (j *JobManager) GetUpgradeToStatusForModel(ctx context.Context, modelUUID s
 }
 
 // ListUpgradeToJobsForModels returns a lightweight per-model status for the most
-// recent relevant upgrade-to supervisor job.
-// The status is fetched from the latest supervisor job for each model, including
-// completed jobs.
+// relevant upgrade-to supervisor job for each requested model.
+// Active jobs are loaded first so in-flight work takes precedence, then the most
+// recent finalized jobs fill in any remaining models.
 func (j *JobManager) ListUpgradeToJobsForModels(ctx context.Context, modelUUIDs []string) (map[string]string, error) {
 	jobsByModelUUID := make(map[string]string, len(modelUUIDs))
 	if len(modelUUIDs) == 0 {
@@ -175,19 +174,19 @@ func (j *JobManager) ListUpgradeToJobsForModels(ctx context.Context, modelUUIDs 
 		requestedModelUUIDs[modelUUID] = struct{}{}
 	}
 
-	jobListResult, err := j.jobQuerier.ListJobs(
+	activeJobListResult, err := j.jobQuerier.ListJobs(
 		ctx,
 		river.NewJobListParams().
 			Kinds(rivertypes.UpgradeToJobKind).
 			First(maxListJobsCount).
-			States((append(activeJobStates, finalizedUpgradeToJobStates...))...),
+			States(activeJobStates...).
+			OrderBy(river.JobListOrderByTime, river.SortOrderDesc),
 	)
 	if err != nil {
 		return nil, err
 	}
 
-	latestJobByModelUUID := make(map[string]*rivertype.JobRow, len(modelUUIDs))
-	for _, job := range jobListResult.Jobs {
+	for _, job := range activeJobListResult.Jobs {
 		if job == nil {
 			continue
 		}
@@ -198,10 +197,36 @@ func (j *JobManager) ListUpgradeToJobsForModels(ctx context.Context, modelUUIDs 
 		if modelUUID == "" {
 			continue
 		}
-		latestJobByModelUUID[modelUUID] = laterJob(latestJobByModelUUID[modelUUID], job)
+		jobsByModelUUID[modelUUID] = UpgradeToModelStatusProgress
 	}
 
-	for modelUUID, job := range latestJobByModelUUID {
+	finalizedJobListResult, err := j.jobQuerier.ListJobs(
+		ctx,
+		river.NewJobListParams().
+			Kinds(rivertypes.UpgradeToJobKind).
+			First(maxListJobsCount).
+			States(finalizedUpgradeToJobStates...).
+			OrderBy(river.JobListOrderByTime, river.SortOrderDesc),
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, job := range finalizedJobListResult.Jobs {
+		if job == nil {
+			continue
+		}
+		modelUUID, err := requestedUpgradeToModelUUID(job, requestedModelUUIDs)
+		if err != nil {
+			return nil, err
+		}
+		if modelUUID == "" {
+			continue
+		}
+		if _, ok := jobsByModelUUID[modelUUID]; ok {
+			continue
+		}
+
 		switch job.State {
 		case rivertype.JobStateCancelled, rivertype.JobStateDiscarded:
 			jobsByModelUUID[modelUUID] = UpgradeToModelStatusError
@@ -229,26 +254,6 @@ func requestedUpgradeToModelUUID(job *rivertype.JobRow, requestedModelUUIDs map[
 		return "", nil
 	}
 	return metadata.ModelUUID, nil
-}
-
-// laterJob returns the later of the two jobs based on their attempted or created time. ù
-// If one of the jobs is nil, it returns the other job.
-func laterJob(current, candidate *rivertype.JobRow) *rivertype.JobRow {
-	if current == nil {
-		return candidate
-	}
-	if jobTime(candidate).After(jobTime(current)) {
-		return candidate
-	}
-	return current
-}
-
-// jobTime returns the attempted time of the job if available, otherwise it returns the created time.
-func jobTime(job *rivertype.JobRow) time.Time {
-	if job.AttemptedAt != nil {
-		return *job.AttemptedAt
-	}
-	return job.CreatedAt
 }
 
 // ListJobs returns a list of jobs based on the provided parameters. It converts the API parameters to the internal river job query parameters and retrieves the job list from the job querier.
--- a/internal/jimm/jobs/jobs.go
+++ b/internal/jimm/jobs/jobs.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"time"
 
 	"github.com/riverqueue/river"
 	"github.com/riverqueue/river/rivertype"
@@ -31,6 +32,21 @@ var finalizedUpgradeToJobStates = []rivertype.JobState{
 	rivertype.JobStateCompleted,
 	rivertype.JobStateDiscarded,
 }
+
+var notCompletedStates = []rivertype.JobState{
+	rivertype.JobStateAvailable,
+	rivertype.JobStatePending,
+	rivertype.JobStateRunning,
+	rivertype.JobStateRetryable,
+	rivertype.JobStateScheduled,
+	rivertype.JobStateCancelled,
+	rivertype.JobStateDiscarded,
+}
+
+const (
+	UpgradeToModelStatusProgress = "progress"
+	UpgradeToModelStatusError    = "error"
+)
 
 // JobQuerier defines the interface for querying and managing jobs in JIMM.
 type JobQuerier interface {
@@ -147,10 +163,11 @@ func (j *JobManager) GetUpgradeToStatusForModel(ctx context.Context, modelUUID s
 	return status, nil
 }
 
-// ListUpgradeToJobsForModels returns the set of model UUIDs that currently have
-// an active upgrade-to supervisor job.
-func (j *JobManager) ListUpgradeToJobsForModels(ctx context.Context, modelUUIDs []string) (map[string]bool, error) {
-	jobsByModelUUID := make(map[string]bool, len(modelUUIDs))
+// ListUpgradeToJobsForModels returns a lightweight per-model status for the most
+// recent relevant upgrade-to supervisor job.
+// The status is fetched from the latest supervisor job for each model.
+func (j *JobManager) ListUpgradeToJobsForModels(ctx context.Context, modelUUIDs []string) (map[string]string, error) {
+	jobsByModelUUID := make(map[string]string, len(modelUUIDs))
 	if len(modelUUIDs) == 0 {
 		return jobsByModelUUID, nil
 	}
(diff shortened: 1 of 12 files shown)


internal/jimm/jobs/jobs.go
```diff
@@ -147,6 +148,41 @@ func (j *JobManager) GetUpgradeToStatusForModel(ctx context.Context, modelUUID s
 	return status, nil
 }
 
+// ListUpgradeToJobsForModels returns the set of model UUIDs that currently have
+// an active upgrade-to supervisor job.
+func (j *JobManager) ListUpgradeToJobsForModels(ctx context.Context, modelUUIDs []string) (map[string]bool, error) {
```
ale8k: s/ListUpgradeToJobsForModels/ListRunningUpgradeToJobs, for models is implied by the signature

internal/jimm/jobs/jobs.go
```diff
@@ -291,6 +327,17 @@ func (j *JobManager) findUpgradeToRootJob(ctx context.Context, modelUUID string)
...
 
+func buildModelUUIDWhereClause(modelUUIDs []string) (string, river.NamedArgs) {
+	clauses := make([]string, 0, len(modelUUIDs))
+	namedArgs := river.NamedArgs{}
+	for i, modelUUID := range modelUUIDs {
+		argName := fmt.Sprintf("model_uuid_%d", i)
+		clauses = append(clauses, fmt.Sprintf("metadata->>'model-uuid' = @%s", argName))
+		namedArgs[argName] = modelUUID
+	}
+	return strings.Join(clauses, " OR "), namedArgs
```
ale8k: Shame river doesn't default to OR / give us a way to set it to OR from AND

internal/jimm/juju/jimm.go
```diff
@@ -425,3 +427,39 @@ func (j *JujuManager) ModelControllerInfo(ctx context.Context, user *openfga.Use
...
+			ControllerUUID: model.Controller.UUID,
+		})
+	}
+
+	slices.SortFunc(info, func(a, b apiparams.ModelControllerInfoListItem) int {
+		if cmp := strings.Compare(a.ModelName, b.ModelName); cmp != 0 {
+			return cmp
+		}
+		return strings.Compare(a.ModelUUID, b.ModelUUID)
+	})
```
kian99: We should sort in the DB query imo. I think we probably already do.

pkg/api/params/params.go
```diff
@@ -839,6 +839,28 @@ type ModelControllerInfo struct {
...
+// model as returned by ListModels.
+type ModelControllerInfoListItem struct {
+	// ModelName is the name of the model.
+	ModelName string `json:"model-name" yaml:"model-name"`
+	// ModelUUID is the UUID of the model.
+	ModelUUID string `json:"model-uuid" yaml:"model-uuid"`
+	// ControllerName is the name of the controller hosting the model.
+	ControllerName string `json:"controller-name" yaml:"controller-name"`
+	// ControllerUUID is the UUID of the controller hosting the model.
+	ControllerUUID string `json:"controller-uuid" yaml:"controller-uuid"`
```
kian99, line 852: I would maybe expect this to embed Juju's struct that is returned when listing models. But at the same time it is possibly preferable to avoid importing Juju's API here as much as possible based on previous discussions.

internal/jimm/juju/jimm.go
```diff
@@ -425,3 +427,39 @@ func (j *JujuManager) ModelControllerInfo(ctx context.Context, user *openfga.Use
...
+	for _, model := range models {
+		info = append(info, apiparams.ModelControllerInfoListItem{
+			ModelName:      model.Name,
+			ModelUUID:      model.UUID.String,
+			ControllerName: model.Controller.Name,
+			ControllerUUID: model.Controller.UUID,
+		})
+	}
+
+	slices.SortFunc(info, func(a, b apiparams.ModelControllerInfoListItem) int {
```
ale8k: Why are we sorting them?

internal/jimm/juju/jimm.go
```diff
@@ -425,3 +427,39 @@ func (j *JujuManager) ModelControllerInfo(ctx context.Context, user *openfga.Use
 		ControllerUUID: model.Controller.UUID,
 	}, nil
 }
+
+// ListModelControllerInfo returns lightweight controller information for all
+// models the user can read.
+func (j *JujuManager) ListModelControllerInfo(ctx context.Context, user *openfga.User) ([]apiparams.ModelControllerInfoListItem, error) {
```
ale8k, line 432: I don't think we need to reimplement list models.

Why not just run ListModels, and if you want it lightweight (I don't think it matters really) adjust it so that it has a new params called "merge" or something, which determines if it merges the controller result in?

Or at the least,extract lines 755->766 internal/jimm/juju/model.go and reuse the func here

internal/jimm/jobs/jobs.go
kian99: Tbh with you, I'm kind of worried about this query. It does a comparison on the JSON field which makes me think it will need to do a table scan and if you have access to e.g. 500 models, this is a lot of "x OR y OR z ....". It might be better served by a "WHERE id IN 1,2,3,....". Wdyt? I doubt this will be a problem at our scale but nonetheless it's worth mentioning.

internal/jimm/juju/jimm.go
SimoneDutto: also, it's useless to sort it by uuid, i'll just remove this

pkg/api/params/params.go
SimoneDutto: i see this api as the plural of `ShowControllerModel`, so i've just used the same fields

internal/jimm/jobs/jobs.go
SimoneDutto: i got an idea, since the number of jobs is always way smaller than the number of models, the smarter thing to do is to fetch all upgrade jobs and filter them in the for loop. So we don't have to mess with river metadata querying, what do you think?

SimoneDutto: changed

internal/jimm/juju/jimm.go
SimoneDutto: removed

internal/jimm/juju/jimm.go
SimoneDutto: i think it's better to have 2 code paths than a flag to switch the implementation.
`ListModels` is doing more than just querying our internal db, it also queries juju and merge the info.

This is way simpler

internal/jimm/jobs/jobs.go
kian99: Yes that's a good idea +1

internal/jimm/jobs/jobs.go
```diff
@@ -147,6 +147,47 @@ func (j *JobManager) GetUpgradeToStatusForModel(ctx context.Context, modelUUID s
...
+	for _, modelUUID := range modelUUIDs {
+		requestedModelUUIDs[modelUUID] = struct{}{}
+	}
+
+	jobListResult, err := j.jobQuerier.ListJobs(
+		ctx,
+		river.NewJobListParams().
+			Kinds(rivertypes.UpgradeToJobKind).
+			First(maxListJobsCount).
+			States(activeJobStates...),
```
kian99: One thing here that relates more to the API / design. If an upgrade-to fails, i.e. all the retries fail, the state will be one of the finalized states, something like `cancelled`. When you run `juju jaas show-model`, you will see the details. But when you run `juju jaas list-models` the `upgrade-to-job-status` status will be empty because we're only listing active jobs. If we wanted we could list all active and failed/cancelled upgrade-to jobs and have 2 states, running/failed.

Maybe the spec speaks to this more.

SimoneDutto: mmm, so just querying everything? dropping this state thing?

SimoneDutto: Now that i think about it though, what if a migration job for a model uuid failed and then we run another one, the result would be confusing non-deterministic right?

Because failed job are persisted for 7 days, right?

kian99: You could do that, or one query of active and another query for finished then instead of a boolean map, maybe some enum that can have 3 states, no migration, actively-migrating, failed-migration.

SimoneDutto: or what if we migrate a model uuid two times in a row, even that would be confusing

kian99: If you do it with 2 queries it will be fairly simple to handle. The active query result takes precedence if the map is already filled. And for the finalized query, you sort by time to return the latest status.

SimoneDutto: i think this all querying the queue is wrong, because the metadata are not unique

kian99: Let's discuss in a few. Grabbing lunch.

SimoneDutto: so we always show the last status? can i migrate the same model at the same time?

kian99: Yeah we can always show the last status. We prevent migration of the same model while one is already in progress.

SimoneDutto: ok, the new implementation sets the status based of the last attempted or scheduled job. Check it out

internal/jimm/jobs/jobs.go
```diff
@@ -147,6 +163,91 @@ func (j *JobManager) GetUpgradeToStatusForModel(ctx context.Context, modelUUID s
...
+	for _, modelUUID := range modelUUIDs {
+		requestedModelUUIDs[modelUUID] = struct{}{}
+	}
+
+	jobListResult, err := j.jobQuerier.ListJobs(
+		ctx,
+		river.NewJobListParams().
+			Kinds(rivertypes.UpgradeToJobKind).
+			First(maxListJobsCount).
+			States(notCompletedStates...),
```
kian99: This is only querying not completed states but presumably we want to see the job details when the job has completed afterwards for some time.

internal/jimm/jobs/jobs.go
```diff
@@ -147,6 +163,91 @@ func (j *JobManager) GetUpgradeToStatusForModel(ctx context.Context, modelUUID s
...
+		}
+		latestJobByModelUUID[modelUUID] = laterJob(latestJobByModelUUID[modelUUID], job)
+	}
+
+	for modelUUID, job := range latestJobByModelUUID {
+		switch job.State {
+		case rivertype.JobStateCancelled, rivertype.JobStateDiscarded:
+			jobsByModelUUID[modelUUID] = UpgradeToModelStatusError
+		case rivertype.JobStateAvailable, rivertype.JobStatePending, rivertype.JobStateRunning, rivertype.JobStateRetryable, rivertype.JobStateScheduled:
+			jobsByModelUUID[modelUUID] = UpgradeToModelStatusProgress
```
kian99, line 236: If we did do 2 queries, one using `activeJobStates` and one using `finalizedUpgradeToJobStates`, you can avoid this extra enumeration of which River state corresponds to active or completed.

kian99 (commented):
Looks fine now. Just 2 comments but potentially we can leave it as is

internal/jimm/jobs/jobs.go
SimoneDutto: i thought that for completed migration there nothing to share in the CLI command, because the model is migrated to the new controller and everything is fine

internal/jimm/jobs/jobs.go
SimoneDutto: In this case, yeah! But, if we add also the "completed" status we would need to do this switch regardless because completed is a "finalized" state.
I don't think it's a big deal though

internal/jimm/jobs/jobs.go
kian99: I'm not sure myself. Probably something to discuss next week. I thought we might want to show the migration status after it was successful, so you can maybe see things like how long it took etc.

For the bootstrap I (maybe wrongly) assumed you don't need to see the bootstrap status after it is done bootstrapping because the goal was just to create the controller. So that code doesn't show the bootstrap status once the job is done.

internal/jimm/jobs/jobs.go
```diff
@@ -32,6 +33,21 @@ var finalizedUpgradeToJobStates = []rivertype.JobState{
...
+	rivertype.JobStatePending,
+	rivertype.JobStateRunning,
+	rivertype.JobStateRetryable,
+	rivertype.JobStateScheduled,
+	rivertype.JobStateCancelled,
+	rivertype.JobStateDiscarded,
+}
+
+const (
+	UpgradeToModelStatusProgress = "progress"
```
alesstimec, line 38: godocs, please

internal/jimm/jobs/jobs.go
```diff
@@ -147,6 +163,91 @@ func (j *JobManager) GetUpgradeToStatusForModel(ctx context.Context, modelUUID s
...
+			jobsByModelUUID[modelUUID] = UpgradeToModelStatusError
+		case rivertype.JobStateAvailable, rivertype.JobStatePending, rivertype.JobStateRunning, rivertype.JobStateRetryable, rivertype.JobStateScheduled:
+			jobsByModelUUID[modelUUID] = UpgradeToModelStatusProgress
+		}
+	}
+
+	return jobsByModelUUID, nil
+}
+
+// requestedUpgradeToModelUUID checks if the job has metadata indicating it is an upgrade-to job for one of the requested ù
```
alesstimec: `ù`??

internal/jimm/jobs/jobs.go
```diff
@@ -147,6 +163,91 @@ func (j *JobManager) GetUpgradeToStatusForModel(ctx context.Context, modelUUID s
...
+	if _, ok := requestedModelUUIDs[metadata.ModelUUID]; !ok {
+		return "", nil
+	}
+	return metadata.ModelUUID, nil
+}
+
+// laterJob returns the later of the two jobs based on their attempted or created time. ù
+// If one of the jobs is nil, it returns the other job.
+func laterJob(current, candidate *rivertype.JobRow) *rivertype.JobRow {
+	if current == nil {
```
alesstimec: what if `candidate == nil`?

internal/jimm/jobs/jobs.go
```diff
@@ -147,6 +163,91 @@ func (j *JobManager) GetUpgradeToStatusForModel(ctx context.Context, modelUUID s
...
+	}
+	if jobTime(candidate).After(jobTime(current)) {
+		return candidate
+	}
+	return current
+}
+
+// jobTime returns the attempted time of the job if available, otherwise it returns the created time.
+func jobTime(job *rivertype.JobRow) time.Time {
+	if job.AttemptedAt != nil {
```
alesstimec: what if `job == nil`?

internal/jimm/jobs/jobs.go
luci1900: There are a few other cases actually.

In general, would we expect to deal with nils in these helper methods, or where the calls happen? It's not entirely clear to me.

internal/jimm/jobs/jobs.go
SimoneDutto: image
in the italian keyboard the ù is dangerously close to the new line, ahah

internal/jimm/jobs/jobs.go
SimoneDutto: those should never be nil, but to simplify the code i've added a:
```go
for _, job := range jobListResult.Jobs {
		if job == nil {
			continue
		}
```

internal/jimm/jobs/jobs.go
SimoneDutto: @kian99 after having a thought, i think it's good to have completed jobs as well.

The status will be shown for 24h (default retention in river for completed job), so it will be clearer the model was recently successfully migrated, ptal

internal/jimm/jobs/jobs.go
```diff
@@ -147,6 +160,97 @@ func (j *JobManager) GetUpgradeToStatusForModel(ctx context.Context, modelUUID s
...
+			continue
+		}
+		modelUUID, err := requestedUpgradeToModelUUID(job, requestedModelUUIDs)
+		if err != nil {
+			return nil, err
+		}
+		if modelUUID == "" {
+			continue
+		}
+		latestJobByModelUUID[modelUUID] = laterJob(latestJobByModelUUID[modelUUID], job)
```
kian99: With 2 queries, I believe you can get rid of some of the helpers like `laterJob` and `jobTime` because you can retrieve the jobs sorted by time and take the first one, if it appears in the map then you can skip it.

kian99 (approved):
Looks good, but do take another look at using 2 queries against River. 1 for jobs in `activeJobStates` and another for jobs in `finalizedUpgradeToJobStates`. 

The fact that we are mapping these states into our API response as plain strings is also problematic because the response is just a plain string when we should really define an enum (for `UpgradeToJobStatus`) and make job states consistent across jobs.

In my head we'd have a field like "active" for a job i.e. busy/done and then a field for the "status" that is success/error/retrying/other/etc. But regardless of whether we do that, I think we need consistency and a better defined set of response types.

internal/jimm/jobs/jobs.go
SimoneDutto: just so you know, we would need to do 3 queries to be extra correct.
Querying for the active jobs by order:
```
// * States `available`, `retryable`, or `scheduled` use `scheduled_at`.
// * State `running` uses `attempted_at`.
// * States `cancelled`, `completed`, or `discarded` use `finalized_at`.
JobListOrderByTime JobListOrderByField = "time"
```

I don't think it's a huge deal, but there might be edge cases where this could bite us.

SimoneDutto: changing it to two queries though

luci1900 (approved):

PR #1998, opened 2026-05-29
Add list models cmd

SimoneDutto:
## Description

Adding the jaas models cmd.

For now the description of the command is very simple, because this is a versitile commands that i guess we will use to show info about models in jaas, happy to add info about upgrade-to status though.

### DIFF
--- a/cmd/jaas/cmd/listmodels.go
+++ b/cmd/jaas/cmd/listmodels.go
@@ -41,6 +41,7 @@ type listModelsCommand struct {
 	out cmd.Output
 }
 
+// Info implements Command.Info.
 func (c *listModelsCommand) Info() *cmd.Info {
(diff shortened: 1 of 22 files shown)


luci1900 (approved):

cmd/jaas/cmd/listmodels.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+// listModelsCommand shows model and controller information for all models the user
+// can access.
+type listModelsCommand struct {
+	jaasCommandBase
+	out cmd.Output
+}
+
+func (c *listModelsCommand) Info() *cmd.Info {
```
alesstimec, line 45: godoc, please

alesstimec (approved):

PR #1999, opened 2026-06-03
feat: add controller details struct

kian99:
## Description
This PR addresses a previous oversight on my part. I added the `BootstrapJobStatus` field to the `ControllerInfo` struct, but we only populate it in calls to `ShowController` and not `ListController` yet both methods return the same type. The newly added `ShowController` struct should use a new type to avoid leaving the `BootstrapJobStatus` field nil.

This PR adds a new type called  `ControllerDetails` that is only returned from `ShowController`. It's best to leave `ControllerInfo` as-is to avoid a breaking change. The newly introduced field (`BootstrapJobStatus`) and the `ShowController` method have not been release yet.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

### DIFF
--- a/internal/dbmodel/controller.go
+++ b/internal/dbmodel/controller.go
@@ -98,9 +98,9 @@ func (c *Controller) SetTag(t names.ControllerTag) {
 	c.UUID = t.Id()
 }
 
-// ToAPIControllerInfo converts a controller entry to a JIMM API
+// ToControllerInfo converts a controller entry to a JIMM API
 // ControllerInfo.
-func (c Controller) ToAPIControllerInfo() apiparams.ControllerInfo {
+func (c Controller) ToControllerInfo() apiparams.ControllerInfo {
 	var ci apiparams.ControllerInfo
 	ci.Name = c.Name
 	ci.UUID = c.UUID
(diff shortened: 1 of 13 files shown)


SimoneDutto (approved):

internal/dbmodel/controller.go
```diff
@@ -132,6 +132,23 @@ func (c Controller) ToAPIControllerInfo() apiparams.ControllerInfo {
 	return ci
 }
 
+// ToAPIShowControllerInfo converts a controller entry to a JIMM API
+// ShowControllerInfo.
```
alesstimec: it seems to me it returns `ControllerDetails`.

internal/dbmodel/controller_bootstrap.go
```diff
@@ -24,7 +24,7 @@ type ControllerBootstrap struct {
 	JobID       sql.NullInt64 `gorm:"column:job_id;uniqueIndex"`
 }
 
-// ToAPIControllerInfo converts a pending bootstrap entry to controller info for list/show APIs.
+// ToAPIControllerInfo converts a pending bootstrap entry to controller info for list-style APIs.
 func (c ControllerBootstrap) ToAPIControllerInfo() apiparams.ControllerInfo {
```
alesstimec: i'd just name this `ToControllerInfo`.. i don't see value in the additional `API` bit

internal/dbmodel/controller.go
```diff
@@ -132,6 +132,23 @@ func (c Controller) ToAPIControllerInfo() apiparams.ControllerInfo {
...
+func (c Controller) ToAPIShowControllerInfo() apiparams.ControllerDetails {
```
alesstimec:
```suggestion
func (c Controller) ToControllerDetails() apiparams.ControllerDetails {
```

internal/dbmodel/controller_bootstrap.go
```diff
@@ -38,3 +38,19 @@ func (c ControllerBootstrap) ToAPIControllerInfo() apiparams.ControllerInfo {
 	}
 	return ci
 }
+
+// ToAPIControllerDetails converts a pending bootstrap entry to controller info for show APIs.
+func (c ControllerBootstrap) ToAPIControllerDetails() apiparams.ControllerDetails {
```
alesstimec:
```suggestion
func (c ControllerBootstrap) ToControllerDetails() apiparams.ControllerDetails {
```

alesstimec (commented):
LGTM modulo a few comments

alesstimec (approved):

PR #2000, opened 2026-06-04
feat: broaden use of JAAS permissions

kian99:
## Description
Allow a subset of JIMM's broader permission endpoints to be used by non-admins provided the user has admin access to the resource they want to modify. E.g. a model admin can use JIMM's AddRelation endpoint to grant access to a model for a group or role or another user.

For groups and roles, the user must still be an admin to list them or view a specific one.

Fixes [JUJU-9926](https://warthogs.atlassian.net/browse/JUJU-9926)

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

[JUJU-9926]: https://warthogs.atlassian.net/browse/JUJU-9926?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/jimm/permissions/relations.go
+++ b/internal/jimm/permissions/relations.go
@@ -25,16 +25,45 @@ import (
 // requests according to the deployed OpenFGA instance configuration.
 const BATCH_SIZE_OPENFGA = 100
 
-// AddRelation checks user permission and add given relations tuples.
-// At the moment user is required be admin.
-func (j *PermissionManager) AddRelation(ctx context.Context, user *openfga.User, tuples []apiparams.RelationshipTuple) error {
-
-	if !user.JimmAdmin {
-		return errors.Codef(errors.CodeUnauthorized, "unauthorized")
+func (j *PermissionManager) authorizeRelationTargetAdmin(ctx context.Context, user *openfga.User, tuple openfga.Tuple) error {
+	if user.JimmAdmin {
+		return nil
 	}
+
+	switch tuple.Target.Kind {
+	case openfga.ControllerType, openfga.ModelType, openfga.ApplicationOfferType, openfga.CloudType:
+		allowed, err := j.authSvc.CheckRelation(ctx, openfga.Tuple{
+			Object:   ofganames.ConvertTag(user.ResourceTag()),
+			Relation: ofganames.AdministratorRelation,
+			Target:   tuple.Target,
+		}, false)
+		if err != nil {
+			return errors.Codef(errors.CodeOpenFGARequestFailed, "%w", err)
+		}
+		if allowed {
+			return nil
+		}
+	case openfga.GroupType, openfga.RoleType:
+		// Membership changes for groups and roles are restricted to JIMM admins.
+	default:
+		// Unsupported relation-management targets are rejected for non-admins.
+	}
+
+	return errors.Codef(errors.CodeUnauthorized, "unauthorized")
+}
+
+// AddRelation checks user permission and adds the given relation tuples.
+// JIMM admins can update any relation, while resource administrators can update
+// relations on supported resource targets.
+func (j *PermissionManager) AddRelation(ctx context.Context, user *openfga.User, tuples []apiparams.RelationshipTuple) error {
 	parsedTuples, err := j.parseTuples(ctx, tuples)
 	if err != nil {
 		return err
+	}
+	for _, tuple := range parsedTuples {
+		if err := j.authorizeRelationTargetAdmin(ctx, user, tuple); err != nil {
+			return err
+		}
 	}
 	for i := 0; i < len(parsedTuples); i += BATCH_SIZE_OPENFGA {
 		end := min(i+BATCH_SIZE_OPENFGA, len(parsedTuples))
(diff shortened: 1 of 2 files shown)


internal/jimm/permissions/relations.go
```diff
@@ -25,17 +25,46 @@ import (
 // requests according to the deployed OpenFGA instance configuration.
 const BATCH_SIZE_OPENFGA = 100
 
-// AddRelation checks user permission and add given relations tuples.
-// At the moment user is required be admin.
-func (j *PermissionManager) AddRelation(ctx context.Context, user *openfga.User, tuples []apiparams.RelationshipTuple) error {
+func (j *PermissionManager) authorizeRelationTargetAdmin(ctx context.Context, user *openfga.User, tuple openfga.Tuple) error {
```
ale8k: s/authorizeRelationTargetAdmin/checkAdminAuthorisationForResource

And also change tuple arg to just target string, otherwise you're making 2 tuples

ale8k (approved):
lgtm just remove the extra tuple please and put target string into arg

internal/jimm/permissions/relations.go
kian99: A string isn't enough as we also use tuple.Target.Kind, I've changed it to just the target object, `*ofga.Entity`.

On the name change, I've just left it as is since it's just an internal function.

internal/jimm/permissions/relations.go
```diff
@@ -25,17 +26,46 @@ import (
 // requests according to the deployed OpenFGA instance configuration.
 const BATCH_SIZE_OPENFGA = 100
 
-// AddRelation checks user permission and add given relations tuples.
-// At the moment user is required be admin.
-func (j *PermissionManager) AddRelation(ctx context.Context, user *openfga.User, tuples []apiparams.RelationshipTuple) error {
+func (j *PermissionManager) authorizeRelationTargetAdmin(ctx context.Context, user *openfga.User, target *ofga.Entity) error {
```
SimoneDutto, line 29: i don't want to broaden the scope of this piece of work too much, but i feel like we are adding another place where we do the auth checks.

Sometimes in jujuapi, sometimes in the business-layer managers, and now in the permission managers.

Should we try to consolidate out auth checks a bit? I'm happy to do all of them here, or always calling the permission manager from everywhere?

Because i feel like this is a special case solution to a bigger problem.

Example: 
Getting a model info is gated by jimm admin
```
func (r *controllerRoot) ModelControllerInfo(ctx context.Context, req apiparams.ModelControllerInfoRequest) (apiparams.ModelControllerInfo, error) {
	if !r.user.JimmAdmin {
		return apiparams.ModelControllerInfo{}, errors.E(errors.CodeUnauthorized, "unauthorized")
	}
```
but this change allows for an non jimm-admin to add permission for the same model.

SimoneDutto: i think this is a good change in the intention, but it's making our future(?) intention of making jimm api more consistent harder

kian99: Couple of things,
- The business-layer managers and this permission manager are the same thing. We don't mix the managers into each other (as much as possible). This manager is the same as the group manager, its role is for permissions. So from that perspective it's not doing anything new.
- While I agree we want a better system of auth checks, I don't think we'll ever end up in a place where the auth checks are purely done at the API layer. For many requests that a non-admin can make, the permission depends on what the user is trying to access/what they are trying to do so some permission checks will need to be lower down the stack.

kian99: > but this change allows for an non jimm-admin to add permission for the same model.

I'm confused on this point btw. That's the point of this change, to allow a model admin to grant permissions to a model even if they aren't a JIMM admin. They can already do that with the `ModifyModelAccess` method but now they can also do it with the JAAS specific method, allowing them to grant access to groups etc.

SimoneDutto: yeah, but isn't it weird you can give permission to a model but you can't `juju show-model`? this was just an example to showcase how inconsistent our API would be in general, i believe there are other ones

kian99: You mean you can't `juju jaas show-model` technically. If you have model access `juju show-model` would work. 
That specific choice was probably made in a vacuum, without looking at the code or thinking about it too much, I'd say why not just do a IsModelReader check there (which I think you'd agree) but we can debate that separately.

SimoneDutto:
in addition, thinking about this change a little more, i think it's doing a big change to the way permissions are assigned.

In sql specifically there is something called `WITH GRANT OPTION` which allows a user to have some permission + being able to give the same permission to another user.

With this pr we would drastically change the way permissions are assigned, because if you are an admin of a model you can create new admins, which i would argue is a breaking change. Once you have a single admin, you can totally bypass the jimm admin.
Expecially for IS this could be not ok. Because once they create an admin for a team, the team could bypass their gitops approach by creating all the other admins themselves.. 

If we want to go down this less restrictive path, maybe we should let users with admin permission assigning permissions lower than `admin` or something hierarchical of this sort.

What do you think?

kian99:
> in addition, thinking about this change a little more, i think it's doing a big change to the way permissions are assigned.
> 
> In sql specifically there is something called `WITH GRANT OPTION` which allows a user to have some permission + being able to give the same permission to another user.
> 
> With this pr we would drastically change the way permissions are assigned, because if you are an admin of a model you can create new admins, which i would argue is a breaking change. Once you have a single admin, you can totally bypass the jimm admin. Expecially for IS this could be not ok. Because once they create an admin for a team, the team could bypass their gitops approach by creating all the other admins themselves..
> 
> If we want to go down this less restrictive path, maybe we should let users with admin permission assigning permissions lower than `admin` or something hierarchical of this sort.
> 
> What do you think?

Nothing really changes with this PR. If you are a model admin you can grant another user with admin permissions with `juju grant ...` but you couldn't use `juju jaas add-permission`, now you can, that's it.

SimoneDutto:
> > in addition, thinking about this change a little more, i think it's doing a big change to the way permissions are assigned.
> > In sql specifically there is something called `WITH GRANT OPTION` which allows a user to have some permission + being able to give the same permission to another user.
> > With this pr we would drastically change the way permissions are assigned, because if you are an admin of a model you can create new admins, which i would argue is a breaking change. Once you have a single admin, you can totally bypass the jimm admin. Expecially for IS this could be not ok. Because once they create an admin for a team, the team could bypass their gitops approach by creating all the other admins themselves..
> > If we want to go down this less restrictive path, maybe we should let users with admin permission assigning permissions lower than `admin` or something hierarchical of this sort.
> > What do you think?
> 
> Nothing really changes with this PR. If you are a model admin you can grant another user with admin permissions with `juju grant ...` but you couldn't use `juju jaas add-permission`, now you can, that's it.

As discussed offline Kian is right, in a sense that ModifyModelAccess, ModifyCloudAccess, and ModifyOfferaccess don't change with this pr.

The only thing that seems to be different is granting admin access to controller. Since we don't implement `ModifyControllerAccess` before this change the only one able to give admin permission to controller would have been jimm admins, now it would be possible for controller admins to create other controller admins.

This is aligned with Juju auth model because when you are an admin in juju you can create new admins in general, but i think it's important that we keep in mind that this pr is indeed loosening the check around creating new admins for controller.

SimoneDutto (approved):
i'm not super happy with the way we do auth check, but this is an improvement for jaas users nonetheless, so i'm happy to +1

PR #2003, opened 2026-06-05
feat: improve river teardown and upgrade-to errors

kian99:
## Description
This PR addresses some bits around River that needed improvement when QA'ing the upgrade-to logic.
1. The upgrade-to process did retries with River's default retry policy [which is](https://riverqueue.com/docs/job-retries) `attempts ^ 4 + rand(±10%) seconds (0s, 1s, 16s, ...)`. This is too quick for the upgrade-to logic.
2. The River client did not get shutdown if JIMM received a stop signal so a running job would not be cleanly killed and could be left in a `running` state in the DB, eventually getting cleaned up by River's rescuer process. Now we call StopAndCancel() on the River client when JIMM is stopped, cancelling any current jobs. Note that StopAndCancel() still waits for active jobs to return, but cancels the active context.
3. The error surfaced in the root job did not include the step that was being executed.

There are still some remaining issues:
- The root job has a retry count of 3 as do the child jobs. This means child jobs can execute up to 9 times, with the first 2 rounds of errors behind masked. 
- If the root job is cancelled (say by a restart) while a child job is waiting to retry, the root job's attempt count will tick up by one but the child job's won't. That means on restart, the root job might never retry but the child job will get picked up again.

These two issues seem to be solved by River's new [resumable-jobs](https://riverqueue.com/docs/resumable-jobs) feature which covers our short-comings.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

### DIFF
--- a/cmd/jaas/cmd/upgradeto.go
+++ b/cmd/jaas/cmd/upgradeto.go
@@ -24,7 +24,7 @@ and upgrades the models to the controller's version.
 `
 	upgradeToExample = `
     juju upgrade-to myController 2cb433a6-04eb-4ec4-9567-90426d20a004
-	juju upgrade-to myController 2cb433a6-04eb-4ec4-9567-90426d20a004 83cf3d62-ab16-4cb2-8e2f-df111fca1a32
+    juju upgrade-to myController 2cb433a6-04eb-4ec4-9567-90426d20a004 83cf3d62-ab16-4cb2-8e2f-df111fca1a32
 `
 )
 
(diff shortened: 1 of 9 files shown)


kian99:
One correction which may reduce the number of changed lines significantly, the docstring for `StopAndCancel` notes,
> // This can also be initiated by cancelling the context passed to Start. There is
// no need to call this method if the context passed to Start is cancelled
// instead.

So I may remove the explicit cancellation.

kian99:
A lot of this is changing in my follow-up PR so need to land this.

PR #2004, opened 2026-06-05
fix(setclouddefaults): fixes a potential panic in SetCloudDefaults

alesstimec:
## Description

Fixes a potential panic in SetCloudDefaults if the defaults map is nil in the DB.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests


SimoneDutto (commented):
Is this actually a bug that can happen? Or in our "Set paths" we always add a non-nil defaults map?

alesstimec:
This is in our "set" path :)

SimoneDutto:
> This is in our "set" path :)

this is a panic in our set path if we have a nil Default map, maybe we can't have a nil Default map in general

alesstimec:
@SimoneDutto still, better safe than sorry..

SimoneDutto (approved):

kian99 (approved):

PR #2005, opened 2026-06-05
fix: avoid race in shared err in SSH tunnel copy goroutines

alesstimec:
## Description

Two goroutines copying data between client and the controller shared the same error variable.


kian99 (approved):

SimoneDutto (approved):
isn't this just a "potential" race condition?

because `err` is not used after the two routines are started.

Happy to approve though.


Edit: the `if err != nil` can be a problem indeed, because it could race

PR #2006, opened 2026-06-05
fix: clean up controllerRoot watchers on websocket disconnect

alesstimec:
## Description

The controllerRoot's cleanup was only reachable via apiServer.Kill, but the WSServer interface never declares Kill and the websocket handler never calls it, so cleanup never ran. 

Any client that started a WatchAllModelSummaries/WatchModelSummaries watcher and disconnected without an explicit stop leaked the modelAccessWatcher goroutine and its subscription permanently. 

Assigning controllerRoot.cleanup to apiServer.cleanup also raced: apiServer is shared across all connections on the handler, so each connection assigned the field concurrently.

Added an test that asserts that the subscriber count decrements after the connection is closed.

## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests

### DIFF
--- a/internal/jujuapi/websocket.go
+++ b/internal/jujuapi/websocket.go
@@ -37,9 +37,8 @@ const (
 
 // An apiServer is a jimmhttp.WSServer that serves the controller API.
 type apiServer struct {
-	jimm    *jimm.JIMM
-	cleanup func()
-	params  Params
+	jimm   *jimm.JIMM
+	params Params
 }
 
 // Authenticate implements jimmhttp.Authenticate and handles browser authentication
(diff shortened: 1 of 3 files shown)


kian99 (commented):
Interesting find, I've double checked that serveRoot only returns when the connection is dead. 
Then I checked whether Kill() was actually called from anywhere and it seems like it might be. See Juju's rpc/server.go lines 350 and lines 321 where it calls `conn.root.Kill()`. So it looks like the Kill() method was already being invoked correctly.

Edit: I'd like to see if the test added in this PR still passes with the changes reverted, and if not, why.

PR #2007, opened 2026-06-05
fix(permissions): restrict resource admins to access-grant relations

alesstimec:
## Description

The "broaden use of JAAS permissions" change let a resource administrator (a non-JIMM-admin who administers a resource) manage relation tuples on that resource. The authorization check, however, looked only at whether the user administered the tuple's *target* — it never constrained the *relation* being written or the *kind* of object it was granted to.

OpenFGA uses tuples on those same targets for two distinct purposes: access grants (administrator/reader/writer/consumer/can_addmodel/ audit_log_viewer, always granted to a user, group or role) and the structural relations that define the resource hierarchy (a model's `controller`, an offer's `model`). Because the relation type was never checked, a mere model administrator could send

    RemoveRelation{Object: controller-X, Relation: "controller",
                   TargetObject: model-Y}

and detach their model from its controller. That breaks any admin whose access is derived through the controller chain, desyncs JIMM's database from OpenFGA, and is not audit-logged (logUserUpdates only records grants whose object kind is `user`). The structural relations are set by JIMM internally and are not part of the user-facing access-grant API.

Tighten the non-admin path with two default-deny allowlists:

  - the relation must be an access-grant relation; structural (controller/model) relations are rejected, and any relation added to the model in future stays JIMM-admin-only until explicitly opted in;
  - the grantee object must be a user, service account, group or role (service accounts and user:* are both kind `user`).

JIMM admins continue to bypass both checks, so they remain able to manage structural relations.

authorizeRelationTargetAdmin now takes the full tuple (rather than just the target) so it can see the relation and object; AddRelation, RemoveRelation and CheckRelation are updated to pass it. As part of this, CheckRelation's "user checking self" shortcut is nil-guarded, fixing a nil-pointer panic when a request omits the tuple Object.

Adds TestStructuralRelationManagementRequiresJimmAdmin covering the detach-model attack, the offer->model structural relation, and granting an access relation to a non-grantee object kind, plus a positive case confirming JIMM admins can still manage structural relations.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests


internal/jimm/permissions/relations.go
```diff
@@ -26,17 +26,62 @@ import (
...
+	openfga.UserType:  true,
+	openfga.GroupType: true,
+	openfga.RoleType:  true,
+}
+
+// authorizeRelationTargetAdmin authorizes a non-JIMM-admin to manage the given
+// relation tuple. JIMM admins may manage any tuple. A non-admin may only grant
+// or revoke an access relation (resourceAdminRelations) to a grantee of an
+// allowed kind (grantableObjectKinds) on a resource target they administer.
+func (j *PermissionManager) authorizeRelationTargetAdmin(ctx context.Context, user *openfga.User, tuple openfga.Tuple) error {
```
kian99, line 63: Please add an example here at the end, like

```suggestion
// relation tuple. JIMM admins may manage any tuple. A non-admin may only grant
// or revoke an access relation (resourceAdminRelations) to a grantee of an
// allowed kind (grantableObjectKinds) on a resource target they administer.
//
// E.g. a non-jimm-admin who is a model admin can grant 'user-alice' with 'write' permission
// to a model but cannot remove the relation between a controller and that model.
func (j *PermissionManager) authorizeRelationTargetAdmin(ctx context.Context, user *openfga.User, tuple openfga.Tuple) error {
```

kian99 (approved):
Good spot

internal/jimm/permissions/relations.go
alesstimec: added

internal/jimm/permissions/relations.go
```diff
@@ -26,17 +26,65 @@ import (
...
 	}
 
-	switch target.Kind {
+	// Non-admins may only manage access-grant relations, never the structural
+	// (controller/model) relations that define the resource hierarchy.
+	if !resourceAdminRelations[tuple.Relation] {
+		return errors.Codef(errors.CodeUnauthorized, "unauthorized")
+	}
+
+	// ...and only grant them to a user, service account, group or role.
```
SimoneDutto, line 74: this comment is tricky, because i guess it's correlated to the one before, but if we add some additional lines in between it wouldn't make sense anymore.

SimoneDutto (approved):
lgtm

internal/jimm/permissions/relations.go
```diff
@@ -26,17 +26,65 @@ import (
...
+// relations are the only ones whose object is a controller/model/cloud, so
+// restricting grantees to these kinds independently prevents a non-admin from
+// writing a structural tuple.
+var grantableObjectKinds = map[openfga.Kind]bool{
+	openfga.UserType:  true,
+	openfga.GroupType: true,
+	openfga.RoleType:  true,
+}
+
+// authorizeRelationTargetAdmin authorizes a non-JIMM-admin to manage the given
```
luci1900, line 56: "non-JIMM-admin" is ambigous. Can mean "admin, but not JIMM's" and also the negation of "JIMM admin".

luci1900: I'd imagine "JIMM non-admin" to be an improvement, in the various places that is referenced.

PR #2009, opened 2026-06-05
feat: refactor upgradeTo river job

kian99:
## Description
Following up from https://github.com/canonical/jimm/pull/2003, there are some bits around River and our UpgradeTo job that needed to be improved, copying from that PR description.
1. The upgrade-to process did retries with River's default retry policy [which is](https://riverqueue.com/docs/job-retries) `attempts ^ 4 + rand(±10%) seconds (0s, 1s, 16s, ...)`. This is too quick for the upgrade-to logic.
2. The River client does shutdown if JIMM received a stop signal because the context passed to `client.Start()` would be cancelled, but we were not waiting for River to cleanup and wait for any pending jobs to resolve so potentially a running job would not be cleanly killed and could be left in a `running` state in the DB, eventually getting cleaned up by River's rescuer process. Now we wait on StopAndCancel() on the River client when JIMM is stopped, cancelling any current jobs. Note that StopAndCancel() still waits for active jobs to return, but cancels the active context.
3. The error surfaced in the root job did not include information on the step that was being executed.

There are still some remaining issues:
- The root job has a retry count of 3 as do the child jobs. This means child jobs can execute up to 9 times, with the first 2 rounds of errors behind masked. 
- If the root job is cancelled (say by a restart) while a child job is waiting to retry, the root job's attempt count will tick up by one but the child job's won't. That means on restart, the root job might never retry but the child job will get picked up again.

These two issues seem to be solved by River's new [resumable-jobs](https://riverqueue.com/docs/resumable-jobs) feature which covers our short-comings. So I have switched to that which allowed for a huge simplification. We no longer need a root job with multiple child jobs. In addition to that I've changed the API response to better align with this change.

Fixes [JUJU-9930](https://warthogs.atlassian.net/browse/JUJU-9930)

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

[JUJU-9930]: https://warthogs.atlassian.net/browse/JUJU-9930?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/jimm/jobs/jobs.go
```diff
@@ -125,38 +125,18 @@ func (j *JobManager) GetUpgradeToStatusForModel(ctx context.Context, modelUUID s
...
-		migration := toJobDetail(migrationJob)
-		status.Migration = &migration
-	}
-
-	if output.UpgradeJobID != nil {
-		upgradeJob, err := j.jobQuerier.GetJobInfo(ctx, *output.UpgradeJobID)
-		if err != nil {
-			return nil, fmt.Errorf("failed to get upgrade job info: %w", err)
+	var output rivertypes.UpgradeToOutput
+	rawOutput := rootJob.Output()
```
ale8k: Probably rename it from root job now? As it's just the upgradeToJob with certain steps that are resumable?

internal/river/upgradetoworker.go
```diff
@@ -5,37 +5,63 @@ package river
...
-		migrateRetries:  migrateRetries,
-		upgradeRetries:  upgradeRetries,
-		awaitCompletion: awaitFunc,
+	upgradeFlowRetryAfterFirstFailure  = 30 * time.Second
+	upgradeFlowRetryAfterSecondFailure = 2 * time.Minute
+)
+
+type upgradeToRetryFunc func(*river.Job[rivertypes.UpgradeToArgs]) time.Time
+
+type stageError struct {
```
ale8k, line 29: This isn't necessary it seems, we aren't checking it is of type stage error and doing anything with the stage?

internal/river/upgradetoworker.go
```diff
@@ -5,37 +5,63 @@ package river
...
-func newUpgradeToWorker(migrateRetries int, upgradeRetries int, awaitFunc awaitCompletionFunc) *upgradeToWorker {
-	return &upgradeToWorker{
-		migrateRetries:  migrateRetries,
-		upgradeRetries:  upgradeRetries,
-		awaitCompletion: awaitFunc,
+	upgradeFlowRetryAfterFirstFailure  = 30 * time.Second
+	upgradeFlowRetryAfterSecondFailure = 2 * time.Minute
+)
+
+type upgradeToRetryFunc func(*river.Job[rivertypes.UpgradeToArgs]) time.Time
```
ale8k, line 27: Does river have a type defined for this already? I'd be surprised if they don't - if they do can we use that instead?

internal/river/upgradetoworker.go
```diff
@@ -45,186 +71,90 @@ func (w *upgradeToWorker) Timeout(*river.Job[rivertypes.UpgradeToArgs]) time.Dur
...
-	if err := persistUpgradeToSupervisorOutput(ctx, client, job.ID, supervisorOutput); err != nil {
-		return err
-	}
-
-	if err := w.awaitCompletion(ctx, migrateInsertResponse, eventCh); err != nil {
-		return err
+// NextRetry implements the [river.Worker] interface.
+// It determines the next retry time based on the attempt number,
+// allowing for a longer delay after the second failure.
+func (w *upgradeToWorker) NextRetry(job *river.Job[rivertypes.UpgradeToArgs]) time.Time {
```
ale8k, line 77: This seems strange to me how we're checking the attempt number and then just adding time. Does River provide internal backoff delay configuration for the job? Or is this how they expect us to do it?

ale8k (commented):
Nice! 🥳 Looks solid to me, just some minor queries

internal/river/upgradetoworker.go
kian99: Yes true, will remove.

internal/river/upgradetoworker.go
kian99: I'll check.

internal/river/upgradetoworker.go
kian99: You can leave it at the default which as I mentioned in the PR description is `attempts ^ 4 + rand(±10%) seconds (0s, 1s, 16s, ...)`. I went with 30s then 2m as that seemed reasonable.

SimoneDutto (commented):
it looks good in general. I have a few questions:

what's the output of getting the job details on one of this?
it shows you the details of the resumable steps or how does it work in general?

In case we'd like to actually do integration testing for these resumable jobs? Can you put the `rivertest` in a middle of a resumable job?

kian99:
> it looks good in general. I have a few questions:
> 
> what's the output of getting the job details on one of this? it shows you the details of the resumable steps or how does it work in general?

The job details don't surface a stable API for retrieving how far along the job is, that's why I have used River's "Recorded Output" feature to track the progress along the way. Internally, it uses River's [middleware](https://riverqueue.com/docs/middleware) feature, which allows us to wrap job insertion and execution, but the middleware used for resumable jobs is inside an internal package "github.com/riverqueue/river/internal/rivermiddleware" so we can't access the internal details about where the resumable steps are at.

> In case we'd like to actually do integration testing for these resumable jobs? Can you put the `rivertest` in a middle of a resumable job?

Yes you can do that, and I've done it in one of our tests where we do `rivertest.ResumableStepAfter(&river.InsertOpts{}, upgradeToMigrationStep)` to simulate starting after the step was completed. 

This is an integration tests in the sense that it is interacting with a real database. I've also add a full integration test to the `testing` package that inserts a River job and verifies the details are returned by the API.

internal/river/upgradetoworker.go
kian99: I think I'll rather leave this, I added it for the easy to use formatting style and when removing it I replaced it with `return errors.New(fmt.Sprintf("%s failed: %v",upgradeToMigrationStep, err))` which imo is a lot harder to read.

internal/river/upgradetoworker.go
ale8k: Sorry I was more asking do they not expose configuration options in a better way that this? A function signature where you implement it yourself seems a bit heavy is all

kian99: I don't see anything else. You either define a default retry policy on the client or implement the method on the worker.

internal/river/upgradetoworker.go
kian99: Doesn't seem like it, it's just a method signature in River's `worker.go` L88.

luci1900 (approved):
Seems like a slightly better abstraction to be operating on.

internal/river/upgradetoworker.go
ale8k: Ah ok... Nvm... I guess a simple back off is OK. Resolving.

internal/river/upgradetoworker.go
ale8k: Yeah it's ok, just expected river to provide us some tuning knobs. Resolving.

ale8k (approved):
lgtm ty!

PR #2010, opened 2026-06-05
feat: dialAll should return on the first successful connection

SimoneDutto:
# Description
before dialAll was returning only after all dial goroutines have finished. The new implementation returns after the first successful one and cancel the other dials. In addition it joins the errors.

This came out of my tracing spike, i've noticed we are spending some time in the dial bit and looking at the code it seemed obvious we shouldn't wait for all the dial go routines to end.

Let me know what you think. This fix might be particular useful in production environment where one of the configure controller address is not reachable/timeout/or something.

# QA

I've used the old dialer with the same test and it was taking >200ms to return a connection, proving this fix is indeed needed.

### DIFF
--- a/internal/rpc/dial.go
+++ b/internal/rpc/dial.go
@@ -6,11 +6,12 @@ import (
 	"context"
 	"crypto/tls"
 	"crypto/x509"
+	stderrors "errors"
 	"fmt"
 	"net/http"
 	"net/url"
 	"path"
-	"sync"
+	"sync/atomic"
 
 	"github.com/gorilla/websocket"
 	"github.com/juju/juju/core/network"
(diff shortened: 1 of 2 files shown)


alesstimec:
Let's talk about this change.. i have some doubts

internal/rpc/dial.go
```diff
@@ -152,40 +157,45 @@ func dialAll(ctx context.Context, dialer *Dialer, urls []string, headers http.He
...
-				errOnce.Do(func() {
-					err = dErr
-				})
-				return
-			}
-			var keep bool
-			clientOnce.Do(func() {
-				res = conn
-				keep = true
-				cancel()
```
kian99, line 179: Here, the first client to finish connecting would call `cancel()` on the context, ensuring the other connections return. So if your change improved the speed, I'd be curious why.

internal/rpc/dial.go
```diff
@@ -152,40 +157,45 @@ func dialAll(ctx context.Context, dialer *Dialer, urls []string, headers http.He
...
-			})
-			if !keep {
-				conn.Close()
+		go func() {
```
luci1900, line 176: Just do I don't forget.

My only concern is with this goroutine not finishing or panicking.

SimoneDutto:
Closing this one because there wasn't much interest behind it

PR #2012, opened 2026-06-08
fix(jujumanager): close controller connections in PollModels

alesstimec:
## Description

PollModels dials each controller once to check its models, but never closes the resulting API connection. PollModels runs on a periodic schedule, so every cycle leaks one connection (a websocket plus its background receive goroutine) per controller, steadily accumulating until file descriptors are exhausted.

Process each controller inside a closure and `defer api.Close()` there, so the connection is released at the end of that controller's iteration. A bare `defer` in the loop body would instead hold every controller's connection open until PollModels returned, which is also undesirable when many controllers are polled.

Adds TestPollModelsClosesControllerConnections, which asserts via the test dialer's open-connection counter that no connection is left open after a poll.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

### DIFF
--- a/internal/jimm/juju/model_poller.go
+++ b/internal/jimm/juju/model_poller.go
@@ -15,6 +15,39 @@ import (
 	"github.com/canonical/jimm/v3/internal/servermon"
 )
 
+func (j *JujuManager) pollModels(ctx context.Context, models []*dbmodel.Model) error {
+	if len(models) == 0 {
+		return nil
+	}
+
+	ctrl := models[0].Controller
+
+	api, err := j.dialController(ctx, &ctrl, nil)
+	if err != nil {
+		zapctx.Error(ctx, "cannot dial controller", zap.String("controller", ctrl.UUID), zap.Error(err))
+		return nil
+	}
+	defer api.Close()
+
+	// Depending the model's migration mode, we either:
+	// - Check if the model exists (MigrationModeNone)
+	// - Check if the model has completed internal migration (MigrationModeMigrateInternal)
+	// - Do nothing if the model is in any other migration mode (MigrationModeImporting, MigrationModeExporting)
+	for _, m := range models {
+		ctx := zapctx.WithFields(ctx,
+			zap.String("model-owner", m.OwnerIdentityName),
+			zap.String("model-name", m.Name),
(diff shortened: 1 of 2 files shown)


internal/jimm/juju/model_poller.go
```diff
@@ -42,27 +42,33 @@ func (j *JujuManager) PollModels(ctx context.Context) (err error) {
...
-				zap.String("model-owner", m.OwnerIdentityName),
-				zap.String("model-name", m.Name),
-				zap.String("migration-mode", string(m.MigrationMode)),
-			)
-
-			_, err := j.modelInfo(ctx, nil, m, api)
+		// Process each controller in a closure so the dialled connection is
+		// closed at the end of the iteration rather than being leaked (or, with
+		// a bare defer, held open for every controller until PollModels returns).
+		func() {
```
kian99: What do you think about moving this to a separate function entirely?

internal/jimm/juju/model_poller_test.go
```diff
@@ -241,6 +241,30 @@ func TestInternalMigrationFailure(t *testing.T) {
 	c.Assert(model.MigrationMode, qt.Equals, dbmodel.MigrationModeNone)
 }
 
+// TestPollModelsClosesControllerConnections ensures that the connection
+// dialled to each controller is closed once its models have been processed.
```
kian99, line 314: Is this test actually doing what the comment says its doing?
It's checking `dialer.IsClosed()` is true at the end. Seems to me it's only checking that we close the connection at some point before `PollModels` returns rather than verify we are closing each controller's connection once its models have been processed.

kian99 (approved):
Okay to land but could use some rewording in the test comment at the minimum.

internal/jimm/juju/model_poller.go
luci1900: It's a limitation of the Go defer pattern that I think only improves with a separate function, agreed.

luci1900 (approved):
This seems fine, especially if we move the logic to a separate function for clearer defer.

I'm not convinced this would've actually led to exhausting file descriptors, though. At least for files, the Go GC finalizer will call `Close`.

PR #2014, opened 2026-06-08
fix(jujuapi): check for models before destroying a controller

alesstimec:
## Description

StartDestroyController guards against tearing down a controller that still hosts models with `if len(ctrl.Models) != 0`. The controller comes from ControllerInfo, which loads it via GetController; that query only preloads CloudRegions, never the Models association. ctrl.Models is therefore always an empty slice, so the guard never fires and a JIMM admin can start a destroy job for a controller that still hosts models.

Add a ControllerModelCount method to the juju manager that queries the models for a controller via GetModelsByController (the same query the RemoveController path already uses), and have StartDestroyController call it instead of inspecting the never-populated association.

TestStartDestroyControllerJob now drives the model count through the manager rather than ctrl.Models, and fails against the old guard.

## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests

### DIFF
--- a/internal/jujuapi/jimm.go
+++ b/internal/jujuapi/jimm.go
@@ -752,7 +752,14 @@ func (r *controllerRoot) StartDestroyController(ctx context.Context, req apipara
 		return apiparams.StartBootstrapResponse{}, fmt.Errorf("failed to fetch controller info: %w", err)
 	}
 
-	if len(ctrl.Models) != 0 {
+	// ControllerInfo does not preload the controller's models, so query them
+	// explicitly; relying on ctrl.Models would always see an empty slice and
+	// silently allow destroying a controller that still hosts models.
+	modelCount, err := r.jimm.JujuManager().ControllerModelCount(ctx, *ctrl)
+	if err != nil {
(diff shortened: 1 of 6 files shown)


kian99 (approved):
Good find

internal/jujuapi/jimm.go
```diff
@@ -752,7 +752,14 @@ func (r *controllerRoot) StartDestroyController(ctx context.Context, req apipara
 		return apiparams.StartBootstrapResponse{}, fmt.Errorf("failed to fetch controller info: %w", err)
 	}
 
-	if len(ctrl.Models) != 0 {
+	// ControllerInfo does not preload the controller's models, so query them
```
SimoneDutto, line 755: can we preload it? so we don't need to query again?

SimoneDutto (approved):
looks good, just suggested an improvement

internal/jujuapi/jimm.go
alesstimec: too many other places where ControllerInfo is called. doesn't make sense to affect all of them for this single instance.

PR #2015, opened 2026-06-08
test: simplify keycloak setup for tests

kian99:
## Description
This change was prompted by a need to create a new local user in our Keycloak instance to debug an issue and having no docs on what the admin credentials were for our Keycloak instance (the admin credentials turned out to be in the docker compose file). The changes include,
1. Simplifies our keycloak "realm" JSON config as per a suggestion from @ale8k. I was trying to create a new user in our local keycloak to debug an issue and simplifies
2. Updated README.md to include relevant and useful info.

~~One change that might affect the web team cc @huwshimi @andogq @Ninfa-Jeon, is that for some reason this "updated" the Keycloak login screen where one enters the username/password from the familiar page without any markup to what you can see below. This may/may not break some of the dashboard's automated login logic.~~ Ignore this, I've kept the plain login page to avoid breaking anyone's tests.


ale8k (approved):
looks solid to me. Could you merge it also into feature/oidc-groups?

luci1900 (approved):
That's a lot simpler, nice.

SimoneDutto (approved):
gucci

PR #2016, opened 2026-06-08
fix: perform local charm uploads with admin token

### LINKED ISSUE #2008: Local charm deployment failing
I cannot deploy local charms as the upload fails, all other operations (`juju status`, `juju config`, etc...) work.
Logs: https://pastebin.canonical.com/p/9J6KHDDTKf/

Versions:

```
❯ juju controllers --format yaml
controllers:
  jaas-ps7:
    current-model: 795798e4-922f-49c7-9169-004ffc17df90@serviceaccount/stg-cpc-vmware
    user: gauthier.jolly@canonical.com
    access: login
    recent-server: jaas.ps7.canonical.com:443/k8s-jaas-ps7-jimm-jimm
    controller-uuid: 63a3e178-2450-8d2d-4ee4-33bc310e6b7a
    api-endpoints: ['jaas.ps7.canonical.com:443/k8s-jaas-ps7-jimm-jimm']
    ca-cert: ""
    cloud: ""
    agent-version: 3.6.21
    model-count: 14
current-controller: jaas-ps7
❯ juju --version
3.6.23-genericlinux-amd64
```

kian99:
## Description
In https://github.com/canonical/jimm/pull/1977 I switched JIMM to mostly use JWTs when authenticating with Juju. This spawned #2008 causing `juju deploy` of a local charm to fail when the user is not a JIMM admin. This is because of a bug in Juju. Specifically Juju creates a struct here https://github.com/juju/juju/blob/3.6/apiserver/authentication/jwt/jwt.go#L104 that acts as a permission delegator but does not include the relevant model tag which then causes the permission check in https://github.com/juju/juju/blob/3.6/apiserver/httpcontext.go#L278 to always fail because `authInfo.ModelTag` is always empty. This affects all Juju versions.

The simplest fix without reverting back to username/password is to mint a JWT with controller superuser permissions.

Checking whether other endpoints are affected,
- `juju debug-log` is seemingly affected as it uses the same authorization checker on Juju's end, but in JIMM this path always mints a token with the user as a controller superuser so in Juju we never reach the model check.
- Other endpoints like those used proxying model migration calls require the user to be a controller superuser so they also do not hit Juju's faulty model check.

The modified integration test uses the non-admin user `bob@canonical.com`. The test fails without the fix and passes with it.

Fixes [JUJU-9988](https://warthogs.atlassian.net/browse/JUJU-9988) and fixes https://github.com/canonical/jimm/issues/2008

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

[JUJU-9988]: https://warthogs.atlassian.net/browse/JUJU-9988?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


SimoneDutto (approved):
good find

luci1900 (approved):
Nice

PR #2018, opened 2026-06-09
feat: extract service configured group claim key and add it to sessio…

ale8k:
## Description

**PLEASE RE-READ - DESC HAS BEEN UPDATED**

This PR updates JIMM group claim handling so groups are extracted from the OAuth access token rather than the ID token, and then carried through browser sessions and JIMM issued device session tokens.

This is intentional for safer service account coverage. We do not currently fall back to the ID token when groups are missing from the access token. I've further configured our keycloak to NOT include it in the ID token. Please note though, our internal IdP WILL include it in the ID token. 

If the configured group claim is missing, authentication still succeeds and JIMM falls back to identity only session data.

The local Keycloak setup has been updated to match this behaviour, including scope wiring for client-credentials and access token group emission used by our e2e paths. I also corrected a client credentials scope issue so machine-to-machine logins no longer send user login scopes and we now provide the ability for operators to explicitly request the group scopes if they are not configured as default in their IdP.

Configuration has been aligned with the new naming:
- JIMM_OAUTH_GROUP_CLAIM_KEY is now used for the configured group claim field name (for lack of a better name honestly).
- JIMM_OAUTH_SCOPES remains the browser/device scope list and is now no longer sent on client credential verification.
- JIMM_OAUTH_CLIENT_CREDENTIAL_SCOPES is used specifically for client credentials flow.

I extended the tests across auth and integration flows to cover group present and group missing cases for browser, device, and client-credentials paths, and Keycloak coupled tests now rely on preseeded identities rather than the generation logic of entities we did previously.

Again, our internal IdP is expected to emit groups on both access and ID tokens, but dual token behaviour is not part of this PR and is not covered here, If it is ever an issue in the future, we can return and add fallback mechanism for handling ID token claim extraction. This may be a moot point though, as I *think* the majority of providers enforce it to be mapped **from** the access token to the ID token (meaning it'd always be present on the access token anyway).

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

## Test instructions
Run tests in the testing package.


internal/auth/auth.go
```diff
@@ -210,6 +231,49 @@ func NewAuthenticationService(ctx context.Context, params AuthenticationServiceP
...
+		return claims, nil
+	}
+
+	groupClaim, ok := rawClaims[as.groupClaimKey]
+	if !ok {
+		zapctx.Warn(ctx, "configured group claim missing from id token", zap.String("claim-key", as.groupClaimKey))
+		return claims, nil
+	}
+
+	if err := json.Unmarshal(groupClaim, &claims.Groups); err != nil {
```
SimoneDutto: why is there a need for json marshal?

SimoneDutto: isn't more robust to just get the string and do the parsing ourselves in a slice? Because this would work only in the case the string is marshalled into a go slice string, right?

internal/jimm/login/login.go
```diff
@@ -51,12 +52,12 @@ type OAuthAuthenticator interface {
...
-	// Email retrieves the users email from an id token via the email claim
-	Email(idToken *oidc.IDToken) (string, error)
+	// IdentityClaims retrieves the user's identity claims from a verified ID token.
+	IdentityClaims(ctx context.Context, idToken *oidc.IDToken) (auth.IdentityClaims, error)
 
 	// MintSessionToken mints a session token to be used when logging into JIMM
-	// via an access token. The token only contains the user's email for authentication.
-	MintSessionToken(email string) (string, error)
+	// via an access token. The token contains the user's email and internal groups claim.
+	MintSessionTokenWithGroups(email string, groups []string) (string, error)
```
SimoneDutto, line 56: Just a suggestion, feel free to ignore it.
if you use:
```
MintSessionToken(email string, group ...string)
```

you can keep the signature agnostic to groups, and potentially less changes all around

internal/testutils/jimmtest/keycloak.go
SimoneDutto, line 1: we can avoid these changes with the seeded approach right?

local/keycloak/jimm-realm.json
```diff
@@ -1506,16 +1506,16 @@
           "id": "8d80ce4b-3a01-48f9-974f-6a5e2074ccc6",
           "name": "groups",
           "protocol": "openid-connect",
-          "protocolMapper": "oidc-usermodel-realm-role-mapper",
+          "protocolMapper": "oidc-group-membership-mapper",
```
SimoneDutto: not familiar with this, can you explain a bit?

SimoneDutto (commented):
looks good in general, just a few questions

internal/testutils/jimmtest/keycloak.go
ale8k: Yup, and that's what I was asking about in MM. I'd prefer if we use seeded **only** users and clients honestly. Wdyt?

local/keycloak/jimm-realm.json
ale8k: It's just a naming label for the claim mappers, we already had a groups one and it had an odd name from when I c/p the original config. Unless you mean what are the protocol mappers?

SimoneDutto: the latter

internal/testutils/jimmtest/keycloak.go
SimoneDutto: seeded only is easier

ale8k: That's 2, will wait for 1 more then majority agrees we just go with seeded. If a 3rd agrees, I've actually already started a merge into this feature branch which incldues the cleanup for keycloak realm. Then I'll add the mappers and users explicitly (renaming them for the scenarios they cover).

internal/auth/auth_test.go
```diff
@@ -76,7 +83,7 @@ func TestAuthCodeURL(t *testing.T) {
 	c.Assert(
 		url,
 		qt.Matches,
-		regexp.MustCompile(`http:\/\/localhost:8082\/realms\/jimm\/protocol\/openid-connect\/auth\?client_id=jimm-device&redirect_uri=http%3A%2F%2Flocalhost%3A8080%2Fauth%2Fcallback&response_type=code&scope=openid\+profile\+email&state=.*`),
```
luci1900, line 79: This test was unreadable before and is still unreadable.

local/keycloak/jimm-realm.json
ale8k: So in the IdP we can create clients (of many kinds, but we care for OAuth and OIDC [brwoser, device and client credential grants])

In the *client* section, you'll see our JIMM client capable of device and auth code and a user facing client, capable of *only* client creds.

Below, you'll see *clientScopes* section, which contains scopings (i.e., ways to interact with the client e.g., add stuff to oidc tokens, display it a consent screen on this page, etc.). We have one called "Microprofile-JWT"  which enables the JWTs for OIDC. This has a bunch of OIDC mappers, but the one we care for is adding the cliam to the token which is our group mapper.

Lastly, you apply the scopes which you can see in [here](https://github.com/ale8k/jimm/blob/50c1c64f7ca129df9d70b1898193fb74ffdfc6b3/local/keycloak/jimm-realm.json#L819-L820) if you scroll down to the mappers for this client it is in the optional client scopes, which means it can be set as a scope for this client.

It is then set as a default under the section [here](https://github.com/ale8k/jimm/blob/50c1c64f7ca129df9d70b1898193fb74ffdfc6b3/local/keycloak/jimm-realm.json#L1582-L1587).

This is convoluted because I c/p'd a big example initially and it includes a bit too much. 
Once I merge v3 in, I'll slim it down and we'll just apply the scope in-line with the client (in this PR).

ale8k: Just realised I may have overcomplicated this lol, in simpler words we create our own scope for applying our own bits to the token. Similar to scopes the "profile" scope or "oidc" scope which add profile details and oidc subject, we're just adding own our. I'm gonna rename it to "canonical-groups" and you'll see it in the URL of our /token endpoint.

luci1900 (approved):

local/keycloak/jimm-realm.json
ale8k: I've updated it to be a loooot clearer now, lmk if you have any more questions

internal/testutils/jimmtest/keycloak.go
ale8k: It is now seeded only - cleaned it up a lot

internal/auth/auth.go
ale8k: Yeah you're right, if it isn't sent as [...groups], it'll fail parsing. Uhm. Leave it with me.

ale8k: Adjusted it to do 2 things:

1. Use "any" on the raw claim map and verify we can stringify the email.
2. Switch the type on the groups key (see code for all formats) & some unit tests for it.

Good shout honestly.

I did think about adding more scopes in our seeded config, but I feel like our unit test is good here.

internal/jimm/login/login.go
ale8k: I really dislike variadic args honestly, so would prefer to leave it as is for this one.

internal/testutils/jimmtest/keycloak.go
```diff
@@ -24,242 +12,14 @@ const (
...
-	keycloakJIMMRealmPath    = "/admin/realms/jimm"
-	keycloakAdminUsername    = "jimm"
-	keycloakAdminPassword    = "jimm"
-	keycloakAdminCLIClientID = "admin-cli"
+	// HardcodedGroupUsername is a hardcoded test keycloak user that pre-exists
+	// and belongs to OIDCGroupsTestGroupName.
+	HardcodedGroupUsername = "jimm-group-user"
+	HardcodedGroupEmail    = "jimm-group-user@canonical.com"
+	HardcodedGroupPassword = "password"
+	HardcodedGroupUserID   = "jimm-group-user-id"
```
SimoneDutto, line 20: it's not the group email, but the username in the hardcoded group?
it's a bit unclear reading the tests

local/keycloak/jimm-realm.json
```diff
@@ -74,6 +96,13 @@
       "directAccessGrantsEnabled": true,
       "serviceAccountsEnabled": false,
       "fullScopeAllowed": true,
+      "defaultClientScopes": [
+        "profile",
+        "email"
+      ],
+      "optionalClientScopes": [
+        "group"
```
SimoneDutto, line 132: nice

local/keycloak/jimm-realm.json
SimoneDutto, line 1: will this new setup mess terraform CI or dashboard CI? we are currently relying on the service account created in keycloak

internal/auth/auth.go
SimoneDutto: yeah, this is better now

internal/testutils/jimmtest/keycloak.go
ale8k: Shall I rename it to subject? You can set the subject for users in keycloak via:
```json
    {
      "id": "jimm-group-user-id", <--- This
      "username": "jimm-group-user",
      "email": "jimm-group-user@canonical.com",
      "emailVerified": true,
      "enabled": true,
      "groups": [
        "/canonical"
      ],
      "credentials": [
        {
          "type": "password",
          "value": "password",
          "temporary": false
        }
      ]
    }
```

local/keycloak/jimm-realm.json
ale8k: Should be ok, I didn't rename / change the client for those purposes because of this
https://github.com/ale8k/jimm/blob/d7f33b3ed935b4066bd8faf96d241d86c48ffa9b/local/keycloak/jimm-realm.json#L74-L79

Bit of a shame, would've liked to give it a better name lol.

ale8k:
Noticed I've named the environment key wrong, should be OIDC not OAuth - updating it.

SimoneDutto (approved):

internal/auth/auth.go
```diff
@@ -210,6 +231,101 @@ func NewAuthenticationService(ctx context.Context, params AuthenticationServiceP
...
+		return claims, nil
+	}
+
+	groupClaim, ok := rawClaims[as.groupClaimKey]
+	if !ok {
+		zapctx.Warn(ctx, "configured group claim missing from id token", zap.String("claim-key", as.groupClaimKey))
+		return claims, nil
+	}
+
+	// As the groups claim could be configured in many waysat the IdP
```
kian99:
```suggestion
	// As the groups claim could be configured in many way at the IdP
```

internal/auth/auth.go
```diff
@@ -359,8 +476,8 @@ func (as *AuthenticationService) MintSessionToken(email string) (string, error)
 // on the user's behalf and migrate the model.
 //
 // Currently the token provides the same level of access as the user, but keeping
-// this as a separate method from `MintSessionToken` allows for future changes to
-// the migration token without affecting the session token.
+// this as a separate method from `MintSessionTokenWithGroups` allows for future
+// changes to the migration token without affecting the session token.
 func (as *AuthenticationService) NewMigrationToken(ctx context.Context, username string) (string, error) {
```
kian99: This token might also need to be changed. Anywhere that we mint a user facing token that does not contain a group could start to fail because the user's groups are dictated by the groups in the token rather than group membership in JIMM.

internal/auth/auth.go
```diff
@@ -420,6 +537,36 @@ func (as *AuthenticationService) VerifySessionToken(token string) (_ jwt.Token,
 	return parsedToken, nil
 }
 
+// SessionGroupsFromToken extracts the stable internal groups claim from a
+// verified JIMM session token.
+func SessionGroupsFromToken(token jwt.Token) ([]string, error) {
```
kian99, line 556: If this is a JIMM session token, why are we decoding multiple types of formats for the claim?

kian99 (commented):
Still reviewing, just some initial minor bits.

kian99 (commented):
Looks fine besides the other comments. But the way we are putting the group info on the JIMM session tokens ties the refresh of the group information to the duration of the user's session. That means,
- For CLI logins, the group info is stale for 24 hours.
- For dashboard logins, group info is stale for at most, the duration of the access token JIMM holds. Once we refresh that in `validateAndUpdateAccessToken`, I assume we'll get new group claims, but I'm struggling to see that is true from the change.
- For service account login via the TF provider, we'd need to make some further changes to pass the group info from the token, see `VerifyClientCredentials`. cc @SimoneDutto just checking we have a card for that?

Overall, I think it's all okay, because you can log out and log back into immediately trigger a refresh of your group info. But in the future we may want to consider always saving a user's access and refresh tokens and using those to retrieve group information as frequently as desired by a separate config value. But that's a bit more complicated and I think this approach is a good start.

ale8k:
> Looks fine besides the other comments. But the way we are putting the group info on the JIMM session tokens ties the refresh of the group information to the duration of the user's session. That means,
> 
> * For CLI logins, the group info is stale for 24 hours.
> * For dashboard logins, group info is stale for at most, the duration of the access token JIMM holds. Once we refresh that in `validateAndUpdateAccessToken`, I assume we'll get new group claims, but I'm struggling to see that is true from the change.
> * For service account login via the TF provider, we'd need to make some further changes to pass the group info from the token, see `VerifyClientCredentials`. cc @SimoneDutto just checking we have a card for that?
> 
> Overall, I think it's all okay, because you can log out and log back into immediately trigger a refresh of your group info. But in the future we may want to consider always saving a user's access and refresh tokens and using those to retrieve group information as frequently as desired by a separate config value. But that's a bit more complicated and I think this approach is a good start.

- CLI Logins: Yup, mentioned this to Ales in our 1:1. Unless we do something around revocation or refresh.
- Browser, it'll be help as long as the cookie is set and then yes, refresh token for *some* providers will mint a new oidc token but this isn't guaranteed and would be provider dependent - so it needs solving another way I believe.
- Svc acc - they can't have groups, they aren't users. Not sure what you mean here? I actually noticed we were mistakenly passing scopes in the client cred req and fixed it (again, a machine doesn't have a profile, email, whatnot)

internal/auth/auth.go
ale8k: This is a bit of paranoia as it returns an []any. We could for sure just trust it is []string. But in any case we'd have to convert it so  better safe than sorry I guess?

cmd/jimmsrv/service/service.go
```diff
@@ -83,9 +83,16 @@ type OAuthAuthenticatorParams struct {
 	// /auth and /token requests.
 	ClientSecret string
 
-	// Scopes holds the scopes that you wish to retrieve.
+	// Scopes holds the scopes requested for browser and device OAuth flows.
 	Scopes []string
 
+	// ClientCredentialScopes holds scopes requested for client-credentials flow.
+	ClientCredentialScopes []string
```
kian99, line 90: Remember to add this config to the charm too.

internal/auth/auth.go
```diff
@@ -210,6 +238,82 @@ func NewAuthenticationService(ctx context.Context, params AuthenticationServiceP
...
+func (as *AuthenticationService) extractGroupsFromAccessToken(ctx context.Context, accessToken *oauth2.Token) ([]string, error) {
+	if accessToken == nil || accessToken.AccessToken == "" {
+		return nil, errors.New("access token is empty")
+	}
+
+	if as.groupClaimKey == "" {
+		return nil, nil
+	}
+
+	parsedToken, err := jwt.ParseInsecure([]byte(accessToken.AccessToken))
```
kian99, line 269: Why parse insecure?

internal/auth/auth.go
kian99: I don't personally like paranoid code, it makes it harder to read and understand. At the very least we should comment why it's like this, or better yet, just do `[]string` and write a test to satisfy our paranoia.

internal/auth/auth_test.go
```diff
@@ -263,13 +301,34 @@ func TestVerifyClientCredentials(t *testing.T) {
...
+		//nolint:gosec // Thinks hardcoded credentials.
+		validClientSecret = "d6JVj6NDXYx56muG6ZmjWMLcJnIYQjD0"
+	)
+
+	authSvc, _, _, cleanup := setupTestAuthSvcWithGroupClaimKey(ctx, c, time.Hour, testGroupClaimKey)
+	defer cleanup()
+
+	groups, err := authSvc.VerifyClientCredentials(ctx, validClientID, validClientSecret)
+	c.Assert(err, qt.IsNil)
+	c.Assert(groups, qt.DeepEquals, []string{"canonical"})
```
kian99: In most places I see `jimmtest.OIDCGroupsTestGroupName`, why hardcoded "canonical" here?

local/keycloak/jimm-realm.json
```diff
@@ -45,6 +45,37 @@
...
       ]
+    },
+    {
+      "id": "jimm-group-user-id",
+      "username": "jimm-group-user",
+      "email": "jimm-group-user@canonical.com",
+      "emailVerified": true,
+      "enabled": true,
+      "groups": [
+        "/canonical"
```
kian99, line 56: Out of interest, why is the leading slash needed?

docker-compose.common.yaml
```diff
@@ -41,7 +41,9 @@ services:
       JIMM_OAUTH_ISSUER_URL: "http://keycloak.localhost:8082/realms/jimm" # Scheme required
       JIMM_OAUTH_CLIENT_ID: "jimm-device"
       JIMM_OAUTH_CLIENT_SECRET: "SwjDofnbDzJDm9iyfUhEp67FfUFMY8L4"
-      JIMM_OAUTH_SCOPES: "openid profile email" # Space separated list of scopes
+      JIMM_OAUTH_SCOPES: "openid profile email group" # Space separated list of scopes
+      JIMM_OAUTH_CLIENT_CREDENTIAL_SCOPES: "group" # Space separated list for client-credentials flow
```
kian99, line 45: This makes me wonder, when we update JIMM to use groups from the IdP, the deployment config will need to set the client_creds scopes like this. Should we just default, in the charm, this value to "groups" or are we concerned that will break with other identity providers.

ale8k: Only if it isn't a default scope at the IdP so for instance - if it is, they'll only set the claim key. I'm gonna explain this in the charm config.

local/keycloak/jimm-realm.json
ale8k: This was produced when I used the admin gui and I believe it is because of this [here](https://www.keycloak.org/docs/latest/server_admin/?utm_source=chatgpt.com#building-hierarchies).

And I think it is how they represent nested groups (which by default we also now supprot in our code).

image

ale8k: (It's also how I added the service account as a member of the group btw, had 0 idea reading the keycloak docs as they're huge and just played)

internal/auth/auth_test.go
ale8k: Ngl, this was a human error, will update soon.

internal/auth/auth.go
ale8k: Comment above explains:
```
// extractGroupsFromAccessToken extracts the configured groups claim from an access token.
// The access token is parsed as a JWT without signature verification because it
// originates from the provider token endpoint and we only need claim extraction.
```
 In client cred flow, we've actually just retrieved it via token and in device/browser, we've just validated the ID token. So my presumption was that we're OK to just parse it insecure.

I think we *can* read the JWKS set from the IdP though and verify it. Not 100% on what is right here. Asked Copilot and GPT and they gave conflicting results... Maybe I'm wrong and we should be verifying it?

internal/auth/auth.go
ale8k: Yeah ok I'll drop it and add lil test, will resolve when updated

internal/auth/auth.go
kian99: Ah right, because we fetch this token directly from the IdP as opposed to being given a token from some other means.

ale8k: Exactly, again the JWKS should be in the well known discovery, so we can verify it pretty easily I believe. Let me have a read and I'll come back to this one.

internal/auth/auth.go
```diff
@@ -210,6 +238,82 @@ func NewAuthenticationService(ctx context.Context, params AuthenticationServiceP
...
+				continue
+			}
+			normalizedGroups = append(normalizedGroups, group)
+		}
+		return normalizedGroups
+	}
+
+	if strings.ContainsAny(trimmedValue, " \t\n\r") {
+		return strings.Fields(trimmedValue)
+	}
```
SimoneDutto: i think this can be done simpler?

if you do strings.split(str, ",\s") maybe or something of this sort?

SimoneDutto (approved):
thanks for the explanation yesterday

kian99 (approved):

internal/auth/auth.go
ale8k: Updated with
```go
	// Replace commas with spaces and split on any whitespace.
	// strings.Fields handles trimming and empty string filtering.
	normalised := strings.ReplaceAll(trimmedValue, ",", " ")
	return strings.Fields(normalised)
```

PR #2019, opened 2026-06-10
refactor: authorizeRelationTargetAdmin leveraging openfga conditions

SimoneDutto:
# Description
For context about `conditions` read this recent pr into our `canonical/openfga` here: https://github.com/canonical/ofga/pull/71
There is also a simpler example of `conditions` to help you understand what's going on here better.

The idea is that before we were using go code to check permission and now the logic has been moved to openfga using conditions.

We have two new relations:
- `request_ok` -> trigger the condition to evaluate the context, basically our switch
- `can_manage` -> check request_ok and administrator at the same time

## Questions you might have
- what happens if these new relations are assigned permanently by an admin instead of being contextual tuples? for `request_ok` is not a problem because it will just reevaluate the condition, but if `can_manage` is assigned permanently it will bypass the condition. We have no way currently from the CLI, even for an admin to assign this tuple inadvertently, so it would be an admin using openfga cli to add this tuple to bypass this check, so i don't think it can be a security issue.
- is this a brand new feature? No, it has been released in 1.4 (read the pr linked to check), and we have been running at least 1.8 for a few months in `stable`


kian99 (commented):
Curious what others think about the readability of this change. While I initially liked the idea that it keeps the permission logic inside the auth model, it feels like the code becomes harder to understand, though I admit that might just be my unfamiliarity with the new feature.

ale8k:
> Curious what others think about the readability of this change. While I initially liked the idea that it keeps the permission logic inside the auth model, it feels like the code becomes harder to understand, though I admit that might just be my unfamiliarity with the new feature.

This is why I wanted to setup a meet, I agree that we are programmatically creating policies and it is now supported natively by OpenFGA, which is better because we can enforce the policy via our model - but I do find it difficult to understand, and again agree it might just be unfamiliarity.

SimoneDutto:
there is consensus around preferring go code rather than this feature.

PR #2021, opened 2026-06-11
feat: allow add permission for idp-group

SimoneDutto:
## Description

add idp-group tag and entity handling, the idea is to allow juju jaas add-permission and the corresponding terraform resource to work with idp-group tags.

So:
`juju jaas add-permission idp-group:engineering audit_log_viewer controller-jimm`

and
```
resource "juju_jaas_access_controller" "devops" {
  access  = "administrator"
  idp_groups  = [local.devops_group_id]
}
```

Adding a new entity and a new tag instead of relying on the preexisting group tag makes it easier to:
- handle the future migrations when we will remove group entirely, from jaas and openfga tuples
- don't special case idp groups in our OpenFgaCleanup logic
- more clarity in the code
- at one point we can remove the group-tag and all tests associated to it

The new UX has been shared with Niko from IS and he said he doesn't change much for them.
See https://docs.google.com/document/d/1hUl5QFyPQ8IP4VZSR1fKMRGZ9e8ATE62nRfJfRmVYFI/edit?disco=AAAB9Ff8l14

### DIFF
--- a/internal/jimm/permissions/tagresolver.go
+++ b/internal/jimm/permissions/tagresolver.go
@@ -112,6 +112,8 @@ func (j *PermissionManager) ToJAASTag(ctx context.Context, tag *ofganames.Tag, r
 			return "", fmt.Errorf("failed to fetch group information: %s: %w", group.UUID, err)
 		}
 		return tagToString(jimmnames.GroupTagKind, group.Name), nil
+	case jimmnames.IdPGroupTagKind:
+		return tagToString(jimmnames.IdPGroupTagKind, tag.ID), nil
 	case jimmnames.RoleTagKind:
 		role := dbmodel.RoleEntry{
 			UUID: tag.ID,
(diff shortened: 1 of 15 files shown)


kian99 (approved):

internal/jimm/permissions/tagresolver.go
```diff
@@ -326,6 +348,8 @@ func resolveTag(jimmUUID string, db *db.Database, tag string) (*ofganames.Tag, e
 		return resolver.userTag(ctx)
 	case jimmnames.GroupTagKind:
 		return resolver.groupTag(ctx, db)
+	case jimmnames.IdPGroupTagKind:
```
ale8k, line 351: Ah nice we did my tag idea : D mucho loveo

ale8k (approved):
lgtm nice!!

PR #2022, opened 2026-06-12
feat: inject contextual tuples from jwt groups

SimoneDutto:
## Description

idp groups are injected into the user obj to then use it as contextual tuples.

The groups are injected in the ctx in the auth path, and the groups are injected in the `openfga.User` at login time.

Then those idp-groups are used as contextual tuples during openfga evaluation.

### DIFF
--- a/internal/auth/auth.go
+++ b/internal/auth/auth.go
@@ -77,6 +77,7 @@ const (
 )
 
 type sessionIdentityContextKey struct{}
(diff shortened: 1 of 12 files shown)


ale8k (approved):
Had an offline call, lgtm. If we've missed any areas, we can ad-hoc fix them.

kian99 (approved):

PR #2023, opened 2026-06-12
Migrate v3

tmihoc:
Part of our docs migration to canonical.com/juju/docs/jaas.

For reference see https://canonical-rtd-proxy.readthedocs-hosted.com/how-to/migrate/ 

The changes to the content files are about spellcheck or linkcheck fixes (so the CI tests pass). The `.customwordlist` temporarily includes spelling errors that occur in the `jaas` plugin and which we should try to fix at source.

### DIFF
--- a/docs/.sphinx/_static/js/overwrite_links.js
+++ b/docs/.sphinx/_static/js/overwrite_links.js
@@ -1,15 +1,37 @@
+// Replace oldDomain with newDomain
 const oldDomain = 'canonical-jaas-1.readthedocs-hosted.com';
 const newDomain = 'canonical.com/juju/docs/jaas';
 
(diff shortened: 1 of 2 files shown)


docs/.sphinx/_static/js/overwrite_links.js
SimoneDutto, line 1: is this file the default way of doing rewrites? it seems the wrong place to do it in js

tmihoc: It's part of the recipe for migrating to the new domain.

kian99 (approved):

SimoneDutto (approved):

PR #2025, opened 2026-06-15
feat: enable tracing

SimoneDutto:
## Description

In this PR we add support for tracing to JAAS and JUJU.

The idea is that JAAS starts the traces and propagates them to the juju backing controller, so for each call you can see the distributed trace.

Tempo is different from Prometheus, so it's JAAS pushing the spans to tempo, not tempo pulling them.

spans for our managers are missing, but i couldn't find a nice way to add them without manually putting the spans everywhere, so i'll leave it for another pr.

This piece of work is divided in 3 commits to be much easier to review:
1. setup of tracer and utils methods to add spans
2. adding the span where i though they would be helpful
3. changes to qa script and docker to enable tracing on a local setup. All the grafana stuff has been completely AI generated, and i've just review the final output (the dashboard) because i don't think it's important what the grafana dashboard looks like since it will be defined by the JAAS operator, not us.

> Note: for testing i was very unsure what to do: the failure mode for spans is to fail to a nil tracer which does nothing, so i've tested the propagation of spans to juju calls. But feel free to suggest.

## QA and Examples

`ENABLE_TRACING=true make qa-lxd`

This is the dashboard with the top-level traces.
image

This is a single trace.

image


internal/rpcproxy/rpcproxy.go
```diff
@@ -399,35 +408,80 @@ func (p *clientProxy) start(ctx context.Context) error {
...
+				msg.span = nil
+			}
 			p.sendError(ctx, p.src, msg, err)
 			p.msgs.removeMessage(msg.RequestID)
 			continue
 		}
 	}
 }
 
+func (p *clientProxy) handleClientAdminMessage(ctx, msgCtx context.Context, msg *message) (*message, bool) {
```
SimoneDutto: i've just put it into a func because the linter was complaining about cognitive difficulty

cmd/jimmsrv/service/service.go
```diff
@@ -558,6 +574,10 @@ func NewServiceFromDependencies(ctx context.Context, deps *ServiceDependencies)
 	if err := deps.Validate(); err != nil {
 		return nil, fmt.Errorf("invalid service dependencies: %w", err)
 	}
+	if deps.Tracer == nil {
+		deps.Tracer = jujuTrace.NoopTracer{}
+	}
+	ctx = jujuTrace.WithTracer(ctx, deps.Tracer)
```
kian99, line 580: Does this just place the tracer onto the context? And is that useful here?

internal/telemetry/juju_adapter.go
```diff
@@ -0,0 +1,142 @@
...
+	jujuTrace "github.com/juju/juju/core/trace"
+	"go.opentelemetry.io/otel/codes"
+	oteltrace "go.opentelemetry.io/otel/trace"
+)
+
+type otelTracer struct {
+	tracer oteltrace.Tracer
+}
+
+func (t otelTracer) Start(ctx context.Context, name string, options ...jujuTrace.Option) (context.Context, jujuTrace.Span) {
```
kian99, line 18: Missing godocs in this file for exported methods

internal/jujuapi/streamcontrollerproxy.go
```diff
@@ -31,7 +31,8 @@ const (
 // Messages are handled by proxying them between the
 // controller and client.
 type streamControllerProxier struct {
-	jimm *jimm.JIMM
+	jimm   *jimm.JIMM
+	params Params
```
kian99, line 35: What about just passing the tracer, since there are a lot of things on `params`.

cmd/jimmsrv/service/service.go
```diff
@@ -802,13 +802,91 @@ func openDB(ctx context.Context, dsn string, logSQL bool) (*gorm.DB, error) {
...
+	if err := database.Callback().Update().After("gorm:update").Register("jimm:trace_after", afterDatabaseTrace("update")); err != nil {
+		return fmt.Errorf("register update after callback: %w", err)
+	}
+	if err := database.Callback().Delete().Before("gorm:delete").Register("jimm:trace_before", beforeDatabaseTrace("delete")); err != nil {
+		return fmt.Errorf("register delete before callback: %w", err)
+	}
+	if err := database.Callback().Delete().After("gorm:delete").Register("jimm:trace_after", afterDatabaseTrace("delete")); err != nil {
+		return fmt.Errorf("register delete after callback: %w", err)
+	}
+	if err := database.Callback().Row().Before("gorm:row").Register("jimm:trace_before", beforeDatabaseTrace("row")); err != nil {
```
kian99, line 846: What's the difference between `row` and `query`?
I assume these lines are capturing all the basic gorm operations.

internal/openfga/openfga.go
kian99, line 1: The changes in this file feel like they highlight the benefits we'd get from an interface between our domain and the auth layer, wdyt? Should we create a card to add an interface on top of the OFGA stuff?

internal/rpcproxy/rpcproxy.go
kian99, line 1: This file definitely wasn't built with much extensibility in mind (I should know, I wrote it). While I have no doubt the additions are correct, there's quite a lot going on here. What do you think about holding back the tracing from this file and doing some rework to make it easier to add in tracing?

cmd/jimmsrv/service/service.go
SimoneDutto: https://github.com/go-gorm/gorm/blob/master/callbacks/row.go#L7

yeah, these lines capture usual gorm operation.
you can take a look at it here:
https://gorm.io/docs/write_plugins.html

internal/openfga/openfga.go
SimoneDutto: agree! the same should go for manager

internal/rpcproxy/rpcproxy.go
SimoneDutto: while i agree that this is most confusing place to put the spans, it's indeed very useful, because you can trace proxied calls, evaluated jimm's overhead when it just proxies stuff. And i'd prefer to leave the spans here, and have a card to refactor it, rather than removing it all together. Wdyt?

internal/telemetry/juju_adapter.go
SimoneDutto: ups

internal/jujuapi/streamcontrollerproxy.go
SimoneDutto: i don't really mind, it felt a good addition to the generic `Params`

cmd/jimmsrv/service/service.go
SimoneDutto: yeah, the higher you place it in the start the sooner you can create spans. Basically here it's top level, so we could potentially add spans in all the methods below.

cmd/jimmsrv/service/service.go
```diff
@@ -494,10 +506,14 @@ func NewServiceDependencies(ctx context.Context, p Params) (*ServiceDependencies
 		Client:                        jimm.NewDialerAdapter(dialer),
 		RiverClient:                   riverClient,
 		OpenFGAClient:                 openFGAclient,
+		Tracer:                        tracer,
 		CredentialStore:               credentialStore,
 		JWTService:                    jwtService,
 		JWKSService:                   jwksService,
 	}
+	deps.cleanupFuncs = append(deps.cleanupFuncs, func() error {
+		return shutdownTracer(context.Background())
```
alesstimec, line 515: if we error out before we start, we'll never call shutdownTracer. is this an issue?

internal/jujuapi/admin.go
```diff
@@ -83,11 +85,19 @@ func (r *controllerRoot) GetDeviceSessionToken(ctx context.Context) (params.GetD
...
+	)
+	defer func() {
+		span.Finish(err)
+	}()
 
 	user, err := r.jimm.LoginManager().LoginWithSessionCookie(ctx, r.identityId)
 	if err != nil {
-		return jujuparams.LoginResult{}, errors.Codef(errors.CodeUnauthorized, "%w", err)
+		err = errors.Codef(errors.CodeUnauthorized, "%w", err)
+		return result, err
```
alesstimec:
```suggestion
		return result, errors.Codef(errors.CodeUnauthorized, "%w", err)
```

internal/rpcproxy/rpcproxy.go
```diff
@@ -399,35 +408,80 @@ func (p *clientProxy) start(ctx context.Context) error {
...
-				msg = toController
-				p.msgs.addLoginMessage(toController)
-			}
+		msgCtx := ctx
+		if msg.TraceID != "" && msg.SpanID != "" {
+			msgCtx = jujuTrace.WithTraceScope(msgCtx, msg.TraceID, msg.SpanID, msg.TraceFlags)
 		}
+		msgCtx, span := telemetry.StartSpan(
+			msgCtx,
+			"jimm.juju-proxy",
```
alesstimec:
```suggestion
			"jimm.model-proxy",
```

internal/rpcproxy/rpcproxy.go
```diff
@@ -244,6 +249,10 @@ func (msgs *inflightMsgs) removeMessage(msgID uint64) {
 	msgs.mu.Unlock()
 
 	if ok {
+		if req.span != nil {
+			req.span.Finish(nil)
```
alesstimec: this looks like an odd place to finish a span.. are you sure?

internal/rpcproxy/rpcproxy.go
alesstimec: since you're passing in a pointer to the message i don't think you need to return it.. it's the same thing you're returning, but you've modified it.. and since it's a pointer, you're modifying what you passed in and then returning it..

internal/telemetry/juju_adapter.go
```diff
@@ -0,0 +1,142 @@
...
+	if t.tracer == nil {
+		return ctx, jujuTrace.NoopSpan{}
+	}
+
+	ctx = remoteParentFromScope(ctx)
+
+	spanOptions := []oteltrace.SpanStartOption{
+		oteltrace.WithSpanKind(oteltrace.SpanKindServer),
```
ale8k, line 26: Do we want 3 kinds?

1. Server for handling the WS entirely
2. Client for our juju client calls
3. Internal for anything we deem worthy of spanning / processing heavy

internal/telemetry/utils.go
```diff
@@ -0,0 +1,61 @@
...
+
+func traceAttributes(options ...jujuTrace.Option) []jujuTrace.Attribute {
+	traceOptions := jujuTrace.NewTracerOptions()
+	for _, option := range options {
+		option(traceOptions)
+	}
+	return traceOptions.Attributes()
+}
+
+func toOTelAttributes(attributes []jujuTrace.Attribute) []attribute.KeyValue {
```
ale8k, line 55: Shame they don't export the trace options & conversion in here https://github.com/juju/juju/blob/main/internal/worker/trace/tracer.go#L308

Maybe we could move it out of internal?

internal/telemetry/tracing.go
```diff
@@ -0,0 +1,97 @@
...
+		return otlpgrpc.New(ctx, grpcEndpointOption(params.Endpoint))
+	case "http/protobuf":
+		return otlphttp.New(ctx, httpEndpointOption(params.Endpoint))
+	default:
+		return nil, fmt.Errorf("unsupported OTLP trace protocol %q", traceProtocol(params))
+	}
+}
+
+func grpcEndpointOption(endpoint string) otlpgrpc.Option {
+	if strings.Contains(endpoint, "://") {
```
ale8k: Bit hacky, could be an invalid URL, URL parse instead

internal/telemetry/tracing.go
```diff
@@ -0,0 +1,97 @@
...
+	zapctx.Info(ctx, "otlp tracing enabled", zap.String("protocol", traceProtocol(params)))
+
+	return otelTracer{tracer: provider.Tracer(tracerName)}, provider.Shutdown, nil
+}
+
+func tracingConfigured(params Params) bool {
+	return strings.TrimSpace(params.Endpoint) != ""
+}
+
+func traceProtocol(params Params) string {
```
ale8k: Not necessary right? Charm will set defaults? Same for all defaults I guess? And if we think it is necessary, can we do the options pattern instead for NewTracer - it'll be cleaner

internal/telemetry/tracing.go
```diff
@@ -0,0 +1,97 @@
...
+		return nil, nil, err
+	}
+
+	providerOptions := []sdktrace.TracerProviderOption{sdktrace.WithBatcher(exporter)}
+	if params.SampleRatio != nil {
+		providerOptions = append(providerOptions, sdktrace.WithSampler(
+			sdktrace.ParentBased(sdktrace.TraceIDRatioBased(*params.SampleRatio)),
+		))
+	}
+	if serviceName := strings.TrimSpace(params.ServiceName); serviceName != "" {
```
ale8k, line 90: Put a func on the params struct called "Verify" or something which checks endpoint is valid, protocol, ratio, etc.

internal/telemetry/tracing.go
```diff
@@ -0,0 +1,97 @@
...
+type Params struct {
+	ServiceName string
+	Endpoint    string
+	Protocol    string
+	SampleRatio *float64
+}
+
+// NewTracer returns a Juju-compatible tracer backed by OpenTelemetry when OTLP
+// export is configured.
+func NewTracer(ctx context.Context, params Params) (jujuTrace.Tracer, func(context.Context) error, error) {
```
ale8k, line 71: I think this needs more detail on the godoc explaining what it is to be Juju compatible

internal/jujuapi/websocket.go
```diff
@@ -222,7 +226,7 @@ func controllerConnectionFunc(s apiModelProxier, jwtGenerator *jujuauth.LoginTok
 				Valid:  uuid != "",
 			},
 		}
-		if err := s.jimm.Database.GetModel(context.Background(), &m); err != nil {
+		if err := s.jimm.Database.GetModel(ctx, &m); err != nil {
```
luci1900, line 229: That's interesting. Is there a reason we were using background before?

internal/openfga/openfga.go
```diff
@@ -183,6 +185,10 @@ func (o *OFGAClient) listObjects(ctx context.Context, user *Tag, relation Relati
 // AddRelation adds given relations (tuples).
 func (o *OFGAClient) AddRelation(ctx context.Context, tuples ...Tuple) (err error) {
 	const op = "openfga.AddRelation"
+	ctx, span := telemetry.StartSpan(ctx, "jimm.openfga", jujuTrace.StringAttr("openfga.operation", strings.TrimPrefix(op, "openfga.")))
```
luci1900, line 188: Can these not be pushed down into the ofga client?

internal/telemetry/juju_adapter.go
```diff
@@ -0,0 +1,142 @@
...
+
+	otelTraceID, err := oteltrace.TraceIDFromHex(traceID)
+	if err != nil {
+		return ctx
+	}
+	otelSpanID, err := oteltrace.SpanIDFromHex(spanID)
+	if err != nil {
+		return ctx
+	}
+	if flags < 0 || flags > 0xff {
```
ale8k: Why not just >255?

internal/telemetry/juju_adapter.go
```diff
@@ -0,0 +1,142 @@
...
+	ctx = jujuTrace.WithSpan(ctx, wrapped)
+
+	return ctx, wrapped
+}
+
+func (t otelTracer) Enabled() bool {
+	return t.tracer != nil
+}
+
+func remoteParentFromScope(ctx context.Context) context.Context {
```
ale8k, line 49: Why not just call it ContextWithRemoteSpanContext same as otel?

luci1900 (commented):
Looks good, tracing is very useful.

I don't mind manager methods not having spans. As we see gaps in traces, we can decide where to add more spans later.

luci1900 (approved):

cmd/jimmsrv/main.go
```diff
@@ -200,6 +201,27 @@ func start(ctx context.Context, s *service.Service) error {
...
+	}
+	traceProtocol := strings.TrimSpace(os.Getenv("OTEL_EXPORTER_OTLP_TRACES_PROTOCOL"))
+	if traceProtocol == "" {
+		traceProtocol = strings.TrimSpace(os.Getenv("OTEL_EXPORTER_OTLP_PROTOCOL"))
+	}
+	traceServiceName := strings.TrimSpace(os.Getenv("OTEL_SERVICE_NAME"))
+	var traceSampleRatio *float64
+	traceSampleRatioRaw := strings.TrimSpace(os.Getenv("OTEL_TRACES_SAMPLE_RATIO"))
+	if traceSampleRatioRaw != "" {
+		ratio, err := strconv.ParseFloat(traceSampleRatioRaw, 64)
```
ale8k, line 216: Should we just rely on charm defaults?

cmd/jimmsrv/service/service.go
```diff
@@ -802,13 +802,91 @@ func openDB(ctx context.Context, dsn string, logSQL bool) (*gorm.DB, error) {
...
+		return nil, err
+	}
+	if err := registerDatabaseTracing(database); err != nil {
+		return nil, fmt.Errorf("failed to register database tracing: %w", err)
+	}
+	return database, nil
+}
+
+func registerDatabaseTracing(database *gorm.DB) error {
+	if err := database.Callback().Create().Before("gorm:create").Register("jimm:trace_before", beforeDatabaseTrace("create")); err != nil {
```
ale8k, line 822: This is an example where it is an internal trace, not server trace right? And if we don't differentiate them you can't do the type queries?

internal/jujuapi/rpc/root.go
```diff
@@ -127,5 +129,12 @@ func (c rootMethodCaller) Call(ctx context.Context, objID string, arg reflect.Va
 	ctx = zapctx.WithFields(ctx, zap.String("facade", c.facadeName))
 	ctx = zapctx.WithFields(ctx, zap.String("method", c.methodName))
 	ctx = zapctx.WithFields(ctx, zap.Int("version", c.version))
-	return c.MethodCaller.Call(ctx, objID, arg)
+	ctx, span := telemetry.StartSpan(ctx, "jimm.facade")
+	result, err := c.MethodCaller.Call(ctx, objID, arg)
+	span.Finish(err,
```
ale8k: Can call panic? Do we wanna defer? Or is it we can't defer because of evaluation? (I don't think it can looking at it but maybe it can and I've missed it)

internal/jujuapi/admin.go
```diff
@@ -83,11 +85,19 @@ func (r *controllerRoot) GetDeviceSessionToken(ctx context.Context) (params.GetD
 //
 // It may be misleading in that it does not interact with cookies at all, but this will only ever
 // be successful upon the http layer login being successful.
-func (r *controllerRoot) LoginWithSessionCookie(ctx context.Context) (jujuparams.LoginResult, error) {
+func (r *controllerRoot) LoginWithSessionCookie(ctx context.Context) (result jujuparams.LoginResult, err error) {
+	ctx, span := telemetry.StartSpan(ctx, "jimm.auth",
```
ale8k: Isn't this captured by our Call() all-in-one spanner?  Why do we need it explicitly for this facade method?

internal/jujuclient/dial.go
```diff
@@ -88,7 +90,15 @@ func (d *Dialer) createLoginRequest(ctx context.Context, ctl *dbmodel.Controller
 }
 
 // Dial implements jimm.Dialer.
-func (d *Dialer) Dial(ctx context.Context, ctl *dbmodel.Controller, modelTag names.ModelTag, user *openfga.User) (*Connection, error) {
+func (d *Dialer) Dial(ctx context.Context, ctl *dbmodel.Controller, modelTag names.ModelTag, user *openfga.User) (_ *Connection, err error) {
+	ctx, span := telemetry.StartSpan(ctx, "jimm.juju-dial")
+	defer func() {
+		span.Finish(err,
```
ale8k, line 96: Sometimes we defer the finish sometimes not, I think cases where we don't can we godoc it? It isn't clear why

internal/jujuclient/dial.go
```diff
@@ -88,7 +90,15 @@ func (d *Dialer) createLoginRequest(ctx context.Context, ctl *dbmodel.Controller
 }
 
 // Dial implements jimm.Dialer.
-func (d *Dialer) Dial(ctx context.Context, ctl *dbmodel.Controller, modelTag names.ModelTag, user *openfga.User) (*Connection, error) {
+func (d *Dialer) Dial(ctx context.Context, ctl *dbmodel.Controller, modelTag names.ModelTag, user *openfga.User) (_ *Connection, err error) {
+	ctx, span := telemetry.StartSpan(ctx, "jimm.juju-dial")
```
ale8k, line 94: Also I'm not sure this is all that useful, we don't know what the call was right? Or do we?  I also see we have client spans which will tell us the exact details?

internal/rpc/client.go
```diff
@@ -169,6 +179,15 @@ func (c *Client) Call(ctx context.Context, facade string, version int, id, metho
 		Request: method,
 		Params:  json.RawMessage(argsb),
 	}
+	if traceID, spanID, flags, ok := jujuTrace.ScopeFromContext(ctx); ok {
```
ale8k: I'm not following this, I'm guessing this is like client > 4 (or w/e version has the tracing), expect it else do our own

If that is it can we godoc it please (and possibly can we make a utility func for this and put it in our tracing package? We'll definitely do this more in the future?)

internal/rpc/client.go
```diff
@@ -11,11 +11,14 @@ import (
 	"time"
```
ale8k, line 12: We're missing the proxy method in httprpoxy or w/e it is called right?

internal/rpc/dial.go
```diff
@@ -38,12 +40,25 @@ func (d Dialer) Dial(ctx context.Context, url string, headers http.Header) (*Cli
 }
 
 // DialWebsocket dials a url and returns a websocket.
-func (d Dialer) DialWebsocket(ctx context.Context, url string, headers http.Header) (*websocket.Conn, error) {
+func (d Dialer) DialWebsocket(ctx context.Context, urlString string, headers http.Header) (_ *websocket.Conn, err error) {
+	endpoint := urlString
+	path := ""
+	if u, parseErr := url.Parse(urlString); parseErr == nil {
```
ale8k, line 46: Is this so we can see the diff between WS initiation and a Call() made?

internal/rpcproxy/rpcproxy.go
```diff
@@ -399,35 +408,80 @@ func (p *clientProxy) start(ctx context.Context) error {
...
-				continue
-			} else if toController != nil {
-				msg = toController
-				p.msgs.addLoginMessage(toController)
-			}
+		msgCtx := ctx
+		if msg.TraceID != "" && msg.SpanID != "" {
+			msgCtx = jujuTrace.WithTraceScope(msgCtx, msg.TraceID, msg.SpanID, msg.TraceFlags)
 		}
+		msgCtx, span := telemetry.StartSpan(
```
ale8k: I see we're making a child context here, but could we call the var childMsgCtx or something, thrown me off a little bit initially lol

internal/rpcproxy/rpcproxy.go
```diff
@@ -399,35 +408,80 @@ func (p *clientProxy) start(ctx context.Context) error {
...
+			msg.TraceID = scope.TraceID()
+			msg.SpanID = scope.SpanID()
+			msg.TraceFlags = scope.TraceFlags()
+		}
+
+		forwardMsg, forward := p.handleClientAdminMessage(ctx, msgCtx, msg)
+		if !forward {
+			continue
+		}
+		msg = forwardMsg
```
ale8k: Why? We make a new message every loop anyway?

internal/rpcproxy/rpcproxy.go
```diff
@@ -399,35 +408,80 @@ func (p *clientProxy) start(ctx context.Context) error {
...
+			msgCtx = jujuTrace.WithTraceScope(msgCtx, msg.TraceID, msg.SpanID, msg.TraceFlags)
 		}
+		msgCtx, span := telemetry.StartSpan(
+			msgCtx,
+			"jimm.juju-proxy",
+			jujuTrace.StringAttr("rpc.facade", msg.Type),
+			jujuTrace.IntAttr("rpc.version", msg.Version),
+			jujuTrace.StringAttr("rpc.method", msg.Request),
+		)
+		msg.span = &span
```
ale8k: Ngl, this reads super confusing how putting the span on the message. It taken me a few reads to realise it's just capturing the admin message with the span details populated so we can end the span

ale8k (commented):
Looks sensible, I've just got a lot of questions...

cmd/jimmsrv/main.go
SimoneDutto: we are, this is just parsing the float, and validating it. Not setting a default

cmd/jimmsrv/service/service.go
SimoneDutto: this is the godoc for shutdown.
```
Shutdown shuts down TracerProvider. All registered span processors are shut down in the order they were registered and any held computational resources are released. After Shutdown is called, all methods are no-ops.
```

So if we error out before we start is not an issue in general, because we don't have any traces started yet.

internal/jujuapi/admin.go
SimoneDutto: yeah, probably we don't need this

internal/jujuapi/rpc/root.go
SimoneDutto: yeah, we can defer it

internal/jujuapi/websocket.go
SimoneDutto: i think it was an oversight

internal/rpcproxy/rpcproxy.go
kian99: I'm not sure I'm desperate enough to have tracing to start sacrificing readability over it instead. We weren't adding tracing to track down a significant issue, so I think it would be worth our time to do a pass over this file and make it easier to hook in the tracing, before we actually add the tracing.

But if others feel differently then we can do the refactoring later.

internal/jujuclient/dial.go
SimoneDutto: it would be a child span in the facade call to juju, and it highlight how much it takes just to dial the controller, without login or anything like that. It's the closest we have to "latency to a controller", i think it's pretty useful

internal/jujuclient/dial.go
SimoneDutto: most of them are in defer, when it's not deferred it's because we want to make sure when we start/end the span manually. I wouldn't add a godoc to it because it would be something like "manually ending the span", and i think it's useless/doesn't matter too much

internal/openfga/openfga.go
SimoneDutto: maybe we do something in our canonical/ofga library, but from jimm code i didn't see any way of doing it

internal/rpc/client.go
SimoneDutto: yeah, i'm putting this one in a separate utils method

internal/rpc/dial.go
SimoneDutto: yes

internal/rpcproxy/rpcproxy.go
SimoneDutto: yes, we finish the span "spanning" from jimm auth to the actual proxing of the message to the backing controller

internal/rpcproxy/rpcproxy.go
SimoneDutto: yeah

internal/rpcproxy/rpcproxy.go
SimoneDutto: yeah, that's the purpose

internal/rpcproxy/rpcproxy.go
SimoneDutto: the code seems to return a new message sometimes here, but then looking at it it seems it's mutating the msg pointer again. However i don't feel too comfortable making this change
`toClient, toController, err := p.handleAdminFacade(ctx, msg)`

internal/telemetry/juju_adapter.go
SimoneDutto: yes, reading this is interesting. https://opentelemetry.io/docs/concepts/signals/traces/#span-kind
However i don't think we have that many spans that we need to have this different now, but i'm happy if we have another pr expanding these utils methods to add the kind as well.

internal/telemetry/juju_adapter.go
SimoneDutto: it's just because the juju trace lib calls them scope

SimoneDutto: `jujuTrace.Scope` i don't know why they are always re-inventing the wheel

internal/telemetry/juju_adapter.go
SimoneDutto: i guess the ai liked the hex format

internal/telemetry/tracing.go
SimoneDutto: there are no details, it's just shimming otel into juju trace to make it more consistent

internal/telemetry/utils.go
SimoneDutto: feel free to create a card for it

cmd/jimmsrv/service/service.go
kian99: Since this context is only used for initialization, I assume you mean that we can add spans to track how long certain things take to start? Does that work properly in the context of spans when the init stuff usually only happens once? Might be valuable for functions that spin off a go routine though.

SimoneDutto: yeah, if we start ssh server or something else and we want to add traces at least we have it available

internal/rpc/client.go
SimoneDutto: ah yeah, we can add it in the future

internal/rpcproxy/rpcproxy.go
SimoneDutto: removed

cmd/jimmsrv/service/service.go
```diff
@@ -781,13 +802,91 @@ func openDB(ctx context.Context, dsn string, logSQL bool) (*gorm.DB, error) {
...
```
alesstimec, line 822: i'd suggest creating consts for:
- gorm:create
- gorm:query
- ...
- jimm:trace_before
- jimm:trace_after

internal/telemetry/tracing.go
```diff
@@ -63,35 +103,78 @@ func tracingConfigured(params Params) bool {
...
 }
 
-func httpEndpointOption(endpoint string) otlphttp.Option {
-	if strings.Contains(endpoint, "://") {
-		return otlphttp.WithEndpointURL(endpoint)
+// TraceScopeFromContext returns trace metadata from either the Juju trace
+// scope or the legacy juju/rpc tracing context.
+//
+// JIMM still needs to accept both context formats so older request paths can
+// continue to propagate trace IDs while the newer trace scope is present.
```
kian99: I don't follow this comment. As if we're trying to keep backwards compatibility with something?

testing/modelupgrader_test.go
```diff
@@ -129,6 +121,7 @@ func TestAbortModelUpgrade(t *testing.T) {
 	c := qt.New(t)
 	s := jimmtest.SetupJimmWithControllers(c)
 	model := s.CreateModelForBob(c)
+	SkipIfControllerAgentVersionGreaterThan(c, model.Controller.AgentVersion, "4.0.0")
```
kian99, line 124: Why have these all been skipped?

kian99 (approved):

testing/modelupgrader_test.go
SimoneDutto: i've put it in the commit message:
"in juju 4 they will not support create a model for version lower than the controller, so these tests
can't be run in juju 4"

so, we would need to upgrade the controller to test this facade in juju, and seems overkill/i'm not sure how to do it

internal/telemetry/juju_adapter.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	return oteltrace.ContextWithRemoteSpanContext(ctx, spanContext)
+}
+
+type otelSpan struct {
+	span oteltrace.Span
+}
+
+func (s otelSpan) Scope() jujuTrace.Scope {
```
alesstimec, line 90: godoc

internal/telemetry/juju_adapter.go
```diff
@@ -0,0 +1,144 @@
...
+	spanContext := s.span.SpanContext()
+	if !spanContext.IsValid() {
+		return jujuTrace.NoopScope{}
+	}
+	return otelScope{spanContext: spanContext}
+}
+
+func (s otelSpan) AddEvent(name string, attributes ...jujuTrace.Attribute) {
```
alesstimec, line 99: godoc

internal/telemetry/juju_adapter.go
```diff
@@ -0,0 +1,144 @@
...
+	attrs := toOTelAttributes(attributes)
+	if len(attrs) == 0 {
+		s.span.AddEvent(name)
+		return
+	}
+	s.span.AddEvent(name, oteltrace.WithAttributes(attrs...))
+}
+
+func (s otelSpan) RecordError(err error, attributes ...jujuTrace.Attribute) {
```
alesstimec, line 109: godoc

internal/telemetry/juju_adapter.go
```diff
@@ -0,0 +1,144 @@
...
+	attrs := toOTelAttributes(attributes)
+	if len(attrs) == 0 {
+		s.span.RecordError(err)
+	} else {
+		s.span.RecordError(err, oteltrace.WithAttributes(attrs...))
+	}
+	s.span.SetStatus(codes.Error, err.Error())
+}
+
+func (s otelSpan) End(attributes ...jujuTrace.Attribute) {
```
alesstimec, line 123: godoc

internal/telemetry/juju_adapter.go
```diff
@@ -0,0 +1,144 @@
...
+		s.span.SetAttributes(attrs...)
+	}
+	s.span.End()
+}
+
+type otelScope struct {
+	spanContext oteltrace.SpanContext
+}
+
+func (s otelScope) TraceID() string {
```
alesstimec, line 136: godocs on exported methods

cmd/jimmsrv/service/service.go
SimoneDutto: why? it's use a single time

internal/telemetry/tracing.go
SimoneDutto: it was wrong. Basically `TracingFromContext` is a rpc method to extract the tracing information from rpc calls inside of juju, it has nothing to do with the mechanism we are implementing here. Removed.

cmd/jimmsrv/service/service.go
alesstimec: i see many occurrences of these strings in this function

PR #2026, opened 2026-06-15
feat: add max of idp groups per user

SimoneDutto:
## Description

openfga supports max of 20 contextual tuples, so we need to return an error if there are >20 idp groups for a user. If somebody will complain about this we will find a solution, but for now we just return an error.

The only 2 path to check are:
- device flow
- client id and secret
- browser flow
Because those are the only place where we check the groups coming from the idp.

### DIFF
--- a/internal/auth/auth.go
+++ b/internal/auth/auth.go
@@ -55,6 +55,7 @@ const (
 
 	// migrationTokenExpiry is the expiry time for migration tokens.
 	migrationTokenExpiry = 3 * time.Hour
+	idpGroupLimit        = 20
 )
 
 // AuthStyle determines how the client credentials are sent to the token endpoint.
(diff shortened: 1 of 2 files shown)


internal/jimm/login/login.go
```diff
@@ -160,6 +165,10 @@ func (j *LoginManager) LoginClientCredentials(ctx context.Context, clientID stri
 		logger.LogFailedLogin(ctx, clientIdWithDomain)
 		return nil, errors.Codef(errors.CodeFatalLoginError, "%w", err)
 	}
+	if len(groups) > maxIDPGroups {
+		logger.LogFailedLogin(ctx, clientIdWithDomain)
+		return nil, errors.Codef(errors.CodeUnauthorized, "authorization denied: IDP group claim contains %d groups, maximum supported is %d", len(groups), maxIDPGroups)
```
ale8k: Can make a constant for the string, also I wouldn't mention IdP here just says groups is too many?

internal/jimm/login/login.go
```diff
@@ -131,6 +133,9 @@ func (j *LoginManager) GetDeviceSessionToken(ctx context.Context, deviceOAuthRes
 	if err != nil {
 		return "", err
 	}
+	if len(claims.Groups) > maxIDPGroups {
```
ale8k: Nil check first maybe...

ale8k (commented):
As discussed browser flow needs covering. if we can do it in verify great - if not just need to cover browser flow

internal/jimm/login/login.go
SimoneDutto: i think it's right to mention the idp claim, it's very explicit

ale8k (approved):
Nice, lgtm. Now we know all 3 flows are covered.

internal/auth/auth.go
```diff
@@ -55,6 +55,7 @@ const (
 
 	// migrationTokenExpiry is the expiry time for migration tokens.
 	migrationTokenExpiry = 3 * time.Hour
+	idpGroupLimit        = 20
```
luci1900, line 58: I would say why 20, someone might one day change it without having the context.

SimoneDutto: right

SimoneDutto: i've added to the contextual tuple method, but here it won't hurt

luci1900 (approved):

internal/auth/auth.go
SimoneDutto: adding it in my next pr

PR #2027, opened 2026-06-16
feat: deprecate AddGroup and AddMemberToGroup

SimoneDutto:
## Description
- deprecate adding new groups and new members to group
- change the tests needing groups to not use client methods anymore by adding groups to the db directly
- add some tests for idp group permission propagation

Honestly for the rebac_admin i was wondering if we need all of that code to begin with, i guess nobody is really using it.

### DIFF
--- a/internal/auth/auth.go
+++ b/internal/auth/auth.go
@@ -55,7 +55,10 @@ const (
 
 	// migrationTokenExpiry is the expiry time for migration tokens.
 	migrationTokenExpiry = 3 * time.Hour
-	idpGroupLimit        = 20
+	// idpGroupLimit is the maximum number of groups that will be accepted from an identity provider
+	// in the configured group claim. Since we are handling idp groups as contextual tuples, and OpenFGA
+	// supports a max of 20 contextual tuples, this is the limit for now.
+	idpGroupLimit = 20
 )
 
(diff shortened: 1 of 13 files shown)


kian99 (commented):
What about moving the deprecation error to the API layer instead of the domain?

SimoneDutto:
> What about moving the deprecation error to the API layer instead of the domain?

this was my first instinct but i feel like a layer down, which is shared between rebac and controllerroot is more correct

SimoneDutto:
> > What about moving the deprecation error to the API layer instead of the domain?
> 
> this was my first instinct but i feel like a layer down, which is shared between rebac and controllerroot is more correct

also it doesn't change much because we'll remove the group code soon

kian99:
> > What about moving the deprecation error to the API layer instead of the domain?
> 
> this was my first instinct but i feel like a layer down, which is shared between rebac and controllerroot is more correct

I don't mind it being duplicated in both API handlers. Mainly because you don't need to change 2 layers of tests.

SimoneDutto:
This is in pause and we might even consider just not landing this at all, and suggest ppl to upgrade to jaas-v4 once they are ready where groups are dropped, since the time when they will adopt idpgroup seems very close to jaas-v4 adoption

PR #2029, opened 2026-06-18
feat: add spans to model-proxy

SimoneDutto:
## Description
Add spans for proxied messages, it turned out simpler than i've originally done it.

# QA
Now you can see model proxy spans in the grafana dashboard
image


internal/rpcproxy/rpcproxy.go
```diff
@@ -225,16 +227,35 @@ func (msgs *inflightMsgs) getLoginMessage() *message {
...
+		jujuTrace.StringAttr("rpc.facade", msg.Type),
+		jujuTrace.IntAttr("rpc.version", msg.Version),
+		jujuTrace.StringAttr("rpc.method", msg.Request),
+	)
+	msg.span = &childSpan
+
+	// Write trace propagation headers so the controller can continue the trace.
+	msg.TraceID = childSpan.Scope().TraceID()
+	msg.SpanID = childSpan.Scope().SpanID()
+	msg.TraceFlags = childSpan.Scope().TraceFlags()
```
kian99: Do we need to set these separately if they're already available on msg.span.X?

kian99 (approved):

internal/rpcproxy/rpcproxy.go
SimoneDutto: the span in the message is to make sure we have the live struct to call `.End` on, the other field are the actual propagated json fields

internal/rpcproxy/message.go
```diff
@@ -5,21 +5,39 @@ package rpcproxy
...
-	Version   int             `json:"version,omitempty"`
-	ID        string          `json:"id,omitempty"`
-	Request   string          `json:"request,omitempty"`
-	Params    json.RawMessage `json:"params,omitempty"`
-	Error     string          `json:"error,omitempty"`
-	ErrorCode string          `json:"error-code,omitempty"`
-	ErrorInfo map[string]any  `json:"error-info,omitempty"`
-	Response  json.RawMessage `json:"response,omitempty"`
+	start      time.Time
+	span       *telemetry.TrackedSpan `json:"-"`
```
alesstimec, line 33: could we add the tracing related fields at the bottom? i rather like the grouping of 
```
type message struct {
   payload fields

   tracing fields
}
```

internal/rpcproxy/message.go
```diff
@@ -5,21 +5,39 @@ package rpcproxy
...
+	ErrorInfo  map[string]any         `json:"error-info,omitempty"`
+	Response   json.RawMessage        `json:"response,omitempty"`
+	TraceID    string                 `json:"trace-id,omitempty"`
+	SpanID     string                 `json:"span-id,omitempty"`
+	TraceFlags int                    `json:"trace-flags,omitempty"`
+}
+
+// finishSpan safely finishes the message's trace span (if any) and records
+// the supplied error.  Safe to call on a nil receiver.
+func (m *message) finishSpan(err error) {
```
alesstimec, line 67: i propose we also add
```
func (m *message) startSpan(ctx context.Context) {
     // Continue the client's trace (no-op when TraceID is empty).
	traceCtx := jujuTrace.WithTraceScope(ctx, msg.TraceID, msg.SpanID, msg.TraceFlags)

	_, childSpan := telemetry.StartSpan(traceCtx, "jimm.model-proxy",
		jujuTrace.StringAttr("rpc.facade", msg.Type),
		jujuTrace.IntAttr("rpc.version", msg.Version),
		jujuTrace.StringAttr("rpc.method", msg.Request),
	)

	msg.span = &childSpan
	// Write trace propagation headers so the controller can continue the trace.
	scope := childSpan.Scope()
	msg.TraceID = scope.TraceID()
	msg.SpanID = scope.SpanID()
	msg.TraceFlags = scope.TraceFlags()
}
```

alesstimec (changes requested):
a suggestion for simplification

internal/rpcproxy/message.go
SimoneDutto: i like it, implemented

luci1900 (approved):

internal/rpcproxy/rpcproxytrace_test.go
```diff
@@ -0,0 +1,133 @@
+// Copyright 2025 Canonical.
```
alesstimec: 2026

internal/rpcproxy/message_test.go
```diff
@@ -0,0 +1,53 @@
+// Copyright 2025 Canonical.
```
alesstimec: 2026

alesstimec (approved):
LGTM

PR #2030, opened 2026-06-18
refactor: re-organise model proxy

kian99:
## Description
The internal/rpcproxy/rpcproxy.go file was becoming a bit lengthy, reaching over 800 lines, but more critically packing a lot of logic into one file. This PR splits things up into several files.
- controller_response.go handles logic for modifying the backing controller response and handles access required errors which require redoing login. We also add tests for two pieces of logic which were not tested before from what I could tell.
- inflight.go extracts the `inflightMsgs` struct into a new file and does some renaming.
- lifecycle.go extracts the logic that was validating params and creating the proxy.
- login_dispatch.go handles the login logic within the model proxy between the client -> JIMM.
- rpcproxy.go the original file, contains mostly removals.

The change is not particularly significant, in the sense that I didn't do any design/architectural changes. If we want to do a larger refactoring with a specific design goal then this change may be unnecessary so I have marked it as a draft and we can discuss. The additional test coverage is nice and the decreased concentration of logic in the single file is a minor positive imo.

I've also targetted the v3 branch but the change will be identical in the v4 branch.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

### DIFF
--- a/internal/rpcproxy/controller_response.go
+++ b/internal/rpcproxy/controller_response.go
@@ -36,7 +36,7 @@ func (h controllerResponseHandler) handle(ctx context.Context, msg *message) err
 		h.handleError(ctx, msg, err)
 		return fmt.Errorf("error modifying controller response: %w", err)
 	}
-	h.proxy.msgs.removeMessage(msg.RequestID)
+	h.proxy.inflight.finish(msg.RequestID)
 	if err := h.proxy.auditLogMessage(msg, true); err != nil {
 		zapctx.Error(context.Background(), "failed to audit log message", zap.Error(err))
 	}
@@ -83,7 +83,7 @@ func (h controllerResponseHandler) processControllerErrors(ctx context.Context, 
 			return false
 		}
 		// Write back to the controller.
-		msg := h.proxy.msgs.getMessage(msg.RequestID)
+		msg := h.proxy.inflight.request(msg.RequestID)
 		if msg != nil {
 			if err := h.proxy.src.writeJson(msg); err != nil {
 				zapctx.Error(context.Background(), "failed to write back to controller", zap.Error(err))
@@ -96,7 +96,7 @@ func (h controllerResponseHandler) processControllerErrors(ctx context.Context, 
 
 func (h controllerResponseHandler) handleError(ctx context.Context, msg *message, err error) {
 	h.proxy.sendError(ctx, h.proxy.dst, msg, err)
-	h.proxy.msgs.removeMessage(msg.RequestID)
+	h.proxy.inflight.finish(msg.RequestID)
 }
 
 // checkPermissionsRequired returns a nil map if no permissions are required.
@@ -151,7 +151,7 @@ func (h controllerResponseHandler) redoLogin(ctx context.Context, permissions ma
 	if h.proxy.anonymousLogin {
 		return errors.Codef(errors.CodeUnauthorized, "Anonymous login does not support re-authentication")
 	}
-	loginMsg := h.proxy.msgs.getLoginMessage()
+	loginMsg := h.proxy.inflight.login()
 	if loginMsg == nil {
 		return errors.Codef(errors.CodeUnauthorized, "Haven't received login yet")
 	}
--- a/internal/rpcproxy/controller_response.go
+++ b/internal/rpcproxy/controller_response.go
@@ -1,0 +1,212 @@
+// Copyright 2025 Canonical.
+
+package rpcproxy
+
+import (
+	"context"
+	"encoding/base64"
+	"encoding/json"
+	"fmt"
+
+	jujuparams "github.com/juju/juju/rpc/params"
+	"github.com/juju/zaputil/zapctx"
+	"go.uber.org/zap"
+
+	"github.com/canonical/jimm/v3/internal/dbmodel"
+	"github.com/canonical/jimm/v3/internal/errors"
+)
+
+// controllerResponseHandler keeps controller->client response handling in one place
+// so the proxy loop stays focused on transport orchestration.
+type controllerResponseHandler struct {
+	proxy *controllerProxy
+}
+
+func newControllerResponseHandler(proxy *controllerProxy) controllerResponseHandler {
+	return controllerResponseHandler{proxy: proxy}
+}
+
+func (h controllerResponseHandler) handle(ctx context.Context, msg *message) error {
+	if !h.processControllerErrors(ctx, msg) {
+		return nil
+	}
+
+	if err := modifyControllerResponse(msg); err != nil {
+		zapctx.Error(ctx, "Failed to modify message", zap.Error(err))
+		h.handleError(ctx, msg, err)
+		return fmt.Errorf("error modifying controller response: %w", err)
+	}
+	h.proxy.msgs.removeMessage(msg.RequestID)
+	if err := h.proxy.auditLogMessage(msg, true); err != nil {
+		zapctx.Error(context.Background(), "failed to audit log message", zap.Error(err))
+	}
+	if err := h.proxy.dst.writeJson(msg); err != nil {
+		zapctx.Error(ctx, "controllerProxy error writing to dst", zap.Error(err))
+		return fmt.Errorf("error writing message to client: %w", err)
+	}
+	return nil
+}
+
+// processControllerErrors checks for errors in the message from the controller
+// and decides how to handle them. It returns true if the message should be
+// returned to the client, false if it should not.
+func (h controllerResponseHandler) processControllerErrors(ctx context.Context, msg *message) bool {
+	// Check if the model is migrating. When it has completed its migration, we will receive
+	// an unauthorized error from the controller and we want to mask that error and inform
+	// clients to try again as JIMM will eventually update the model's controller.
+	// See internal/jimm/juju/model_poller.go.
+	modelMigrating := h.proxy.modelMigrationMode == dbmodel.MigrationModeMigrateInternal || h.proxy.modelMigrationMode == dbmodel.MigrationModeExporting
+	if modelMigrating && msg.ErrorCode == string(errors.CodeUnauthorized) {
+		msg.ErrorCode = string(errors.CodeModelMigrating)
+		msg.Error = "model is finishing migration, please retry later"
+		return true
+	}
+
+	// Next we check for permission required errors where the Juju controller is informing us
+	// that the user needs more permissions to perform the requested operation.
+	// If so, we attempt to redo the login with the required permissions (if the user has them)
+	// and then resend the original login request.
+	//
+	// It's ideally unlikely we'll hit this code path often because JIMM sets a broad scope
+	// of permissions on the initial login request.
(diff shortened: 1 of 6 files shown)


internal/rpcproxy/controller_response.go
```diff
@@ -0,0 +1,212 @@
+// Copyright 2025 Canonical.
```
alesstimec, line 1: 2026

internal/rpcproxy/inflight.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+func (tracker *inflightTracker) setControllerUUID(uuid string) {
+	tracker.mu.Lock()
+	defer tracker.mu.Unlock()
+
+	tracker.controllerUUID = uuid
+}
+
+func (tracker *inflightTracker) controllerID() string {
```
alesstimec, line 33:
```suggestion
func (tracker *inflightTracker) controllerUUID() string {
```

internal/rpcproxy/inflight.go
```diff
@@ -0,0 +1,90 @@
...
+	tracker.mu.Lock()
+	defer tracker.mu.Unlock()
+
+	return tracker.controllerUUID
+}
+
+func (tracker *inflightTracker) rememberLogin(msg *message) {
```
alesstimec, line 40:
```suggestion
func (tracker *inflightTracker) setLoginMessage(msg *message) {
```

internal/rpcproxy/inflight.go
```diff
@@ -0,0 +1,90 @@
...
+	tracker.mu.Lock()
+	defer tracker.mu.Unlock()
+
+	tracker.loginMessage = msg
+}
+
+func (tracker *inflightTracker) login() *message {
```
alesstimec, line 47:
```suggestion
func (tracker *inflightTracker) loginMessage() *message {
```

internal/rpcproxy/inflight.go
```diff
@@ -0,0 +1,90 @@
...
+	tracker.mu.Lock()
+	defer tracker.mu.Unlock()
+
+	return tracker.loginMessage
+}
+
+func (tracker *inflightTracker) track(msg *message) {
```
alesstimec, line 54:
```suggestion
func (tracker *inflightTracker) addMessage(msg *message) {
```

internal/rpcproxy/inflight.go
```diff
@@ -0,0 +1,90 @@
...
+	tracker.mu.Lock()
+	defer tracker.mu.Unlock()
+
+	msg.start = time.Now()
+	tracker.requests[msg.RequestID] = msg
+}
+
+func (tracker *inflightTracker) request(requestID uint64) *message {
```
alesstimec, line 62:
```suggestion
func (tracker *inflightTracker) getMessage(requestID uint64) *message {
```

internal/rpcproxy/inflight.go
```diff
@@ -0,0 +1,90 @@
...
+
+	msg, ok := tracker.requests[requestID]
+	if !ok {
+		return nil
+	}
+	return msg
+}
+
+// finish deletes the request message that corresponds to the response message ID.
+func (tracker *inflightTracker) finish(requestID uint64) {
```
alesstimec, line 74:
```suggestion
func (tracker *inflightTracker) deleteMessage(requestID uint64) {
```

internal/rpcproxy/inflight.go
```diff
@@ -0,0 +1,90 @@
+// Copyright 2025 Canonical.
```
alesstimec, line 1: 2026

internal/rpcproxy/lifecycle.go
```diff
@@ -0,0 +1,92 @@
+// Copyright 2025 Canonical.
```
alesstimec, line 1: 2026

internal/rpcproxy/login_dispatch.go
```diff
@@ -0,0 +1,187 @@
...
+		if err != nil {
+			return errorFnc(err)
+		}
+
+		return controllerLoginMessageFnc(user)
+	case "Login":
+		controllerMessage, err := h.handleLegacyLogin(ctx, msg)
+		return nil, controllerMessage, err
+	default:
+		return nil, nil, nil
```
alesstimec, line 126: should we not return an error on an unknown login method?

internal/rpcproxy/login_dispatch.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+	tag, err := names.ParseTag(request.AuthTag)
+	if err != nil {
+		return nil, fmt.Errorf("invalid user tag: %v", err)
+	}
+	switch tag := tag.(type) {
+	case names.UserTag:
+		if tag.Id() == api.AnonymousUsername {
+			h.proxy.anonymousLogin = true
+			// return the client's login message verbatim to the controller.
```
alesstimec, line 158:
```suggestion
			// forward the client's login message verbatim to the controller.
```

alesstimec (commented):
i quite like this proposed change.. a few comment, though

luci1900 (commented):
I personally find it harder to read than before.

Nice to have more tests, though.

internal/rpcproxy/controller_response.go
```diff
@@ -0,0 +1,212 @@
...
+type controllerResponseHandler struct {
+	proxy *controllerProxy
+}
+
+func newControllerResponseHandler(proxy *controllerProxy) controllerResponseHandler {
+	return controllerResponseHandler{proxy: proxy}
+}
+
+func (h controllerResponseHandler) handle(ctx context.Context, msg *message) error {
+	if !h.processControllerErrors(ctx, msg) {
```
ale8k, line 30: So if false, it means an error was handled / sent to client or we need to redo the login right? Maybe worth a godoc as it isn't clear this means both

internal/rpcproxy/controller_response.go
```diff
@@ -0,0 +1,212 @@
...
+// the provided permissions. This is sometimes necessary if Juju requires
+// extra permission checks for an operation. If the client performed anonymous
+// login, an error is always returned since we cannot authorize an anonymous user.
+func (h controllerResponseHandler) redoLogin(ctx context.Context, permissions map[string]any) error {
+	if h.proxy.anonymousLogin {
+		return errors.Codef(errors.CodeUnauthorized, "Anonymous login does not support re-authentication")
+	}
+	loginMsg := h.proxy.inflight.login()
+	if loginMsg == nil {
+		return errors.Codef(errors.CodeUnauthorized, "Haven't received login yet")
```
ale8k, line 156: Goerrors aren't capitalised usually

internal/rpcproxy/controller_response.go
```diff
@@ -0,0 +1,212 @@
...
+		return err
+	}
+	return nil
+}
+
+// addJWT adds a JWT token to the the provided message.
+func addJWT(ctx context.Context, msg *message, permissions map[string]any, tokenGen TokenGenerator) error {
+	// First we unmarshal the existing LoginRequest.
+	if msg == nil {
+		return errors.New("nil messsage")
```
ale8k, line 173: s/messsage/message

internal/rpcproxy/login_dispatch.go
```diff
@@ -0,0 +1,187 @@
...
+			AuthTag: names.NewUserTag(user.Name).String(),
+			Token:   base64.StdEncoding.EncodeToString(jwt),
+		})
+		if err != nil {
+			return errorFnc(err)
+		}
+		m := *msg
+		m.Type = "Admin"
+		m.Request = "Login"
+		m.Version = 3
```
ale8k, line 54: Given we're rewriting, be good to explain why we're locking it down to 3

internal/rpcproxy/controller_response.go
```diff
@@ -0,0 +1,212 @@
...
+		zapctx.Error(ctx, "Access Required error")
+		if err := h.redoLogin(ctx, permissionsRequired); err != nil {
+			zapctx.Error(ctx, "Failed to redo login", zap.Error(err))
+			h.handleError(ctx, msg, err)
+			return false
+		}
+		// Write back to the controller.
+		msg := h.proxy.inflight.request(msg.RequestID)
+		if msg != nil {
+			if err := h.proxy.src.writeJson(msg); err != nil {
```
ale8k, line 88: If this fails - which granted is like super unlikely as we've just tested the connection logging in, but still, we'll silently continue and the client will end up waiting on a req that isn't gonna complete right - meaning the client will just sit waiting until it times out? I guess maybe logging is good but is it all we can do? 

I'm saying this becaue of this code:
```go
// WriteJSON is part of base.Stream.
func (s *DeadlineStream) WriteJSON(v interface{}) error {
	// This uses a real clock rather than trying to use a clock passed
	// in because the websocket will use a real clock to determine
	// whether the deadline has passed anyway.
	deadline := time.Now().Add(s.Timeout)
	if err := s.Conn.SetWriteDeadline(deadline); err != nil {
		return errors.Annotate(err, "setting write deadline")
	}
	return errors.Trace(s.Conn.WriteJSON(v))
}
```
In the Juju pkg.

ale8k (commented):
LGTM as a refactor and happy to land as is, but I feel like this moves code around without addressing the two bits that (for me at least) make the code hard to follow. (I disagree it is harder than before though, this makes it at least navigable in rpcproxy now for me)

1. I would get rid of dst/src naming. Both routines embed our modelProxy but mean opposite things, so the src/dst flips depending where you are looking.  I'd just name them client/controller so it's really clear.
2. The login dispatchers triple return is like am ini encoded state machine state retriever, we could type this out so that depending on the combination of results it determines an exact next step for the caller to do (i.e., sentClient, sendController, reauth, drop, fail) I think this would greatly clear up that part of the code (lmk wyt)
3. The retry logic is split between two routines but it could be 1 right? I streuggled to word this and had AI word what I *think* we can do (this code even confused AI when asking it to verify my suggestion haha), lmk what you think:
```
Your suggestion:
The retry logic is split between two workers when it could be one.

Two goroutines run at once:

- Worker A reads messages from the client and sends them to the controller.
- Worker B reads replies from the controller and sends them back to the client.
When someone logs in, Worker A stashes the login message in a shared box (the tracker).

Later, if the controller says "access required", Worker B reaches into that same shared box, grabs the stashed login, re-logs-in with the right permissions, and replays the original request.

So the "re-login and retry" story is told in two places, and the two workers hand a message to each other through shared, mutex-protected state.

The suggestion: Worker B is the one that actually notices the error and does the re-login and does the replay. Worker A only stores the login for B to find later. Since B does all the real work, the login doesn't need to live in a shared box that both workers fight over — it could just be a value B owns. Then the whole "error → re-login → retry" would read top-to-bottom in one place instead of being a handoff between two workers.

The catch I'm unsure about: "this only works cleanly though right if the client can't send a second login while B is mid retry???". I haven't confirmed that, so it's a "worth checking" idea.
```
I haven't scoured Juju to see what happens if you "double login", but if you can, then I think we can simplify it further maybeeee.

kian99:
Closing for now to de-clutter.

PR #2031, opened 2026-06-18
feat(jujuapi): serve ModelManager v10 for Juju 3.6 clients

alesstimec:
## Description

Register the ModelManager facade at version 10 (Juju 3.6) alongside the existing version 11 (Juju 4.x), so a 3.6 client negotiates v10 and talks to JIMM in the owner-tag wire format.

The five methods whose wire format changed at v11 (the model owner-tag -> qualifier rename) get dedicated v10 handlers (CreateModelLegacy, ModelInfoLegacy, ListModelsLegacy, ModelStatusLegacy, ListModelSummariesLegacy). Each adapts the 3.6 owner-tag format to the v11 handler via the params36 conversion package, reusing the same business logic. The other nine methods are wire-identical and register the existing v11 handlers at v10. facadeInit now returns {10, 11}.

Placement of a 3.6 client's new model onto a compatible controller is left for future work; CreateModelLegacy is the natural seam for that constraint and carries a TODO marking it.

Tests cover the advertised version set and owner-tag round-tripping for CreateModel, ListModels and ModelInfo via mocked JujuManager results.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

### DIFF
--- a/testing/modelmanager_legacy_test.go
+++ b/testing/modelmanager_legacy_test.go
@@ -1,0 +1,227 @@
+// Copyright 2026 Canonical.
+
+package testing
+
+import (
+	"strings"
+	"testing"
+
+	petname "github.com/dustinkirkland/golang-petname"
+	qt "github.com/frankban/quicktest"
+	jujuparams "github.com/juju/juju/rpc/params"
+	"github.com/juju/names/v6"
+
+	"github.com/canonical/jimm/v3/internal/testutils/jimmtest"
+)
+
+// These tests exercise the ModelManager facade version 10 (Juju 3.6) handlers
+// end to end against a real backing controller. A single Go module pins one
+// Juju version (4.x), so we cannot import a 3.6 API client; instead we make
+// hand-crafted version-pinned calls with conn.APICall and the juju 4.x
+// "*Legacy" param structs, which are the 3.6 owner-tag wire shapes. This
+// confirms a real JIMM advertises and dispatches v10 and that the owner-tag
+// conversion holds on real controller data.
+
+var bobOwnerTag = names.NewUserTag("bob@canonical.com")
+
+// TestModelManagerV10CreateModel covers the v10 CreateModel handler: the
+// owner-tag happy path, owner defaulting, authorization, duplicate names, and
+// owner tags that cannot be converted to a qualifier.
+func TestModelManagerV10CreateModel(t *testing.T) {
+	c := qt.New(t)
+	s := jimmtest.SetupJimmWithControllers(c)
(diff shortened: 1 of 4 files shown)


kian99 (commented):
If we want to write integration tests in the `testing` package, I imagine it will be quite difficult since we can only import the latest Juju client package, it will always negotiate to use v11 of the facacde.

SimoneDutto (commented):
we need to define a testing framework, this would be very dangerous to release no?

kian99:
I was thinking about my previous comment some more. What if we add a param to the API handler that defines the highest version of a facade the server is allowed to use.
It would only be used in tests to specify e.g. ["ModelManager"]: 10, then our integration tests would work as intended.

SimoneDutto:
> I was thinking about my previous comment some more. What if we add a param to the API handler that defines the highest version of a facade the server is allowed to use. It would only be used in tests to specify e.g. ["ModelManager"]: 10, then our integration tests would work as intended.

i'll give you another idea, but maybe it's not possible.

What if `testing` is it's own package with its own go.mod, and we basically just create a `testing-v3` where we import the 3 client.

It might be a lot of work though

testing/modelmanager_legacy_test.go
```diff
@@ -0,0 +1,227 @@
...
+	s := jimmtest.SetupJimmWithControllers(c)
+	model := s.CreateModelForBob(c)
+
+	conn := s.Open(c, nil, bobOwnerTag.Id(), nil)
+	defer conn.Close()
+
+	var list jujuparams.UserModelListLegacy
+	err := conn.APICall(t.Context(), "ModelManager", 10, "", "ListModels", jujuparams.Entity{}, &list)
+	c.Assert(err, qt.IsNil)
+	c.Assert(len(list.UserModels) > 0, qt.IsTrue)
```
kian99, line 158: This works too. My only question would be, do we or Juju actually then test anywhere, the logic in the 4.0 client against a 3 controller/v10 ModelManager facade?

SimoneDutto: it's tested everywhere in our e2e suite here and in the terraform code.
Here particularly there is a workflow to run jaas v4 against a 3 backing controller.

And in terraform too.

SimoneDutto (approved):

kian99 (approved):

testing/modelmanager_legacy_test.go
kian99: Right, good point, ty.

PR #2033, opened 2026-06-19
feat(jujuapi): place Juju 3.6 client models on compatible controllers

alesstimec:
## Description

A Juju 3.6 client cannot operate a 4.x controller: JIMM brokers the controller-level facades, but model-level facades are proxied raw, so a 3.6 client talking to a 4.x-hosted model breaks. When a 3.6 client creates a model (the ModelManager v10 handler), its model must therefore be placed on a controller of major version <= 3.

- Add MaxControllerMajorVersion to the internal ModelCreateArgs (0 means unconstrained); the v10 CreateModelLegacy handler sets it to 3, the v11 handler leaves it unset.
- Add modelBuilder.WithMaxControllerMajorVersion and a controllerCompatible helper; apply it in WithAnyController (excludes incompatible controllers and, when that empties the candidate set, returns a version-specific error) and in the explicit WithController path (errors if the named controller is too new). Controllers with an empty or unparseable agent version are treated as ineligible under a constraint, and logged.
- Refactor CreateModel into a shared createModel core so v10 and v11 reuse the same logic, differing only in wire format and the constraint.

Unit-tests the version-decision helper (controllerCompatible) across equal/lower/higher majors and empty/unparseable versions.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

### DIFF
--- a/internal/jujuapi/modelmanager.go
+++ b/internal/jujuapi/modelmanager.go
@@ -214,14 +214,21 @@ func (r *controllerRoot) ModelInfo(ctx context.Context, args jujuparams.Entities
 
 // CreateModel implements the ModelManager facade's CreateModel method.
 func (r *controllerRoot) CreateModel(ctx context.Context, args jujuparams.ModelCreateArgs) (jujuparams.ModelInfo, error) {
-
-	ctx, cancel := context.WithCancel(ctx)
-	defer cancel()
-
 	mca, err := toAddModelArgs(args, r.user.ResourceTag())
 	if err != nil {
 		return jujuparams.ModelInfo{}, err
 	}
+	return r.createModel(ctx, mca)
+}
+
+// createModel adds the model described by mca and converts the result to the
+// current (qualifier) wire format. It is shared by the v11 CreateModel handler
+// and the v10 CreateModelLegacy handler, which differ only in their wire format
+// and (for v10) the controller-version placement constraint set on mca.
+func (r *controllerRoot) createModel(ctx context.Context, mca *juju.ModelCreateArgs) (jujuparams.ModelInfo, error) {
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
 	info, err := r.jimm.JujuManager().AddModel(ctx, r.user, mca)
 	if err != nil {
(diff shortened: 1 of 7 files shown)


testing/modelmanager_legacy_test.go
```diff
@@ -24,12 +24,25 @@ import (
 
 var bobOwnerTag = names.NewUserTag("bob@canonical.com")
 
+// setupLegacyModelManagerTest sets up the JIMM e2e environment and skips the
+// test unless a Juju 3.x backing controller is available. The 3.6-client
+// (ModelManager version 10) handlers place new models on a controller of major
+// version <= 3 (see JAAS-6), so these tests are only meaningful when the
+// backing fleet includes such a controller.
+func setupLegacyModelManagerTest(c *qt.C) jimmtest.JimmWithControllers {
```
kian99, line 32: How come this helper was only needed now? I don't fully follow that.

internal/jujuapi/modelmanager.go
```diff
@@ -214,14 +214,21 @@ func (r *controllerRoot) ModelInfo(ctx context.Context, args jujuparams.Entities
...
+	return r.createModel(ctx, mca)
+}
+
+// createModel adds the model described by mca and converts the result to the
+// current (qualifier) wire format. It is shared by the v11 CreateModel handler
+// and the v10 CreateModelLegacy handler, which differ only in their wire format
+// and (for v10) the controller-version placement constraint set on mca.
+func (r *controllerRoot) createModel(ctx context.Context, mca *juju.ModelCreateArgs) (jujuparams.ModelInfo, error) {
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
```
kian99, line 230: What's with the added cancellation? I see it was there already, but looks unused?

kian99 (approved):

testing/modelmanager_legacy_test.go
alesstimec: because we just introduced the placement logic in this PR. and we're running e2e test matrix which at some point runs tests with a juju 4 backing controller - at that point we expect these tests to fail so we just skip them

SimoneDutto (approved):
LGTM

PR #2038, opened 2026-06-22
doc: add upgrade-to model management info

kian99:
## Description
Add a new section to our model management how-to on the JAAS `upgrade-to` feature.

Fixes [JUJU-9992](https://warthogs.atlassian.net/browse/JUJU-9992)

## Engineering checklist

- [x] Documentation updated

[JUJU-9992]: https://warthogs.atlassian.net/browse/JUJU-9992?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


docs/howto/manage-models.md
```diff
@@ -276,6 +276,94 @@ If the model migration fails, then no further user input is required and the mod
...
 
+(upgrade-a-model)=
+## Upgrade a model
+
+Use the `jaas upgrade-to` command to move one or more models to another JAAS-managed controller and upgrade those models to the target controller's Juju version.
+
+This is useful when you have bootstrapped or registered a newer controller in JAAS and want a model to run on that controller without manually combining the model migration and model upgrade steps.
+
+```{note}
+This feature is complemtary to Juju [model-upgrades](https://canonical.com/juju/docs/juju-cli/latest/howto/manage-models/#upgrade-a-model) which is preferable when upgrading across patch versions.
```
tmihoc:
```suggestion
This feature is complementary to Juju [model-upgrades](https://canonical.com/juju/docs/juju-cli/latest/howto/manage-models/#upgrade-a-model) which is preferable when upgrading across patch versions.
```

That's true only if all you're trying to do is upgrade the model. But this procedure as written is specifically about migrate+upgrade. And I'm not sure we can actually rewrite it to be just about upgrades, because this will cause a migration to happen too. Essentially, `jaas upgrade-to` is a bit of a misnomer, imo. 

BTW I considered whether it's similar to what we have in the `juju` CLI but it's not -- there `migrate` actually just means `migrate` -- you must follow up with `upgrade-model --agent-version=...` to actually upgrade. So I guess what this note is trying to say is that if you _only_ mean to upgrade the model to a new patch version you should use `juju upgrade-model`. But that feels redundant if the section title here is, as it should be based on what's happening, "Migrate and upgrade..." 

_Unless_ you're trying to say that the whole point of migrate here is to upgrade. If that's the case then we should follow the logic of https://canonical.com/juju/docs/juju-cli/latest/howto/manage-models/#upgrade-a-model -- i.e., have a fork in the procedure to distinguish between patch upgrades vs. minor or major upgrades, and in the later feature `upgrade-to`. Then this section will actually finally deserve to be called "Upgrade a model" and the doc could like something like this (update the versions etc.):

````
## Upgrade a model in JAAS

How you upgrade a model depends on whether you’d be crossing patch versions (e.g., v.2.9.25 -> v.2.9.26) or rather minor (e.g., v.2.7 -> v.2.8) or major versions (v.2 -> v.3).

- To upgrade a model's patch version, use `juju upgrade-model`.

> See more: {external+juju:ref}`upgrade-a-model`

- To upgrade a model's major or minor version, you must migrate your model to a controller of the target version and upgrade the model to that version. In JAAS this means you get the model's UUID, then use `juju jaas upgrade-to` followed by the name of the target JAAS-managed controller and the model's UUID, then verify that the upgrade has succeeded. For example:

```
# Get the model UUID:
MODEL_NAME=my-model
MODEL_UUID=$(juju show-model $MODEL_NAME --format yaml | yq .$MODEL_NAME.model-uuid)

# Start the migrate + upgrade:
MODEL_NAME=my-model
MODEL_UUID=$(juju show-model $MODEL_NAME --format yaml | yq .$MODEL_NAME.model-uuid)
juju jaas upgrade-to juju-3-6-controller $MODEL_UUID

# Verify that the procedure has succeeded 
# (output should show the target controller's details):
juju jaas show-model $MODEL_UUID --format yaml

```


> See more: {external+juju:ref}`Juju | juju show-model <command-show-model>`, {ref}`command-jaas-upgrade-to`

File: docs/howto/manage-models.md [hunk 2]
```diff
@@ -276,6 +276,94 @@ If the model migration fails, then no further user input is required and the mod
 
 To inspect the reason for failure, consult the output from `juju debug-log` and `juju status`.
 
+(upgrade-a-model)=
+## Upgrade a model
+
```
tmihoc, line 281: **UPDATE: IGNORE THIS IN FAVOR OF https://github.com/canonical/jimm/pull/2038#discussion_r3452904191 ** (though read below for the How the upgrade works bit).

------

The steps seem to be:

1. Get the model UUID.
2. Trigger the upgrade.
3. Check upgrade status.

I'd make this more visible and everything else less so.

For example:

````
## Migrate and upgrade a model in one fell swoop [because _this_ is what this command seems to actually do]

To migrate a model from one JAAS-managed controller to another and also upgrade it to the target controller's version in one fell swoop: [this makes the prerequisites redundant; the detail about JAAS admin can be appended at the beginning: "If you are a JAAS admin and want to..." but we don't usually comment on this -- if someone without the right permissions tries they'll fail -- so we could drop this from here too.]

1. Get the model UUID using `juju show-model`. For example:

```
MODEL_NAME=my-model
MODEL_UUID=$(juju show-model $MODEL_NAME --format yaml | yq .$MODEL_NAME.model-uuid)
```

> See more: {ref}`the anchor for juju show-model`

2. Start the migration and upgrade with `juju jaas upgrade-to`, passing as arguments the name of the target controller followed by the model UUID. For example:

```
MODEL_NAME=my-model
MODEL_UUID=$(juju show-model $MODEL_NAME --format yaml | yq .$MODEL_NAME.model-uuid)
juju jaas upgrade-to juju-3-6-controller $MODEL_UUID
```

> See more: {ref}`the anchor for jaas upgrade-to` [the detail about how to migrate more than one model at once will be covered there]

3. Check the migration and upgrade status. For example:

```
juju jaas show-model $MODEL_UUID --format yaml
```

> See more: {ref}`the anchor for juju show-model` [demonstrating with --format etc. should already help]

The upgrade is complete when the model's controller details reflect the target controller.
````

I'm not sure who the "How the upgrade works" is going to be useful for, but it feels like optional, zoom-in kinda detail, so I'd put it into a dropdown, not sure where. Maybe under step 2? Though, looking at the details there, none feels particularly insightful. I mean, we'd expect JIMM to validate the inputs to `upgrade-to`, or for the workflow to migrate the model to the target controller, etc. Correct me if I'm wrong but I don't see any critical detail there. If I'm write, then I'd skip this section entirely.

**Review by tmihoc — CHANGES_REQUESTED** [review]
Thank you for this. Left some comments. 

tl;dr it seems we're trying to replicate https://canonical.com/juju/docs/juju-cli/latest/howto/manage-models/#upgrade-a-model in JAAS. If that's the case I suggest we follow that same structure. I've left the concrete Markdown you could use, maybe with some small further adjustments.

File: docs/howto/manage-models.md [hunk 3]
```diff
@@ -276,6 +276,94 @@ If the model migration fails, then no further user input is required and the mod
...
+Use the `jaas upgrade-to` command to move one or more models to another JAAS-managed controller and upgrade those models to the target controller's Juju version.
+
+This is useful when you have bootstrapped or registered a newer controller in JAAS and want a model to run on that controller without manually combining the model migration and model upgrade steps.
```
luci1900:
```suggestion
This is useful when you have bootstrapped or registered a newer controller in JAAS and want a model to run on that controller without having to manually do both the model migration and model upgrade steps.
```

docs/howto/manage-models.md
```diff
@@ -276,6 +276,94 @@ If the model migration fails, then no further user input is required and the mod
...
+> See more: {ref}`command-jaas-show-model`
+
+When an upgrade is running, the output includes the model's current controller information and the status of the upgrade workflow. When the upgrade completes, the model's controller information reflects the target controller.
+
+### How the upgrade works
+
+JIMM coordinates the upgrade as a background workflow:
+
+1. JIMM validates the requested model UUIDs and target controller.
+2. JIMM starts one workflow per model that will includes retry logic for each step.
```
luci1900:
```suggestion
2. JIMM starts one task per model that will includes retry logic for each step.
```

docs/howto/manage-models.md
```diff
@@ -276,6 +276,94 @@ If the model migration fails, then no further user input is required and the mod
...
+3. Each workflow migrates the model to the target controller.
```
luci1900:
```suggestion
3. Each task migrates the model to the target controller.
```

docs/howto/manage-models.md
```diff
@@ -276,6 +276,94 @@ If the model migration fails, then no further user input is required and the mod
...
+juju jaas show-model $MODEL_UUID --format json
+```
+
+> See more: {ref}`command-jaas-show-model`
+
+When an upgrade is running, the output includes the model's current controller information and the status of the upgrade workflow. When the upgrade completes, the model's controller information reflects the target controller.
+
+### How the upgrade works
+
+JIMM coordinates the upgrade as a background workflow:
```
luci1900, line 310:
```suggestion
JIMM coordinates the upgrade as a background task:
```

luci1900 (approved):

docs/howto/manage-models.md
```diff
@@ -276,6 +276,45 @@ If the model migration fails, then no further user input is required and the mod
...
+MODEL_UUID=$(juju show-model $MODEL_NAME --format yaml | yq .$MODEL_NAME.model-uuid)  
+juju jaas upgrade-to juju-3-6-controller $MODEL_UUID  
+
+# Verify that the procedure has succeeded 
+# (output should show the target controller's details):  
+juju jaas show-model $MODEL_UUID --format yaml
+```  
+
+
+> See more: {external+juju:ref}`Juju | juju show-model <command-show-model>`, {ref}`command-jaas-upgrade-to`
```
tmihoc:
```suggestion
> See more: {external+juju:ref}`Juju | juju show-model <command-juju-show-model>`, {ref}`command-jaas-upgrade-to` 
```

tmihoc (approved):
One small anchor fix.

Check also `make linkcheck` -- the CI tests seems to be failing.

kian99:
> One small anchor fix.
> 
> Check also `make linkcheck` -- the CI tests seems to be failing.

Fixed, there was a link returning a 404. And thank you for the re-review!

PR #2040, opened 2026-06-23
feat: add trace_id to sampled traces in the log

SimoneDutto:
## Description

If the span is sample we add the span_id as a Field in the zap log.
We need to store something in the context to avoid putting the same field multiple times, sadly zap doesn't support to check if a field has been inserted already, so this is the only way.

# QA
`ENABLE_TRACING=true make qa-lxd`

And you can see
```
13:05:40.508	DEBUG	user is resource administrator	{"facade": "Admin", "method": "LoginWithSessionToken", "version": 4, "trace_id": "ce8e28ecbb178dbb13aaed92f8880bca", "user": "user-jimm-test@canonical.com", "resource": "controller-3217dbc9-8ea9-4381-9e97-01eab0b3f6bb"}
```

which you can then check in grafana.

### DIFF
--- a/internal/telemetry/juju_adapter.go
+++ b/internal/telemetry/juju_adapter.go
@@ -6,9 +6,29 @@ import (
 	"context"
 
 	jujuTrace "github.com/juju/juju/core/trace"
+	"github.com/juju/zaputil/zapctx"
 	"go.opentelemetry.io/otel/codes"
 	oteltrace "go.opentelemetry.io/otel/trace"
+	"go.uber.org/zap"
 )
+
+const (
+	// TraceIDLogField is the zap field key for the trace ID in log entries.
+	// Uses the OpenTelemetry convention so Loki/Grafana can automatically
+	// correlate logs to traces in Tempo.
+	traceIDLogField = "trace_id"
+)
+
+// traceLoggerKey marks a context as already having the trace_id injected into
+// its logger. A single context chain shares one trace ID, so we only need to
+// add it once — at the first sampled span.
+type traceLoggerKey struct{}
+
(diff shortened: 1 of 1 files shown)


ale8k (approved):
nice ty!

internal/telemetry/juju_adapter.go
```diff
@@ -35,6 +55,16 @@ func (t otelTracer) Start(ctx context.Context, name string, options ...jujuTrace
...
 		ctx = jujuTrace.WithTraceScope(ctx, spanContext.TraceID().String(), spanContext.SpanID().String(), int(spanContext.TraceFlags()))
+
+		// If the span is sampled, we add the trace ID to the logger context.
+		// We don't add it again for child spans, because they all relate to
+		// the same trace ID.
+		if spanContext.TraceFlags().IsSampled() && !traceIDAdded(ctx) {
+			ctx = zapctx.WithFields(ctx,
+				zap.String(traceIDLogField, spanContext.TraceID().String()),
+			)
+			ctx = context.WithValue(ctx, traceLoggerKey{}, true)
```
kian99, line 66: Out of curiosity, what happens if you re-add it?

kian99 (approved):

internal/telemetry/juju_adapter.go
SimoneDutto: nothing crazy just
```
13:05:40.508	DEBUG	user is resource administrator	{"trace_id": "ce8e28ecbb178dbb13aaed92f8880bca", "trace_id": "ce8e28ecbb178dbb13aaed92f8880bca"}
```
so it gets added multiple times depending on how many children the trace has

PR #2044, opened 2026-06-26
feat: run dry run on migrate

kian99:
## Description
This PR runs dry-run logic before we perform a `upgrade-to` operation. Because this feature landed in 3.6.13 without a bump in facade version, we return an error if the target controller version is <3.6.13. 

Fixes [JUJU-9061](https://warthogs.atlassian.net/browse/JUJU-9061)

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

[JUJU-9061]: https://warthogs.atlassian.net/browse/JUJU-9061?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/jimm/juju/controller.go
+++ b/internal/jimm/juju/controller.go
@@ -28,22 +28,6 @@ import (
 	ofganames "github.com/canonical/jimm/v3/internal/openfga/names"
 )
 
-// ControllerClient defines an interface of the juju controller api client
-// used by JIMM to interact with the Controller facade of Juju controllers.
-type ControllerClient interface {
-	// InitiateMigration attempts to begin the migration of one or
-	// more models to other controllers.
-	InitiateMigration(controller.MigrationSpec, bool) (string, error)
-	// Close closes the connection to the API server.
-	Close() error
-}
-
-var (
-	newControllerClient = func(api base.APICallCloser) ControllerClient {
-		return controller.NewClient(api)
-	}
-)
-
 // convertJujuCloudsToDbClouds converts all of the incoming Juju clouds (from a map) into
 // a slice of dbmodel Clouds.
 func convertJujuCloudsToDbClouds(clouds map[names.CloudTag]jujucloud.Cloud) []dbmodel.Cloud {
@@ -716,10 +700,9 @@ func (j *JujuManager) DryRunInternalMigration(ctx context.Context, user *openfga
 	if err != nil {
 		return fmt.Errorf("failed to dial source controller: %w", err)
 	}
-	client := newControllerClient(api)
-	defer client.Close()
+	defer api.Close()
 
-	if _, err = client.InitiateMigration(spec, true); err != nil {
+	if _, err = api.InitiateMigration(spec, true); err != nil {
 		return fmt.Errorf("migration precheck failed: %w", err)
 	}
 	return nil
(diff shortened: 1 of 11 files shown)


internal/jimm/juju/controller.go
```diff
@@ -660,6 +644,70 @@ func (j *JujuManager) UpdateMigratedModel(ctx context.Context, user *openfga.Use
 	return nil
 }
 
+// minDryRunMigrateVersion is the minimum Juju controller version that supports
+// the dry-run flag on InitiateMigration (introduced in Juju 3.6.13).
+var minDryRunMigrateVersion = version.Number{Major: 3, Minor: 6, Patch: 13}
```
SimonRichardson, line 649: Juju failed here, we should have bumped the facade so you can detect it using the facade version and not the juju version.

luci1900 (approved):
Looks reasonable.

internal/jimm/juju/controller.go
luci1900: It makes little difference, we would still at least need the version number hard-coded in the error message.

SimonRichardson: Sure, but the modelling of facades was broken here and that's not great. You can write what you want in the error message to inform the user, but the building blocks failed.

internal/jimm/juju/controller.go
```diff
@@ -660,6 +644,70 @@ func (j *JujuManager) UpdateMigratedModel(ctx context.Context, user *openfga.Use
...
+	targetControllerTag, err := names.ParseControllerTag(migrationTarget.ControllerTag)
+	if err != nil {
+		return fmt.Errorf("failed to parse target controller tag: %w", err)
+	}
+	targetUserTag, err := names.ParseUserTag(migrationTarget.AuthTag)
+	if err != nil {
+		return fmt.Errorf("failed to parse target auth tag: %w", err)
+	}
+
+	spec := controller.MigrationSpec{
```
ale8k, line 690: Would prefer if spec was instantiated just above where it is used (line 705)

internal/jimm/juju/controller.go
```diff
@@ -660,6 +644,70 @@ func (j *JujuManager) UpdateMigratedModel(ctx context.Context, user *openfga.Use
...
+		TargetCACert:         migrationTarget.CACert,
+		TargetUser:           targetUserTag.Id(),
+		TargetPassword:       migrationTarget.Password,
+	}
+
+	api, err := j.dial(ctx, &model.Controller, names.ModelTag{}, nil)
+	if err != nil {
+		return fmt.Errorf("failed to dial source controller: %w", err)
+	}
+	defer api.Close()
```
ale8k, line 703: Should we defer and log the err here?

internal/jimm/juju/controller.go
```diff
@@ -660,6 +644,70 @@ func (j *JujuManager) UpdateMigratedModel(ctx context.Context, user *openfga.Use
...
+var minDryRunMigrateVersion = version.Number{Major: 3, Minor: 6, Patch: 13}
+
+// DryRunInternalMigration runs a migration precheck (dry run) against the
+// source controller without actually initiating the migration.
+//
+// If the source controller is older than [minDryRunMigrateVersion], the call
+// returns an error asking the user to upgrade the controller first. Otherwise
+// it calls InitiateMigration with dryRun=true; any precheck failure is
+// returned as an error.
+func (j *JujuManager) DryRunInternalMigration(ctx context.Context, user *openfga.User, modelNameOrUUID string, targetControllerName string) error {
```
ale8k, line 658: This function is doing a lot other than "DryRun..."? I would've expected DryRun to just be a wrapper that sets a dryrun flag. Maybe we should do all this stuff in `internal/jimm/upgrade/upgrade.go`? It's mostly validation what we're doing here, plus we're already parsing the target version there.

kian99:
Test failed `prechecks failed: controller: machine 0 not running` which raised an interesting problem, our upgrade-to server logic retries and with the addition of a dry-run, we fail early when waiting and retrying would eventually.

Discussed with @ale8k, since the dry-run might fail due to a short-lived error or an error requiring operator intervention, calling dry-run seems like an overall worse experience.

If in the future we want to introduce an API that does something like "migrate all models from controller A -> controller B" then a dry run over all the models, stopping on the first failure might prove more useful. For now, this is a slightly lower level API, allowing migration of individual models.

PR #2045, opened 2026-06-29
fix: fix authcode state cookie verification path

ale8k:
### Why
The browser OAuth2 login flow stores a CSRF `state` value in a short-lived cookie (`jimm-oauth-state`) and compares it against the `state` query parameter on the `/auth/callback` request. The cookie's `Path` was hardcoded to `/auth/callback`.

However, the redirect URI sent to the identity provider is built from `PublicDNSName` (`JIMM_DNS_NAME`), which may include a sub-path. With `JIMM_DNS_NAME = https://jimm.workshop/jimm-jimm`, the IdP redirects the browser to `https://jimm.workshop/jimm-jimm/auth/callback`. Per [RFC 6265](https://datatracker.ietf.org/doc/html/rfc6265#section-5.1.4), a cookie is only sent when the request path matches (or is a subpath of) the cookie's `Path`. `/jimm-jimm/auth/callback` is not under `/auth/callback`, so the browser withheld the cookie and the callback failed with `no state cookie present`.

This only affected deployments where JIMM is not hosted at the root (e.g., charmed deployments behind Traefik with a path prefix, domain would be unaffected).

## Options considered

1. Derive the base path from our preexisting `PublicDnssName`
2. Set cookie to just `/`
3. Add an explicit config option for this path 

I decided to go with option 1 because it is the most correct (from an operational POV) and keeps the cookie tightly scoped, while reusing the existing `PublicDNSName` as the single source of truth — exactly how the redirect URI is already derived. Option 2 was a viable quick fix but needlessly widens the cookie scope, right now it doesn't matter but it may in the future. Option 3 adds redundant configuration and user drift risk.

Behaviour is unchanged for root hosted/doman deployments (cookie path remains `/auth/callback`).

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

## Test instructions

`TestLoginStateCookiePath` covers the cookie-path derivation for: no base path, a base path, and a base path with a trailing slash. So I think we're all good here.

I didn't verify manually, as we can deploy to staging but... To verify manually, deploy JIMM behind an ingress with a path prefix (e.g. `JIMM_DNS_NAME=https://<host>/jimm-jimm`), perform a browser logi, and confirm the flow completes through `/jimm-jimm/auth/callback` without a `no state cookie present` error. And then in the browser we could inspect the `jimm-oauth-state` cookie and confirm its `Path` is `/jimm-jimm/auth/callback` (which if you successfully logged in, it obviously will be.

### DIFF
--- a/internal/jimmhttp/auth_handler.go
+++ b/internal/jimmhttp/auth_handler.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"encoding/json"
 	"net/http"
+	"strings"
 
 	"github.com/go-chi/chi/v5"
 	"github.com/juju/zaputil/zapctx"
@@ -34,6 +35,11 @@ type OAuthHandler struct {
 	Router                    *chi.Mux
 	authenticator             BrowserOAuthAuthenticator
 	dashboardFinalRedirectURL string
+	// cookieCallbackPath is the path used to scope the OAuth state cookie so
+	// that the browser sends it back on the callback request. It includes any
+	// external base path under which JIMM is hosted (e.g. when behind an
+	// ingress that prefixes requests with a sub-path).
+	cookieCallbackPath string
 }
 
 // OAuthHandlerParams holds the parameters to configure the OAuthHandler.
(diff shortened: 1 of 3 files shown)


internal/jimmhttp/auth_handler.go
```diff
@@ -44,6 +50,13 @@ type OAuthHandlerParams struct {
 	// DashboardFinalRedirectURL is the final redirection URL to send users to
 	// upon completing the authorisation code flow.
 	DashboardFinalRedirectURL string
+
+	// BasePath is the external base path under which JIMM is hosted, derived
+	// from JIMM's public DNS name (e.g. "/jimm-jimm" when served behind an
+	// ingress that prefixes requests with a sub-path). It is prepended to the
+	// OAuth state cookie path so the browser sends the cookie back on the
+	// callback request. It may be empty when JIMM is hosted at the root.
+	BasePath string
```
kian99, line 59: There is a simpler fix than extending this. We already create the `redirectURL` in `Login`, that redirect URL will contain any necessary path. I threw my agent at this and there is some code here that demonstrates the change https://github.com/canonical/jimm/compare/v3...kian99:jimm:adjust-cookie-path

kian99 (approved):

luci1900 (approved):

PR #2046, opened 2026-06-30
feat: jaas upgrade-controller

luci1900:
## Description

Add a `jaas upgrade-controller <controller-name>` command that upgrades the Juju agent running on a named backing controller registered with JIMM.

Uses a new `UpgradeController` RPC methods. The controller model UUID can't be directly exposed by JIMM since there is one for each backing controller, hence the need for a `jaas` command.

JIMM also doesn't have any (other) need to store the UUID of this model, so JIMM calls `ListModels` and picks the model named `"controller"`. It then calls the existing `API.UpgradeModel` with that UUID, keeping the upgrade logic entirely on the backing controller.

The command is restricted to JIMM admins, consistent with `AddController`, `RemoveController`, and other controller-level operations.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

## Test instructions

```bash
make dev-env

# bootstrap a controller making sure newer patch is available
AGENT_VERSION=3.6.21 ./local/jimm/setup-controller.sh

./local/jimm/add-controller.sh

jaas upgrade-controller qa-lxd --dry-run
jaas upgrade-controller qa-lxd
juju show-controller qa-lxd # will soon see the new version
jaas show-controller qa-lxd # will eventually see the new version
```

### DIFF
--- a/cmd/jaas/cmd/upgradecontroller.go
+++ b/cmd/jaas/cmd/upgradecontroller.go
@@ -1,0 +1,113 @@
+// Copyright 2026 Canonical.
+
+package cmd
+
+import (
+	"fmt"
+
+	"github.com/juju/cmd/v3"
+	"github.com/juju/gnuflag"
+	jujucmd "github.com/juju/juju/cmd"
+	"github.com/juju/juju/cmd/modelcmd"
+	"github.com/juju/juju/jujuclient"
+	"github.com/juju/version/v2"
+
+	apiparams "github.com/canonical/jimm/v3/pkg/api/params"
+)
+
+const (
+	upgradeControllerDoc = `
+Upgrades the Juju agent running on the named backing controller to the next
+available patch release. If --target-version is specified, upgrades to that
+exact version instead.
+
+The command requires the caller to be a JIMM admin.
+`
+	upgradeControllerExamples = `
+    jaas upgrade-controller mycontroller
+    jaas upgrade-controller mycontroller --target-version 3.6.8
+    jaas upgrade-controller mycontroller --dry-run
+`
+)
+
+// NewUpgradeControllerCommand returns a command to upgrade a backing controller's agent.
+func NewUpgradeControllerCommand() cmd.Command {
+	cmd := &upgradeControllerCommand{}
+	cmd.SetClientStore(jujuclient.NewFileClientStore())
+	return modelcmd.WrapBase(cmd)
+}
+
+type upgradeControllerCommand struct {
+	jaasCommandBase
+
+	controllerName      string
+	targetVersionStr    string
+	targetVersion       version.Number
+	agentStream         string
+	ignoreAgentVersions bool
+	dryRun              bool
+}
+
+func (c *upgradeControllerCommand) Info() *cmd.Info {
+	return jujucmd.Info(&cmd.Info{
+		Name:     "upgrade-controller",
+		Args:     "<controller-name>",
+		Purpose:  "Upgrades the agent of a backing Juju controller",
+		Doc:      upgradeControllerDoc,
+		Examples: upgradeControllerExamples,
+	})
+}
+
+func (c *upgradeControllerCommand) SetFlags(f *gnuflag.FlagSet) {
+	c.CommandBase.SetFlags(f)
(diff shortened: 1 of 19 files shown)


internal/jimm/juju/model.go
```diff
@@ -629,6 +629,46 @@ func (j *JujuManager) AbortModelUpgrade(ctx context.Context, user *openfga.User,
...
+		return version.Number{}, errors.Codef(errors.CodeServerError, "failed to list models on controller %q: %w", controllerName, err)
+	}
+
+	var controllerModelUUID string
+	for _, m := range models {
+		if m.Name == "controller" {
+			controllerModelUUID = m.UUID
+			break
+		}
+	}
```
kian99: This is fine, there is also this code in the Terraform provider that we use to get the controller model's UUID if preferred.
```go
	conn, err := connr.Connect()
	if err != nil {
		return version.Number{}, err
	}
	defer conn.Close()

	modelCfgClient := modelconfig.NewClient(conn)
	defer modelCfgClient.Close()

	modelAttrs, err := modelCfgClient.ModelGet()
	if err != nil {
		return version.Number{}, err
	}

	cfg, err := config.New(config.NoDefaults, modelAttrs)
	if err != nil {
		return version.Number{}, err
	}

	controllerModelUUID := cfg.UUID()
```

kian99 (approved):
A note about this feature in the "how-to manage controller" doc would also be a good addition.

internal/jimm/juju/model.go
luci1900: I'll have a look, I already don't like using list.

cmd/jaas/cmd/upgradecontroller.go
```diff
@@ -0,0 +1,113 @@
...
+		Doc:      upgradeControllerDoc,
+		Examples: upgradeControllerExamples,
+	})
+}
+
+func (c *upgradeControllerCommand) SetFlags(f *gnuflag.FlagSet) {
+	c.CommandBase.SetFlags(f)
+	f.StringVar(&c.targetVersionStr, "target-version", "", "Upgrade to this specific version")
+	f.StringVar(&c.agentStream, "agent-stream", "", "Check this agent stream for upgrades")
+	f.BoolVar(&c.ignoreAgentVersions, "ignore-agent-versions", false, "Don't check if all agents have already reached the current version")
```
ale8k, line 65: What does this mean?

snaps/jaas/snapcraft.yaml
```diff
@@ -89,4 +89,5 @@ parts:
       ln -sf jaas bin/juju-update-controller-profile
       ln -sf jaas bin/juju-list-models
       ln -sf jaas bin/juju-models
-      ln -sf jaas bin/juju-show-controller      
+      ln -sf jaas bin/juju-show-controller
```
ale8k, line 92: Weird it thinks this line has been deleted and readded with the exact same content? The link for show is still there right / is this a bug?

ale8k (approved):
Spot on to me, nice ty!

SimoneDutto (commented):
Aren't we doing this because we don't expose the controller model in jaas? because otherwise you could just run the upgrade-model on the backing controller model?

snaps/jaas/snapcraft.yaml
luci1900: Whitespace

cmd/jaas/cmd/upgradecontroller.go
luci1900: It's just matching the Juju CLi behaviour https://canonical.com/juju/docs/juju-cli/3.6/reference/juju-cli/list-of-juju-cli-commands/upgrade-controller/#options

luci1900:
> Aren't we doing this because we don't expose the controller model in jaas? because otherwise you could just run the upgrade-model on the backing controller model?

As discussed, the problem is about exposing the `controller` model for all backing controllers. We'd need to figure out UX on how to refer to this, which would require changing the juju cli command.

My approach is a compromise which avoid that, although it is indeed not the cleanest long term solution.

PR #2047, opened 2026-06-30
refactor: remove group from JAAS

SimoneDutto:
# Description
groups are removed from JAAS, with that the tuples will be eventually cleaned up by the cleanup routine.

The strategy to remove this:
- group removed from jimm, together with group tag and group's API, and commands
- all the tests involving groups have been refactored to use idpgroups to keep maximum coverage
- no bump in the facade, because bumping the facade would require us to update the jimm-sdk and the terraform provider, which forces ppl to upgrade their terraform if they want to use jimm v4, which is exactly what we don't want. So the group's method are still in the sdk to support using jimm-sdk@v4 against jaas v3.
- the tuples left in openfga will be cleaned up by our cleanup routine because tuples with the group tag will always return a `NotFound` error


internal/jimmhttp/rebac_admin/identities_test.go
```diff
@@ -198,19 +176,12 @@ func TestGetIdentityGroups(t *testing.T) {
...
 	c.Assert(res, qt.IsNotNil)
-	c.Assert(res.Data, qt.HasLen, 1)
-	c.Assert(*res.Data[0].Id, qt.Equals, "my-group-id")
-	c.Assert(res.Data[0].Name, qt.Equals, "fake-group-name")
-	c.Assert(*res.Next.PageToken, qt.Equals, "continuation-token")
-
-	listTuplesErr = errors.New("foo")
-	_, err = idSvc.GetIdentityGroups(ctx, username, &resources.GetIdentitiesItemGroupsParams{})
-	c.Assert(err, qt.ErrorMatches, "foo")
+	// JAAS groups have been removed, so the result is always empty.
```
ale8k: I think I would've expected an error on the rebac API for group related stuff explaining groups don't exist within JAAS and to contact admin or w/e.

internal/jimmhttp/rebac_admin/identities.go
```diff
@@ -196,98 +196,40 @@ func (s *identitiesService) PatchIdentityRoles(ctx context.Context, identityId s
...
-			return false, v1.NewUnknownError(err.Error())
-		}
-	}
-	if len(deletions) > 0 {
-		err = s.jimm.PermissionManager().RemoveRelation(ctx, user, deletions)
-		if err != nil {
-			zapctx.Error(context.Background(), "cannot remove relations", zap.Error(err))
-			return false, v1.NewUnknownError(err.Error())
-		}
+	if len(groupPatches) > 0 {
```
ale8k, line 228: Referring to my other comment, like you did here

internal/jimmhttp/rebac_admin/identities.go
```diff
@@ -196,98 +196,40 @@ func (s *identitiesService) PatchIdentityRoles(ctx context.Context, identityId s
...
-		Meta: resources.ResponseMeta{
-			Size:      len(groups),
-			PageToken: &originalToken,
-		},
-	}
-	if cNextToken != "" {
-		res.Next.PageToken = &cNextToken
-	}
-	return &res, nil
+	return &resp, nil
```
ale8k: Why not return an error like patch?

internal/jimm/permissions/tagresolver.go
```diff
@@ -123,6 +114,13 @@ func (j *PermissionManager) ToJAASTag(ctx context.Context, tag *ofganames.Tag, r
 			return "", fmt.Errorf("failed to fetch role information: %s: %w", role.UUID, err)
 		}
 		return tagToString(jimmnames.RoleTagKind, role.Name), nil
+	case ofganames.GroupTagKind:
+		// The JAAS group type has been removed from the OpenFGA
```
ale8k, line 118: This doc should  probably explain that they've been replaced with IdP groups rathenr than simply removed.

internal/openfga/names/names.go
```diff
@@ -48,14 +48,20 @@ var allRelations = []cofga.Relation{MemberRelation, AdministratorRelation, Contr
 // EveryoneUser is the username representing all users and is treated uniquely when used in OpenFGA tuples.
 const EveryoneUser = "everyone@external"
 
+// GroupTagKind is the kind string for JAAS groups. The group type has been
+// removed from the OpenFGA authorisation model and the groups table has been
```
ale8k, line 52: Same here, they've been replaced rather

ale8k (approved):
lgtm, some very minor nit comments, ty!

kian99 (approved):

internal/jimmhttp/rebac_admin/identities.go
SimoneDutto: yeah, you are right

internal/jimm/permissions/tagresolver.go
SimoneDutto: well, here we don't exactly care what's the replacement for this, also because it's just a comment to explain why we still have a group tag kind here

PR #2048, opened 2026-06-30
fix: also clean up dangling offer on creation

luci1900:
## Description

Application offers in JIMM can dangle, where the record exists in JIMM's database and OpenFGA but the offer no longer exists on the Juju controller. The way this can happen is the destroying via JIMM where the request is made against the controller and succeeds, but JIMM never receives the response and times out.

Previously, dangling offer cleanup was only triggered on **read** (`GetApplicationOffer`). If a user tried to re-create an offer with the same URL without reading it first, JIMM would find the stale DB record and return `already exists`, leaving the dangling state in place indefinitely.

This PR adds the same cleanup on **create** (`Offer`). When the duplicate-URL check finds an existing record, JIMM now verifies whether the offer still exists on the controller.

Addresses https://github.com/canonical/jimm/issues/1942

## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests

## Test instructions

```bash
make qa-lxd

# create an offer through jimm
juju switch jimm-dev:jimm-test
juju deploy juju-qa-dummy-source dummy-source
juju offer dummy-source:sink myoffer
juju find-offers jimm-dev:            # myoffer listed

# simulate dangling
juju switch qa-lxd:jimm-test@canonical.com/test-lxd
juju remove-offer jimm-test@canonical.com/test-lxd.myoffer --force -y
juju find-offers jimm-dev:             # myoffer not seen

# reacreate same named offer
juju switch jimm-dev:test-lxd
juju offer dummy-source:sink myoffer
juju show-offer jimm-test@canonical.com/test-lxd.myoffer  # returns details, no error
```

### DIFF
--- a/internal/jimm/juju/applicationoffer.go
+++ b/internal/jimm/juju/applicationoffer.go
@@ -74,7 +74,26 @@ func (j *JujuManager) Offer(ctx context.Context, user *openfga.User, offer AddAp
 	offerCheck.URL = offerURL.String()
 	err = j.Database.GetApplicationOffer(ctx, &offerCheck)
 	if err == nil {
-		return errors.Codef(errors.CodeAlreadyExists, "offer %s already exists, please use a different name", offerURL.String())
+		// The offer exists in JIMM's database. Check whether it still
+		// exists on the controller — it may be dangling (e.g. the
+		// controller-side offer was destroyed but JIMM's cleanup did not
+		// complete). If the controller reports it as not found, remove
+		// the stale record and continue with the creation.
+		checkAPI, dialErr := j.dial(ctx, &model.Controller, names.ModelTag{}, user)
+		if dialErr != nil {
+			return dialErr
+		}
+		_, controllerErr := checkAPI.GetApplicationOffer(ctx, offerURL.String())
+		checkAPI.Close()
+		if controllerErr != nil && errors.ErrorCode(controllerErr) == errors.CodeNotFound {
+			// Dangling offer: clean it up and proceed to (re-)create it.
+			if cleanupErr := j.deleteApplicationOffer(ctx, &offerCheck); cleanupErr != nil {
+				zapctx.Error(ctx, "error cleaning up dangling offer on create", zap.Error(cleanupErr))
+				return cleanupErr
+			}
+		} else {
+			return errors.Codef(errors.CodeAlreadyExists, "offer %s already exists, please use a different name", offerURL.String())
+		}
 	} else if errors.ErrorCode(err) != errors.CodeNotFound {
 		// Anything besides Not Found is a problem.
 		return err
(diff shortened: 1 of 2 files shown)


internal/jimm/juju/applicationoffer.go
```diff
@@ -74,7 +74,27 @@ func (j *JujuManager) Offer(ctx context.Context, user *openfga.User, offer AddAp
...
+		// The offer exists in JIMM's database, check against the Juju controller.
+		// It's possible for an offer record in JIMM to dangle.
+		checkAPI, dialErr := j.dial(ctx, &model.Controller, names.ModelTag{}, user)
+		if dialErr != nil {
+			return dialErr
+		}
+		_, controllerErr := checkAPI.GetApplicationOffer(ctx, offerURL.String())
+		checkAPI.Close()
+		if controllerErr == nil {
+			// Actual duplicate offer
```
kian99: The comment is a bit weird, why call it a duplicate?
I would just delete it (or say "Not a dangling offer"), the error message is clear enough.

kian99 (approved):
Ty for this

internal/jimm/juju/applicationoffer.go
luci1900: It's a wrong comment from a previous iteration of this, indeed.

luci1900: "duplicate" as in collision between a remaining non-dangling offer and one we're trying to create

ale8k (approved):

internal/jimm/juju/applicationoffer.go
```diff
@@ -74,9 +74,29 @@ func (j *JujuManager) Offer(ctx context.Context, user *openfga.User, offer AddAp
...
+		}
+		if errors.ErrorCode(controllerErr) != errors.CodeNotFound {
+			// Any other error
+			return controllerErr
+		}
+		// Dangling offer, clean it up and continue
+		if cleanupErr := j.deleteApplicationOffer(ctx, &offerCheck); cleanupErr != nil {
+			zapctx.Error(ctx, "error cleaning up dangling offer on create", zap.Error(cleanupErr))
+			return cleanupErr
+		}
```
SimoneDutto, line 97: in `ModelInfo` we have a this pattern:
```
if errFromAPI == nil {
		return j.reactToModelInfoSuccess(ctx, model, modelInfo)
	} else {
		return j.reactToModelInfoError(ctx, user, errFromAPI, model)
	}
```

since you've mentioned you have multiple occurences of this react code, maybe you want to replicate this

luci1900: Yes, I had considered that even originally.

Both of my cases so far are quite different, though.

PR #2049, opened 2026-07-06
refactor: model deletion tweaks

kian99:
## Description
This change slightly adjusts how we keep JIMM's database in sync with Juju, specifically with regards to models. Currently whenever certain calls like ModelInfo or ModelStatus return a NotFound or Unauthorized error, we would remove the model from JIMM's database assuming that this was because it was removed from the backing controller. That assumption is accurate, calls like `ModelInfo` return an Unauthorized error for a model that doesn't exist even when the user is a controller admin.

While this is usually works fine, there was one incident on the production JAAS where a model dissapeared from JIMM's database but was still present in the backing controller. It's hard to "re-sync" this model to JIMM because we lose all the authorization information when we delete resources.

This change makes the sync more conservative, ensuring we only remove models from JIMM that were marked as "dying", i.e. a user had previously run a `juju destroy-model` against JIMM. The only additional issue this revealed is that the Juju CLI does a `ModelStatus` call before doing a `DestroyModel` call, meaning that with JIMM's currently logic, if the model does not exist on a backing controller but does exist in JIMM, it is impossible, using the Juju CLI, to delete it from JIMM. So, we adjust `ModelStatus` to return a minimal status if the model does not exist on the backing controller.

This PR is mainly tests beyond the changes described above, unit tests, integration tests and an e2e test with the Juju CLI.

Fixes [JUJU-9182](https://warthogs.atlassian.net/browse/JUJU-9182)

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

[JUJU-9182]: https://warthogs.atlassian.net/browse/JUJU-9182?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/jimm/juju/model.go
+++ b/internal/jimm/juju/model.go
@@ -13,6 +13,7 @@ import (
 	"time"
 
 	"github.com/juju/juju/api/base"
+	"github.com/juju/juju/core/life"
 	jujuparams "github.com/juju/juju/rpc/params"
 	"github.com/juju/juju/state"
 	"github.com/juju/names/v5"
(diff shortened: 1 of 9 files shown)


internal/jimm/juju/model.go
```diff
@@ -439,29 +440,58 @@ func (j *JujuManager) mergeModelInfo(ctx context.Context, user *openfga.User, mo
...
+		}
+		return base.ModelStatus{}, err
 	}
 	return ms, nil
 }
 
+func fallbackModelStatus(m *dbmodel.Model, err error) base.ModelStatus {
+	modelLife := life.Value(m.Life)
+	if modelLife == "" {
+		modelLife = life.Alive
```
SimoneDutto: since this fix is targeting a model not-found on the backing controller, should we put "dying" here?

SimoneDutto (approved):
in this way we are making the juju CLI idempotent with deletion, which i think it's good.

I'd just add some documentation somewhere to explain very simply "what to do if i delete a model from a controller behind jaas", so we can just link it when it happens.

What do you think?

internal/jimm/juju/model.go
kian99: I was thinking to remove the fallback entirely since the Life field is unlikely to be empty since it would be either "Alive" if you've never run `juju destroy-model` and then will transition to "Dying" after running `juju destroy-model`.

kian99:
> in this way we are making the juju CLI idempotent with deletion, which i think it's good.
> 
> I'd just add some documentation somewhere to explain very simply "what to do if i delete a model from a controller behind jaas", so we can just link it when it happens.
> 
> What do you think?

Yeah I'll add a section on it in the docs.

kian99:
> in this way we are making the juju CLI idempotent with deletion, which i think it's good.

Btw, this change doesn't make deletion idempotent, at least in the sense that `juju destroy-model` still returns an error if the model doesn't exist but that matches Juju's behaviour. Only if the model exists in JIMM's DB but not Juju, then there is no error, the deletion just works. Calling `juju destroy-model` again, still returns an error.

internal/jimm/juju/model.go
luci1900: My first thought was also that it should fall back to "Dying".

Not needing the fallback seems better.

internal/jimm/juju/model_poller.go
```diff
@@ -129,6 +130,18 @@ func (j *JujuManager) maybeCleanupModel(ctx context.Context, errFromAPI error, m
 	// Some versions of juju return unauthorized for models that cannot be found.
```
luci1900, line 130: As much as I dislike branching on versions, I keep having this niggling worry that a future Juju 4 version might change how auth works in a way that is otherwise compatible except for this case.

Perhaps not worth it, but I'd imagine it would be safer to stuff this check in a helper that branches on version first.

luci1900 (approved):

internal/jimm/juju/model_poller.go
kian99: Yeah this comment is pretty obscure and I initially changed it in this PR but reverted the change. What i found was that ModelInfo always returned Unauthorized when the model didn't exist, even for an admin while DestroyModel returned a NotFound for a model that doesn't exist so while it may be true that some versions of Juju return one or the other, it also seems true that some endpoints return differing errors for similar operations where a resource does not exist.

internal/jimm/juju/model.go
```diff
@@ -439,29 +440,54 @@ func (j *JujuManager) mergeModelInfo(ctx context.Context, user *openfga.User, mo
...
 	}
 	return ms, nil
 }
 
+func fallbackModelStatus(m *dbmodel.Model, err error) base.ModelStatus {
+	return base.ModelStatus{
+		UUID:  m.UUID.String,
+		Life:  life.Value(m.Life),
+		Owner: m.OwnerIdentityName,
+		Error: err,
```
ale8k, line 487: Ah nice, I like this similar to what I was trying to suggest last week

internal/jimm/juju/model.go
```diff
@@ -439,29 +440,54 @@ func (j *JujuManager) mergeModelInfo(ctx context.Context, user *openfga.User, mo
...
+	}
+	if !hasAccess {
+		return base.ModelStatus{}, errors.Codef(errors.CodeUnauthorized, "unauthorized")
+	}
+
+	api, err := j.dial(ctx, &m.Controller, names.ModelTag{}, user)
+	if err != nil {
+		return base.ModelStatus{}, err
+	}
+	defer api.Close()
```
ale8k, line 461: We often defer close like this but I think it should be in a callback where we log the error  otherwise if this ever has issues, we'll struggle to know why

tests/modeldeletion/delete-from-backing-controller.sh
```diff
@@ -0,0 +1,75 @@
...
+# from JAAS via the Juju CLI.
+#
+# This test covers the case where JAAS still has a database record for a
+# model that no longer exists on its backing controller. In that case, a
+# Juju CLI destroy-model request through JAAS should clean up the JAAS model
+# record successfully.
+#
+# Usage: ./tests/modeldeletion/delete-from-backing-controller.sh
+
+set -euo pipefail
```
ale8k, line 17: Do we need this test? Feels like our modelmanager_test is sufficient?

ale8k (approved):
Nice work on this some minor comments

tests/modeldeletion/delete-from-backing-controller.sh
kian99: It's just verifying the behaviour with the Juju CLI because we change ModelStatus specifically for the Juju CLI to still work deleting a model that doesn't exist in the backing controller.

internal/jimm/juju/model.go
kian99: Seems reasonable but can do in a follow-up.

PR #2050, opened 2026-07-07
test(jujuapi): golden-file guard for 3.6 facade versions and wire shapes

alesstimec:
## Description

Add CI guards against silent drift in the surface JIMM presents to Juju 3.6 clients:

- TestSupportedFacadesGolden pins the advertised facade/version matrix (SupportedFacades), so a version unintentionally added or removed fails CI.
- TestLegacyWireShapesGolden pins the field/json-tag signatures of the 3.6 wire structs (the juju *Legacy twins, plus ModelConfigResults), so a juju dependency bump that silently renames a field (e.g. owner-tag -> qualifier) or drops a *Legacy struct fails CI loudly. This is the one drift class facade-version negotiation cannot catch.
- TestAdvertisedVersionsAreDispatchable asserts every advertised facade version has at least one registered method (absorbed from the dropped JAAS-2), backed by a re-added rpc.Root.HasMethods introspection helper.

Golden files live in internal/jujuapi/testdata and are regenerated with `go test ./internal/jujuapi/ -run Golden -update-golden`. They reflect the current feature/juju-4 state; landing the Controller v12 / ApplicationOffers v5 / ModelConfig v3 work will intentionally change the advertised-versions golden, forcing a conscious regeneration.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests


internal/jujuapi/legacywire/legacywire.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+// structSignature returns a deterministic, one-level description of a struct's
+// fields: name, json tag and type. A renamed/retyped/added/removed field
+// changes the signature.
+func structSignature(t reflect.Type) string {
+	fields := make([]string, 0, t.NumField())
+	for i := 0; i < t.NumField(); i++ {
+		f := t.Field(i)
+		fields = append(fields, fmt.Sprintf("%s `json:%q` %s", f.Name, f.Tag.Get("json"), f.Type.String()))
```
luci1900, line 94: Is there a reason the output isn't an existing format like JSON?

internal/jujuapi/testdata/legacy_wire_shapes.txt
```diff
@@ -0,0 +1,117 @@
+# Code generated by cmd/legacywire; DO NOT EDIT.
+#
+# Field/json-tag signatures of the Juju 3.6 wire structs JIMM serves. A renamed,
+# retyped, added or removed field (most importantly a silent owner-tag ->
+# qualifier rename on a juju bump) changes this file and fails CI.
+#
+# Regenerate with: make update-golden
+
+=== ModelConfigResults ===
```
luci1900, line 9: Yes, I do think this would be better as either JSON or INI/TOML.

PR #2051, opened 2026-07-07
refactor: adjust model import

kian99:
## Description
This PR fixes an issue in JIMM's model import logic. Originally this logic was written to be able to take a set of models created outside of JAAS and import them. That was never used and actually has issues where e.g. creating an offer for a model where the owner in JIMM != owner in Juju fails. Where this logic is useful is rebuilding JIMM's state if a model is present in the backing controller but not in JIMM.

One issue in the import logic was that JIMM associated a cloud-credential with a model, during import it picked any credential in JIMM that was A) Appropriate for the model's cloud and B) had the same owner. This breaks down when using a reconciliation tool like Terraform which may try to reconcile a model's credential. Specifically in a scenario with IS, they had a model in Juju that was not present in JIMM, imported the model back into JIMM and then Terraform notices that the cloud-credential that JIMM reports for the model is not the same as the one it expects.

To resolve this, we:
1. Require the same credential in JIMM as the importing model.
2. Drop support for changing model owner, this causes issues when creating offers and doesn't align with 1. when importing a model using a credential owned by a local user.

Fixes [JUJU-10117](https://warthogs.atlassian.net/browse/JUJU-10117)

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

[JUJU-10117]: https://warthogs.atlassian.net/browse/JUJU-10117?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/jimm/juju/controller.go
+++ b/internal/jimm/juju/controller.go
@@ -370,11 +370,9 @@ func (j *JujuManager) EarliestControllerVersion(ctx context.Context) (version.Nu
 }
 
 type modelImporter struct {
-	jimm      *JujuManager
-	model     dbmodel.Model
-	modelInfo base.ModelInfo
-	// newOwner may be nil if the user wants to keep the original owner.
-	newOwner      *names.UserTag
+	jimm          *JujuManager
+	model         dbmodel.Model
+	modelInfo     base.ModelInfo
 	originalOwner names.UserTag
 	offersToAdd   []*crossmodel.ApplicationOfferDetails
 }
@@ -383,14 +381,9 @@ func newModelImporter(jimm *JujuManager, newOwner string) (modelImporter, error)
 	modelImporter := modelImporter{
 		jimm: jimm,
 	}
-	if newOwner == "" {
-		return modelImporter, nil
+	if newOwner != "" {
+		return modelImporter, errors.Codef(errors.CodeBadRequest, "switching the imported model owner is no longer supported")
 	}
-	if !names.IsValidUser(newOwner) {
-		return modelImporter, errors.Codef(errors.CodeBadRequest, "invalid new username for new model owner")
-	}
-	newOwnerTag := names.NewUserTag(newOwner)
-	modelImporter.newOwner = &newOwnerTag
 	return modelImporter, nil
 }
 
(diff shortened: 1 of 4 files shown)


internal/jimm/juju/controller.go
```diff
@@ -383,14 +381,9 @@ func newModelImporter(jimm *JujuManager, newOwner string) (modelImporter, error)
 	modelImporter := modelImporter{
 		jimm: jimm,
 	}
-	if newOwner == "" {
-		return modelImporter, nil
+	if newOwner != "" {
+		return modelImporter, errors.Codef(errors.CodeBadRequest, "switching the imported model owner is no longer supported")
```
luci1900, line 385: I suppose since it was broken, we don't care about existing usage?

kian99: That was my thinking

luci1900 (approved):
Reasonable.

Even if we did somehow have usage, we can revive it based on the field that we now error on.

internal/jimm/juju/controller_test.go
```diff
@@ -783,15 +737,15 @@ func TestImportModel(t *testing.T) {
...
-			info.Owner = "local-user"
+			info.CloudCredential = "invalid-cloud/alice@canonical.com/unknown-credential"
+			info.Owner = "alice@canonical.com"
 			return info, nil
 		},
-		expectedError: `cannot import model from local user, try --owner to switch the model owner`,
+		expectedError: `Failed to find cloud credential "unknown-credential" for user alice@canonical.com on cloud invalid-cloud`,
 	}, {
-		about:          "cloud credentials not found",
+		about:          "matching cloud credential name not found",
```
SimoneDutto, line 748: what's this test about?

SimoneDutto (commented):
seems weird what we were doing before :D

SimoneDutto (approved):

internal/jimm/juju/controller_test.go
kian99: If the cloud-credential for the model you are importing is not in JIMM, it should fail.

PR #2052, opened 2026-07-07
test(jujuapi): golden-file guard for the Juju 3.6 wire formats

alesstimec:
## Description

Adds a CI guard for the one drift class that facade-version negotiation cannot catch: a juju dependency bump silently changing a wire struct at a fixed facade version (e.g. an owner-tag → qualifier rename), which would break 3.6 clients without failing any build.

## Engineering checklist

- [x] Documentation updated
- [x] Covered by unit tests


cmd/legacywire/README.md
```diff
@@ -0,0 +1,41 @@
...
+make update-golden
+```
+
+Or run the generator directly:
+
+```bash
+go run ./cmd/legacywire -o internal/jujuapi/testdata/legacy_wire_shapes.json
+```
+
+It is also wired into `go generate ./internal/jujuapi`, so `make generate`
```
ale8k, line 33: I think this is something we'd run very explicitly in its own action and probably not via generate. I'd expect a solid output on the action telling us exactly where the drift happened. I think you can also pick which generator to run too I guess.

cmd/legacywire/README.md
```diff
@@ -0,0 +1,41 @@
+# legacywire tool
+
+This tool emits a golden snapshot of the **Juju 3.6 ("legacy") wire-struct
+signatures** JIMM serves to, or accepts from, 3.6 clients
```
ale8k, line 4: So we'll produce a file of all versions JIMM consumes. But now we need toe Juju side to make the comparison right?

ale8k (commented):
I think in principle this is the right approach but it'd run in Juju right and this would mean it needs either:

1. Importing from here - not possible
2. An action making to create this file

A juju side component to run the check on their end and it is required? 

Perhaps I'm misunderstanding so lmk if I'm missing it

luci1900 (approved):
The checker seems fine.

Where would it get run, though?

kian99 (commented):
I see we're generating `internal/jujuapi/testdata/legacy_wire_shapes.json` based on the Juju API params, which will be based on the version of Juju that we import.

This guard will only trigger a warning once we update our Juju dependency in the go.mod which may/may not be too late depending on when Juju releases.

To avoid that we could run this generator daily against the tip of Juju to catch breaking changes to these APIs early.

We could also have some human-check/code/generator in Juju that verifies any change to these APIs triggers a warning.

What do you think?
1. Keep this change as is
2. Run daily/weekly against the tip of Juju
3. Do the guard in Juju verifying against JIMM
4. Do nothing and assume that people won't make breaking changes to these API params

alesstimec:
I am not too attached to this change. It's just something AI suggested in the plan. If anything i'd run this check daily against the tip of juju.

alesstimec:
@ale8k @luci1900 To clarify where this runs: entirely in this repo's CI — no juju-side component needed. The golden file is generated from the wire structs of the juju version pinned in our go.mod, and CI (verify_generated + TestLegacyWireShapesGolden) compares the committed file against a fresh regeneration. Day to day it's inert; it fires exactly when someone bumps the juju dependency and one of the 3.6 wire structs changed shape — so a silent owner-tag-style break becomes a red CI check whose diff the bump author has to consciously review and regenerate past.

@kian99 yes, the guard alone only fires at bump time, so I've now added your option 2 to this PR as well — the two mechanisms are complementary rather than alternatives:

- The in-repo guard (option 1) is enforcement: JIMM can never ship changed 3.6 wire shapes unnoticed, regardless of when we bump. That's the part that protects users.
- The new legacy-wire-drift workflow (option 2, last commit) is early warning: a daily scheduled job that moves the juju dependency to the tips of the 4.0 and main branches, regenerates the snapshot, and fails on any difference — giving us lead time to raise drift with the juju team before it ships in a release. A build failure of the generator against a tip is treated as drift too, since it means a pinned struct was removed upstream (the *Legacy params are temporary there). One caveat: GitHub only fires cron from the default branch, so the daily run arms itself once this reaches it; until then it's runnable via workflow_dispatch.

I've dry-run the workflow logic locally against both tips: the generator builds and the golden file matches on today's 4.0 and main — so we're starting from a clean baseline.

internal/jujuapi/legacywire/legacywire.go
```diff
@@ -0,0 +1,169 @@
...
+	enc.SetEscapeHTML(false)
+	enc.SetIndent("", "  ")
+	if err := enc.Encode(document{Comment: comment, Types: types}); err != nil {
+		// The document is plain strings and bools; Encode cannot fail.
+		panic(err)
+	}
+	return buf.Bytes()
+}
+
+// collect adds to seen every juju-owned named struct type reachable from t
```
kian99, line 121:
```suggestion
// collect adds to seen every juju-owned named struct type reachable
```

kian99 (commented):
Conceptually I think this is fine and good. Practically, I slightly dislike like the extra code, although it's not very long but it seems like a marginal gain. How worried are we that Juju will change these legacy structs?

It also seems that we could accomplish something similar by running our tests against a Juju 3 controller daily, though of course that doesn't give us the exact same guarantee that the API shape hasn't changed.

Curious overall what others think.

luci1900:
I also don't like that it's a bunch of code we're adding instead of removing, but it is isolated and fairly easy to remove.

The main thing I like about this check is that it's fast to run. I'm unconvinced it will ever catch anything, but I don't think there's any harm.

PR #2053, opened 2026-07-07
feat: show controller models

SimoneDutto:
## Description

Show controller models in jaas.
The idea is to allow controller admin to switch to the controller model to run command (ex. juju debug-log), and also allow in the future to integrate COS with the controller application via terraform using JAAS.

The implementation is pretty straight forward:
- store the controller facing owner, which is the owner we will use when we talk to the backing controller
  - this will solve contextually the problem with import models with local users, or importing models with another external user
- the translation is done everytime we interact with offers in jaas, so during get, find, list, destroy. Since it's done just in JAAS, and the source of truth is kept in the backing controller i don't see potential issues we can't solve.

Potential problems for drifting terraform states(please think about scenarios, and we can qa/test it together)
- import model didnt' work before for local users, so there shouldn't be any offer created for those models that might drift with this change
- controller models were not shown so no drift there as well
- the only potential drift if somebody was listing all models to do something and now it will find an additional model, but i don't think that's a problem

# QA (controller model)

`make qa-lxd`

now `juju models` shows:
```
[4:21:31] ➜  jimm git:(backing-controller) juju models
Controller: jimm-dev

Model                          Cloud/Region         Type  Status     Machines  Units  Access  Last connection
qa-lxd@controller/controller*  localhost/localhost  lxd   available         1      1  admin   never connected
test-lxd                       localhost/localhost  lxd   available         1      1  admin   never connected
b                              localhost/localhost  lxd   available         1      1  admin   never connected
```

you should be able to switch to it, create an offer and consume it on another model.
For example:
```
juju switch qa-lxd@controller/controller
juju offer controller dashboard
juju add-model c
juju deploy juju-dashboard
juju consume <offer-url>
juju integration juju-dashboard controller
```

# QA (import model with local user)

```
juju switch qa-lxd
juju add-model c
```

```
juju switch jimm-dev
./jaas import-model qa-lxd 0893f41c-c129-48a2-812d-5c55a6471fce --owner jimm-test@canonical.com
juju deploy juju-qa-dummy-sink
juju offer dummy-sink:source 
juju switch b                               
juju consume jimm-test@canonical.com/c.dummy-sink                                       
```

### DIFF
--- a/internal/dbmodel/model.go
+++ b/internal/dbmodel/model.go
@@ -55,6 +55,17 @@ type Model struct {
 	OwnerIdentityName string   `gorm:"uniqueIndex:unique_model_names;not null"`
 	Owner             Identity `gorm:"foreignkey:OwnerIdentityName;references:Name"`
 
+	// ControllerOwnerName is the model owner as known by the backing Juju
+	// controller. JAAS can alias a model's owner locally (imported models with a
+	// new owner, or backing controller models), so this preserves the owner the
+	// backing controller actually knows. When empty, OwnerIdentityName is also
+	// the controller-facing owner.
+	ControllerOwnerName string
+
+	// IsControllerModel records whether this model is a backing controller's
+	// model tracked by JAAS.
+	IsControllerModel bool `gorm:"not null;default:false"`
+
 	// Controller is the controller that is hosting the model.
 	ControllerID uint
 	Controller   Controller
@@ -77,9 +88,35 @@ type Model struct {
 	MigrationMode MigrationMode `gorm:"default:''"`
 }
 
+const (
+	// ControllerModelOwnerSuffix is the suffix of the synthetic identity that
+	// owns a backing controller's model within JAAS. The owner is
+	// "<controller-name>@controller", which lets us alias the (otherwise
+	// identically named) "controller" model of each backing controller to a
+	// unique "<controller-name>@controller/controller" address.
+	ControllerModelOwnerSuffix = "@controller"
+
+	// ControllerModelName is the name a controller model always has on its
+	// backing Juju controller.
+	ControllerModelName = "controller"
+)
+
 // Tag returns a names.Tag for the model.
 func (m Model) Tag() names.Tag {
 	return m.ResourceTag()
+}
+
+// ControllerFacingOwner returns the model owner as known by the backing Juju
+// controller. JAAS can alias a model's owner locally (imported models with a
+// new owner, or backing controller models), so operations that address the
+// model on the backing controller (e.g. building offer URLs) must use this
+// owner rather than OwnerIdentityName. When no aliased owner is recorded, the
+// JAAS owner is also the controller-facing owner.
+func (m Model) ControllerFacingOwner() string {
+	if m.ControllerOwnerName != "" {
+		return m.ControllerOwnerName
+	}
+	return m.OwnerIdentityName
 }
 
 // ResourceTag returns a tag for the model.  This method
--- a/internal/jimm/juju/applicationoffer.go
+++ b/internal/jimm/juju/applicationoffer.go
@@ -28,6 +28,35 @@ import (
 	ofganames "github.com/canonical/jimm/v3/internal/openfga/names"
 )
 
+// controllerFacingOfferURL translates a JAAS offer URL into the URL the backing
+// controller knows. If the URL's owner matches a tracked model's JAAS owner but
+// differs from its controller-facing owner (imported or controller models), the
+// owner is rewritten. If the model cannot be resolved the URL is returned
+// unchanged so that offers stored under their original URL still work.
+func (j *JujuManager) controllerFacingOfferURL(ctx context.Context, offerURL string) (string, error) {
+	ourl, err := crossmodel.ParseOfferURL(offerURL)
+	if err != nil {
+		// Pass the URL through unchanged so the DB lookup can still
+		// match offers stored under a URL that ParseOfferURL can't
+		// parse (e.g. legacy data).
+		//nolint:nilerr // Intentional: we deliberately ignore the parse error
+		// to preserve backward compatibility.
+		return offerURL, nil
+	}
+	if ourl.User == "" || ourl.ModelName == "" {
+		return offerURL, nil
+	}
+	m := dbmodel.Model{Name: ourl.ModelName, OwnerIdentityName: ourl.User}
+	if err := j.Database.GetModel(ctx, &m); err != nil {
+		if errors.ErrorCode(err) == errors.CodeNotFound {
+			return offerURL, nil
+		}
+		return "", err
+	}
+	ourl.User = m.ControllerFacingOwner()
+	return ourl.String(), nil
+}
+
 // AddApplicationOfferParams holds parameters for the Offer method.
 type AddApplicationOfferParams struct {
 	ModelTag               names.ModelTag
@@ -62,7 +91,9 @@ func (j *JujuManager) Offer(ctx context.Context, user *openfga.User, offer AddAp
 	}
 
 	offerURL := crossmodel.OfferURL{
-		User:      model.OwnerIdentityName,
+		// The offer URL must use the owner the backing controller knows, which
+		// can differ from the JAAS owner (see Model.ControllerFacingOwner).
+		User:      model.ControllerFacingOwner(),
 		ModelName: model.Name,
 		// Confusingly the application name in the offer URL is
 		// actually the offer name.
@@ -201,8 +232,15 @@ func (j *JujuManager) Offer(ctx context.Context, user *openfga.User, offer AddAp
 // the rest of the details.
 func (j *JujuManager) GetApplicationOfferConsumeDetails(ctx context.Context, user *openfga.User, details *jujuparams.ConsumeOfferDetails, v bakery.Version) error {
 
+	// Translate the JAAS offer URL to the controller-facing URL the offer is
+	// stored under and the backing controller expects.
+	controllerURL, err := j.controllerFacingOfferURL(ctx, details.Offer.OfferURL)
+	if err != nil {
+		return err
+	}
+
 	offer := dbmodel.ApplicationOffer{
-		URL: details.Offer.OfferURL,
+		URL: controllerURL,
 	}
 	if err := j.Database.GetApplicationOffer(ctx, &offer); err != nil {
 		if errors.ErrorCode(err) == errors.CodeNotFound {
@@ -233,7 +271,7 @@ func (j *JujuManager) GetApplicationOfferConsumeDetails(ctx context.Context, use
 	}
 	defer api.Close()
 
-	consumeDetails, err := api.GetApplicationOfferConsumeDetails(ctx, details.Offer.OfferURL)
+	consumeDetails, err := api.GetApplicationOfferConsumeDetails(ctx, controllerURL)
 	if err != nil {
 		return err
 	}
@@ -359,10 +397,16 @@ func (j *JujuManager) enrichOfferDetails(ctx context.Context, user *openfga.User
 // GetApplicationOffer returns details of the offer with the specified URL.
 func (j *JujuManager) GetApplicationOffer(ctx context.Context, user *openfga.User, offerURL string) (*crossmodel.ApplicationOfferDetails, error) {
 
-	offer := dbmodel.ApplicationOffer{
-		URL: offerURL,
+	// Translate the JAAS offer URL to the controller-facing URL.
+	controllerURL, err := j.controllerFacingOfferURL(ctx, offerURL)
+	if err != nil {
+		return nil, err
 	}
-	err := j.Database.GetApplicationOffer(ctx, &offer)
+
+	offer := dbmodel.ApplicationOffer{
+		URL: controllerURL,
+	}
+	err = j.Database.GetApplicationOffer(ctx, &offer)
 	if err != nil {
 		if errors.ErrorCode(err) == errors.CodeNotFound {
 			return nil, errors.Codef(errors.CodeNotFound, "application offer not found")
@@ -391,7 +435,7 @@ func (j *JujuManager) GetApplicationOffer(ctx context.Context, user *openfga.Use
 	}
 	defer api.Close()
 
-	offerDetails, err := api.GetApplicationOffer(ctx, offerURL)
+	offerDetails, err := api.GetApplicationOffer(ctx, controllerURL)
 	if err != nil {
 		if errors.ErrorCode(err) != errors.CodeNotFound {
 			return nil, err
@@ -416,8 +460,8 @@ func (j *JujuManager) GetApplicationOffer(ctx context.Context, user *openfga.Use
 
 // DestroyOffer removes the application offer.
 func (j *JujuManager) DestroyOffer(ctx context.Context, user *openfga.User, offerURL string, force bool) error {
-	err := j.doApplicationOfferAdmin(ctx, user, offerURL, func(offer *dbmodel.ApplicationOffer, api API) error {
-		if err := api.DestroyApplicationOffer(ctx, offerURL, force); err != nil {
+	err := j.doApplicationOfferAdmin(ctx, user, offerURL, func(offer *dbmodel.ApplicationOffer, api API, controllerURL string) error {
+		if err := api.DestroyApplicationOffer(ctx, controllerURL, force); err != nil {
 			return err
 		}
 		return j.deleteApplicationOffer(ctx, offer)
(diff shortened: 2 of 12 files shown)


ale8k:
> DO NOT REVIEW THIS


OK

internal/jimm/juju/controller.go
```diff
@@ -329,6 +329,108 @@ func (j *JujuManager) AddController(ctx context.Context, user *openfga.User, ctl
 		)
 	}
 
+	// Track the backing controller's model as a first-class JAAS model.
+	if err := j.ensureControllerModel(ctx, ctl); err != nil {
```
ale8k, line 333: Arguably if we cannot ensure the controller model we shouldn't be adding it?

internal/dbmodel/model.go
```diff
@@ -77,11 +88,37 @@ type Model struct {
 	MigrationMode MigrationMode `gorm:"default:''"`
 }
 
+const (
+	// ControllerModelOwnerSuffix is the suffix of the synthetic identity that
+	// owns a backing controller's model within JAAS. The owner is
+	// "<controller-name>@controller", which lets us alias the (otherwise
+	// identically named) "controller" model of each backing controller to a
+	// unique "<controller-name>@controller/controller" address.
+	ControllerModelOwnerSuffix = "@controller"
```
ale8k, line 97: I think @controller is a bit light here. Just in the rare off chance someone has a domain "controller", maybe "juju-controller"

internal/jimm/juju/model_poller.go
```diff
@@ -46,6 +46,11 @@ func (j *JujuManager) pollModels(ctx context.Context, models []*dbmodel.Model) e
 			zapctx.Error(ctx, "error getting model info", zap.Error(err))
 		}
 	}
+
+	// Ensure the controller model is tracked, reusing the open connection.
+	if err := j.ensureControllerModelWithAPI(ctx, &ctrl, api); err != nil {
+		zapctx.Error(ctx, "failed to ensure controller model", zap.String("controller", ctrl.Name), zap.Error(err))
+	}
 	return nil
```
ale8k, line 54: This func never actually returns an error, remove it from the signature

internal/jimm/juju/controller.go
```diff
@@ -329,6 +329,108 @@ func (j *JujuManager) AddController(ctx context.Context, user *openfga.User, ctl
...
+
+	existing := dbmodel.Model{UUID: sql.NullString{String: modelUUID, Valid: true}}
+	if err := j.Database.GetModel(ctx, &existing); err == nil {
+		return nil
+	} else if errors.ErrorCode(err) != errors.CodeNotFound {
+		return err
+	}
+
+	// Read the model's owner as the backing controller reports it rather than
+	// assuming a fixed value, so we stay correct if Juju changes it.
```
ale8k, line 379: What does this mean? Why would Juju change the owner of the controller model?

internal/jimm/juju/controller.go
```diff
@@ -329,6 +329,108 @@ func (j *JujuManager) AddController(ctx context.Context, user *openfga.User, ctl
...
+	modelInfo, err := api.ModelInfo(ctx, names.NewModelTag(modelUUID))
+	if err != nil {
+		return fmt.Errorf("failed to get controller model info: %w", err)
+	}
+	backingOwner := modelInfo.Owner
+
+	owner, err := dbmodel.NewIdentity(controllerModelOwner(ctl.Name))
```
ale8k, line 386: This is super critical right as it determines what the model is displayed as. Can we delete the func controller model owner, do it in-line, and have a clear godoc?

internal/jimm/juju/controller.go
```diff
@@ -329,6 +329,108 @@ func (j *JujuManager) AddController(ctx context.Context, user *openfga.User, ctl
...
+	if err := j.Database.GetCloud(ctx, &cloud); err != nil {
+		return fmt.Errorf("failed to get controller cloud: %w", err)
+	}
+	region := cloud.Region(ctl.CloudRegion)
+	if region.ID == 0 {
+		return fmt.Errorf("controller cloud region %q not found", ctl.CloudRegion)
+	}
+
+	// The controller model has no real cloud credential, so attach a placeholder
+	// "empty" credential to avoid a NULL foreign key.
```
ale8k, line 404: This isn't right is it? It does indeed have a credential but:

1. JIMM manages credentials on behalf of controllers for users - so that when a user adds a model, we can push to a controller that has the said cloud they wish to deploy to.
2. In this scenario, it isn't a user managed model so it by-passes our usual flow.

This makes me wonder... Something related/unrelated - If someone somehow had the same credential name as the name of the credential the controller model is using in another model, and then updated said credential for that model, it'd update the controllers model credential - despite you having set an empty credential here.

internal/jimm/juju/jimm.go
```diff
@@ -76,13 +76,19 @@ func (j *JujuManager) ControllerInfo(ctx context.Context, user *openfga.User, na
 }
 
 // ControllerModelCount returns the number of models hosted on the given
-// controller.
+// controller, excluding backing controller models.
```
ale8k, line 79: I'd explain we exclude it explicitly because this is used to determine if a controller is ready to be deleted.

internal/jimm/juju/jimm.go
```diff
@@ -166,12 +172,20 @@ func (j *JujuManager) RemoveController(ctx context.Context, user *openfga.User,
 		if err != nil {
 			return err
 		}
-		if len(models) > 0 && !force {
+		// Exclude controller models from the "still has models" check: they are
+		// always present on a tracked controller and are removed below.
+		userModels := make([]dbmodel.Model, 0, len(models))
+		for _, m := range models {
```
ale8k, line 178: Don't do this here, update our GetmodelsByController query to exclude the controller model by default and possibly add a bool to include it. Just in case we have side effects anywhere else / our future selves. We've never expected the controller model to come through and if we need it, we'll know and can explicitly request it.

internal/jimm/juju/jimm.go
```diff
@@ -76,13 +76,19 @@ func (j *JujuManager) ControllerInfo(ctx context.Context, user *openfga.User, na
...
+// controller, excluding backing controller models.
 func (j *JujuManager) ControllerModelCount(ctx context.Context, ctl dbmodel.Controller) (int, error) {
 	models, err := j.Database.GetModelsByController(ctx, ctl)
 	if err != nil {
 		return 0, err
 	}
-	return len(models), nil
+	count := 0
+	for _, m := range models {
+		if !m.IsControllerModel {
```
ale8k, line 87: On my previous comment, it would remove this duplication here too.

internal/jimm/juju/model_poller.go
```diff
@@ -70,15 +75,21 @@ func (j *JujuManager) PollModels(ctx context.Context) (err error) {
...
-			continue
+	// Iterate every known controller once. Controllers with user models are
+	// polled (and their controller model ensured within the same dial).
+	// Controllers without user models are handled here to ensure their
+	// controller model is tracked too.
+	err = j.Database.ForEachController(ctx, func(c *dbmodel.Controller) error {
+		if models, ok := controllerModels[c.UUID]; ok && len(models) > 0 {
+			_ = j.pollModels(ctx, models)
+		} else {
+			if err := j.ensureControllerModel(ctx, c); err != nil {
```
ale8k, line 86: And then after a while we can delete this because it shouldn't ever happen to any new deployments?

internal/jimm/juju/controller.go
```diff
@@ -329,6 +329,108 @@ func (j *JujuManager) AddController(ctx context.Context, user *openfga.User, ctl
...
+		ControllerID:        ctl.ID,
+		CloudRegionID:       region.ID,
+		CloudCredentialID:   credential.ID,
+		Life:                "alive",
+	}
+	if err := j.Database.AddModel(ctx, &model); err != nil {
+		return fmt.Errorf("failed to add controller model: %w", err)
+	}
+
+	if err := j.OpenFGAClient.AddControllerModel(ctx, ctl.ResourceTag(), model.ResourceTag()); err != nil {
```
ale8k, line 430: Welp, naming isn't great anymore here haha.

internal/dbmodel/model.go
```diff
@@ -77,11 +88,37 @@ type Model struct {
...
+	ControllerModelName = "controller"
+)
+
 // Tag returns a names.Tag for the model.
 func (m Model) Tag() names.Tag {
 	return m.ResourceTag()
 }
 
+// ControllerFacingOwner returns the model owner as known by the backing Juju
+// controller. JAAS can alias a model's owner locally (imported models with a
```
ale8k, line 110: This doesn't read great. 

ControllerFacingOwner returns either:
- The original owner who created the controller model (external or otherwise) in the case of blah blah.
And if it is not present
- The sythetic owner derived from <controller-name>@controller in the case of blah blah.

internal/jimm/juju/applicationoffer.go
```diff
@@ -28,6 +28,35 @@ import (
 	ofganames "github.com/canonical/jimm/v3/internal/openfga/names"
 )
 
+// controllerFacingOfferURL translates a JAAS offer URL into the URL the backing
+// controller knows. If the URL's owner matches a tracked model's JAAS owner but
+// differs from its controller-facing owner (imported or controller models), the
```
ale8k, line 33: So we are just converting it from a URL that may contain our faked user url into the owner that is actually on the backing controller. That's all we mean here right? And we fall back just in case the controller model hasn't been imported?

ale8k (commented):
I don't really like the whole synthetic user approach honestly. And separately I find the godocs on the Controller owner stuff really confusing.


Maybe I need a mini run over the offer stuff, but is it just:
1. If they didn't import model, use what we have.
2. If they did, we'll be displaying it differently under our fake user owner, so convert that back into the original owner so that the offer works properly.

Also my make qa-lxd env isn't working on this PR, it is complaining about model qualifiers

SimoneDutto:
> I don't really like the whole synthetic user approach honestly. And separately I find the godocs on the Controller owner stuff really confusing.
> 
> Maybe I need a mini run over the offer stuff, but is it just:
> 
> 1. If they didn't import model, use what we have.
> 2. If they did, we'll be displaying it differently under our fake user owner, so convert that back into the original owner so that the offer works properly.
> 
> Also my make qa-lxd env isn't working on this PR, it is complaining about model qualifiers

1. import model is not for the controller models, it's just when you want to import a model into jaas. I've fixed 2 things in one pr, one is the controller model and the other is importing models. That might be the source of confusion, sorry about that
2. the synthetic user is just for controller models

make qa-lxd should work, if you are talking model qualifiers i assume you are using juju v4 instead of juju v3, so that may be the issue. jimm v3 doesnt' work with juju v4

internal/jimm/juju/controller.go
SimoneDutto: the credential here could just not exist in jimm, because the controller was not bootstrapped via jaas for example.

internal/jimm/juju/jimm.go
SimoneDutto: well, at the end of the day we always want the controller model to be there aside from this place, no?

internal/jimm/juju/model_poller.go
SimoneDutto: yeah, the poller is here specifically just for ps7

internal/jimm/juju/controller.go
SimoneDutto: yeah, right?

internal/jimm/juju/applicationoffer.go
SimoneDutto: y

internal/jimm/juju/controller.go
ale8k: Discussed on call, from old iteration and needs removing.

internal/dbmodel/model.go
luci1900, line 97: We're certain that won't collide with a domain we might use in auth?

luci1900 (approved):

internal/jimm/juju/applicationoffer.go
```diff
@@ -28,6 +28,35 @@ import (
...
+// owner is rewritten. If the model cannot be resolved the URL is returned
+// unchanged so that offers stored under their original URL still work.
+func (j *JujuManager) controllerFacingOfferURL(ctx context.Context, offerURL string) (string, error) {
+	ourl, err := crossmodel.ParseOfferURL(offerURL)
+	if err != nil {
```
kian99, line 38: Why would parsing an offer URL fail? Even if the original owner was changed we wouldn't generate an invalid offer URL?

internal/jimm/juju/applicationoffer.go
```diff
@@ -416,8 +460,8 @@ func (j *JujuManager) GetApplicationOffer(ctx context.Context, user *openfga.Use
 
 // DestroyOffer removes the application offer.
 func (j *JujuManager) DestroyOffer(ctx context.Context, user *openfga.User, offerURL string, force bool) error {
-	err := j.doApplicationOfferAdmin(ctx, user, offerURL, func(offer *dbmodel.ApplicationOffer, api API) error {
-		if err := api.DestroyApplicationOffer(ctx, offerURL, force); err != nil {
+	err := j.doApplicationOfferAdmin(ctx, user, offerURL, func(offer *dbmodel.ApplicationOffer, api API, controllerURL string) error {
+		if err := api.DestroyApplicationOffer(ctx, controllerURL, force); err != nil {
```
kian99, line 464: Calling this `controllerURL` is confusing, it reads as if it's the controller's API address.

kian99 (commented):
I'll be honest I'm not a big fan of this change. It feels like a lot of non-obvious code to get this to work.
If we think it's critical we make this possible then it's okay to land but I would hope there's a different path we could take. Some thoughts that came to mind while reviewing,
- Use `juju jaas debug-log --target-controller` for checking controller model logs through JIMM.
- Use `juju jaas model-config --target-controller` for backing controller's model config.
- Not sure how to handle offers through JAAS. Maybe we make a change to Juju so that the controller model is different if you configure the `--login-token-refresh-url`.
- Some other wacky ideas I was thinking were that we could have a synthetic `admin/controller` model in JIMM and all the different controllers are apps in that model and then when you deploy into that model to relate apps to controllers, JIMM will do the necessary magic to perform this operation across the backing controllers. This would also be a lot of non-obvious code.
- Would be good to get @manadart to review this.

SimoneDutto:
> * --target-controller

if we lean more into this --target-controller i think we can achieve also the exposing of controller model via another `juju_jaas_controller_model` resource, or something of that sort, but it requires additional machinery and another approach to jaas UX.

SimonRichardson (commented):
I feel like we're papering over the cracks, and that we're not dealing with the concept of controllers well in JIMM. I either think we should expose the controller alias (the name of the controller for which the model resides) and put that front and center for users, or we should hide the controller models behind a new command, purely for JIMM and in turn JAAS.

We already know that offers aren't easy to deal with in JIMM. Offers are laid out as `[<controller name>:][<model qualifier>/]<model name>.<application name>`, though getting the controller name isn't easy, and is re-written to make it work.

The use of model qualifier (model owner is old terminology) as a way to identify a controller seems less than ideal. I personally think it would be better to expose the controller name with the controller suffix `<controller name>/controller`. Though, that requires flushing through the controller name, which seems counter to what JIMMs purpose is.

internal/jimm/juju/applicationoffer.go
```diff
@@ -62,7 +91,9 @@ func (j *JujuManager) Offer(ctx context.Context, user *openfga.User, offer AddAp
 	}
 
 	offerURL := crossmodel.OfferURL{
```
manadart, line 93: If we have 2 controller models with synthetic names:
- controller-a@controller/controller
- controller-b@controller/controller

When we create an offer we throw those away and use the backing name so they both become admin/controller.dashboard.

And there is a global uniqueness constraint on that identifier.

internal/jimm/juju/applicationoffer.go
manadart, line 108: And here, if we offer from A then B, B finds the A record, assumes it's dangling, deletes it and creates for B.

The A offer is still there, but lost to JAAS.

internal/jimm/juju/jimm.go
manadart, line 87: Various special casing is added for controller models here, but is there anything that the previous omission from JAAS prevented from going through to the backing controller directly?

One case: `ListModels` excludes controllers; `ListModelSummaries` includes them. Even if the behaviour is unchanged, there is now an obligation to reconcile behaviour in every case for consistency.

testing/backing_controller_model_test.go
```diff
@@ -0,0 +1,211 @@
...
+func containsModelUUID(summaries []base.UserModelSummary, uuid string) bool {
+	for _, s := range summaries {
+		if s.UUID == uuid {
+			return true
+		}
+	}
+	return false
+}
+
+// TestImportModelWithOwnerCreateOffer imports a model into JAAS under a new
```
manadart, line 115: This is a very awkward contract. A model is displayed under a new JAAS owner, but returns an offer URL with the old backing owner.

manadart (commented):
I don't think this should merge.

Having landed https://github.com/canonical/jimm/pull/2051, which disabled owner switching, we're bringing it back here after a fashion.

We should work out the specific requirements and consider them separately if we can ensure their non-intersection.

Do we need to support `switch` for controller models?
What set of commands do we need to support `-m` for?
Should we bite the bullet and consider a change to the offer URL schema and back-port it to 3.6?

SimoneDutto:
> I don't think this should merge.
> 
> Having landed #2051, which disabled owner switching, we're bringing it back here after a fashion.
> 
> We should work out the specific requirements and consider them separately if we can ensure their non-intersection.
> 
> Do we need to support `switch` for controller models? What set of commands do we need to support `-m` for? Should we bite the bullet and consider a change to the offer URL schema and back-port it to 3.6?

Doing a schema change for offer url would definitely be the clearest path forward for the offers, but it would still not solve the controller model name clash when having multiple backing controller.

I think, since nobody asked for this feature in juju 3, we can find a nice solution for juju 4 only. For example saying: you can consume or create offer for juju 4 controller model.
Potentially we can even leave behind the `admin/controller` hardcoded name for controller models, which would help removing also the aliasing for controller models.

It's generally something we should invest in finding a solution if we want to enable controller models in JAAS for the future.

SimoneDutto:
Closing this one because it served its purpose of showing the controller model limitation in JAAS, and it will not be merged as is.

Issue #2055, opened 2026-07-07
Feature Request: Support reader permissions at the controller level (cascading to all models)


nsakkos:
Currently, in the JAAS authorization model `(openfga/authorisation_model.fga)`, there is no way to blanket assign reader permissions across all models. To grant a user or group read-only access to all models, the permission has to be applied to each model individually, which add administrative burden when new models are added, e.g.

```
juju models --format json \
  | jq -r '.models[]."model-uuid"' \
  | while read uuid; do
      juju add-permission role-model-readers#assignee reader "model-${uuid}"
    done
```

I propose adding a `reader`  relation to the `controller` type and then update the `model` type to inherit it. 

e.g.

```

type controller
   relations
     define controller: [controller]
     define administrator: [user, user:*, group#member, idpgroup#member, role#assignee] or administrator from controller
+    define reader: [user, user:*, group#member, idpgroup#member, role#assignee] or administrator
     define audit_log_viewer: [user, user:*, group#member, idpgroup#member, role#assignee] or administrator
     define can_addmodel: [user, user:*, group#member, idpgroup#member, role#assignee] or administrator

...

type model
   relations
     define controller: [controller]
     define administrator: [user, user:*, group#member, idpgroup#member, role#assignee] or administrator from controller
-    define reader: [user, user:*, group#member, idpgroup#member, role#assignee] or writer
+    define reader: [user, user:*, group#member, idpgroup#member, role#assignee] or writer or reader from controller
     define writer: [user, user:*, group#member, idpgroup#member, role#assignee] or administrator

```

ale8k:
Yep, this makes sense to me. It might be a *little* controversial as we don't usually stray from modelling the Juju RBAC model but I think in this case it makes a lot of sense and given it is only reading, *should* have no side effects.

With this auth model, you'd only be able to set it via the add-relation JIMM specific API and not via Juju (where we create tuples underneath masking the RBAC). 

I'll put up an initial PR and discuss it with the team.

alesstimec:
@nsakkos can you please describe the exact use case for this change? i'm not entirely convinced we need the reader relation.

nsakkos:
The specific use-case is to enable an AI agent to access juju models in a safe way (read-only access).

I think there might be other use cases, e.g. a security team that wants to audit all the models

ale8k:
[...]

PR #2057, opened 2026-07-07
feat: controller reader relation

ale8k:
## Description
In #2055 it was expressed adding reader permission to each model individually is excessive and an operator burden, and the only laternative they have is setting superuser to the user which may not be ideal in the case of simply wishing to read all models on a controller. 

I liked the suggestion proposed in the issue and just did the suggestion.

As I explained in my response on the issue, we often mirror the Juju RBAC model so that it is possible to add the permissions either via our AddRelation API or Juju's built in grants (which we internally convert to ofga calls). Obviously in this case Juju only models `superuser` and `login` and does not have a concept of `reader` for controllers.

This may be a point of contention (as explained in my response to the issue), but as it is a reading permission, it should have no side effects and is simply additive to our auth model, which I believe is grounds to stray away from the Juju RBAC in [this] reading cases.

As it stands, I haven't added anything extra to enable operators to set this permission. I did think about adding an explicit JIMM command, but I don't believe it is necessary given we have our AddRelation API.

Now the following is possible:
`juju add-permission group-<uuid>#member reader controller-<uuid>` 

## Engineering checklist

- [x] Documentation updated
- [x] Covered by unit tests

### DIFF
--- a/docs/howto/manage-permissions.md
+++ b/docs/howto/manage-permissions.md
@@ -57,6 +57,14 @@ juju add-permission group-mygroup#member can-addmodel cloud-mycloud
 
 For any given resource, permissions are currently hierarchical and some permissions are implicit -- e.g., given a cloud associated with a controller and a model associated with the cloud, a controller `administrator` entails cloud `administrator` entails cloud `can_addmodel`.
 
+Granting the `reader` permission on a controller cascades read access to every model on that controller. For example, to give a user, all assignees of a role, or all members of an IdP group read access to all models on controller `mycontroller`:
+
+```text
+juju add-permission user-alice@canonical.com reader controller-mycontroller
+juju add-permission role-myrole#assignee reader controller-mycontroller
+juju add-permission idpgroup-myidpgroup#member reader controller-mycontroller
+```
+
 > See more: {doc}`juju add-permission <../reference/jaas-plugin>`
 
 
--- a/testing/accesscontrol_test.go
+++ b/testing/accesscontrol_test.go
@@ -1491,6 +1491,77 @@ func TestCheckRelationControllerAdministratorFlow(t *testing.T) {
 	c.Assert(results, qt.DeepEquals, expected)
 }
 
+// TestCheckRelationControllerReaderCascadesToModel verifies that granting a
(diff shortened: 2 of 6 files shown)


kian99 (commented):
Ty for this, I'd like to see our docs updated to mention this and a test in `internal/jimm/`. In the docs we should also mention that you can set a user/group/role as a read of `controller-jimm` which will give them read access to all models/offers/etc.

testing/accesscontrol_test.go
```diff
@@ -1491,6 +1491,77 @@ func TestCheckRelationControllerAdministratorFlow(t *testing.T) {
...
+				Object:       userJAASKey,
+				Relation:     "writer",
+				TargetObject: modelJAASKey,
+			},
+			want: false,
+		},
+	}
+
+	for _, tc := range tests {
+		req := apiparams.CheckRelationRequest{Tuple: tc.input}
```
kian99, line 1558: Since the test is inside the integration style `testing` package, I would've expected to see a test exercising actually reading a model, though this technically works too and it's still totally an integration test, the latter would be preferable imo.

docs/howto/manage-permissions.md
```diff
@@ -57,6 +57,14 @@ juju add-permission group-mygroup#member can-addmodel cloud-mycloud
 
 For any given resource, permissions are currently hierarchical and some permissions are implicit -- e.g., given a cloud associated with a controller and a model associated with the cloud, a controller `administrator` entails cloud `administrator` entails cloud `can_addmodel`.
 
+Granting the `reader` permission on a controller cascades read access to every model on that controller. For example, to give a user, all assignees of a role, or all members of an IdP group read access to all models on controller `mycontroller`:
```
kian99, line 60: I'd add another sentence at the end to say something like "To extend this all models across any controller, use the special-case `controller-jimm`".

kian99 (approved):

alesstimec (changes requested):
Please hold off landing this until we understand the exact use case

ale8k:
Rejected as we'll handle when we approach the fine grained permission model.

PR #2058, opened 2026-07-08
feat(jujuapi): derive model placement cap from the reported client version

alesstimec:
## Description

Per the approved JIMM/Juju interoperability spec, model placement is constrained by the client version reported in the X-Juju-ClientVersion websocket header, not by the negotiated facade version: a new model may only be placed on a controller whose major version is <= the client's reported major, and a client that reports no (or an unparseable) version is treated as a Juju 3.6 client (fail closed).

Previously the ModelManager v10 (legacy) CreateModel handler hardcoded a <=3 cap and the v11 handler was unconstrained. Both handlers, and the JIMM facade's AddModelToController, now derive the same cap from the connection's reported client version. The builder machinery is unchanged; only the cap derivation moves.

Note this changes v11 behavior: until a Juju client that reports its version on the main API websocket dial is released, all clients look unversioned and new models are placed only on <=3.x controllers. This is the spec's intended fail-closed default; API-driven model-creation tests therefore require a 3.6 backing controller (as CI already uses).

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

### DIFF
--- a/internal/jujuapi/controllerroot.go
+++ b/internal/jujuapi/controllerroot.go
@@ -7,14 +7,30 @@ import (
 	"sync"
 
 	"github.com/juju/names/v6"
+	"github.com/juju/version/v2"
 	"github.com/rogpeppe/fastuuid"
 	"golang.org/x/oauth2"
 
 	"github.com/canonical/jimm/v3/internal/errors"
+	"github.com/canonical/jimm/v3/internal/jimmhttp"
 	"github.com/canonical/jimm/v3/internal/jujuapi/rpc"
 	"github.com/canonical/jimm/v3/internal/openfga"
 	jimmnames "github.com/canonical/jimm/v3/pkg/names"
 )
+
+// highestControllerVersionForClient returns the model-placement cap for the client on
+// this connection, per the JIMM/Juju interoperability spec: a client may only
+// host new models on controllers whose major version is <= the client's
+// reported major version (the X-Juju-ClientVersion websocket header), and a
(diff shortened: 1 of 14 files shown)


internal/jujuapi/controllerroot.go
```diff
@@ -7,15 +7,31 @@ import (
...
 
+// highestControllerVersionForClient returns the model-placement cap for the client on
+// this connection, per the JIMM/Juju interoperability spec: a client may only
+// host new models on controllers whose major version is <= the client's
+// reported major version (the X-Juju-ClientVersion websocket header), and a
+// client that reports no, or an unparseable, version is treated as a Juju 3.6
+// client (fail closed).
+func highestControllerVersionForClient(ctx context.Context) int {
+	v, err := version.Parse(jimmhttp.ClientVersionFromContext(ctx))
+	if err != nil {
```
SimoneDutto, line 29: maybe we should have 2 paths here:
- unparsable client version, in case some mistakes were made
- empty client version, juju 3

SimoneDutto (commented):
two comments:
- i'd like to see an integration test in the `testing` pkg to make sure we are parsing everything correctly, i think you landed the pr in juju, so you can use "4/edge" and update the juju dep to have it working
- with this change jimm will only work for juju > 4.0.11 at least, should we put it in the doc somewhere?

kian99 (approved):
Looks good, but I agree it would be good to see an integration style test.

SimoneDutto (approved):

PR #2060, opened 2026-07-10
perf: improve ListModelSummaries by querying openfga a single time

SimoneDutto:
## Description

Making ListModelSummaries a little bit faster.
Before: using ForEachModel that was iterating over each model and making a call to openfga to check access O(n)

Now: 3 calls to fetch the access level for a user, this is closer to the implementation for `ListModels`.

It is slightly faster (30-40ms in a local setup), so i would expect something to be gained on ps7, not the 20seconds IS was complaining about.

The real performance problem is that:
- we get the model uuids a user has access too
- we call `AllModels` on the backing controllers with an admin user
- so even if a user has a single model on a controller we are still fetching the whole list

For that we would need to work out a `ListSummaries` with filters on passing the model-uuids in the jwt, i don't know.

I guess there is where the 20seconds are spent, but this is better than nothing.
 
With 100 models.

Traces before:
image

Traces now:
image

# QA

`make qa-lxd`

...create 100 models...

`juju models` should feel faster between the 2 implementation.

### DIFF
--- a/internal/jimm/juju/model.go
+++ b/internal/jimm/juju/model.go
@@ -315,11 +315,52 @@ func (m *modelSummariesMap) addModelSummary(summary base.UserModelSummary) {
 	m.modelSummaries[summary.UUID] = summary
 }
 
+// listUserModelAccess returns a map from model UUID to the user's highest
+// access level ("admin", "write" or "read") for every model the user can
+// read.
+func (j *JujuManager) listUserModelAccess(ctx context.Context, user *openfga.User) (map[string]string, error) {
+	access := make(map[string]string)
+	for _, relation := range []openfga.Relation{
+		ofganames.ReaderRelation,
+		ofganames.WriterRelation,
+		ofganames.AdministratorRelation,
+	} {
+		uuids, err := user.ListModels(ctx, relation)
+		if err != nil {
+			return nil, err
+		}
+		accessStr := permissions.ToModelAccessString(relation)
+		for _, uuid := range uuids {
+			access[uuid] = accessStr
+		}
+	}
+	return access, nil
+}
+
 // ListModelSummaries returns the list of modelsummary the user has access to.
 // It queries the controllers and then merge the info from the JIMM db.
 func (j *JujuManager) ListModelSummaries(ctx context.Context, user *openfga.User, maskingControllerUUID string) ([]base.UserModelSummary, error) {
 	modelSummariesSafeMap := modelSummariesMap{}
 	modelSummaryResults := []base.UserModelSummary{}
+
+	modelAccess, err := j.listUserModelAccess(ctx, user)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(modelAccess) == 0 {
+		return modelSummaryResults, nil
+	}
+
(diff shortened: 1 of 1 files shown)


luci1900 (approved):
Looks like O(n**2) to O(n), nice.

internal/jimm/juju/model.go
```diff
@@ -315,34 +315,71 @@ func (m *modelSummariesMap) addModelSummary(summary base.UserModelSummary) {
 	m.modelSummaries[summary.UUID] = summary
 }
 
+// listUserModelAccess returns a map from model UUID to the user's highest
+// access level ("admin", "write" or "read") for every model the user can
+// read.
+func (j *JujuManager) listUserModelAccess(ctx context.Context, user *openfga.User) (map[string]string, error) {
```
ale8k: LGTM, but this shouldn't live here, it should be in `internal/jimm/permissions/access.go` and then you can call the permission manager in list model summaries

ale8k (commented):
LGTM just move listUserModelAccess please

internal/jimm/juju/model.go
SimoneDutto: layers layers mocks layers mocks

SimoneDutto: image

ale8k: image

SimoneDutto: done my guy

ale8k (approved):
Thanks! Now can you just create another service layer method in Juju Manager to expose this functionality to the entire manager, with a godoc that ... Just kidding haha ty! SHIP IT

internal/jimm/juju/model.go
ale8k: Let's not resolve this comment, it deserves to be seen.

kian99: As a counterpoint, `auth` is a pretty cross cutting concern and the only places we use the permission manager in the Juju manager is for 2 methods `
Those 2 methods don't actually do much that the `openfga.User` struct can't do. They literally do,
```go
// GetUserCloudAccess returns users access level for the specified cloud.
func (j *PermissionManager) GetUserCloudAccess(ctx context.Context, user *openfga.User, cloud names.CloudTag) (string, error) {
	accessLevel := user.GetCloudAccess(ctx, cloud)
	return ToCloudAccessString(accessLevel), nil
}

// GetUserModelAccess returns the access level a user has against a specific model.
func (j *PermissionManager) GetUserModelAccess(ctx context.Context, user *openfga.User, model names.ModelTag) (string, error) {
	accessLevel := user.GetModelAccess(ctx, model)
	return ToModelAccessString(accessLevel), nil
}
```
I'm a fan of the idea of avoiding putting managers in other managers.

SimoneDutto: after serious consideration offline, i'm keeping the func in the model manager

kian99 (approved):

kian99 (commented):
One thing worth pointing out.
Because we don't have a way to mock the OpenFGA client and/or DB in the Juju manager, we can't write a test that verifies we only make 1 call to ofga and 1 call to the DB now.

PR #2061, opened 2026-07-10
ci: add tf e2e tests

ale8k:
## Description
Adds another set of e2e for JIMM v4 that:

1. Spin up a 3 controller via 3 CLI
2. Spin up JIMM v4
3. Pulls TF v1.5 (3 compatible TF)
4. Runs TF v1.5 tests in an effort to verify compatibility with a **3**(tf)**4**(jimm)**3**(juju) setup.

I kept the naming consistent but prefixed it with TF, so it is understood the default E2E test is JIMM.

I also moved all the setup fluff into a composite action for them to share and get their go versions independently from their own repo mods.


.github/workflows/tf-e2e-test.yaml
```diff
@@ -34,12 +34,20 @@ jobs:
...
         uses: actions/setup-go@v6
         with:
-          go-version-file: terraform-provider-juju/go.mod
+          go-version-file: jimm/go.mod
           cache: false
 
+      # test-server's `make dev-env` uses `docker compose --profile`, which
+      # requires the Compose v2 plugin. Install it explicitly.
+      - name: Set up Docker Compose
+        uses: docker/setup-compose-action@v1
```
kian99: Latest for this action is v2 btw

ale8k: I just realised, gonna update and add caching in.

.github/workflows/tf-e2e-test.yaml
```diff
@@ -0,0 +1,97 @@
...
+    timeout-minutes: 67
+    runs-on: [self-hosted-linux-amd64-jammy-large]
+    strategy:
+      matrix:
+        include:
+          # JIMM v4 (built from the current checkout via the "dev" version) ->
+          # backing controller v3, exercised by terraform-provider-juju v1.5.
+          - name: "JIMM v4, backing v3, TF v1.5"
+            juju-channel: "3/stable"
+            tf-ref: "v1.5"
```
kian99, line 24: I don't think we need a matrix here since we're only running 1 set of variables.

.github/workflows/tf-e2e-test.yaml
```diff
@@ -0,0 +1,97 @@
...
+    steps:
+      - name: Checkout JIMM repo
+        uses: actions/checkout@v4
+        with:
+          path: jimm
```
kian99, line 29: Do we need the `with` part?

kian99 (commented):
Looks solid to me

.github/workflows/tf-e2e-test.yaml
ale8k: I like it because if we ever debug this, the two repos are adjacent rather than tf pulled into JIMM root

.github/workflows/tf-e2e-test.yaml
ale8k: Added it ahead of time, just in case we ever fancy running 2.0 tf vs jimm 4 and N juju. Doesn't do any harm here running a single set either :)

kian99: *Screams in premature optimisation* Fair enough.

.github/workflows/tf-e2e-test.yaml
kian99: What I mean is, the repo is already called `jimm` so I thought it would already be called `jimm` and you'd only specify this if you wanted to change the folder name.

.github/workflows/tf-e2e-test.yaml
```diff
@@ -0,0 +1,108 @@
...
+          echo "JUJU_CLIENT_ID=${{ steps.jimm.outputs.client-id }}" >> "$GITHUB_ENV"
+          echo "JUJU_CLIENT_SECRET=${{ steps.jimm.outputs.client-secret }}" >> "$GITHUB_ENV"
+          echo "JUJU_CA_CERT<<EOF" >> "$GITHUB_ENV"
+          echo "${{ steps.jimm.outputs.ca-cert }}" >> "$GITHUB_ENV"
+          echo "EOF" >> "$GITHUB_ENV"
+          echo "JUJU_SKIP_FAILED_DELETION=true" >> "$GITHUB_ENV"
+
+      - name: Run terraform provider acceptance tests
+        working-directory: terraform-provider-juju
+        run: go test ./internal/provider/... -parallel 10 -v -timeout 40m
```
kian99: Something worth changing here. This is currently also running the provider's own internal tests when we do this rather than only runnning the integration tests in `internal/provider/`. The parallelism is also quite high. We previously had troubles in the TF repo increasing the value above 4 iirc.

.github/workflows/tf-e2e-test.yaml
```diff
@@ -0,0 +1,108 @@
...
+      # raw binary is on PATH for the test framework to discover.
+      - name: Install Terraform
+        uses: hashicorp/setup-terraform@v3
+        with:
+          terraform_version: "1.14.*"
+          terraform_wrapper: false
+
+      # Derive the provider configuration from the running JIMM controller.
+      # The Juju API endpoint must come from the CLI (not the traefik
+      # https://jimm.localhost:443 web endpoint) or the client gets a 404.
```
kian99: What does this mean that the endpoint must come from the CLI? That doesn't sound right. The endpoint without the scheme should work i.e. `jimm.localhost:443`.

.github/workflows/tf-e2e-test.yaml
```diff
@@ -0,0 +1,108 @@
...
+jobs:
+  run-tf-e2e-test:
+    name: Run (${{ matrix.name }})
+    timeout-minutes: 67
+    runs-on: [self-hosted-linux-amd64-jammy-large]
+    strategy:
+      matrix:
+        include:
+          # JIMM v4 (built from the current checkout via the "dev" version) ->
+          # backing controller v3, exercised by terraform-provider-juju v1.5.
```
kian99: If you can make these more full sentences I think it will help reading this in a few months time, like,

```suggestion
          # Use the latest changes in JIMM by specifying the special string "dev"
          # in the `test-server` action below, use a Juju 3 backing controller, and use
          # the terraform-provider-juju v1.5 as a client.
```

kian99 (approved):

.github/workflows/tf-e2e-test.yaml
```diff
@@ -0,0 +1,108 @@
...
+          path: terraform-provider-juju
+
+      # Use JIMM's Go version: building JIMM (make dev-env) requires the higher
+      # version declared in jimm/go.mod. A newer Go still builds and runs the
+      # terraform-provider-juju tests, so a single toolchain covers both repos.
+      - name: Setup Go
+        uses: actions/setup-go@v6
+        with:
+          go-version-file: jimm/go.mod
+          cache: false
```
SimoneDutto: why cache false? is this setting the read-only cache?

.github/workflows/tf-e2e-test.yaml
SimoneDutto: yeah, i don't think 10 will ever work reliably, with 4 we have a lot of flakyness already

SimoneDutto (approved):
seems good!

.github/workflows/tf-e2e-test.yaml
ale8k: My understanding of this (from the [action gh page](https://github.com/actions/checkout/tree/v4/)):
```
    # Relative path under $GITHUB_WORKSPACE to place the repository
    path: ''
```

Is it'll just check it out into root of workspace? Haven't manually verified but IIRC always noticed it dumps it into the root.

.github/workflows/tf-e2e-test.yaml
ale8k: My wrong presumption it doesn't handle multi-repo roots because it uses the go.mod as part of the key, just checked and they have a doc [here](https://github.com/actions/setup-go/blob/main/docs/advanced-usage.md#caching)  which can glob over sums. Will update it to the multg repo root caching strategy as they suggest.

.github/workflows/tf-e2e-test.yaml
ale8k: Mhm I can try again, but that was the source of my 404s. Updated comment a bit more to be clearer

ale8k:
Merging as is. Currently add group is failing and a follow up will need to handle this. Either deleting the test, injecting skips, etc etc.

PR #2064, opened 2026-07-16
fix: recover credential for a model

SimoneDutto:
## Description

This is a new command to fix the PS7 problem, so the credential is still in jimm's db but the actual secret content is not added to vault.

This pr adds a commnad to:
- fetch the credential for jimm's db
- get the model
- get the credential
- set the content in vault again

# QA

It works against LXD as well, and you can see the content is restored.

Run this, the output should be success, and vault should be populated with: `[client-cert,client-key,server-cert]`

```
#!/bin/bash

set -euo pipefail
source ./local/jimm/detect-jaas.sh

# Force the local dev Vault creds (a stale VAULT_TOKEN in the shell causes 403s).
export VAULT_ADDR="http://localhost:8200"
export VAULT_TOKEN="root"

OWNER="$(juju whoami --format json | jq -r .user)"
CLOUD="${CLOUD:-localhost}"                                # k8s cloud on the JIMM controller
CRED_NAME="${CRED_NAME:-localhost}"                        # cloud credential name
CRED_TAG="${CLOUD}/${OWNER}/${CRED_NAME}"                 # cloud/owner/name
VAULT_PATH="jimm-kv/creds/${CLOUD}/${OWNER}/${CRED_NAME}" # JIMM's vault layout

# Reads the credential secret content (the attribute map) from Vault as JSON.
secret_content() { vault kv get -format=json "${VAULT_PATH}" | jq -S .data.data; }

echo "0. Capture the secret content currently in Vault (before the outage)"
BEFORE="$(secret_content)"
echo "${BEFORE}"

echo "1. Delete the secret from Vault (simulate the outage)"
vault kv metadata delete "${VAULT_PATH}"

echo "2. Prove the content is gone: reading it from Vault must FAIL"
if secret_content >/dev/null 2>&1; then
    echo "   UNEXPECTED: secret still present in Vault after deletion" >&2
    exit 1
fi
echo "   secret is gone from Vault, as expected"

echo "3. Recover it from the controller via JIMM"
${JAAS} recover-model-credential "${CRED_TAG}"

echo "4. Confirm the secret content is restored AND identical to before"
AFTER="$(secret_content)"
echo "${AFTER}"
if [ "${BEFORE}" != "${AFTER}" ]; then
    echo "   ERROR: recovered secret content differs from the original" >&2
    diff <(echo "${BEFORE}") <(echo "${AFTER}") || true
    exit 1
fi
echo "   recovered content matches the original"

echo "SUCCESS"
```


internal/db/cloudcredential.go
```diff
@@ -106,6 +106,31 @@ func (d *Database) ForEachCloudCredential(ctx context.Context, identityName, clo
 	return nil
 }
 
+// GetAllCloudCredentials returns every cloud credential known to JIMM,
+// regardless of owner or cloud. This is primarily intended for
+// administrative/disaster-recovery operations.
+func (d *Database) GetAllCloudCredentials(ctx context.Context) (_ []dbmodel.CloudCredential, err error) {
```
alesstimec, line 112: this method is missing tests

ale8k (approved):

internal/jimm/juju/cloudcredential.go
```diff
@@ -384,3 +384,168 @@ func (j *JujuManager) CopyCredential(ctx context.Context, originalUser *openfga.
...
+	if err := j.Database.GetCloudCredential(ctx, &credential); err != nil {
+		return err
+	}
+
+	return j.recoverCredential(ctx, &credential, dryRun)
+}
+
+// RecoverModelCredentialResult holds the outcome of recovering a single cloud
+// credential.
+type RecoverModelCredentialResult struct {
```
alesstimec, line 417: is this the place to define this type?

internal/jimm/juju/cloudcredential.go
```diff
@@ -384,3 +384,168 @@ func (j *JujuManager) CopyCredential(ctx context.Context, originalUser *openfga.
...
+	tag := credential.ResourceTag()
+
+	// Find the models (and therefore controllers) that use this credential so
+	// we know where to fetch the secrets from.
+	models, err := j.Database.GetModelsUsingCredential(ctx, credential.ID)
+	if err != nil {
+		return err
+	}
+	if len(models) == 0 {
+		return errors.Codef(errors.CodeNotFound, "no models use credential %q, cannot recover secrets", tag.String())
```
alesstimec:
```suggestion
		return errors.Codef(errors.CodeNotFound, "no models use credential %q, cannot recover secret attributes", tag.String())
```

internal/db/cloudcredential.go
SimoneDutto: there are no tests anywhere, right?

kian99 (approved):

luci1900 (approved):

PR #2065, opened 2026-07-16
feat(jujuapi): gate model proxy on client/model version compatibility

alesstimec:
## Description

Per the approved JIMM/Juju interoperability spec, apply the compatibility rule at model-connection setup, before proxying: a client may interact with models hosted on controllers of major version <= its reported major version (X-Juju-ClientVersion websocket header), and a client that reports no (or an unparseable) version is treated as a Juju 3.6 client. An unversioned or 3.6 client is therefore refused a connection to any 4.x-hosted model with a clear error naming the model and its version, instead of failing deep into the session with obscure facade errors. Connections to 3.6-hosted models are unrestricted; discovery (listing, switching) is deliberately not filtered.

A controller whose agent version is unknown cannot be established as compatible and is likewise rejected (fail closed), consistent with version-constrained model placement.

Deployment notes: any L7 ingress in front of JIMM must forward the X-Juju-ClientVersion header on the websocket upgrade, or every client will look unversioned and be refused 4.x models. Do not deploy this to production before a Juju 4.x client that reports its version on the main API dial (juju/juju#22794) is released; the header is client-asserted and gates UX/routing, not authorization.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

## Test instructions

- sudo snap refresh juju --channel 4.0/edge
- make qa-lxd
- juju deploy juju-qa-dummy-source
- juju config dummy-source token="hello world"
- juju offer dummy-source:sink
- sudo snap refresh juju --channel 3.6/stable
- CONTROLLER_NAME=qa-lxd-36 ./local/jimm/setup-controller.sh 
- CONTROLLER_NAME=qa-lxd-36 ./local/jimm/add-controller.sh 
- juju add-model test
- juju deploy juju-qa-dummy-sink
- juju consume jimm-test@canonical.com/test-lxd.dummy-source
- juju relate dummy-sink dummy-source
- verify that juju status output shows `Token is hello world` under `Message` for dummy-sink
- juju switch test-lxd 
- juju status -> expect an error message because you are using a 3.6 client to talk to 4.0 model
- sudo snap refresh juju --channel 4.0/edge
- verify that you can run juju status on both models


internal/jujuapi/websocket.go
```diff
@@ -209,6 +210,34 @@ func (s apiModelProxier) ServeWS(ctx context.Context, clientConn *websocket.Conn
 	}
 }
 
+// checkClientModelCompatibility rejects a model connection when the Juju
+// client cannot operate the model's hosting controller, per the JIMM/Juju
+// interoperability spec: a client may interact with models hosted on
+// controllers of major version <= the client's reported major version, and a
+// client that reports no (or an unparseable) version is treated as a Juju 3.6
```
luci1900: So this means the Juju 4.0 CLI will be treated as a 3.6 client, right?

Do we want to patch that?

alesstimec: this was merged in juju 4: https://github.com/juju/juju/pull/22794 so juju 4 client are reporting the version now.

luci1900 (approved):

internal/jujuapi/websocket.go
```diff
@@ -196,19 +197,78 @@ func (s apiModelProxier) ServeWS(ctx context.Context, clientConn *websocket.Conn
...
-		AuthenticatedIdentityID: auth.SessionIdentityFromContext(ctx),
-		RedirectInfo:            redirectInfo,
+		ConnClient:               clientConn,
+		TokenGen:                 &jwtGenerator,
+		ConnectController:        connectionFunc,
+		AuditLog:                 auditLogger,
+		LoginService:             s.jimm.LoginManager,
+		AuthenticatedIdentityID:  auth.SessionIdentityFromContext(ctx),
+		RedirectInfo:             redirectInfo,
+		ClientCompatibilityCheck: clientCompatibilityCheckFunc(s),
```
ale8k: We aren't mocking this and it likely isn't going to change right? So we could just have it on the proxyhelper struct

internal/jujuapi/websocket.go
```diff
@@ -196,19 +197,78 @@ func (s apiModelProxier) ServeWS(ctx context.Context, clientConn *websocket.Conn
...
 		zapctx.Error(ctx, "failed to start jimm model proxy", zap.Error(err))
 	}
 }
 
+// clientCompatibilityCheckFunc returns the check the model proxy runs when a
+// user logs in over a proxied model connection. It resolves the model (and
+// with it the hosting controller) from the request path and applies
+// checkClientModelCompatibility. It deliberately runs at user login, not at
+// connection setup: the same websocket endpoint also serves the model's own
+// agents (which report no client version), and gating the connection itself
```
ale8k: The comments on these funcs are pretty hard to read and we've split them into two funcs with huge comments. 

All it really is saying is:
<func name> verifies a client call to a model behind JIMM is compatible. Compatibility is decided based on whether the client version is NEWER than the controller version.

Can we just make it one func, and put a doc similar to the above on? 

Regarding the call location of the func and that it needs to be run after login, just comment above where it is run IMO. Doesn't help me understand this func any further having it here.

internal/jujuapi/websocket.go
```diff
@@ -196,19 +197,78 @@ func (s apiModelProxier) ServeWS(ctx context.Context, clientConn *websocket.Conn
...
+// checkClientModelCompatibility rejects a user's model connection when their
+// Juju client cannot operate the model's hosting controller, per the
+// JIMM/Juju interoperability spec: a client may interact with models hosted
+// on controllers of major version <= the client's reported major version,
+// and a client that reports no (or an unparseable) version is treated as a
+// Juju 3.6 client (fail closed). A controller whose agent version is unknown
+// cannot be established as compatible and is likewise rejected. The rule
+// applies to user logins only — the model's agents are exempt.
+func checkClientModelCompatibility(ctx context.Context, m *dbmodel.Model) error {
+	controllerVersion, err := version.Parse(m.Controller.AgentVersion)
```
ale8k: I think this is worth adding a metric for. It'll help us understand if users use the wrong client often enough to see if it is a problem.

internal/rpcproxy/rpcproxy.go
```diff
@@ -690,6 +698,18 @@ func (p *clientProxy) handleAdminFacade(ctx context.Context, msg *message) (clie
 		return nil, nil, err
 	}
 	controllerLoginMessageFnc := func(user *openfga.User) (*message, *message, error) {
+		// Only authenticated user logins funnel through here, so this is
```
ale8k: This isn't really clear to me, what is it trying to say? We keep talking about model agent logins etc but we're not explaining why? If anything I think this comment has made me more confused than the code itself

testing/modelproxy_test.go
```diff
@@ -0,0 +1,203 @@
...
+	Params    json.RawMessage `json:"params,omitempty"`
+	Error     string          `json:"error,omitempty"`
+	ErrorCode string          `json:"error-code,omitempty"`
+	ErrorInfo map[string]any  `json:"error-info,omitempty"`
+	Response  json.RawMessage `json:"response,omitempty"`
+}
+
+// legacyAgentLogin dials the model's /api endpoint through JIMM's model
+// proxy exactly as a Juju agent does — no X-Juju-ClientVersion header — and
+// performs a legacy Admin.Login with the given agent tag, returning the
```
ale8k, line 172: But whyyyyyy do agents try login through JIMM, whyyyy are we doing this? 😆

ale8k (commented):
I think I've forgotten agents login through JIMM and don't know why anymore. If you could clear that up for me I think I can re-review then and understand this PR better. 

It makes sense to me from the juju CLIc version trying to login and being checked but the rest of it in the comments isn't clearing it up for me.

testing/modelproxy_test.go
kian99: I can answer this one. It's because JIMM acts as a controller, so when you migrate a model to JIMM via `juju migrate`, controller A speak to JIMM as if it was another controller and all the agents running on machines do the same. The agents eventually get redirected to a valid backing controller.

internal/jujuapi/websocket.go
alesstimec: it's easier to unit test checkClientModelCompatibility in isolation..

internal/jujuapi/websocket.go
alesstimec: we can add it in a followup

testing/modelproxy_test.go
alesstimec: correct

go.mod
```diff
@@ -1,13 +1,13 @@
...
-go 1.26.4
+go 1.26.5
 
 // Juju based dependencies
 require (
 	github.com/juju/description/v9 v9.0.0
 	github.com/juju/errors v1.0.0
 	github.com/juju/gnuflag v1.0.0
-	github.com/juju/juju v0.0.0-20260616131348-a11d3fb1da2a
+	github.com/juju/juju v0.0.0-20260721130522-0573e8bfc882
```
SimoneDutto: latest juju? what's this?

internal/jujuapi/websocket.go
```diff
@@ -196,19 +197,78 @@ func (s apiModelProxier) ServeWS(ctx context.Context, clientConn *websocket.Conn
...
+		if err := s.jimm.Database.GetModel(ctx, &m); err != nil {
+			return errors.Codef(errors.CodeNotFound, "failed to find model: %w", err)
+		}
+		return checkClientModelCompatibility(ctx, &m)
+	}
+}
+
+// checkClientModelCompatibility rejects a user's model connection when their
+// Juju client cannot operate the model's hosting controller, per the
+// JIMM/Juju interoperability spec: a client may interact with models hosted
```
SimoneDutto: i don't think we need to say the spec in the godoc

internal/rpcproxy/rpcproxy.go
```diff
@@ -695,6 +703,13 @@ func (p *clientProxy) handleAdminFacade(ctx context.Context, msg *message) (clie
 		return nil, nil, err
 	}
 	controllerLoginMessageFnc := func(user *openfga.User) (*message, *message, error) {
+		// User logins funnel through here, so this is
+		// where client/model compatibility is enforced.
+		if p.clientCompatibilityCheck != nil {
+			if err := p.clientCompatibilityCheck(ctx); err != nil {
+				return errorFnc(err)
+			}
+		}
```
kian99: We don't often write code like this elsewhere, where we pass in an anonymous function just for the sake of testing. Usually we'd make an interface for that or just call the logic directly and the test becomes a bit more integration-style.

I was playing around with this code locally. Wdyt about adjusting the `controllerConnectionFunc` to insert into the context a `modelVersion` field that holds the model's version then here you can grab the model's version and the client's version and do the comparison check. And then avoid adding another anonymous function and just test it all via the tests in internal/rpcproxy/rpcproxylogin_test.go or internal/rpcproxy/rpcproxy_test.go.

testing/modelproxy_test.go
```diff
@@ -0,0 +1,203 @@
...
+// response frame.
+func legacyAgentLogin(c *qt.C, s jimmtest.JimmWithControllers, mt names.ModelTag, authTag string) rpcMessage {
```
kian99, line 174: I can see why this is useful, but the fact that we have to copy/paste this `rpcMessage` and drop down to such a low level is a huge smell imo. I'd prefer if we didn't add this.

alesstimec: you'd remove the whole test?

kian99: Looking at it again, I don't still feel that strongly about it and it's just for tests so seems okay.

internal/jujuapi/websocket.go
```diff
@@ -191,7 +191,25 @@ func (s apiModelProxier) ServeWS(ctx context.Context, clientConn *websocket.Conn
...
+
+	// Resolve the model (and with it the hosting controller) from the
+	// request path once, at connection setup. The result feeds two
+	// consumers: the client/model compatibility inputs stored in the
+	// context, which the proxy applies at user login (agent and anonymous
+	// logins are never gated — a model's agents may be older than a hosting
+	// controller they migrated to and must still reach it), and the
+	// controller dial in the connection func. A lookup failure is not
+	// handled here: the connection func returns it so that the error
+	// reaches the client as the response to its first message.
```
kian99: Can you reword the comments across this PR to be less verbose please.

kian99 (approved):
Looks good, just one request for simplification of comments.

internal/jujuapi/websocket.go
```diff
@@ -191,7 +191,19 @@ func (s apiModelProxier) ServeWS(ctx context.Context, clientConn *websocket.Conn
 	ctx = jujuTrace.InjectTracerIfRequired(ctx, s.params.Tracer)
 	redirectInfo := redirectInfoAdapter{jimm: s.jimm}
 	jwtGenerator := s.jimm.NewJujuAuthenticator()
-	connectionFunc := controllerConnectionFunc(s, &jwtGenerator)
+
+	model, finalPath, lookupErr := s.modelFromPath(ctx)
+	if lookupErr == nil {
```
SimoneDutto, line 196: in case lookupErr is not nil should we at least print a warning, this is the model proxier after all, it should always have that populated no?

SimoneDutto: ah, i see we do that in the controllerCOnnectionFunc

testing/modelproxy_test.go
```diff
@@ -0,0 +1,193 @@
...
+
+// TestModelProxyClientCompatibility exercises the model proxy's
+// client/model compatibility gate end to end: a client may interact with
+// models hosted on controllers of major version <= its reported
+// X-Juju-ClientVersion, an unversioned client is treated as Juju 3.6 (fail
+// closed), and discovery is deliberately not filtered. A model is pinned to
+// each backing controller and expectations are derived from that
+// controller's agent version, so the test works with 3.x controllers, 4.x
+// controllers, or a mix.
+func TestModelProxyClientCompatibility(t *testing.T) {
```
SimoneDutto: i get the test is trying to be smart such as it will always work if run against jimm with 3  and 4 backing controllers, but i would prefer an explicit easier to read test only running when the backing controller is 3 or 4. Like this is hard to read/reason about.

What do you think?

SimoneDutto (approved):
i just have a suggestion for test readability, but looks good.

go.mod
alesstimec: 4.0.13

internal/jujuapi/websocket.go
alesstimec: removed

internal/jujuapi/websocket.go
alesstimec: done

testing/modelproxy_test.go
alesstimec: done

PR #2066, opened 2026-07-17
feat: restricted permissions when dialling controller on behalf of user

luci1900:
## Description

Currently, when JIMM dials a backing Juju controller on behalf of a user it always mints a `superuser` JWT, relying on an OpenFGA permission check having been performed in a higher layer.

This PR changes all dials made on behalf of a user to only include JWT claims for precisely the permissions of the calling user.

JIMM still does calls against controllers with `superuser` when it's not representing a user or when the semantics it overlays though proxying don't allow it.

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests


luci1900:
Closing for now, potentially to be looked at again later.

PR #2068, opened 2026-07-23
fix: fix UpdateCredentials to check against the args.Credentials

SimoneDutto:
## Description

In this pr I've addressed a bug where we were checking credentials using the current attributes and not the new attributes.
While doing this i've noticed we were using this method with a closure, and it is just unclear to read.
I've created a specialized method, which fetch the attrs if they are not provided.

All calls site now have:
- the correct credential + attr in jimm's state, so we don't need to provide them from the args
- the attrs arg specified

### DIFF
--- a/internal/jimm/juju/cloudcredential.go
+++ b/internal/jimm/juju/cloudcredential.go
@@ -176,7 +176,7 @@ func (j *JujuManager) UpdateCloudCredential(ctx context.Context, user *openfga.U
 
 	if !args.SkipCheck {
 		err := j.forEachController(ctx, controllers, func(ctl *dbmodel.Controller, api API) error {
-			models, err := j.updateControllerCloudCredential(ctx, &credential, api.CheckCredentialModels)
+			models, err := j.checkControllerCloudCredential(ctx, &credential, args.Credential.Attributes, api)
 			resultMu.Lock()
 			defer resultMu.Unlock()
 			result = append(result, models...)
@@ -204,7 +204,7 @@ func (j *JujuManager) UpdateCloudCredential(ctx context.Context, user *openfga.U
 	}
 
 	err = j.forEachController(ctx, controllers, func(ctl *dbmodel.Controller, api API) error {
-		models, err := j.updateControllerCloudCredential(ctx, &credential, api.UpdateCloudsCredentialForce)
+		models, err := j.forceUpdateControllerCloudCredential(ctx, &credential, args.Credential.Attributes, api)
 		if err != nil {
 			return err
 		}
@@ -234,29 +234,25 @@ func (j *JujuManager) updateCredential(ctx context.Context, credential *dbmodel.
 	return nil
 }
 
-func (j *JujuManager) updateControllerCloudCredential(
+// checkControllerCloudCredential validates that the given credential would be
+// accepted by all models currently using it on the controller reachable via
+// api. The attributes passed are the NEW ones being validated.
+func (j *JujuManager) checkControllerCloudCredential(
 	ctx context.Context,
 	cred *dbmodel.CloudCredential,
-	f func(context.Context, jujuparams.TaggedCredential) ([]jujuparams.UpdateCredentialResult, error),
+	attrs map[string]string,
+	api API,
 ) ([]jujuparams.UpdateCredentialModelResult, error) {
-
-	attr, err := j.getCloudCredentialAttributes(ctx, cred)
-	if err != nil {
-		return nil, err
-	}
-
-	out, err := f(ctx, jujuparams.TaggedCredential{
+	out, err := api.CheckCredentialModels(ctx, jujuparams.TaggedCredential{
 		Tag: cred.Tag().String(),
 		Credential: jujuparams.CloudCredential{
 			AuthType:   cred.AuthType,
-			Attributes: attr,
+			Attributes: attrs,
 		},
 	})
-
 	if err != nil {
 		return nil, err
 	}
-
 	// Shouldn't happen, the Juju client presumes that
 	// the returned slice will always contain a result
 	// for each credential passed in, but handle it just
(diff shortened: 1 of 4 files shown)


internal/jimm/juju/cloudcredential.go
```diff
@@ -176,7 +176,7 @@ func (j *JujuManager) UpdateCloudCredential(ctx context.Context, user *openfga.U
 
 	if !args.SkipCheck {
 		err := j.forEachController(ctx, controllers, func(ctl *dbmodel.Controller, api API) error {
-			models, err := j.updateControllerCloudCredential(ctx, &credential, api.CheckCredentialModels)
+			models, err := j.checkControllerCloudCredential(ctx, &credential, args.Credential.Attributes, api)
```
SimoneDutto, line 179: here was the problem, because we were doing a check with the credentials stored in jimm, not the one we were receiving from the args

internal/jimm/juju/cloudcredential.go
```diff
@@ -204,7 +204,7 @@ func (j *JujuManager) UpdateCloudCredential(ctx context.Context, user *openfga.U
 	}
 
 	err = j.forEachController(ctx, controllers, func(ctl *dbmodel.Controller, api API) error {
-		models, err := j.updateControllerCloudCredential(ctx, &credential, api.UpdateCloudsCredentialForce)
+		models, err := j.forceUpdateControllerCloudCredential(ctx, &credential, args.Credential.Attributes, api)
```
SimoneDutto, line 207: here potentially we could leave `nil` because we are updating the credential in the store a few lines higher

internal/jimm/juju/cloudcredential.go
```diff
@@ -234,41 +234,74 @@ func (j *JujuManager) updateCredential(ctx context.Context, credential *dbmodel.
...
+// forceUpdateControllerCloudCredential force-updates the credential on the
+// controller reachable via api. If attrs is nil the attributes are fetched
+// from the credential store; otherwise the supplied attrs are applied.
+func (j *JujuManager) forceUpdateControllerCloudCredential(
+	ctx context.Context,
+	cred *dbmodel.CloudCredential,
+	attrs map[string]string,
+	api API,
+) ([]jujuparams.UpdateCredentialModelResult, error) {
+	if attrs == nil {
```
kian99: I'd prefer to see one approach or another, especially since this is internal code.

internal/jimm/juju/cloudcredential.go
```diff
@@ -234,41 +234,74 @@ func (j *JujuManager) updateCredential(ctx context.Context, credential *dbmodel.
...
+		},
+	})
 	if err != nil {
 		return nil, err
 	}
+	// Shouldn't happen, the Juju client presumes that
+	// the returned slice will always contain a result
+	// for each credential passed in, but handle it just
+	// in case.
+	if len(out) == 0 {
```
luci1900, line 256: Shouldn't we still return some kind of error? A nil check succeeding would lead to a nil panic most likely.

internal/jimm/juju/cloudcredential_test.go
```diff
@@ -1866,3 +1866,91 @@ func TestCopyCredentialWithMissingCredential(t *testing.T) {
...
+	_, err := j.UpdateCloudCredential(ctx, user, juju.UpdateCloudCredentialArgs{
+		CredentialTag: credTag,
+		Credential: jujuparams.CloudCredential{
+			AuthType:   "test-auth-type",
+			Attributes: newAttributes,
+		},
+	})
+	c.Assert(err, qt.IsNil)
+
+	// The check must have validated the NEW auth-type and attributes,
```
ale8k, line 1952: OK 😆

luci1900 (approved):
Nice.

Agreed that splitting into two methods is much clearer and would've likely avoided the bug in the first place.

internal/jimm/juju/cloudcredential.go
```diff
@@ -234,41 +234,74 @@ func (j *JujuManager) updateCredential(ctx context.Context, credential *dbmodel.
...
+			AuthType:   cred.AuthType,
+			Attributes: attrs,
+		},
+	})
 	if err != nil {
 		return nil, err
 	}
+	// Shouldn't happen, the Juju client presumes that
+	// the returned slice will always contain a result
+	// for each credential passed in, but handle it just
```
ale8k: Can delete this comment, they do this in a bunch a places and we're used to it

kian99 (commented):
Looks good just one thing

internal/jimm/juju/cloudcredential.go
```diff
@@ -234,41 +234,74 @@ func (j *JujuManager) updateCredential(ctx context.Context, credential *dbmodel.
...
+		return nil, nil
+	}
+	if out[0].Error != nil {
+		return out[0].Models, out[0].Error
+	}
+	return out[0].Models, nil
+}
 
-	out, err := f(ctx, jujuparams.TaggedCredential{
+// forceUpdateControllerCloudCredential force-updates the credential on the
```
ale8k: If we're breaking this into two methods, doc them like a little story I'd say. Something like:

checkControllerCloudCredential checks that a given credential is safe to update across all of the models it is currently used. This is useful for doing a cross-controller check credential update safety.

forceUpdateControllerCloudCredential updates a cloud credential for a given controller. It presumes the caller has run [checkControllerCloudCredential] prior.

internal/jimm/juju/cloudcredential.go
```diff
@@ -234,41 +234,74 @@ func (j *JujuManager) updateCredential(ctx context.Context, credential *dbmodel.
...
+// controller reachable via api. If attrs is nil the attributes are fetched
+// from the credential store; otherwise the supplied attrs are applied.
```
ale8k: Prefer not to have AI prose as comments, comments are for us and AI doesn't speak human well. Don't ever wanna see `;` in a comment 😆

internal/jimm/juju/cloudcredential.go
ale8k: Fine as is IMO.

internal/jimm/juju/cloudcredential.go
SimoneDutto: it was preexisting code, i'm happy to remove it

internal/jimm/juju/cloudcredential.go
SimoneDutto: what do you mean? Fetching the attrs before calling the methods, so attrs is always not nil?

internal/jimm/juju/cloudcredential.go
SimoneDutto: what do you mean? it's a simple sentence, it's not like 50 lines comments.
About the ";" i don't know

internal/jimm/juju/cloudcredential.go
kian99: Yes, fetch the attributes and pass them in, or have the method fetch the attribute based on the credential. I don't see why it should have 2 approaches.

SimoneDutto: because sometimes the attrs is coming from the args and not from the cred store.

I can fetch the attributes at the caller side, if you prefer

kian99: Yeah just pass in the attributes from the caller unless there's a reason not too. Lmk what you think.

SimoneDutto: it was just not the best code because it was a fetch scattered around the code, and this looks a lil better, but i don't mind too much

SimoneDutto: done

internal/jimm/juju/cloudcredential.go
SimoneDutto: removed

internal/jimm/juju/cloudcredential.go
SimoneDutto: done

internal/jimm/juju/cloudcredential.go
SimoneDutto: done

kian99 (approved):

ale8k (approved):

PR #2069, opened 2026-07-27
docs: idp-groups setup and migration

SimoneDutto:
## Description
Add doc for idp-groups setup and migration from local to idp group.


docs/howto/manage-groups.md
```diff
@@ -10,6 +10,16 @@ myst:
 >
 > See also: {ref}`group`
 
+```{important}
+**Deprecation notice — migration to IDP groups**
+
+JAAS local groups (created with `juju add-group`) are deprecated and will be **removed** in the next major release of JAAS on the `v4` track. They are replaced by **IDP groups**, which are groups that your Identity Provider (IdP) already manages and sends to JAAS as an extra claim in the user's access token at login time.
```
alesstimec:
```suggestion
JAAS groups (created with `juju add-group`) are deprecated and will be **removed** in the next major release of JAAS on the `v4` track. They are replaced by **groups managed by the Identity Provider**.
```

docs/howto/manage-groups.md
```diff
@@ -10,6 +10,16 @@ myst:
...
+
+With IDP groups you no longer need to create groups in JAAS or assign users to them — user-to-group membership is owned by the IdP. In JAAS you only need to **assign permissions to an IDP group**, and every user who logs in with that group claim will inherit the permission automatically.
```
alesstimec: either user IDP or IdP (i prefer this one), but not both.

docs/howto/manage-groups.md
```diff
@@ -94,3 +106,91 @@ juju remove-group TeamB
...
+Local groups are deprecated and will be removed in the `v4` release of JAAS. To migrate, move group membership management to your IdP and replace local-group permission assignments with IDP-group permission assignments.
+
+The high-level migration steps are:
+
+1. **Configure your IdP** to emit a group claim (e.g. `groups`) that contains the group names you want to use in JAAS. The exact configuration depends on your IdP (e.g. Canonical Identity Platform / Keycloak, Auth0, etc.).
+2. **Map the existing local groups** in JAAS to the group names the IdP will emit. For each local group, make sure the IdP emits a group with a matching name, and assign the same users to that group in the IdP.
+3. **Re-assign permissions** from the local group to the IDP group. For every permission previously granted to `group-<name>#member`, grant the same permission to `idpgroup-<name>#member`.
+4. **Remove the local group** permission assignments and the local group itself once the IDP group is in use and users have confirmed they inherit the expected access.
+
+### Terraform migration example
```
kian99, line 152: I'm not sure this Terraform example belongs here. I don't mind it personally but could we link to the Terraform docs instead?

docs/howto/manage-groups.md
alesstimec:
```suggestion
With Identity Provider (IdP) groups you no longer need to create groups in JAAS or assign users to them — group membership is managed by the IdP. In JAAS you only need to **assign permissions to an IdP group**, and every member of group inherits the permission automatically.
```

docs/howto/manage-groups.md
```diff
@@ -10,6 +10,16 @@ myst:
...
 > See also: {ref}`group`
 
+```{important}
+**Deprecation notice — migration to IDP groups**
+
+JAAS local groups (created with `juju add-group`) are deprecated and will be **removed** in the next major release of JAAS on the `v4` track. They are replaced by **IDP groups**, which are groups that your Identity Provider (IdP) already manages and sends to JAAS as an extra claim in the user's access token at login time.
+
+With IDP groups you no longer need to create groups in JAAS or assign users to them — user-to-group membership is owned by the IdP. In JAAS you only need to **assign permissions to an IDP group**, and every user who logs in with that group claim will inherit the permission automatically.
+
+Local groups remain fully supported on the current `v3` track. If you are starting a new deployment, prefer IDP groups from the outset. If you already use local groups, see {ref}`migrate-to-idp-groups` below for the migration path.
```
alesstimec:
```suggestion
JAAS groups remain fully supported on the current `v3` track. If you are starting a new deployment, prefer IdP groups from the outset. If you already use JAAS groups, see {ref}`migrate-to-idp-groups` below for the migration path.
```

docs/howto/manage-groups.md
```diff
@@ -37,6 +47,8 @@ juju add-permission group-A#member member group-B
 ```
 ````
 
+The sections below describe the **local groups** workflow, which is still supported on the `v3` track. For the IDP groups workflow and the migration from local groups to IDP groups, see {ref}`idp-groups` and {ref}`migrate-to-idp-groups`.
```
alesstimec: i would not talk about "local" group, but use "JAAS groups" instead.

docs/howto/manage-groups.md
```diff
@@ -94,3 +106,91 @@ juju remove-group TeamB
 ```
 
 > See more: {doc}`juju remove-group <../reference/jaas-plugin>`
+
+(idp-groups)=
+## About IDP groups
+
+> See also: {ref}`group`
+
+An **IDP group** is a group that is owned and managed by your Identity Provider (IdP), not by JAAS. When a user logs in to JAAS, the IdP includes an extra claim in the access token that lists the groups the user belongs to. JAAS reads these claims and treats each value as an IDP group the user is a member of.
```
alesstimec: group being "owned" by and IdP is a bit weird.. "manged" is a better word

kian99 (commented):
We're missing information on some of the things we are losing with IdP groups - particularly:
1. `juju jaas check-permission` no longer works for checking a user's indirect access since the user's groups are not known.
2. `juju show-model` will no longer show users that have access through idp-group membership.
3. In general, any command that relies on another user's permissions may not work as expected.

alesstimec (changes requested):
a few comments:
- use "JAAS groups", not "local groups"
- unify IdP vs IDP vs idp
- let's not mix JIMM and JAAS - use either one or the other for consistency
- don't mix in too many implementation details - claims and such - groups are managed by the IdP and in JAAS you only assign permissions to these groups (or users)

docs/howto/manage-groups.md
luci1900: — could be ,

docs/howto/manage-groups.md
```diff
@@ -94,3 +106,91 @@ juju remove-group TeamB
...
+
+Compared to local groups, the key difference is that **JAAS never creates IDP groups or assigns users to them** — that is entirely the IdP's responsibility. In JAAS you only assign permissions (e.g. `administrator` on a model) to an IDP group, and every user who logs in with that group claim inherits the permission automatically.
```
luci1900:
```suggestion
Compared to JAAS groups, the key difference is that **JAAS no longer creates groups or assigns users to them**. In JAAS you only assign permissions (e.g. `administrator` on a model) to an IdP group, then every user who logs in with that group claim inherits the permission automatically.
```

docs/howto/manage-groups.md
luci1900: It's a bit difficult to find a better place for it, though.

It's also tricky to come up with better syntax for expressing the migration.

luci1900: I do agree that it feels a bit weird. If we weren't also working on the provider and so familiar with it, we wouldn't even consider this.

luci1900:
> 1. `juju jaas check-permission` no longer works for checking a user's indirect access since the user's groups are not known.

This is particularly a case where some amount of describing the implementation is useful. The fact that JAAS *doesn't* replicate the groups in its database from the IdP (either lazily or explicitly) has UX-impacting consequences. I think we should mention it.

docs/howto/manage-groups.md
SimoneDutto: At the end of the day this seems public doc but it's indeed a guide for IS to migrate to idp group, and they specifically use the terraform provider. So it's useful to have an example here, on how to do it.

We can even remove this page at one point, since nobody will use the old jaas-groups

docs/howto/manage-groups.md
```diff
@@ -94,3 +106,96 @@ juju remove-group TeamB
...
+### Limitations when checking indirect access
+
+IdP group membership is provided in the group claim when a user logs in. As a result, commands that need to evaluate another user's permissions cannot determine that user's indirect access through an IdP. 
+
+For example, `juju jaas check-permission` cannot reliably check a user's access inherited from an IdP group, and `juju show-model` does not list users whose access comes through IdP group membership. The same limitation applies to any command that relies on another user's permissions.
+
+(configure-jimm-IdP-groups)=
+## Configure JIMM's charm to accept group claims
+There are two ways to configure JIMM's charm for IdP group claims, and it depends on the way the Identity Platform was configured.
+If the identity platform was configured to emit the group claim by default you will need to change the `oauth-group-claim-key` charm config with the key JIMM will find the group claim in the extra claims (e.g. `groups`).
```
kian99:
```suggestion
If the identity platform was configured to emit the group claim by default you will need to change the `oauth-group-claim-key` charm config with the key where JIMM will find the group claim (e.g. `groups`).
```

docs/howto/manage-groups.md
```diff
@@ -94,3 +106,96 @@ juju remove-group TeamB
...
+If the identity platoform was configured to offer the group claim as an option, in addition to previous configuration you'll need to configure `oauth-client-credential-scopes` and `oauth-optional-scopes` (e.g. `groups`) to inform the Identity Platform to emit the optional scopes for users and service accounts.
```
kian99:
```suggestion
If the identity platform is configured to _optionally_ offer the group claim then additionally configure both `oauth-client-credential-scopes` and `oauth-optional-scopes` (e.g. `groups`) to request group claims when authenticating client-credentials and users, respectively.
```

kian99 (approved):

SimoneDutto:
> a few comments:
> 
> * use "JAAS groups", not "local groups"
> * unify IdP vs IDP vs idp
> * let's not mix JIMM and JAAS - use either one or the other for consistency
> * don't mix in too many implementation details - claims and such - groups are managed by the IdP and in JAAS you only assign permissions to these groups (or users)

i've addressed all the comments, about the implementation details i think it makes sense what i've mentioned to give context about the configuration.

I've used JIMM when referring the actual charm, that it's called jimm

alesstimec (approved):
Thanks

PR #2071, opened 2026-07-30
feat: improve juju jaas query-models

SimoneDutto:
## Description

Niko from IS complain `juju jaas query-models` was very slow for a query he was doing.

Indeed, we were querying sequentially each model, which in IS case is ~400.

This change is very simple and it just tries to parallellize the querying with an error group.

The behavior should be the exact same as before, just with a 10 parallelism.

We can go higher than 10 parallelism for sure, but i think it's a good start. I personally wouldn't expose it as a configurable option for now.

I'm aware this change means we have 10 jq queries potentially running at the same time, but i think it shouldn't be a problem. Or not less of a problem of 10 people running the same query, or 1 person running a very expensive query.

# QA

`make qa-lxd`

<create 100 models>

Before the patch:
```
[9:15:27] ➜  jimm-v3 git:(v3) ✗ time ./jaas query-models '.models'
<outpu>
./jaas query-models '.models'  0.02s user 0.01s system 1% cpu 2.010 total
```

After the patch:
```
[9:16:03] ➜  jimm-v3 git:(make-jaas-query-models-faster) time ./jaas query-models '.models'
<output>
./jaas query-models '.models'  0.02s user 0.02s system 7% cpu 0.464 total
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,3 +37,5 @@ docs/.sphinx/.wordlist.dic
 docs/.sphinx/.doctrees/
 docs/.sphinx/update/
 docs/.sphinx/node_modules/
+
+jaas
--- a/internal/jimm/juju/model_status_parser.go
+++ b/internal/jimm/juju/model_status_parser.go
@@ -7,6 +7,7 @@ import (
 	"encoding/json"
 	stderrors "errors"
 	"fmt"
+	"sync"
 
 	"github.com/itchyny/gojq"
 	jujucmd "github.com/juju/cmd/v3"
@@ -16,11 +17,14 @@ import (
 	"github.com/juju/names/v5"
 	"github.com/juju/zaputil/zapctx"
 	"go.uber.org/zap"
+	"golang.org/x/sync/errgroup"
 
 	"github.com/canonical/jimm/v3/internal/dbmodel"
 	"github.com/canonical/jimm/v3/internal/errors"
 	"github.com/canonical/jimm/v3/pkg/api/params"
 )
+
+const maxConcurrentModelQueries = 10
 
 // QueryModels queries every specified model in modelUUIDs.
 //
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -37,3 +37,5 @@ docs/.sphinx/.wordlist.dic
 docs/.sphinx/.doctrees/
 docs/.sphinx/update/
 docs/.sphinx/node_modules/
+
+jaas
```
SimoneDutto, line 41: for the jaas binary

alesstimec (commented):
I would suggest we parallelize getting the model status, which is the slow bit. We can still run queries sequentially - idk if gojq is safe for concurrent use.

SimoneDutto:
> I would suggest we parallelize getting the model status, which is the slow bit. We can still run queries sequentially - idk if gojq is safe for concurrent use.

this is the godoc:
```
// Run the query.
//
// It is safe to call this method in goroutines, to reuse a parsed [*Query].
// But for arguments, do not give values sharing same data between goroutines.
func (e *Query) Run(v any) Iter {
	return e.RunWithContext(context.Background(), v)
}
```

So it should be completely safe

internal/jimm/juju/model_status_parser.go
```diff
@@ -39,71 +43,90 @@ func (j *JujuManager) QueryModelsJq(ctx context.Context, modelUUIDs []string, jq
...
 
-		for {
-			v, ok := queryIter.Next()
-			if !ok {
-				break
+			formattedStatus, err := formatter.Format()
+			if err != nil {
+				zapctx.Error(ctx, "failed to format status", zap.String("model-uuid", modelUUID))
+				addItem(modelUUID, nil, err)
+				return nil
```
alesstimec, line 88: why do we return nil here and on L95, but on L99 we return an error?

it might be a bit clearer to read if we deferred the addItem call

alesstimec (approved):

internal/jimm/juju/model_status_parser.go
SimoneDutto: i didn't use the defer because the code before was very purposeful on where to return an error or a result

internal/jimm/juju/model_status_parser.go
```diff
@@ -39,71 +43,90 @@ func (j *JujuManager) QueryModelsJq(ctx context.Context, modelUUIDs []string, jq
...
 				}
-				results.Errors[modelUUID] = append(results.Errors[modelUUID], "jq error: "+err.Error())
-				continue
-			}
 
-			results.Results[modelUUID] = append(results.Results[modelUUID], v)
-		}
+				// Jq errors can range from one failure in an iterative query to an entirely broken
+				// query. As such, we simply append all to the errors field and continue to collect
+				// both erreoneous and valid query results.
```
luci1900, line 116: erroneous

luci1900 (approved):
I agree on hard-coding the concurrency for now. It's not necessarily something an operator will be able to reason about.

Locking around adding results should be fine, the parallelism is for IO.

internal/jimm/juju/model_status_parser.go
```diff
@@ -39,71 +43,90 @@ func (j *JujuManager) QueryModelsJq(ctx context.Context, modelUUIDs []string, jq
...
 	if err != nil {
 		return results, errors.New("failed to get models for user")
 	}
 
-	for _, model := range models {
-		modelUUID := model.UUID.String
-		params, err := retriever.GetParams(ctx, model)
+	g, _ := errgroup.WithContext(ctx)
+	g.SetLimit(maxConcurrentModelQueries)
+	mux := sync.Mutex{}
```
kian99: You'd usually see the word `mux` associated with a networking mux in Go. A mutex is commonly just `m` or more descriptive for the use case. I'd just go with `m`.

kian99 (approved):

kian99:
The linter complaining about the cognitive complexity should be address without add an ignore if possible.

SimoneDutto:
> The linter complaining about the cognitive complexity should be address without add an ignore if possible.

i was about to ask, i think it's best to just ignore

internal/jimm/juju/model_status_parser.go
SimoneDutto: will do

internal/jimm/juju/model_status_parser.go
SimoneDutto: kept the same behavior as before, with the return nil or error

PR #2073, opened 2026-08-18
feat: use batch checks

SimoneDutto:
## Description

Following https://github.com/canonical/ofga/pull/79 i'm using batch checks.
By taking a look at the code it should improve:
- the CLI command to check permissions
- the redoLogin path in case of juju controller login errors

Just a note, if we used https://github.com/canonical/jimm/pull/2019 i wouldn't have to extract and call parseAndAuthorizeRelationCheck from both `CheckRelation` and `CheckRelations`, because that kind of logic would have been enforced in the authorization model, but i know that you didn't like it at that time.

# QA

If tests are passing should be fine.


internal/jimm/permissions/relations.go
```diff
@@ -162,45 +163,86 @@ func (j *PermissionManager) RemoveRelation(ctx context.Context, user *openfga.Us
...
+			CorrelationId:    correlationID,
+			ContextualTuples: contextualTuples,
+		}
+		parsedTuples = append(parsedTuples, check)
+	}
+
+	if len(parsedTuples) == 0 {
+		return results, nil
+	}
+	allowed, err := j.authSvc.BatchCheckRelations(ctx, parsedTuples)
```
kian99, line 235: Does the batch check have a max limit? We should check that we don't allow the input to exceed that limit.

internal/jimm/permissions/relations.go
```diff
@@ -162,45 +163,86 @@ func (j *PermissionManager) RemoveRelation(ctx context.Context, user *openfga.Us
...
 		return allowed, errors.Codef(errors.CodeOpenFGARequestFailed, "%w", err)
 	}
 	return allowed, nil
 }
 
+// parseAndAuthorizeRelationCheck parses a relation check and verifies that the
+// authenticated user is allowed to inspect it. JIMM administrators can inspect
+// any relation; other users can inspect their own relation or a relation on a
+// resource they administer.
+func (j *PermissionManager) parseAndAuthorizeRelationCheck(ctx context.Context, user *openfga.User, tuple apiparams.RelationshipTuple) (*openfga.Tuple, error) {
```
kian99, line 186: Why was this reorganization of the code made?

internal/jimm/permissions/relations.go
SimoneDutto: yeah, i moved the batching to the limit inside the client method itself. Take a look at internal/openfga/openfga.go

Ideally we should receive that limit from the charm, but since it's not even possible to configure it at this time, i've just added a constant.
https://charmhub.io/openfga-k8s/configurations

internal/jimm/permissions/relations.go
SimoneDutto: because i need to call this from both `CheckRelation` and `CheckRelations` so i've extracted this into a method to avoid duplicating the same code

internal/jimm/permissions/access.go
```diff
@@ -207,15 +214,37 @@ func (j *PermissionManager) CheckPermission(ctx context.Context, user *openfga.U
...
+	batch := make([]openfga.TupleWithCorrelationId, len(checks))
+	for i, check := range checks {
+		batch[i] = openfga.TupleWithCorrelationId{
+			Tuple: &openfga.Tuple{
+				Object:   ofganames.ConvertTag(user.ResourceTag()),
+				Relation: check.relation,
+				Target:   ofganames.ConvertGenericTag(check.tag),
+			},
+			CorrelationId:    strconv.Itoa(i),
+			ContextualTuples: contextualTuples,
```
alesstimec, line 236: we are sending the same contextual tuples with every tuple we want to check.. that seems rather wasteful.

SimoneDutto: we could improve it by making the assumption that the contextual tuple in this checks are useful only when we are checking something with the admin user, but i don't think it's worth it.

Plus, we were adding the contextual tuple to each single request before

internal/openfga/openfga.go
```diff
@@ -246,6 +255,27 @@ func (o *OFGAClient) CheckRelation(ctx context.Context, tuple Tuple, trace bool,
 	return o.cofgaClient.CheckRelation(ctx, tuple, contextualTuples...)
 }
 
+// BatchCheckRelations verifies multiple relations in a single OpenFGA request.
+// The returned map is keyed by the correlation ID supplied with each tuple.
+func (o *OFGAClient) BatchCheckRelations(ctx context.Context, tuples []TupleWithCorrelationId) (_ map[string]bool, err error) {
```
kian99, line 260: We don't seem to have a test for this, particularly for the case where the len(tuples) > BatchCheckSize. Specifically, what are the keys of the `results` map?

kian99 (commented):
Looks good, just one final thing.

internal/openfga/openfga.go
SimoneDutto: `// The returned map is keyed by the correlation ID supplied with each tuple.`

you mean this or something else?

i'll add the tests yeah, i though i did, but it was just in the lib which isn't enough

kian99: Ah okay so the keys will be from 0 -> len(tuples).

SimoneDutto: added a tests for it

kian99 (approved):

internal/jimm/permissions/access.go
```diff
@@ -207,15 +214,37 @@ func (j *PermissionManager) CheckPermission(ctx context.Context, user *openfga.U
...
+		}
+	}
+	allowed, err := j.authSvc.BatchCheckRelations(ctx, batch)
```
alesstimec, line 239: what is the maximum batch size? are we likely to hit the limit?

internal/jimm/permissions/access.go
```diff
@@ -192,7 +193,13 @@ func (j *PermissionManager) RevokeAuditLogAccess(ctx context.Context, user *open
...
+	type permissionCheck struct {
+		key      string
+		value    string
+		tag      names.Tag
+		relation openfga.Relation
+	}
+	checks := make([]permissionCheck, 0, len(desiredPerms))
 	for key, val := range desiredPerms {
 		if _, ok := cachedPerms[key]; !ok {
 			stringVal, ok := val.(string)
```
alesstimec, line 205: if we expect map[string]string, then why isn't desiredPerms (i'd name it requiredPermissions, but that's just me) a map[string]string?

alesstimec: or better yet map[names.Tag]openfga.Relation?

internal/jimm/permissions/access.go
alesstimec: aren't contextual tuples specific to a user? so they'd always be the same for the entire batch?

internal/jimm/permissions/access.go
SimoneDutto: the default is 50, and we do the batching inside the client method, so even if somebody sends a 51 tuples check is fine.

The only actual problem could be if somebody changes the limit, but since the charm doesn't expose any option to do that, it's not very likely.

SimoneDutto: see https://github.com/canonical/jimm/pull/2073#discussion_r3803382264

internal/jimm/permissions/access.go
SimoneDutto: i didn't touch this code, i've just adapted it to the new batching code

internal/jimm/permissions/access.go
SimoneDutto: in this case yes, but the API in openfga has contextual tuples for each item, not generic contextual tuples for the whole batch. 
And it makes sense, because you might want to use different contextual tuples.

We don't, but the API is shaped like that, so i need to duplicate it for each batch item

alesstimec (approved):

PR #2074, opened 2026-08-19
feat: enable cloud updates

### LINKED ISSUE #2072: Allow updating a cloud
In a discussion with IS, Terraform was trying to reconcile its state with JIMM and running into the following issue,
```terraform
Terraform will perform the following actions:

  # juju_kubernetes_cloud.cluster_k8s_cloud will be updated in-place
  ~ resource "juju_kubernetes_cloud" "cluster_k8s_cloud" {
        id                            = "k8s-prod-launchpad-job-farm-ps7:k8s-prod-launchpad-job-farm-ps7"
      ~ kubernetes_config             = (sensitive value)
        name                          = "k8s-prod-launchpad-job-farm-ps7"
        # (4 unchanged attributes hidden)
    }

....
╷
│ Error: Not Supported
│ 
│   with juju_kubernetes_cloud.cluster_k8s_cloud,
│   on main.tf line 158, in resource "juju_kubernetes_cloud" "cluster_k8s_cloud":
│  158: resource "juju_kubernetes_cloud" "cluster_k8s_cloud" {
│ 
│ Cloud Update is not supported in JAAS.
╵
Releasing state lock. This may take a few moments...
```
The error comes from the Terraform provider which specifically forbids updating a cloud when connected to JAAS, because JAAS does not support updating a cloud definition.

We should understand why this is / is not possible and how to remedy it.

SimoneDutto:
## Description
Fix #2072 

I'll write what I've found here, even if it might be trivial.

`juju add-cloud` calls the `AddCloud` facade, which we can't use to add clouds in JAAS unless it's a hosted cloud.
The reason is that without the hosted cloud info we can't resolve which controller to add the cloud to.

And that's why we have a limitation in our `AddCloud` code to only allow kubernetes clouds, which are always hosted.

For the Update path is different, the cloud tag is used to update a cloud, so we can allow updating hosted and non-hosted clouds because we can figure out the controllers by getting the cloud first.

One important thing to know is that currently in terraform we only support adding `juju_kubernetes_cloud` to JAAS, there is nothing preventing us for allowing using the `juju_cloud` resource as well, we just need to use the right `AddCloudToController` JIMM facade.

# QA

the e2e tests, and i've also locally tested the kubernetes_cloud resource with terraform.


kian99 (dismissed):
I am a little apprehensive about this just because the deleted todo makes me pause and think what else might we be missing. But in theory this shouldn't be any different from running `juju update-cloud` on a standalone controller.

SimoneDutto:
> I am a little apprehensive about this just because the deleted todo makes me pause and think what else might we be missing. But in theory this shouldn't be any different from running `juju update-cloud` on a standalone controller.

i'm also working on the terraform pr to add clouds to jaas, so it should improve the coverage over this cloud update feature.

luci1900 (dismissed):

alesstimec (commented):
I am also a bit worried about the todo that's being removed.. Could we get input from the core juju team about cloud updates and what dragons we could potentially find there?

ale8k:
To add some historical context as to why update is the way it is: 

The reason we have hosted clouds additions allowed is because of BYOC for K8S scenarios in the previous jaas.ai solution. And all other clouds pre-existed via the admins on the controller. So there was never a reason to update the cloud via JIMM as the admin could go to the controller directly. 

I'm not sure if you can update the k8s cloud via add-k8s but I think you can (if someone could verify this?) 

I'm not 100% sure I follow martins todo though, we previously had in house RBAC within our Postgres before OpenFGA so they were able to model clouds and in theory allow updates even pre openfga. 

My gut thinks this is probably ok, but perhaps I'm missing something.

SimoneDutto:
After running the terraform tests with this newly Update path i've found 2 inconsistencies:
- regions in JAAS were added, and not replaced. The juju implementation replaces the regions, see here: https://github.com/juju/juju/blob/78832a31ad6b/state/cloud.go#L86-L121
- skipTLSVerify was not persisted in JAAS, so it was always returned to client zero-valued. So we need to persist it.

I have removed the gorm model from cloud_region because gorm soft deletes entries so ON CASCADE DELETE doesnt' work (see [here](https://github.com/juju/juju/blob/78832a31ad6b/state/cloud.go#L86-L121)). 

Now the terraform acc tests pass.

kian99 (approved):
Good find

SimoneDutto:
> I am also a bit worried about the todo that's being removed.. Could we get input from the core juju team about cloud updates and what dragons we could potentially find there?

asked, and there shouldn't be anything particularly problematic about allowing it in JAAS, since it's allowed in juju

luci1900 (approved):

PR #2077, opened 2026-08-24
test(e2e): refactors e2e test since juju api reports the juju version

alesstimec:
## Description

Refactors e2e tests since the Juju API reports the its version.

## Engineering checklist

- [ ] Documentation updated
- [ ] Covered by unit tests


SimoneDutto (approved):
so this is just an improvement, not a change right? because before it was already reporting the right juju version

kian99 (approved):

PR #2078, opened 2026-08-25
perf: find-offers

SimoneDutto:
# Description

This pr is a way to realize the way we are performing `find-offers`.
Before this pr we were always dialing every controller with the filter and then we were checking permission to avoid returning offer a user doesn't have access to.

Now it fetches the offer a user has at least reader access to, get the controllers where the offers live, and dial only the subset of controllers.

This should be a net improvement for whoever doesn't have a offer on a model on each controller.
It is slight slower for people having access to offers on each controller, because now it also list all offers from openfga.
But i think it's a good trade off.

Initially, i though i could improve this even more by:
- listing offers
- figuring model/controller pair for each
- build the filters to ask only the right offers from the controllers

But the code was harder to read, and this improvement is better in terms of not dialing unnecessary controllers without complicating the code too much.

### DIFF
--- a/internal/jimm/juju/applicationoffer.go
+++ b/internal/jimm/juju/applicationoffer.go
@@ -7,6 +7,8 @@ import (
 	"database/sql"
 	stderrors "errors"
 	"fmt"
+	"maps"
+	"slices"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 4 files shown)


internal/jimm/juju/applicationoffer.go
kian99, line 559: This change highlights something worth changing imo. If memory serves, the reason we need a *db.Controller to pass into j.dial is that the dial logic updates some controller details after the dial succeeds.

Basically, the `ctl := &controllers[i]` is a bit weird and it would be great if we didn't have this oddness.

kian99 (commented):
Isn't the "real" problem that by default everyone has read access to all offers?

internal/jimm/juju/applicationoffer.go
SimoneDutto: i think i can skip this assignment

luci1900: It's certainly surprising.

SimoneDutto:
> Isn't the "real" problem that by default everyone has read access to all offers?

is it? everybody has read access to all offers? i didn't know that

luci1900 (approved):
Nice approach.

I also think one more round-trip in the worst case is worth many fewer round-trips in the common case.

SimoneDutto:
Everybody has read access to all offers as Kian pointed out, and this is not a jaas thing, it's a juju thing. See https://github.com/juju/juju/blob/957e67d923d5/apiserver/facades/client/applicationoffers/applicationoffers.go#L148

So this pr will not improve performance because we will always query all controllers.

One way to improve this would be to actually implement the filters in JAAS such as if you filter by model we don't query all controllers, however juju is not sending the filters from the CLI right now. So there is no reason to fix it right now.

PR #2080, opened 2026-08-26
feat: dial with user permissions

luci1900:
## Description

Scopes the JWTs JIMM mints for backing-controller connections to the caller's real OpenFGA permissions, instead of always asserting controller `superuser`. This provides additional defense in depth, since Juju would also be checking permissions for calls made on behalf of users. JIMM failing to correctly check a user's permissions may not necessarily be a security event.

Application-offer operations (consume details, get offer, offer admin) are now authorized via offer-level access even when the caller has no model access: caller tokens are built from a set of target resources (model and/or application-offer tags) rather than a single model tag.

Controllers < 3.6.24 don't correctly honour model-level JWT claims, so we fall back to a superuser token for them.

Operations that genuinely need more than the caller has still dial as JIMM's service identity. Most are legitimate, but others are due to limitations in JIMM or Juju. These exceptions are documented at `internal/jimm/juju/dialing.md`.

Builds on the dialer split from https://github.com/canonical/jimm/pull/2090.

## Engineering checklist

- [x] Documentation updated
- [x] Covered by unit tests

## Test instructions

Requires the e2e backing-controller setup described in `testing/README.md`.
Requires a backing controller on Juju **3.6.24+** to exercise the scoped-token path (older controllers hit the superuser fallback).

```
RUN_E2E_TESTS=1 JIMM_BACKING_CONTROLLER_CONFIG="$PWD/controllers.yaml" \
  go test ./testing/
```


internal/jimm/jimm.go
```diff
@@ -370,9 +371,28 @@ func NewDialerAdapter(dialer *jujuclient.Dialer) *DialerAdapter {
...
+// It uses the underlying jujuclient.Dialer to establish a model-scoped
+// connection to the Juju controller and returns a juju.API connection.
+func (d *DialerAdapter) DialModel(ctx context.Context, ctl *dbmodel.Controller, modelTag names.ModelTag, user *openfga.User) (juju.API, error) {
+	return d.dialer.DialModel(ctx, ctl, modelTag, user)
+}
+
+// DialController implements the juju.Dialer interface for the DialerAdapter.
+// It dials the controller with a controller-scoped connection whose JWT
+// carries the user's access claims for the given target resources.
+func (d *DialerAdapter) DialController(ctx context.Context, ctl *dbmodel.Controller, targets []names.Tag, user *openfga.User) (juju.API, error) {
```
SimoneDutto: what are the targets used for when dialing a controller?

SimoneDutto: Can you expand with examples the godoc of these methods? So it's clearer when to use one or the other

internal/jimm/juju/cloud.go
```diff
@@ -343,8 +343,11 @@ func (j *JujuManager) AddHostedCloud(ctx context.Context, user *openfga.User, ta
 // addControllerCloud returns the definition of the cloud retrieved from
 // the controller. No error will be returned if the cloud already exists on
 // the controller or the user already has access to the cloud.
-func (j *JujuManager) addControllerCloud(ctx context.Context, ctl *dbmodel.Controller, user *openfga.User, tag names.CloudTag, cloud jujucloud.Cloud, force bool) (*jujucloud.Cloud, error) {
-	api, err := j.dial(ctx, ctl, names.ModelTag{}, user)
+//
+// Dials as the JIMM service identity after enforcing authorisation, since
+// users lack the required permission on the backing controller.
+func (j *JujuManager) addControllerCloud(ctx context.Context, ctl *dbmodel.Controller, tag names.CloudTag, cloud jujucloud.Cloud, force bool) (*jujucloud.Cloud, error) {
+	api, err := j.dialControllerAsService(ctx, ctl)
```
SimoneDutto, line 353: if i get it, everytime we dial as a service it means there is something we could potentially address in juju

internal/jimm/juju/jimm.go
```diff
@@ -206,7 +206,7 @@ func (j *JujuManager) FullModelStatus(ctx context.Context, user *openfga.User, m
 		return nil, err
 	}
 
-	api, err := j.dial(ctx, &model.Controller, modelTag, nil)
+	api, err := j.dialModelAsService(ctx, &model.Controller, modelTag)
```
SimoneDutto, line 209: why can't you use DialModel here?

internal/jimm/juju/migrationtarget.go
```diff
@@ -47,7 +47,7 @@ func (j *JujuManager) AbortMigration(ctx context.Context, user *openfga.User, mo
 		return fmt.Errorf("failed to get model migration %q: %w", modelUUID, err)
 	}
 
-	api, err := j.dialController(ctx, &incomingModel.TargetController, user)
+	api, err := j.dialControllerAsService(ctx, &incomingModel.TargetController)
```
SimoneDutto, line 50: why do we need service here? all things migration related needs a service? Can we encode this info somewhere in the code such as for the future we know?

internal/jimm/juju/model.go
```diff
@@ -708,7 +708,7 @@ func (j *JujuManager) UpgradeController(ctx context.Context, user *openfga.User,
 		return version.Number{}, err
 	}
 
-	api, err := j.dialController(ctx, controller, user)
+	api, err := j.dialControllerAsService(ctx, controller)
```
SimoneDutto, line 711: sometimes you've used the diasasservice with a godoc, and sometimes you didn't.
So i'm sure if it's actually needed to use the service or not.

I'd prefer a markdown file somewhere with a list of operations needing service access, so we can monitor/fix

internal/jimm/juju/modelbuilder.go
```diff
@@ -592,7 +604,15 @@ func (b *modelBuilder) CreateControllerModel() *modelBuilder {
 	// attempts to create a model with the same name again.
 	// TODO(JUJU-8869): We need to keep this despite encoding permissions in
```
SimoneDutto: what about this TODO? seems related to the jwt scoping thing

internal/jimm/jujuauth/logintokens.go
```diff
@@ -107,6 +109,93 @@ func (auth *LoginTokenGenerator) makeSuperuserToken(ctx context.Context, user *o
...
+			hasRealAccess = true
+		}
+	}
+
+	// Juju rejects a present-but-empty claim as invalid, so empty claims are
+	// only omitted once another target grants real access; with a single
+	// target (e.g. a direct model login), an empty claim is kept so Juju
+	// denies it.
+	for target, access := range targetAccess {
+		if access == "" && hasRealAccess {
```
SimoneDutto, line 172: i don't get this one, hasRealAccess is a flag that can be set by any of the target.

Can you explain it to me?

SimoneDutto (commented):
this is very good, and it makes very easy to understand the permission level we are dialing a controller and that's amazing.

I have just a bunch of questions to document what you've understood so far, so we know what to do in the future.

internal/jimm/jimm.go
luci1900: `targets` is a generalisation of `modelTag` so you can also pass in offers.

I'll explain better.

luci1900: FWIW it's explained more in the interface, but slop https://github.com/canonical/jimm/pull/2080/changes#diff-c4b815b876dd30895b62ab4a6118a44f1de2ccbf476eb47886bb400b20a23427R39

internal/jimm/juju/model.go
luci1900: I've basically started with no such calls, then kept adding them to make tests pass.

I'll do an inventory.

internal/jimm/jujuauth/logintokens.go
luci1900: Juju rejects if there are claims present but empty. There's a few places working around this.

internal/jimm/juju/cloud.go
luci1900: Mostly.

For now, there are also some places where it's merely more convenient.

We'll probably always keep some, at least for anything we might do scheduled and not triggered by a user.

internal/jimm/jujuauth/logintokens.go
luci1900: I've improved the wording a bit.

internal/jimm/jimm.go
luci1900: Improved the wording on the interface godoc.

internal/jimm/juju/modelbuilder.go
luci1900: Sort of. The permissions are in controller state and needed by later JIMM calls like migration.

It's one of the cases where this PR doesn't try to change the existing behaviour.

internal/jimm/juju/model.go
luci1900: I've added an inventory.

internal/jimm/juju/jimm.go
luci1900: A JIMM admin isn't required to have permissions on the model.

Maybe this could change? I don't really understand the consequences.

internal/jimm/juju/migrationtarget.go
luci1900: Yes, I think lots of migration-related things will need JIMM to dial as itself.

In this case, the target controller won't know about the user, or at least not yet.

SimoneDutto (commented):
have you added the inventory in a PR before? I can't remember.

In this PR what are the place where we are dialing with less permissions than before? So we can check if we test that path e2e. 

Other than than looks good

luci1900:
@SimoneDutto I've added this https://github.com/luci1900/jimm/blob/aa4ce220f1df4e55cff9e569b7df6700bc8bf7d4/internal/jimm/juju/dialing.md to explain dialing.

Everything that *isn't* on that list is being dialed with the user's OpenFGA permissions, so reduced from the `superuser` before this PR.

internal/jimm/jujuauth/factory.go
```diff
@@ -79,6 +81,11 @@ func (f *Factory) NewCallerScopedLoginToken(ctx context.Context, resourceTags []
 	ctlWithClouds := dbmodel.Controller{}
 	ctlWithClouds.SetTag(ctl.ResourceTag())
 	if err := f.db.GetController(ctx, &ctlWithClouds); err != nil {
+		// Fall back only for not-found (e.g. controller not persisted
+		// yet during AddController). Propagate transient DB errors.
+		if errors.ErrorCode(err) != errors.CodeNotFound {
+			return nil, fmt.Errorf("failed to fetch controller for caller token: %w", err)
+		}
```
SimoneDutto: this is new, why?

luci1900:
Splitting up in several smaller PRs.

PR #2081, opened 2026-08-27
feat: implement SSHServerHostKey

SimoneDutto:
# Description

juju ssh --jump requires SSHServerHostKey to fetch the jump host key, it is a controller method we need to implement in jimm

Contextually i had to update to latest juju from main, to have the new api params for the new facade.

# QA 
Since we dopnt' have a way to add keys to jaas you need to manually edit the code to accept an arbitrary key you can hard code
```
// VerifyPublicKey lists the key for a user and compares the key to find a match.
func (sm *SSHKeyManager) VerifyPublicKey(ctx context.Context, claimUser string, publicKey []byte) (bool, error) {
	// Temporary local jump-SSH test bypass. Remove before committing.
	if claimUser == temporaryJumpTestIdentity {
		temporaryKey, _, _, _, err := gossh.ParseAuthorizedKey([]byte(temporaryJumpTestPublicKey))
		if err != nil {
			return false, err
		}
		presentedKey, err := gossh.ParsePublicKey(publicKey)
		if err != nil {
			return false, err
		}
		if ssh.KeysEqual(temporaryKey, presentedKey) {
			return true, nil
		}
	}
```

`make qa-lxd` using latest juju from main

`juju ssh --jump 0` should work.


kian99 (approved):

internal/jujuapi/controller.go
```diff
@@ -63,7 +65,21 @@ func init() {
 		r.AddMethod("Controller", 14, "WatchAllModelSummaries", watchAllModelSummariesMethod)
 		r.AddMethod("Controller", 14, "InitiateMigration", initiateMigrationMethod)
 
-		return []int{12, 14}
+		// Controller facade version 15 adds SSHServerHostKey.
```
luci1900, line 68: I don't get why we're now adding all the other ones. Is it because we're adding them for this facade version?

SimoneDutto: because the whole facade was bumbed to 15.
So we are implementing all methods, otherwise a client with facade 14 (juju 4.0 cli) couldn't talk with this version of jaas

luci1900 (approved):

PR #2087, opened 2026-08-31
feat: intercept key manager to populate jimm's db

SimoneDutto:
## Description

Intercept keymanager to populate jimm's db.

- for juju 4 we intercept the call, and we don't proxy to the controller
- for juju 3 we just forward the call

## QA

`make qa-lxd`

```
juju add-model a
juju add-machine

juju ssh --jump 0
```

### DIFF
--- a/internal/db/sshkeys.go
+++ b/internal/db/sshkeys.go
@@ -32,6 +32,32 @@ func (d *Database) AddSSHKey(ctx context.Context, sshKey *dbmodel.SSHKey) (err e
 	return nil
 }
 
+// RemoveSSHKeys removes a user's keys matching any provided fingerprint or
+// comment. Missing targets are ignored.
+func (d *Database) RemoveSSHKeys(ctx context.Context, identityName string, model SSHKeyModelFilter, targets ...string) (err error) {
+	const op = "db.RemoveSSHKeys"
+	if err := d.ready(); err != nil {
+		return err
+	}
+	if len(targets) == 0 {
+		return nil
+	}
+
+	durationObserver := servermon.DurationObserver(servermon.DBQueryDurationHistogram, op)
+	defer durationObserver()
+	defer servermon.ErrorCounter(servermon.DBQueryErrorCount, &err, op)
+
+	query := d.DB.WithContext(ctx).
+		Where("identity_name = ?", identityName).
+		Where("model_uuid = ?", model.ModelUUID).
+		Where("md5_fingerprint IN ? OR key_comment IN ?", targets, targets).
+		Delete(&dbmodel.SSHKey{})
+	if query.Error != nil {
+		return dbError(query.Error)
+	}
+	return nil
+}
+
 // RemoveSSHKeyByFingerprint removes a user's ssh key identified by its fingerprint.
 func (d *Database) RemoveSSHKeyByFingerprint(ctx context.Context, identityName string, model SSHKeyModelFilter, fingerprint string) (err error) {
 	const op = "db.RemoveSSHKeyByFingerprint"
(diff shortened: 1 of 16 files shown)


internal/rpcproxy/keymanager.go
kian99, line 1: I know this is the same way it was done before, but I wonder if we should put this code into `internal/websocket/` and route the requests from the model proxy through to there somehow. But feel free to ignore this comment.

tests/ssh/verify-ssh-server-running.sh
kian99, line 1: I presume this test already works since we are using the Juju edge snap?

kian99 (approved):

tests/ssh/verify-ssh-server-running.sh
SimoneDutto: i hope, yes

internal/db/sshkeys.go
```diff
@@ -32,6 +32,32 @@ func (d *Database) AddSSHKey(ctx context.Context, sshKey *dbmodel.SSHKey) (err e
...
+	}
+	if len(targets) == 0 {
+		return nil
+	}
+
+	durationObserver := servermon.DurationObserver(servermon.DBQueryDurationHistogram, op)
+	defer durationObserver()
+	defer servermon.ErrorCounter(servermon.DBQueryErrorCount, &err, op)
+
+	query := d.DB.WithContext(ctx).
```
luci1900, line 50: Maybe we should test that you can't actually delete another user's key.

SimoneDutto: in jimm?

luci1900 (approved):

internal/db/sshkeys.go
SimoneDutto: added

internal/rpcproxy/keymanager.go
SimoneDutto: i did some tests and none looked particularly better than this, so i'm leaving as is

internal/rpcproxy/rpcproxy.go
```diff
@@ -420,6 +438,25 @@ func (p *clientProxy) start(ctx context.Context) error {
 				p.msgs.addLoginMessage(toController)
 			}
 		}
+		if msg.Type == "KeyManager" {
+			intercept, err := shouldInterceptKeyManager(p.controllerVersion)
```
alesstimec, line 442: shouldn't this be decided when the clientProxy is created, not for every message?

internal/rpcproxy/rpcproxy.go
```diff
@@ -789,6 +828,80 @@ func (p *clientProxy) handleAdminFacade(ctx context.Context, msg *message) (clie
 	}
 }
 
+// handleKeyManagerFacade processes a model-scoped Juju 4 KeyManager facade
+// call using JIMM's key store.
+func (p *clientProxy) handleKeyManagerFacade(ctx context.Context, msg *message) (*message, error) {
```
alesstimec, line 833: i'd rather not have this live in this file..

alesstimec (commented):
idk if this is the last time we'll have to intercept a facade or a bunch of facade methods.. should we consider building a general purpose interception mechanism?

internal/rpcproxy/rpcproxy.go
SimoneDutto: p.controllerVersion = connWithMetadata.ControllerVersion
the controller version is set by 		err := p.makeControllerConnection(ctx) on each message, so i guess we decide everytime based on the target controller.

SimoneDutto:
> idk if this is the last time we'll have to intercept a facade or a bunch of facade methods.. should we consider building a general purpose interception mechanism?

i hope we don't do this anymore, because it certaintly not good to intercept messages

internal/rpcproxy/rpcproxy.go
alesstimec: that doesn't sound very safe in case multiple messages come in at roughly the same time

SimoneDutto: nono, it's safe for that reason

kian99:
Let's wait on landing this btw until we see what the resolution is after the issue I found last week.

SimoneDutto:
> Let's wait on landing this btw until we see what the resolution is after the issue I found last week.

Which one? The ssh when a controller is bouncing? Or what else?

SimoneDutto:
> > Let's wait on landing this btw until we see what the resolution is after the issue I found last week.
> 
> Which one? The ssh when a controller is bouncing? Or what else?

Edit:
I remember, presenting the right key for the right model.

PR #2092, opened 2026-09-04
feat: add v8 cloud facade method version

ale8k:
## Description
JIMM lagged behind Juju on the Cloud facade (missing v8). This registers Cloud v8 and implements its one singularar new method, ModelConfigSchema. I plumbed it through properly, despite being a passthrough. 

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests


internal/jimm/juju/cloud.go
```diff
@@ -597,6 +597,38 @@ func (j *JujuManager) RemoveCloudFromController(ctx context.Context, user *openf
...
+// type. The schema is provider-static (it comes from the provider registry
+// compiled into the controller) so it is independent of any particular model
+// or cloud. JIMM therefore forwards the request to the first reachable
+// controller, trying each in turn until one responds.
+func (j *JujuManager) ModelConfigSchema(ctx context.Context, user *openfga.User, providerType string) (map[string]jujuparams.ModelConfigSchemaField, error) {
+	var controllers []dbmodel.Controller
+	err := j.Database.ForEachController(ctx, func(ctl *dbmodel.Controller) error {
+		controllers = append(controllers, *ctl)
+		return nil
+	})
```
kian99, line 610: This is trickier than it looks right? Because a Juju 4 controller could have a different schema from Juju 3 and we'll return a different schema depending on which controller we reach first.

ale8k: It only exists in v8 onwards, which is in 4 only - so this isn't a problem? Or maybe I'm misunderstanding sorry

kian99: Ah fair point. Then the only concern is whether the schema can change between Juju 4.0 -> 4.1 etc. That's just another problem of JAAS masking controllers.

kian99: The problem then also extends towards Juju 5 one day, etc.

kian99 (commented):
The code seems fine, just not sure about the issue raised.

internal/testutils/jimmtest/api.go
```diff
@@ -327,6 +328,13 @@ func (a *API) ListApplicationOffers(ctx context.Context, f []crossmodel.Applicat
 	return a.ListApplicationOffers_(ctx, f)
 }
 
+func (a *API) ModelConfigSchema(ctx context.Context, providerType string) (map[string]jujuparams.ModelConfigSchemaField, error) {
```
alesstimec, line 331: godoc

internal/testutils/jimmtest/mocks/jimm_juju_mock.go
```diff
@@ -176,6 +177,12 @@ func (j *JujuManager) GetCloudCredentialAttributes(ctx context.Context, u *openf
 	}
 	return j.GetCloudCredentialAttributes_(ctx, u, cred, hidden)
 }
+func (j *JujuManager) ModelConfigSchema(ctx context.Context, user *openfga.User, providerType string) (map[string]jujuparams.ModelConfigSchemaField, error) {
```
alesstimec, line 180: godoc

internal/jimm/juju/cloud.go
```diff
@@ -597,6 +597,38 @@ func (j *JujuManager) RemoveCloudFromController(ctx context.Context, user *openf
...
+	}
+
+	var dialErr error
+	for i := range controllers {
+		api, err := j.dialController(ctx, &controllers[i], user)
+		if err != nil {
+			dialErr = err
+			continue
+		}
+		schema, err := api.ModelConfigSchema(ctx, providerType)
```
alesstimec, line 625: this feels wrong.. at least we should be returning the first result that returns a nil error

internal/testutils/jimmtest/mocks/jimm_juju_mock.go
ale8k: it is a mock

internal/jimm/juju/cloud.go
ale8k: It now dials descending, capturing the final error (not really much else I can do here other than concatenate them)

internal/jimm/juju/cloud.go
```diff
@@ -597,6 +599,59 @@ func (j *JujuManager) RemoveCloudFromController(ctx context.Context, user *openf
...
+		switch {
+		case erri != nil && errk != nil:
+			return false
+		case erri != nil:
+			return false
+		case errk != nil:
+			return true
+		default:
+			return vi.Compare(vk) > 0
+		}
```
kian99, line 653: The sorting code is a bit unfortunate. Would it be easier to just sort the AgentVersion as a string?

internal/jimm/juju/cloud.go
```diff
@@ -597,6 +599,59 @@ func (j *JujuManager) RemoveCloudFromController(ctx context.Context, user *openf
...
+	})
+
+	var dialErr error
+	for i := range controllers {
+		api, err := j.dialController(ctx, &controllers[i], user)
+		if err != nil {
+			dialErr = err
+			continue
+		}
```
kian99, line 663: We are always masking some errors here which may be useful to debug an issue where one controller is not returning the schema. What if we collect all the errors, log the slice with `errors.Join()` and then return the last one. I suggest returning the last one because I believe clients don't handle multi-line errors well.

internal/jimm/juju/cloud.go
```diff
@@ -597,6 +599,59 @@ func (j *JujuManager) RemoveCloudFromController(ctx context.Context, user *openf
...
+		return nil, errors.New("no controllers registered")
+	}
+
+	// TODO: We prefer the highest reachable versioned controller when
+	// responding to config schema calls. Naturally, this means we may
+	// respond to an older version controller with a newer schema.
+	//
+	// We do this at least until we have a more stable solution, which may involve
+	// better handling of controller versions and schema compatibility.
+	sort.SliceStable(controllers, func(i, k int) bool {
```
kian99, line 641: We could also remove Juju 3 controllers from this slice since we know they don't implement the method

kian99 (approved):

internal/jimm/juju/cloud.go
ale8k: Depends, I think the lexicographic sorting would do something like 3.6.50 comes before 3.6.6 right? Could be quite the jump.

internal/jimm/juju/cloud.go
ale8k: I said this to Ales too, that I could join them [here](https://github.com/canonical/jimm/pull/2092#discussion_r4003731812) - happy to do that. 

Also, with your suggestion we don't need to join them we can just log them as they happen - right?

internal/jimm/juju/cloud.go
ale8k: Actually yeah, that makes so much more sense - I was originally skipping the test on 3. I'll add this idea ty!

internal/jimm/juju/cloud.go
kian99: Yeah could just log them as they happen.

internal/jimm/juju/cloud.go
```diff
@@ -597,6 +599,75 @@ func (j *JujuManager) RemoveCloudFromController(ctx context.Context, user *openf
 	return nil
 }
 
+// ModelConfigSchema returns the model config schema for the given provider
+// type. The schema is provider-static (it comes from the provider registry
+// compiled into the controller) so it is independent of any particular model
+// or cloud. JIMM therefore forwards the request to the first reachable
+// controller, trying each in turn until one responds.
+func (j *JujuManager) ModelConfigSchema(ctx context.Context, user *openfga.User, providerType string) (map[string]jujuparams.ModelConfigSchemaField, error) {
```
SimoneDutto, line 607: i feel like this method is even doing too much, can we just simply return the result of dialing the latest 4 controller, if it errors out we return the error.

If the controller is in an error state, we have bigger problem than model schema

testing/cloud_test.go
```diff
@@ -76,6 +76,30 @@ func TestClouds(t *testing.T) {
...
+	conn := s.Open(c, nil, "test@canonical.com", nil)
+	defer conn.Close()
+
+	client := cloudapi.NewClient(conn)
+	schema, err := client.ModelConfigSchema(t.Context(), jimmtest.TestE2EProviderType)
+	// If we have an error, it is likely we're testing specifically against Juju 3 only backing controller.
+	if err != nil {
+		c.Assert(err, qt.ErrorMatches, ".*no controllers registered.*")
+		return
+	}
```
SimoneDutto, line 92: what do you mean "likely"? :laughing: are we putting a likely flaky test?

SimoneDutto (commented):
minor comments but it looks good

testing/cloud_test.go
ale8k: Haha fair, I guess always actually unless Juju reports version wrongly

internal/jimm/juju/cloud.go
ale8k: Are you suggesting remove the descending reachable controller? Idm removing it, but we'd still remove vers 3 controllers right?

SimoneDutto: yes, just pick the latest 4 and that's it

SimoneDutto (approved):
ty

PR #2094, opened 2026-09-07
feat: ssh tunnel over http upgrade

luci1900:
# Description

Replaces the SSH-plus-JWT-password dial to the controller with an HTTP upgrade request to `GET /ssh-relay/:virtualHostname`. JIMM authenticates with a bearer JWT. The controller hijacks the connection after `101` and terminates the user's SSH session in an embedded server. JIMM blind-relays the bytes and holds no key.

## Key changes

- `internal/jimm/ssh`: `DialController` performs an HTTPS upgrade with a bearer JWT and returns the raw `net.Conn`. `BasicDialer.DialRelay` dials the controller's API port over TLS (replacing `InsecureIgnoreHostKey`). `DialInfo` drops `Port`, gains `TLSConfig`. `prefixConn` preserves buffered bytes past the 101 and forwards `CloseWrite`.
- `internal/ssh`: `DialController` takes a virtual hostname instead of a user and returns `net.Conn` instead of `*gossh.Client`. `directTCPIPHandler` passes the hostname through to the relay request.

## Checklist

- [x] Code style
- [x] Comments saying why
- [x] Go unit tests
- ~Integration tests~
- ~doc.go~

## QA steps

TODO

### Bootstrap

```
make dev-env
juju unregister jimm-dev
./local/jimm/setup-controller.sh
./local/jimm/add-controller.sh
```

### JIMM relay

```
juju add-model test
juju deploy ubuntu
juju add-unit ubuntu -n 1
juju add-ssh-key "$(cat ~/.ssh/id_ed25519.pub)"
juju ssh 0
juju scp ./foo.txt <machine>:/tmp/foo.txt
juju ssh <machine> cat /tmp/foo.txt
```

## Links

Requires https://github.com/juju/juju/pull/23228

### DIFF
--- a/internal/jimm/ssh/ssh.go
+++ b/internal/jimm/ssh/ssh.go
@@ -3,17 +3,21 @@
 package ssh
 
 import (
+	"bufio"
 	"context"
+	"crypto/tls"
 	"encoding/base64"
 	goerr "errors"
 	"fmt"
+	"io"
 	"net"
+	"net/http"
+	"net/url"
+	"strings"
 	"time"
 
 	"github.com/gliderlabs/ssh"
-	jujucontroller "github.com/juju/juju/controller"
 	"github.com/juju/zaputil/zapctx"
-	gossh "golang.org/x/crypto/ssh"
 
 	"github.com/canonical/jimm/v3/internal/dbmodel"
 	"github.com/canonical/jimm/v3/internal/errors"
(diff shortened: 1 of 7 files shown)


internal/jimm/ssh/ssh.go
```diff
@@ -43,25 +58,124 @@ type IdentityManager interface {
...
+	if err != nil {
+		_ = conn.Close()
+		return nil, fmt.Errorf("building relay upgrade request: %w", err)
+	}
+	req.Header.Set("Connection", "Upgrade")
+	req.Header.Set("Upgrade", relayUpgradeToken)
+	req.Header.Set("Authorization", "Bearer "+bearerToken)
+
+	if err := req.Write(conn); err != nil {
+		_ = conn.Close()
```
alesstimec, line 121: can we change the method signature to
```
func (d *BasicDialer) DialRelay(ctx context.Context, addr string, tlsConfig *tls.Config, virtualHostname, bearerToken string) (conn net.Conn, err error)
```

then instead of doing `_ = conn.Close()` we could do 
```
defer func() { 
  if err != nil {
    _ = conn.Close()
  }
}
```

wdyt?

luci1900: I suppose.

I don't generally like named return values.

PR #2095, opened 2026-09-08
feat: upgrade model upgrader to v2

ale8k:
## Description
In Juju 4.X ModelUpgrader has been updated to V2 but the wire params are the same and no new methods have been added. As such I've simply reregistered our current version from V1 onto V2. I also added a small test to verify we support the versions the test states we should. 

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests


SimoneDutto (commented):
if they are completely identical, why did they bump it?

SimoneDutto:
> if they are completely identical, why did they bump it?

i answer myself, because they've moved the facade to model level, and back to controller

SimoneDutto (approved):

ale8k:
> > if they are completely identical, why did they bump it?
> 
> i answer myself, because they've moved the facade to model level, and back to controller

Is this it? I think it is more they have a multi-state layer (there's controller and model there) and previously it was just model I think

(Although why is this a new facade version? Params and logic looks same to me so idk)

luci1900 (approved):

PR #2097, opened 2026-09-09
feat: add the ability to dial as the calling user

luci1900:
## Description
No dial callers are yet updated to do this.

Depends on https://github.com/canonical/jimm/pull/2096

## Engineering checklist

- [ ] Documentation updated
- [x] Covered by unit tests

### DIFF
--- a/internal/jimm/juju/interface.go
+++ b/internal/jimm/juju/interface.go
@@ -41,12 +41,12 @@ type Dialer interface {
 	DialControllerAsUser(ctx context.Context, user *openfga.User, ctl *dbmodel.Controller, resourceTags ...names.Tag) (API, error)
 
 	// DialModelAsSuperuser creates a model-scoped connection for the
-	// given user with forced superuser permissions.
+	// given user with superuser permissions.
 	// Use as a workaround.
 	DialModelAsSuperuser(ctx context.Context, user *openfga.User, ctl *dbmodel.Controller, modelTag names.ModelTag) (API, error)
 
 	// DialControllerAsSuperuser creates a controller-scoped connection
-	// for the given user with forced superuser permissions. Pass
+	// for the given user with superuser permissions. Pass
 	// resourceTags for operations tied to specific models or offers.
 	// Use as a workaround.
 	DialControllerAsSuperuser(ctx context.Context, user *openfga.User, ctl *dbmodel.Controller, resourceTags ...names.Tag) (API, error)
--- a/internal/jimm/juju/interface.go
+++ b/internal/jimm/juju/interface.go
@@ -42,13 +42,13 @@ type Dialer interface {
 
 	// DialModelAsSuperuser creates a model-scoped connection for the
 	// given user with superuser permissions.
-	// Use as a workaround.
+	// Deprecated: Use as a workaround.
 	DialModelAsSuperuser(ctx context.Context, user *openfga.User, ctl *dbmodel.Controller, modelTag names.ModelTag) (API, error)
 
 	// DialControllerAsSuperuser creates a controller-scoped connection
 	// for the given user with superuser permissions. Pass
 	// resourceTags for operations tied to specific models or offers.
-	// Use as a workaround.
+	// Deprecated: Use as a workaround.
 	DialControllerAsSuperuser(ctx context.Context, user *openfga.User, ctl *dbmodel.Controller, resourceTags ...names.Tag) (API, error)
 
 	// DialModelAsService creates a model-scoped API connection using
--- a/internal/jimm/juju/interface.go
+++ b/internal/jimm/juju/interface.go
@@ -42,12 +42,14 @@ type Dialer interface {
 
 	// DialModelAsSuperuser creates a model-scoped connection for the
 	// given user with superuser permissions.
+	//
 	// Deprecated: Use as a workaround.
 	DialModelAsSuperuser(ctx context.Context, user *openfga.User, ctl *dbmodel.Controller, modelTag names.ModelTag) (API, error)
 
 	// DialControllerAsSuperuser creates a controller-scoped connection
 	// for the given user with superuser permissions. Pass
 	// resourceTags for operations tied to specific models or offers.
+	//
 	// Deprecated: Use as a workaround.
 	DialControllerAsSuperuser(ctx context.Context, user *openfga.User, ctl *dbmodel.Controller, resourceTags ...names.Tag) (API, error)
 
--- a/internal/jimm/juju/interface.go
+++ b/internal/jimm/juju/interface.go
@@ -42,15 +42,13 @@ type Dialer interface {
 
 	// DialModelAsSuperuser creates a model-scoped connection for the
 	// given user with superuser permissions.
-	//
-	// Deprecated: Use as a workaround.
+	// Use as a workaround.
 	DialModelAsSuperuser(ctx context.Context, user *openfga.User, ctl *dbmodel.Controller, modelTag names.ModelTag) (API, error)
 
 	// DialControllerAsSuperuser creates a controller-scoped connection
 	// for the given user with superuser permissions. Pass
 	// resourceTags for operations tied to specific models or offers.
-	//
-	// Deprecated: Use as a workaround.
+	// Use as a workaround.
 	DialControllerAsSuperuser(ctx context.Context, user *openfga.User, ctl *dbmodel.Controller, resourceTags ...names.Tag) (API, error)
 
 	// DialModelAsService creates a model-scoped API connection using
--- a/internal/jimm/juju/interface.go
+++ b/internal/jimm/juju/interface.go
@@ -28,32 +28,27 @@ import (
 // dialing the controller the UUID, AgentVersion and HostPorts fields in the
 // given controller should be updated to the values provided by the
 // controller.
-//
-// When to use which method:
-//
-//   - DialModelAsSuperuser / DialControllerAsSuperuser: dial on behalf of a real user. Prefer
-//     these wherever possible.
-//
-//   - DialModelAsService / DialControllerAsService: dial as JIMM's
-//     service identity. Use only for internal housekeeping with no real
-//     user. Every AsService call is a potential gap to close in Juju's
-//     permission model.
 type Dialer interface {
-	// DialModelAsSuperuser creates a model-scoped API connection on behalf of a real
-	// user. The user and modelTag must not be zero-valued.
+	// DialModelAsUser creates a model-scoped connection for the given
+	// user with the caller's real permissions.
+	// Prefer whenever possible.
+	DialModelAsUser(ctx context.Context, user *openfga.User, ctl *dbmodel.Controller, modelTag names.ModelTag) (API, error)
+
+	// DialControllerAsUser creates a controller-scoped connection for
+	// the given user with the caller's real permissions. Pass
+	// resourceTags for operations tied to specific models or offers.
+	// Prefer whenever possible.
+	DialControllerAsUser(ctx context.Context, user *openfga.User, ctl *dbmodel.Controller, resourceTags ...names.Tag) (API, error)
+
+	// DialModelAsSuperuser creates a model-scoped connection for the
+	// given user with forced superuser permissions.
+	// Use as a workaround.
 	DialModelAsSuperuser(ctx context.Context, user *openfga.User, ctl *dbmodel.Controller, modelTag names.ModelTag) (API, error)
 
-	// DialControllerAsSuperuser creates a controller-scoped API connection on behalf
-	// of a real user. This is used for operations that call
-	// controller-level facades (e.g. ModelManager) with a model UUID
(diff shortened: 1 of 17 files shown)


internal/jimm/juju/interface.go
```diff
@@ -28,32 +28,27 @@ import (
...
-	// argument, or offer-related operations tied to a single offer: the
-	// connection must be controller-scoped, but the Juju permission
-	// check requires the user's access to the relevant resource(s).
-	//
-	// resourceTags is a generalisation of DialModel's modelTag: pass any
-	// mix of model and application-offer tags the operation is tied to.
-	// Pass no resourceTags when the operation is not tied to a specific
-	// resource (e.g. cloud or controller-level calls).
+	// DialControllerAsSuperuser creates a controller-scoped connection
+	// for the given user with forced superuser permissions. Pass
```
kian99:
```suggestion
	// for the given user with superuser permissions. Pass
```

internal/jimm/juju/interface.go
```diff
@@ -28,32 +28,27 @@ import (
...
+	DialModelAsUser(ctx context.Context, user *openfga.User, ctl *dbmodel.Controller, modelTag names.ModelTag) (API, error)
+
+	// DialControllerAsUser creates a controller-scoped connection for
+	// the given user with the caller's real permissions. Pass
+	// resourceTags for operations tied to specific models or offers.
+	// Prefer whenever possible.
+	DialControllerAsUser(ctx context.Context, user *openfga.User, ctl *dbmodel.Controller, resourceTags ...names.Tag) (API, error)
+
+	// DialModelAsSuperuser creates a model-scoped connection for the
+	// given user with forced superuser permissions.
```
kian99:
```suggestion
	// given user with superuser permissions.
```

internal/jimm/jujuauth/callertoken_test.go
```diff
@@ -0,0 +1,316 @@
...
+	ofganames "github.com/canonical/jimm/v3/internal/openfga/names"
+	"github.com/canonical/jimm/v3/internal/testutils/jimmtest"
+)
+
+// TestCallerScopedLoginTokenForNonAdminUser verifies that NewCallerScopedLoginToken
+// mints a JWT carrying the caller's real OpenFGA-derived permissions
+// (controller: login, cloud: add-model, model: write) rather than the
+// old hardcoded superuser claim. This is the core behavioural guarantee
+// of the de-proxying change: non-admin users must not receive
+// controller-superuser tokens.
```
kian99:
```suggestion
// TestCallerScopedLoginTokenForNonAdminUser verifies that NewCallerScopedLoginToken
// mints a JWT carrying the caller's real OpenFGA-derived permissions
// (controller: login, cloud: add-model, model: write).
```

internal/jimm/jujuauth/callertoken_test.go
```diff
@@ -0,0 +1,316 @@
...
+	bobIdentity, err := dbmodel.NewIdentity(bobEmail)
+	c.Assert(err, qt.IsNil)
+	err = env.JIMM.Database.GetIdentity(ctx, bobIdentity)
+	c.Assert(err, qt.IsNil)
+	bob := env.NewUser(bobIdentity)
+
+	// Add a controller to the DB so buildAccessMap can fetch it.
+	// A cloud row must exist first to satisfy the controllers_cloud_name_fkey.
+	cloudName := "test-cloud"
+	err = env.JIMM.Database.AddCloud(ctx, &dbmodel.Cloud{Name: cloudName, Type: "lxd"})
```
kian99: You can also use a yaml definition to populate the DB which is more common in `internal/jimm/` but not necessarily stricly required. Ctrl-f for `const testControllerConfigEnv` for example.

internal/jimm/jujuauth/logintokens.go
```diff
@@ -107,6 +109,88 @@ func (auth *LoginTokenGenerator) makeSuperuserToken(ctx context.Context, user *o
...
+
+	// Juju rejects logins with empty access claims. Drop empty claims
+	// when the user holds real access to at least one tag. Otherwise,
+	// keep one so Juju itself denies the login.
+	for target, access := range targetAccess {
+		if access == "" && hasRealAccess {
+			continue
+		}
+		accessMap[target.String()] = access
+	}
```
kian99: This is quite weird. Can we get rid of `hasRealAccess` and at the end of this loop just call `len(accessMap)` and do something if its empty? It seems like lower down, we're adding the user's controller access so the map will never be empty?

internal/jimm/jujuauth/logintokens.go
```diff
@@ -107,6 +109,88 @@ func (auth *LoginTokenGenerator) makeSuperuserToken(ctx context.Context, user *o
...
+		return access, nil
+	case names.ApplicationOfferTag:
+		access, err := accessChecker.GetUserApplicationOfferAccess(ctx, user, tag)
+		if err != nil {
+			zapctx.Error(ctx, "application offer access check failed", zap.Error(err))
+			return "", err
+		}
+		return access, nil
+	default:
+		return "", nil
```
kian99: It seems misleading that a caller can specify that they want a token scoped for access to tags beyond models/application-offers and yet here we silently swallow that without an error. I would expect this to be an error.

kian99 (commented):
I think I'm finally able to articulate my thoughts on this change a bit better.

When we added JWT support between Juju and JIMM there was a question of what permissions we should put in the JWT. Especially for the model proxy, it seemed a difficult task to try and inspect each request and determine every authorisation check that will be needed. So the model proxy just mints a fairly broad-scoped permission and Juju returns a certain kind of error asking for more permissions if necessary.

For the controller methods that JIMM does terminate, we would authorize the user in JIMM then make an "admin" call to the backing controller. This was a limited set of methods and it was more feasible. In this change, now that we're being more fine-grained about the JWT we mint and send to Juju, we are in a sense achieving a tighter coupling between JIMM and Juju's permissions. That can be tricky if a method in Juju 4 requires a different set of permissions to Juju 3. But on the positive side it means we have a better security posture and are working towards de-proxying JIMM and returning the tokens to the user.

I'd like to suggest a different path towards the de-proxy objective that we've sort of discussed before. If JIMM were to act purely as a reverse proxy that doesn't hide the fact that there are multiple controllers, it could perform only authentication. You have e.g. `jimm.com/controller-1`, `jimm.com/controller-2`, etc. All controllers are separate, losing us some of the capabilities of JIMM, but you connect through JIMM which does something like Traefik's [forwardAuth](https://doc.traefik.io/traefik/reference/routing-configuration/http/middlewares/forwardauth/) and JIMM just issues a token when forwarding the call to the backing controller that **only** holds your identity. If we then want delegated authorisation where multiple controllers rely on the same source of truth for permissions, we'd need to think about the backwards compatibility story across multiple Juju versions with different sets of permissions but scoped down to just permissions rather than being backwards compatible for the API, etc.

Anyway, rambling aside. Just some comments. The scoping down of the JWT we send Juju does I think make us more tightly coupled to Juju's permission model and that might prove to be complicating in the future. We should weigh that against the improved security.

internal/jimm/jujuauth/logintokens.go
```diff
@@ -107,6 +109,75 @@ func (auth *LoginTokenGenerator) makeSuperuserToken(ctx context.Context, user *o
...
+		}
+	}
+
+	controllerAccess, err := accessChecker.GetUserControllerAccess(ctx, user, ct)
+	if err != nil {
+		return nil, err
+	}
+	accessMap[ct.String()] = controllerAccess
+
+	clouds := make(map[names.CloudTag]bool)
```
kian99: More idiomatic way of representing a set in Go. I believe we have sets coming in the stdlib in an upcoming Go release.
```suggestion
	clouds := make(map[names.CloudTag]struct{})
```

kian99 (approved):
LGTM, let's also discuss the increased coupling with Juju's permission models in stdup to make sure we're all aligned on what that means.

internal/jimm/juju/interface.go
```diff
@@ -28,32 +28,27 @@ import (
...
+
+	// DialControllerAsUser creates a controller-scoped connection for
+	// the given user with the caller's real permissions. Pass
+	// resourceTags for operations tied to specific models or offers.
+	// Prefer whenever possible.
+	DialControllerAsUser(ctx context.Context, user *openfga.User, ctl *dbmodel.Controller, resourceTags ...names.Tag) (API, error)
+
+	// DialModelAsSuperuser creates a model-scoped connection for the
+	// given user with superuser permissions.
+	// Use as a workaround.
```
ale8k: Can mark it as deprecated too, that way our editors will strike it through :):

```
	// DialModelAsSuperuser creates a model-scoped connection for the
	// given user with superuser permissions.
    //
	// Deprecated: Use as a workaround.
```

internal/jimm/jujuauth/logintokens.go
```diff
@@ -107,6 +109,75 @@ func (auth *LoginTokenGenerator) makeSuperuserToken(ctx context.Context, user *o
 	})
 }
 
+// resolveTargetAccess resolves the caller's access level for a single
+// target tag.
+func resolveTargetAccess(ctx context.Context, user *openfga.User, target names.Tag, accessChecker GeneratorAccessChecker) (string, error) {
+	switch tag := target.(type) {
+	case names.ModelTag:
+		access, err := accessChecker.GetUserModelAccess(ctx, user, tag)
```
ale8k, line 117: As fine grained is on its way this is fine, but maybe in future we generate this resvoling code from our auth model

internal/jimm/jujuauth/logintokens.go
```diff
@@ -107,6 +109,75 @@ func (auth *LoginTokenGenerator) makeSuperuserToken(ctx context.Context, user *o
...
+		return access, nil
+	case names.ApplicationOfferTag:
+		access, err := accessChecker.GetUserApplicationOfferAccess(ctx, user, tag)
+		if err != nil {
+			zapctx.Error(ctx, "application offer access check failed", zap.Error(err))
+			return "", err
+		}
+		return access, nil
+	default:
+		return "", fmt.Errorf("unsupported resource tag type: %T", target)
```
ale8k, line 131: Whilst the happy path of login is tested via `NewCallerLoginToken` which ultimately boils down to this, the default path here isn't tested. I think we should have explicit tests for both `buildAccessMap` and `resolveTargetAccess` to cover the none happy path cases

internal/jujuclient/dial.go
```diff
@@ -87,10 +101,49 @@ func (d *Dialer) createLoginRequest(ctx context.Context, ctl *dbmodel.Controller
 	}, nil
 }
 
+// createUserLoginRequest creates a login request carrying the caller's
+// real access claims. Used by AsUser dials.
```
ale8k: Remove used by doc plz, this is subject to change and can be misleading

internal/jujuclient/dial_test.go
```diff
@@ -0,0 +1,77 @@
...
+	qt "github.com/frankban/quicktest"
+	"github.com/google/uuid"
+	"github.com/juju/names/v5"
+
+	"github.com/canonical/jimm/v3/internal/dbmodel"
+	"github.com/canonical/jimm/v3/internal/jimmjwx"
+	"github.com/canonical/jimm/v3/internal/openfga"
+)
+
+type fakeTokenMinter struct {
```
ale8k, line 20: Why a whole new file? Why not just dial_test.go?

internal/jujuclient/dial_test.go
```diff
@@ -0,0 +1,77 @@
...
+
+func (f *fakeTokenMinter) NewCallerLoginToken(ctx context.Context, resourceTags []names.Tag, ctl *dbmodel.Controller, user *openfga.User) ([]byte, error) {
+	f.called = true
+	f.resourceTags = resourceTags
+	f.controller = ctl
+	f.user = user
+	return []byte("fake-jwt"), nil
+}
+
+func TestDialModelAsUserRejectsNilUser(t *testing.T) {
```
ale8k, line 39: Can also test that correctly reports createLoginRequestErrors, would be a doozy if we didn't haha

ale8k (approved):
Really like this PR, just minor comments honestly. Nice work! Just asking for some QoL bits and extra tests as it is an auth based PR and good to have coverage in these areas.

internal/jujuclient/dial_test.go
luci1900: Different package.

luci1900: Different package.

internal/jujuclient/dial_test.go
luci1900: Added

internal/jimm/juju/interface.go
luci1900: The linter complains a lot if i do that, so I'll revert.

PR #2100, opened 2026-09-15
fix: leaking ssh connection to controller

SimoneDutto:
## Description

We weren't closing the client on which we are creating ssh channels.

Now the copy routines are waited and the controller client is always closed at the end of the handler.

# QA

`make qa-lxd`

`juju add-ssh-key <>`
`juju ssh 0`

`docker exec jimm -it /bin/bash`
`ss -o state established '( dport = :17022 )'`

Before on each juju ssh 0 connection the list keeps growing. Now it should be empty everytime an `juju ssh` is exited.


internal/ssh/ssh.go
```diff
@@ -182,21 +186,29 @@ func directTCPIPHandler(sshManager SSHManager) func(srv *ssh.Server, conn *gossh
 		// Since we only need the raw data to redirect, we can discard them.
 		go gossh.DiscardRequests(reqs)
 
+		// Copy data in both directions until the tunnel is torn down.
+		// Closing either connection unblocks the other copy, and the
+		// deferred client.Close above then tears down the connection
+		// to the controller.
+		var wg sync.WaitGroup
+		wg.Add(2)
 		go func() {
```
kian99: Use the newer wg.Go here.

kian99 (approved):
Great job finding this

internal/ssh/ssh.go
SimoneDutto: right!

ale8k (approved):
Ah nice find, we were only closing channel

PR #2101, opened 2026-09-15
test: add harness to test for leaked conns

SimoneDutto:
# Description

The idea is to add an automated way to test leaked connections to backing controllers in our e2e testing suite.
The idea is pretty simple:
- we keep track of connection we create, with info about the target controller, model and where in the code the conn was created
- the Dial method returns a tracked connection which is then closed by our code
- at the end of each test we check for the tracked connection map in search of leaking connections, we fail the test and print the locations of leaked connections

This is only enabled in tests, and it doesn't affect production. And i've caught two interesting places where we weren't closing the log streams in the tests, as you can see from the diff.

If you want to see it in action, revove the `api.Close` from the ModelInfo call, and see the traces:

```
leaked connection to controller "test-e2e" (model ""), dialed from:
        github.com/canonical/jimm/v3/internal/jujuclient.(*Dialer).dial
          /home/duttos/jaas/jimm/internal/jujuclient/dial.go:208
        github.com/canonical/jimm/v3/internal/jujuclient.(*Dialer).DialControllerAsSuperuser
          /home/duttos/jaas/jimm/internal/jujuclient/dial.go:170
        github.com/canonical/jimm/v3/internal/jimm.(*DialerAdapter).DialControllerAsSuperuser
          /home/duttos/jaas/jimm/internal/jimm/jimm.go:390
        github.com/canonical/jimm/v3/internal/jimm/juju.(*JujuManager).dialControllerAsSuperuser
          /home/duttos/jaas/jimm/internal/jimm/juju/juju.go:111
        github.com/canonical/jimm/v3/internal/jimm/juju.(*JujuManager).ModelInfo
          /home/duttos/jaas/jimm/internal/jimm/juju/model.go:219
        github.com/canonical/jimm/v3/internal/jujuapi.(*controllerRoot).ModelInfo
          /home/duttos/jaas/jimm/internal/jujuapi/modelmanager.go:187
```

Let me know what you think.


luci1900 (approved):
This is nice.

I would prefer that we do this using tracing APIs, which would make it disappear between tests/production. But that can be done later.

SimoneDutto:
> This is nice.
> 
> I would prefer that we do this using tracing APIs, which would make it disappear between tests/production. But that can be done later.

what do you mean using tracing API? do you have any doc?

internal/rpc/dial.go
```diff
@@ -22,6 +25,81 @@ import (
...
+	registry = map[*websocket.Conn]ConnInfo{}
+)
+
+// EnableConnTracking enables recording controller connections opened
+// via Dial, so tests can detect leaked connections. Test setups only.
+func EnableConnTracking() {
+	connTracking.Store(true)
+}
+
+func trackConn(conn *websocket.Conn, info ConnInfo) {
```
alesstimec, line 63:
```suggestion
func registerConn(conn *websocket.Conn, info ConnInfo) {
```

internal/rpc/dial.go
```diff
@@ -22,6 +25,81 @@ import (
...
+	if !connTracking.Load() {
+		return
+	}
+	info.DialStack = captureDialStack()
+	regMu.Lock()
+	defer regMu.Unlock()
+	registry[conn] = info
+}
+
+func untrackConn(conn *websocket.Conn) {
```
alesstimec, line 73:
```suggestion
func unregisterConn(conn *websocket.Conn) {
```

luci1900:
> > This is nice.
> > I would prefer that we do this using tracing APIs, which would make it disappear between tests/production. But that can be done later.
> 
> what do you mean using tracing API? do you have any doc?

I mean distibuted tracing like otel or similar. We *could* have a custom backend that runs during tests and only extracts spans/calls to this and counts, etc.

I'm not sure it'd actually be simpler, so it can be explored some other time.

alesstimec (commented):
we are adding a bunch of code just for testing purposes..

PR #2102, opened 2026-09-16
feat: group fetcher for offer authorizer

SimoneDutto:
## Description

Adding the plumbing for the group fetcher in offer authorizer, plus some additional test converage.

Ofc `internal/jimm/idpgroupfetcher/idpgroupfetcher.go` this will be changed in the future to whatever is needed for the actual hook service implementation, but it was to have it all wired.


cmd/jimmsrv/service/service_test.go
kian99, line 1: If you want to clean this up in a follow-up, these tests and the offer discharger logic shouldn't be mixed in here where we mostly just wire up dependencies.

kian99 (approved):

luci1900 (approved):

cmd/jimmsrv/service/service_test.go
SimoneDutto: right, however i was scared of removing coverage on this path

PR #2105, opened 2026-09-17
feat: hook service api to fetch groups for the user

SimoneDutto:
## Description

This pr introduces the hook service client and group fetcher implementation to fetch the groups for a user at discharge time.

A nice addition is that the hook service can be run with only with postgres, so i've added it to our docker compose.

What's missing is a way for us to authenticate to the hook service, because right now it's simply disabled.
But i think it's something we'll come out of the meeting with the identity team.

> NOTE: the identity client has an AGPL licence, so we need to evaluate if it's ok to use it for JIMM.

# QA
The tests.


internal/jimm/idpgroupfetcher/hookservice.go
```diff
@@ -0,0 +1,73 @@
...
+	// token is sent as a Bearer token on every call; the hook-service
+	// gRPC interceptor requires the header even when authentication is
+	// disabled.
+	token string
+}
+
+// NewHookService returns a HookService fetcher connected to the
+// hook-service gRPC API at the given address.
+func NewHookService(address, token string) (*HookService, error) {
+	conn, err := grpc.NewClient(address, grpc.WithTransportCredentials(insecure.NewCredentials()))
```
alesstimec, line 36: we never close this connection

internal/jimm/idpgroupfetcher/hookservice.go
```diff
@@ -0,0 +1,73 @@
...
+		token:  token,
+	}, nil
+}
+
+// FetchGroups implements offer.IdPGroupFetcher. It fails closed: any
+// error from the hook-service denies access.
+func (h *HookService) FetchGroups(ctx context.Context, username string) ([]string, error) {
+	userID := strings.TrimSuffix(username, serviceAccountSuffix)
+
+	ctx = metadata.AppendToOutgoingContext(ctx, "authorization", "Bearer "+h.token)
```
alesstimec, line 51: can we add a sensible timeout to this context, please.. otherwise the discharge might have to wait for a long long time

internal/jimm/idpgroupfetcher/hookservice.go
```diff
@@ -0,0 +1,73 @@
...
+	stream, err := h.client.GetGroupsForUser(ctx, &pb.GetGroupsForUserReq{
+		UserId: userID,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch groups for user %q: %w", username, err)
+	}
+
+	var groups []string
```
alesstimec, line 59:
```suggestion
	var groupNames []string
```

internal/jimm/idpgroupfetcher/hookservice.go
```diff
@@ -0,0 +1,73 @@
...
+	for {
+		mapping, err := stream.Recv()
```
alesstimec, line 61:
```suggestion
		group, err := stream.Recv()
```

internal/jimm/idpgroupfetcher/hookservice.go
```diff
@@ -0,0 +1,73 @@
...
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			return nil, fmt.Errorf("failed to fetch groups for user %q: %w", username, err)
+		}
+		if name := mapping.GetName(); name != "" {
```
alesstimec, line 68:
```suggestion
		if name := group.GetName(); name != "" {
```

internal/jimm/idpgroupfetcher/hookservice.go
```diff
@@ -0,0 +1,73 @@
...
+			groups = append(groups, name)
```
alesstimec, line 69:
```suggestion
			groupNames = append(groupNames, name)
```

internal/jimm/idpgroupfetcher/hookservice.go
```diff
@@ -0,0 +1,73 @@
...
+		}
+	}
+	return groups, nil
```
alesstimec, line 72:
```suggestion
	return groupNames, nil
```

internal/jimm/idpgroupfetcher/hookservice.go
alesstimec, line 36:
```
conn, err := grpc.NewClient(address,
    grpc.WithTransportCredentials(...),
    grpc.WithPerRPCCredentials(oauth.TokenSource{
        TokenSource: oauth2.StaticTokenSource(&oauth2.Token{AccessToken: token, TokenType: "Bearer"}),
    }),
)
```

SimoneDutto: the conn here is lazy and shared across multiple calls to fetch groups. For each different call a "stream" is created and closed once Recv() receives an EOF.

So it's never leaked, but we have a long running connection for the lifecycle of jimm

internal/jimm/idpgroupfetcher/hookservice.go
SimoneDutto: i think the timeout should come from the callers right? it's the discharge who should decide how much it wants to wait.

docker-compose.yaml
```diff
@@ -151,6 +151,34 @@ services:
...
 
+  # Minimal hook-service (Canonical IDP group store) for integration testing
+  # the IdP group fetcher. Only needs its own postgres; Kratos/Hydra/OpenFGA
+  # are not required (authentication and authorization are disabled).
+  # Groups and members are seeded by local/hook-service/entrypoint.sh.
+  hook-service-db:
+    image: postgres:16-alpine
+    container_name: hook-service-db
+    environment:
+      POSTGRES_DB: groups
```
ale8k, line 162: Can just use the postgres we already have why add another?

local/hook-service/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+WORKDIR /app
+
+COPY entrypoint.sh /app/entrypoint.sh
+RUN chmod +x /app/entrypoint.sh
+
+ENTRYPOINT [ "/app/entrypoint.sh" ]
+
+# Opt to check for a file rather than the hook-service itself
+# because we perform migrations and seeding after starting the server.
+HEALTHCHECK \
```
ale8k, line 23: You can put healthcheck in the compose and check for the file still. Doesn't have to go here I don't think & then it'd be consistent with other healthchecks in our compose.

local/hook-service/entrypoint.sh
```diff
@@ -0,0 +1,51 @@
...
+
+# Starts the hook-service, migrates its database and seeds IdP groups and
+# members used by JIMM integration tests.
+#
+# To add users or groups, edit GROUP_NAME/GROUP_MEMBERS below and
+# recreate the container: docker compose up -d --build hook-service
+
+set -e
+
+DSN="postgres://groups:groups@hook-service-db:5432/groups?sslmode=disable"
```
ale8k, line 11: Pass this as an arg? So it doesn't look like magic in the compose?

internal/jimm/offer/authorizer_test.go
```diff
@@ -170,6 +171,50 @@ func TestIsUserConsumerForOffer(t *testing.T) {
 	}
 }
 
+// TestIsUserConsumerForOfferViaHookServiceGroupFetcher verifies the
+// authorization path with the real hook-service IdPGroupFetcher
+// (docker compose up -d hook-service).
+func TestIsUserConsumerForOfferViaHookServiceGroupFetcher(t *testing.T) {
```
ale8k, line 177: TestUserIsConsumer* ?

internal/jimm/idpgroupfetcher/idpgroupfetcher.go
```diff
@@ -10,22 +8,25 @@ import (
...
+	HookServiceAddress string
+	// HookServiceToken is the Bearer token sent to the hook-service.
+	HookServiceToken string
+}
 
 // New returns an offer.IdPGroupFetcher configured by the given params.
-//
-// TODO: construct the identity provider implementation from p when the
-// integration lands.
 func New(ctx context.Context, p Params) (offer.IdPGroupFetcher, error) {
```
ale8k, line 21: This feels wrong, we're returning an interface for "something" to use our group fetcher, when they should be definiung the interface on the consuming side, plus the params it is taking are very specific anyway

 why do we need this?

internal/jimm/idpgroupfetcher/hookservice.go
```diff
@@ -0,0 +1,73 @@
...
+	"context"
+	"fmt"
+	"io"
+	"strings"
+
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+	"google.golang.org/grpc/metadata"
+
+	pb "github.com/canonical/hook-service/gen/hook/groups/v1"
```
ale8k, line 15: pb? Protobuf?

internal/jimm/idpgroupfetcher/hookservice_test.go
```diff
@@ -0,0 +1,50 @@
...
+	c.Assert(groups, qt.HasLen, 0)
+
+	// The @serviceaccount suffix is stripped before lookup.
+	groups, err = fetcher.FetchGroups(ctx, "jimm-group-client@serviceaccount")
+	c.Assert(err, qt.IsNil)
+	c.Assert(groups, qt.DeepEquals, []string{"canonical"})
+}
+
+// TestNewNoOp verifies the factory returns the NoOp fetcher when no
+// hook-service address is configured.
```
ale8k, line 41: We can define the no op logic on the consuming side this doesn't seem right

internal/jimm/idpgroupfetcher/hookservice.go
```diff
@@ -0,0 +1,73 @@
...
+	ctx = metadata.AppendToOutgoingContext(ctx, "authorization", "Bearer "+h.token)
+	stream, err := h.client.GetGroupsForUser(ctx, &pb.GetGroupsForUserReq{
+		UserId: userID,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch groups for user %q: %w", username, err)
+	}
+
+	var groups []string
+	for {
```
ale8k, line 60: We have no means to stop streaming when the context is cancelled, can we add a select here for context cancellation plz

internal/jimm/idpgroupfetcher/hookservice.go
ale8k, line 68: If name is "" we are doing nothing - why?

internal/jimm/idpgroupfetcher/hookservice.go
```diff
@@ -0,0 +1,73 @@
...
+	"google.golang.org/grpc/credentials/insecure"
+	"google.golang.org/grpc/metadata"
+
+	pb "github.com/canonical/hook-service/gen/hook/groups/v1"
+)
+
+// serviceAccountSuffix is appended to service account usernames by the
+// identity platform. The hook-service keys service accounts by their raw
+// client ID, so the suffix must be stripped before lookup.
+const serviceAccountSuffix = "@serviceaccount"
```
ale8k, line 21: Good spot here nice, totally forgot about svc accs

internal/jimm/idpgroupfetcher/hookservice_test.go
```diff
@@ -0,0 +1,50 @@
...
+	"context"
+	"testing"
+
+	qt "github.com/frankban/quicktest"
+
+	"github.com/canonical/jimm/v3/internal/jimm/idpgroupfetcher"
+	"github.com/canonical/jimm/v3/internal/testutils/jimmtest"
+)
+
+// TestHookServiceFetchGroups tests the HookService fetcher against a
```
ale8k, line 15: Can we add cases for:

1. Failed hook service constructor creation
2. The two error scenarios in fetch groups  (getgroupsforuser & stream failure)
3. Context cancellation (I've made a suggestion to add it)

